PR #2, opened 2022-01-20
trymito.io: add media to landing page

naterush:


naterush:
@aarondr77 wanna mess about with the media sizing? It should be offset from the background, but I've been trying on it for a few hours and I've lost steam...

Also, peep how I prefix my commits with the area of the codebase it touches. Let's try and do this going forward, for git cleanliness.

PR #3, opened 2022-01-20
address website pr review

aarondr77:


naterush:
Onboarding Program is still capital

naterush:
I think we should go through and compress all of our images. The website on mobile loads real slow... Or at least we should put alts on them or something.

naterush:
The width of the dropdown on mobile for the plans is a tiny bit too small.

PR #6, opened 2022-01-21
trymito: fix minor mobile and formatting bugs

naterush:


aarondr77:
Did your video tell you how to make the SVG take up more space? It looks amazing in the mockups IMO. It either needs to be displayed in the padding/margin or we need to remove the padding and just space with div that has the image on it

Screen Shot 2022-01-21 at 10 24 02 AM

naterush:
@aarondr77 I fucked around with it for a bit, and it's tough. I think it's ok for now - if you wanna come and fuck around with it more go ahead, but I'm calling it done here!

PR #25, opened 2022-01-27
mitosheet: turn off usage triggered feedback

naterush:


mitosheet/mitosheet/mito_widget.py
```diff
@@ -245,9 +245,13 @@ def set_usage_triggered_feedback_id(self):
         previous_step_type = self.steps_manager.steps[self.steps_manager.curr_step_idx - 1].step_type
         
         if self.steps_manager.curr_step.step_type != previous_step_type and previous_step_type != 'bulk_old_rename':
+            """
+            Turn off usage triggered feedback
+
             if random.randint(1, 400) == 1:
                 self.usage_triggered_feedback_id = STEP_TYPE_TO_USAGE_TRIGGERED_FEEDBACK_ID[previous_step_type]
                 return
+            """
```
naterush, line 254: This isn't valid Python; if you have an if statement, you need something inside of it (like `pass`). This must just be getting caught by some error handling code.

Also, let's not keep old code around commented out. Can you just delete the if statement entirely?

naterush (commented):
Minor change, then LGTM.

naterush:
@aarondr77 try and keep your commit messages informative. So instead of the last two commits, you could have done

mitosheet: remove unnecessary check for displaying in app feedback
mitosheet: final cleanup of documentation for in app feedback

PR #26, opened 2022-01-27
mitosheet: add deployment scripts and workflow

naterush:
# Description

Adds deployment scripts, and deployment action. 

Before merging this, we need to add the secrets to this Github repo that allow these actions to post to PyPi. Since I don't have access to PyPi rn, @aarondr77 the task falls to you. To do so:
1. Go to the Github secrets for our org here: https://github.com/organizations/mito-ds/settings/secrets/actions
2. Create a new org secret called `PYPI_API_TOKEN`. Set it equal to the value you get from PyPi (instructions here: https://pypi.org/help/#apitoken). Make sure to give it enough permissions to upload packages.
3. Add another secret, called `TEST_PYPI_API_TOKEN`, which you can get from test pypi. Otherwise, the same.

Then, you should be good to merge, and this should run correctly.

# Testing

We'll have to test this once we merge it for the first time!

# Documentation

No.

### DIFF
--- a/.github/workflows/deploy-mitosheet-mitoinstaller.yml
+++ b/.github/workflows/deploy-mitosheet-mitoinstaller.yml
@@ -1,0 +1,156 @@
+name: Deploy mitosheet and mitoinstaller
+on:
+  push:
+    branches:
+      - dev
+      - main
+
+jobs:
+  deploy-mitosheet:
+    name: Deploy mitosheet and mitosheet3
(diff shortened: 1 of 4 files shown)


naterush:
Before merging this, can you add the PYPI secrets here: https://github.com/organizations/mito-ds/settings/secrets/actions

TODO: add instructions

Issue #17, opened 2022-01-26
Add types to remaining functions


naterush:
We still have bugs in our Python code as a result of no types on our functions. This is a bigger issue in the API where there is no type on the step manager (which we need to fix, because circular imports).

naterush:
Ok, I opened #29 PR, which does a lot of this work. There are two major changes I want to do still before calling this done:
1. Add an explicit ColumnHeader and ColumnID type; I think this will _greatly_ improve clarity of the codebase, and make clear what we're working with.
2. Add an explicitly Event type. For the same reason.

aarondr77:
Yeah, I think having a ColumnHeader and ColumnID type would be incredibly helpful. Both in Python and Typescript. Trying to figure out if the string type in an object is the columnHeader or columnID forces me to go back to the comments in the types.tsx file all the time. Enforcing that the type is correct in the type system would be super helpful.

naterush:
@aarondr77 PR #32 does this on the Python side; we can and will add explicit type aliases for ColumnIDs in Typescript as well, before closing this issue.

PR #29, opened 2022-01-28
Add mypy types

naterush:
# Description

Closes issue #17 

Also makes us more aggressive about types going forward. I think we can continually refactor these to make them useful, but I want good types everywhere always! They don't only protect us - in many places they make the code a lot more readable (and so good for contributions).

# Testing

For any places real code changed, make sure it's good.

# Documentation

No.


naterush (commented):
Ruff, but ok.

PR #32, opened 2022-01-29
mitosheet: add explicit types for ColumnHeader and ColumnID, for clarity

naterush:
# Description

This makes no functional changes; it just edits types to make things more clear. Seeing List[ColumnID] is much more clear than List[str]!

In making these changes, this PR also catches a few bugs where we were treating column headers as strings when they had no right to be! It will continue to catch bugs for us in the future as well :-)

# Testing

# Documentation


naterush (commented):
Minor

PR #37, opened 2022-02-02
mitosheet: have error modal display an traceback

naterush:
# Description

@aarondr77 this is something you've been talking about for a while, so feast your eyes. Feel free to make it prettier / look better. 

Note that this is a workaround as we continue to improve our in place error handling; this does not want to stay forever!

# Testing

Try as many errors as you can think of:
1. Invalid formulas
2. Invalid merge
3. Writing any formula in a sheet with the column header `0` <- this is in `small-datasets/small-excel`
4. And any more you can think of!

Note the goal here is to make sure none of our error infra broke; so some of the above don't trigger the error modal at all.

# Documentation

### DIFF
--- a/mitosheet/mitosheet/errors.py
+++ b/mitosheet/mitosheet/errors.py
@@ -36,6 +36,7 @@ class MitoError(Exception):
         self.header = header
         self.to_fix = to_fix
         self.error_modal = error_modal
+        self.traceback = get_recent_traceback() # record the most recent error when the error is created
 
 def make_no_sheet_error(sheet_indexes: Set[int]) -> MitoError:
     """
--- a/mitosheet/mitosheet/mito_widget.py
+++ b/mitosheet/mitosheet/mito_widget.py
@@ -204,14 +204,16 @@ class MitoWidget(DOMWidget):
                 'id': event['id'],
                 'type': e.type_,
                 'header': e.header,
-                'to_fix': e.to_fix
+                'to_fix': e.to_fix,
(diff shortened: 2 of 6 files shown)


aarondr77:
See in this video, when I changed the imported dataframe, I expected to see an error because the column that the analysis relied on did not exist anymore. However, there was no error displayed to the user. Instead the sheet was out of sync with generated code and a JupyterLab error was thrown. 

https://user-images.githubusercontent.com/18709905/152701666-d1ce23e0-9ae5-42f6-8a5b-5374fc95d647.mov

aarondr77:
I think this PR is def a big step in the right direction of helping users understand how to recover from bad states. Not for now, but at a later point, it would be nice to improve our communication even more. ie: See this error traceback for writing a formula in a dataset that has a column header of `0`. This error isn't helpful. 

But lets worry about that at a later date!

Screen Shot 2022-02-06 at 4 22 13 PM

aarondr77:
Whatever you want to do with that comment, then LGTM.

naterush:
@aarondr77 opening an issue for that first error you reported. You can see it here: #48 

Also, I personally don't think displaying raw tracebacks to users is _ever_ going to be a particularly good solution. It will _always_ likely display more context on our codebase than is helpful in resolving the issue (for anyone who isn't a dev on the codebase). 

The path forward for error handing IMO is a) stopping users from generating them whereever we can with better upfront communication and b) giving actually informative, specific errors that we manually catch on the backend.

PR #38, opened 2022-02-02
error boundary: add basic error boundary to Mito to stop violent sheet crashes

naterush:
# Description

See title.

# Testing

Add this line somewhere in the toolbar, before the return statement: 
```
throw new Error('I crashed!');
```
Check that the log appears in Mixpanel!

# Documentation

No.

### DIFF
--- a/mitosheet/src/components/toolbar/Toolbar.tsx
+++ b/mitosheet/src/components/toolbar/Toolbar.tsx
@@ -51,6 +51,8 @@ const Toolbar = (
         void props.mitoAPI.checkoutStepByIndex(props.lastStepIndex);
     }
 
+    throw new Error('I crashed!');
+
     return (
         <div className='toolbar-container'>
             <div className='toolbar-left-half'>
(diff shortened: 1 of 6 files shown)


mitosheet/src/components/toolbar/Toolbar.tsx
```diff
@@ -51,6 +51,8 @@ const Toolbar = (
         void props.mitoAPI.checkoutStepByIndex(props.lastStepIndex);
     }
 
+    throw new Error('I crashed!');
```
aarondr77, line 54: Remove

aarondr77:
A potential improvement to make to this PR is to log more information about what the error was. Although that might not be super easy to do. However, for example, if we received this log we wouldn't know exactly how to go about fixing it. We'd just have to look at the code in the drop duplicates taskpane and figure out what could have caused it. 

Screen Shot 2022-02-06 at 5 08 22 PM

aarondr77:
LGTM.

naterush:
@aarondr77 good callout; I checked and we actually can get more information about the error! Thank god for typescript types making it so easy + safe :-)

PR #40, opened 2022-02-03
[INVESTIGATION] mitosheet: optimize out unneeded code after deletes

naterush:
# Description

This PR implements optimizing out unneeded code after a delete. See the core algorithm for the optimization in the compiler utilities. Generally, pretty happy with how it _works_, but not the actual implementation (it's very messy). As such, **let's not merge this.**

What I liked from this implementation:
1. The general approach of having a set of input and output indexes, that a step performer knows about. 
2. The modification of the step class to be aware of this data (so input and output indexes are wrapped by the step, as well as the newly created indexes makes a _lot_ of sense). 

What I didn't like from this implementation:
1. The lack of symmetry between the return types of input sheet indexes
2. The existence of a sheet id as being really useful for the algorithms we implemented. It kinda makes me want to... move everything from sheet_index to sheet id. I wonder how this refactor would go!
3. How our steps manager stores and reports steps. I think the fact that all consumers of the steps in the step manager have to filter them to valid ones is... terrible UX. I want to move to hiding the invalid steps unless you want them; the list of steps should be auto filtered!

# Testing

Do some imports, merges, and pivots, and then delete data frames. If the data frame you delete was not used in any later step, it's creation and edits should disappear as well!

# Documentation

No. We're also not merging.

### DIFF
--- a/mitosheet/mitosheet/step.py
+++ b/mitosheet/mitosheet/step.py
@@ -121,6 +121,43 @@ class Step:
         """
         return self.post_state if self.post_state is not None else \
             (self.prev_state if self.prev_state is not None else State([]))
+        
+    @property
+    def new_sheet_indexes_this_step(self) -> Set[int]:
+        """
+        Returns the set of sheet indexes for sheets that were created 
+        during this step. This is useful for depedency detection
+        """
+        if self.prev_state is None:
+            return {i for i in range(len(self.post_state.dfs))}
+        
+        if self.post_state is None:
+            return {}
+
+        return {
+            i for i in range(
+                len(self.prev_state.dfs),
+                len(self.post_state.dfs)
+            )
+        }
+    
+    @property
+    def input_sheet_indexes(self) -> Set[int]:
+        """
+        TODO
+        """
+        return self.step_performer.get_input_dataframe_indexes(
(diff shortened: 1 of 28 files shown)


aarondr77:
# Bug: Improperly skipping transpiled code required for duplicate sheet step

### Steps to Reproduce
1. Import a dataframe
2. Create a pivot table from the dataframe
3. Duplicate the pivot table
4. Delete the original pivot table
5. Notice that the transpiled code removes the creation of the pivot table, but still uses the pivot table in the code.

Screen Shot 2022-02-06 at 5 23 18 PM

aarondr77:
This exploration is great. This touches on the primary place that we get feedback about the lack of optimization of the generated code -- still displaying code for deleted dataframes. So I'm glad you started investigating it!

I think just from a semantic perspective, it would be nice to make the move to sheet_ids instead of just creating them in the transpiled step cause it would be helpful in other places as well (ie: we currently have some small bugs that occur from using sheet indexes) But obviously that’s a bigger move.

aarondr77:
Doing the `sheet_id` refactor will also be nice so that we could keep the tabs of graphs and dataframes in the same bar.

PR #41, opened 2022-02-03
Create LICENSEs

naterush:
# Description

This PR adds licenses. Our main license is AGPL, and our Enterprise/Pro license from PostHog; please read both thoroughly so you understand what's in them!

I did AGPL rather than GPL. See why here: https://www.gnu.org/licenses/why-affero-gpl.en.html I think it's what we want. I wonder if the AGPL works for our litmus test company, I think it probably does.

The other thing is that the `package.json` now has the license mentioned in it. The two options for GPL here are "AGPL-3.0-only", "AGPL-3.0-or-later". I am pretty sure we have the first, so picked the first. You can see documentation for them all here: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#license (check me). **NOTE: on PyPi, this should update after we deploy this!**

# Testing

No. But read the damn thing.

# Documentation

No.

### DIFF
--- a/mitosheet/mitosheet/pro/LICENSE.txt
+++ b/mitosheet/mitosheet/pro/LICENSE.txt
@@ -1,0 +1,35 @@
+The Mito Pro and Enterprise license (the “Enterprise License”)
+Copyright (c) 2022 Mito Inc.
+
+With regard to the Mito Software:
+
+This software and associated documentation files (the "Software") may only be
+used in production, if you (and any entity that you represent) have agreed to,
+and are in compliance with, the Mito Subscription Terms of Service, or other
+agreement governing the use of the Software, as agreed by you and Mito,
+and otherwise have a valid Mito Pro license or Mito Enterprise license for the
+correct number of user seats. Subject to the foregoing sentence, you are free to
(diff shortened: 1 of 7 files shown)


mitosheet/mitosheet/pro/LICENSE.txt
```diff
@@ -0,0 +1,35 @@
+The Mito Pro and Enterprise license (the “Enterprise License”)
+Copyright (c) 2022 Mito Inc.
```
aarondr77, line 2: There is no Mito Inc. Should this be Saga Inc.

mitosheet/src/pro/LICENSE.txt
```diff
@@ -0,0 +1,35 @@
+The Mito Pro and Enterprise license (the “Enterprise License”)
+Copyright (c) 2022 Mito Inc.
```
aarondr77, line 2: Also no Mito Inc.

aarondr77 (commented):
This looks good. Two small changes. 

Do we need to put any copyright stuff at the top of the files? I think currently some files might have them and others don't. But not sure if any files have the correct copyright.

PR #44, opened 2022-02-04
Full screen graph

naterush:
# Description

Does some refactoring of task panes to make them better components (they look oh so much better now, and I think will be much more maintainable and reusable going forward). Also fixes a few sheet crashing bugs I found in task panes while testing this.

Also, makes graphing full screen, which I think gets all the way to fixing the display issues that we see in graphing; it doesn't fix lifecycle stuff, though, and we can address that in the specification if we still desire to do so.

# Testing

Test all of the taskpanes, both with data and without data!

# Documentation

No. We can leave graphs as is for now, I think, and just update it after the documentation is updated.


mitosheet/src/components/taskpanes/DefaultTaskpane/DefaultTaskpane.tsx
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Mito
+
+import React, { ReactNode } from 'react';
+import "../../../../css/taskpanes/DefaultTaskpane.css"
+
+export enum TaskpaneSize {
+    ENTIRE = 'entire',
+    WIDE = 'wide',
+    NARROW = 'narrow'
+}
```
naterush, line 10: Remove this. Add doc strings to all of this

mitosheet/src/components/taskpanes/DefaultTaskpane/DefaultTaskpaneBody.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import React, { ReactNode } from 'react';
+import "../../../../css/taskpanes/DefaultTaskpane.css"
+import { classNames } from '../../../utils/classNames';
+
+export enum TaskpaneSize {
+    ENTIRE = 'entire',
+    WIDE = 'wide',
+    NARROW = 'narrow'
+}
```
naterush, line 11: And here

naterush (commented):
Cleanup!

naterush:
The one usability concern on this is users won't realize it's a taskpane, and so won't know how to close it; they'll get trapped in the graphing interface.

aarondr77:
Yeah, I've thought a bit more about the full screen taskpane since we last talked. I think the primary reason that I'm not a fan of this solution is because this doesn't respect our graphing design principle: **Respect the relationship between graphing and transformations** Specifically, I think this makes it extra difficult to understand the data while creating the graph because you can't even switch between the two, you have to close the graph to get to the data. I think that this is a downgrade from what we currently have. 

In general, I think when applying the 80/20 rule to design we should continue to remind ourselves that we want to always be designing really high quality product. I'm more persuaded by the 80/20 rule when its about scope of features rather than quality. So using the 80/20 rule to say that we don't want to add the ability to create dual axis graphs because it introduces a lot of added complexity for not much benefit I'm onboard with! We're always trying to take vertical slices from the product pyramid. This full screen slide is a horizontal one.

aarondr77:
> The one usability concern on this is users won't realize it's a taskpane, and so won't know how to close it; they'll get trapped in the graphing interface.


Yeah this is a good point. I agree with you. 

I've thought a bit more about the full screen taskpane since we last talked. I think the primary reason that I'm not a fan of this solution is because this doesn't respect our graphing design principle: **Respect the relationship between graphing and transformations** Specifically, I think this makes it extra difficult to understand the data while creating the graph because you can't even switch between the two, you have to close the graph to get to the data. I think that this is a downgrade from what we currently have. 

In general, I'm more persuaded by the 80/20 rule when its about scope of features rather than quality. So using the 80/20 rule to say that we don't want to add the ability to create dual axis graphs because it introduces a lot of added complexity for not much benefit I'm onboard with! We're always trying to take vertical slices from the product pyramid. This full screen slide is a horizontal one. 

Let's discuss this in the context of the graphing spec where we have the full context.

aarondr77:
I got rid of the full screen graph stuff. If that looks good to you, then LGTM.

PR #45, opened 2022-02-04
mitosheet: add more types for better type checking

naterush:
# Description

See description.

# Testing

# Documentation

### DIFF
--- a/mitosheet/mitosheet/saved_analyses/save_utils.py
+++ b/mitosheet/mitosheet/saved_analyses/save_utils.py
@@ -12,7 +12,7 @@ in analyses.
 from mitosheet.step import Step
 import os
 import json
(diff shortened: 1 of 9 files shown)


naterush:
@cla-bot check

aarondr77:
Not sure if this is ready for review, but this looks good to me. 

Can you explain what the TYPE_CHECKING conditional where we define the type of the step manager is. Why do we have that check for the step manager, but not others?

naterush:
@aarondr77 StepsManager imports files that in turn import the files where we're trying to import StepsManager; as such, we can't just import StepsManager without causing circular import errors. 

Since we only want StepsManager to type check, we can just use that solution, which is the MyPy solution for type checking!

PR #55, opened 2022-02-09
Dogfood

naterush:
# Description

- graphing: add a correct border back to the taskpane for now
- formatting: fixed a bug where default formatting would only display 1 decimal (on backend and front-end)
- docs: adds the release notes to these docs (shout out monorepo!)
- signup: makes it so signup only is requested if we're collecting logs

# Testing

Make sure to test all of the taskpanes, as they have been refactored. Graphing should work as it currently does as well. Check that the dtypes load everywhere (looks good for me!). 

Also, make sure the number formatting changes I made work (especially on Export, which requires you going pro). Can you test on Excel to make sure export is working and check me?

Also make sure the new signup display logic is working (not on pro, or not if telemetry is turned off). 

# Documentation

No. Release notes are written.

### DIFF
--- a/docs/misc/release-notes.md
+++ b/docs/misc/release-notes.md
@@ -4,6 +4,15 @@ description: Want to see what is new in the Mitosheet? Check it out below.
 
 # Release Notes
 
+## 2022-2-9 <a href="#2022-2-9" id="2022-2-9"></a>
+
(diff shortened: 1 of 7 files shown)


naterush (commented):
Looks good!

Issue #58, opened 2022-02-10
Remove all unnecessary metadata from steps


naterush:
Currently, we track all sorts of metadata in steps that is duplicated or otherwise can be gotten in other way. We should do our best to remove this data, as it:
1. Requires extra work when adding any new step.
2. Has led to multiple bugs as it gets out of date. 

Ideally, we'd be able to come up with an algorithm that maintains this metadata automatically, but I figure this is not possible unless we subscribe to changes in pandas.

naterush:
Some thoughts on further meta-data reduction:

1. dfs: we need to keep tracking these. 
1. df_names: these also need to be tracked.
2. df_sources: this is not much effort to track
3. column_ids: this we need to track, I think.
4. column_spreadsheet_code: seemingly the only way to get rid of this is to get rid of live-updating column formulas. As this seems like it would have a huge benefit otherwise, I think we should investigate this more.
5. column_filters: I am not sure we need to track these. Simply put, we can "go get the filters" like we do with the pivot table values. Notably like the pivot table editing, this works _much better_ with sheet ids, as it makes it easy to figure out when a specific data frame was created. 
6. column_format_types: these _could_ be gone and got, but I wonder if we just want to be able to track stuff like this. Hmm.

PR #57, opened 2022-02-10
Remove MitoType

naterush:
# Description

This PR works on issues explored here: https://github.com/mito-ds/monorepo/issues/54 and https://github.com/mito-ds/monorepo/issues/58.

Specifically, the Column Mito Type was an example of a piece of meta data that:
1. Literally served no purpose. The current PR implements the exact same functionality while saving 200 lines of code.
2. Was confusing. What is the difference between a Mito type and a dtype? God only knows. We 

As such, I removed it from the codebase. No functionality changes are expected or should be present, with the exception that we don't handle setting cell values exactly the same in the presence of ints and floats. I think this is totally not a big deal!

For more context on what I want to do to fully separate floats and ints, see here: https://github.com/mito-ds/monorepo/issues/61

# Testing

NOTE: the other open PR is a superset of this PR. Review this PR separately, and test it, but you can do the bulk of the manual testing on that PR!

Do thorough testing of the app. I'd recommend testing the filter code heavily, and honestly all the other steps could use some testing.. This PR is not going to fix any issues we find with filtering, thought there are some.

# Documentation

No changes!

PR #62, opened 2022-02-13
State refactor

naterush:
# Description

Building on top of the PR #57 and the issue #58 , this PR attempts to remove as much state as it possibly can from the Mitosheet backend. 

To do so, it removes:
1. The column meta type (we didn't really use this anywhere, anyways).
2. The column Python code. We can easily recompute this, and keeping it around just made renames harder.
3. The column eval graph. We can also recompute this easily, and in doing so we also make renames easier.

The new theory of what metadata to keep in the current state is just _if you need it in the current state_ and _cannot easily recompute it from other metadata we already have_. So formatting, and column formulas, and data frame sources are all needed! Applying this leads to >200 lines of code saved, dramatically simpler logic in some cases, and 0 change in functionality!

I also ripped out the code that allows renames to work with multiple column headers. As it existed, it was just technical debt (since we aren't using it), and I wrote it without knowing much about multi-header renames (aka, it was probably trash code).

# Testing

Big testing, as with the other PR. This touches pretty much every step, and so making sure that it doesn't break anything is key.

Since this subsumes the other PR, you can do most of your testing here!

# Documentation

No. Internal only.


mitosheet/mitosheet/evaluation_graph_utils.py
```diff
@@ -84,6 +117,10 @@ def subgraph_from_starting_column_id(column_evaluation_graph: Dict[ColumnID, Set
 
     return column_evaluation_subgraph
 
+def topological_sort_dependent_columns(state: State, sheet_index: int, column_id: ColumnID) -> List[ColumnID]:
```
naterush, line 120: Check spelling of defendant, doc string

mitosheet/mitosheet/step_performers/column_steps/set_column_formula.py
```diff
@@ -113,10 +106,13 @@ def execute( # type: ignore
         _, _, old_dependencies_column_headers = parse_formula(old_formula, column_header, column_headers)
         old_dependencies = set(prev_state.column_ids.get_column_ids(sheet_index, old_dependencies_column_headers))
 
+        column_evaluation_graph = create_column_evaluation_graph(prev_state, sheet_index)
+        print(column_evaluation_graph)
```
naterush, line 110: Remove

mitosheet/mitosheet/step_performers/column_steps/set_column_formula.py
```diff
@@ -183,50 +179,6 @@ def get_modified_dataframe_indexes( # type: ignore
         return {sheet_index}
 
 
-def _update_column_formula_in_step(
```
naterush, line 186: This entire function disappears. How nice!

mitosheet/mitosheet/tests/test_evaluation_graph.py
```diff
@@ -7,8 +7,22 @@
...
+    df = pd.DataFrame({'A': [123]})
+    mito = create_mito_wrapper_dfs(df)
+    mito.set_formula('=A', 0, 'B', add_column=True)
+    mito.set_formula('=B', 0, 'C', add_column=True)
+
+    assert create_column_evaluation_graph(
+        mito.curr_step.post_state,
+        0
+    ) == {'A': {'B'}, 'B': {'C'}, 'C': set()}
+
```
naterush, line 25: Maybe one more complex test.

naterush (commented):
Minor cleanup, mostly looks great!

aarondr77:
## Bug with error traceback 
The traceback for the column dependent error is not working. 

### To reproduce:
1. Import data
2. Add a new column and set its formula equal to another column, ex: `=A`
3. Delete column A
4. Click to display the error traceback and notice that it doesn't provide any information

Screen Shot 2022-02-13 at 2 38 51 PM

aarondr77:
Okay, I think I've tested all of the steps and only found these two bugs (which I think existed prior to this PR). Good work doing such a big refactor without introducing any (easily findable) bugs. 

Seems like this is the type of PR that deserves a more thorough testing process though. Do you want to merge into dev and then hire some upwork testers? + we can create a quick full workflow to test on. Thoughts?

naterush:
@aarondr77 RE: these bugs - I think they belong in issues and not in this PR. I opened issues here #66 #65  - so nbd. Good catches on both of them. In the future, if the bugs are not relevant to the specifics of the PR, then open an issue for them and just link to them here. 

Re: testing. Yes! I was gonna suggest the same thing. That being said, we have a bunch of other stuff this week, so let's just put this on the back burner for now. 

It's ready to merge whenever we want to do a big testing session. Let's aim for next week.

PR #63, opened 2022-02-14
mitoinstaller: automatically move from mitosheet3 -> mitosheet

naterush:
# Description

See doc string.

# Testing

# Documentation


naterush:
Merging so I can test in the wild!

PR #99, opened 2022-02-22
Add mitosheet private + deployment workflow

naterush:
# Description

This adds a package `mitosheet-private`, that disables telemetry. The workflow that deploys this is manual, as we're planning on taking down this package after we distribute to relevant clients.

# Testing

We'll have to merge this and make sure it works, but please review to make sure it all makes sense!

# Documentation

No.


mitosheet/setup.py
```diff
@@ -148,7 +148,7 @@
             """,
         long_description_content_type='text/markdown'
     )
-elif name == 'mitosheet' or name == 'mitosheet3':
+elif name == 'mitosheet' or name == 'mitosheet3' or name == 'mitosheet-priv':
```
naterush, line 151: Wrong name

naterush (commented):
Wrong name!

mitosheet/switch.py
```diff
@@ -392,9 +283,11 @@ def switch(new_package):
     if new_package == 'mitosheet2':
       open('package.json', 'w').write(MITOSHEET_TWO_PACKAGE_JSON)
     elif new_package == 'mitosheet':    
-      open('package.json', 'w').write(MITOSHEET_PACKAGE_JSON)
+      open('package.json', 'w').write(MITOSHEET_PACKAGE_JSON.replace('REPLACE_WITH_PACKAGE_NAME_WITH_REPLACE', 'mitosheet'))
```
aarondr77, line 286: This text doesn't make sense to me. 

Should it say: "REPLACE_WITH_PACKAGE_NAME" instead of "REPLACE_WITH_PACKAGE_NAME_WITH_REPLACE"?

PR #104, opened 2022-02-23
mitosheet: improve analysis id to be private when telemetry is off

naterush:
# Description

Per clients request, since when telemetry is off we don't have anything, we use just a comment in the analysis code block.

# Testing

```
!python -m mitosheet turnofflogging
import mitosheet
mitosheet.sheet()
```

Then so some things, and make sure it replays. Then, turn on logging [!python -m mitosheet turnonlogging], and make sure it switches back + reads in correctly. Maybe switch back to logging off one more time, and make sure that direction works as well.

# Documentation

N/A

### DIFF
--- a/mitosheet/src/plugin.tsx
+++ b/mitosheet/src/plugin.tsx
@@ -62,7 +62,8 @@ function getCellAtIndex(cells: IObservableUndoableList<ICellModel> | undefined, 
 
 function codeContainer(
     analysisName: string,
-    code: string[]
+    code: string[],
+    telemetryEnabled: boolean
 ): string {
 
     if (code.length == 0) {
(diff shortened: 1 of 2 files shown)


mitosheet/src/plugin.tsx
```diff
@@ -129,6 +141,17 @@ function getAnalysisName(codeblock: string): string | undefined {
             codeblock.indexOf('register_analysis(\'') + 'register_analysis(\''.length,
             codeblock.indexOf('\n') - 2
         );
+    } else if (codeblock.includes('# Analysis:')) {
+        console.log("case 4", codeblock.substring(
```
naterush, line 145: Remove

naterush (commented):
Minor cleanup

PR #107, opened 2022-02-24
Graph step performer

aarondr77:
# Description

This PR took longer than expected because I misunderstood how to add new state. See https://github.com/mito-ds/monorepo/issues/110

Using `redo` updates the displayed graph, but it doesn't update the graph configuration UI. We don't currently have infrastructure for recognizing that a redo event just occured. This is an existing problem with the Pivot Table. See issue https://github.com/mito-ds/monorepo/issues/120. Btw: something I've been thinking about is around this type of issue and when to fix it in the PR vs not. In general, something we've been talking about a lot recently is how creating small PR's that are bounded on what parts of the codebase they touch have been helpful for me in wrapping up PR's quickly. So I'm operating under the framework that if I run into an issue like this that requires a refactor outside of the new code that I'm writing and it is not critical to the PR, then I should open an issue and not fix it right away. ie: If we already have the infrastructure to easily solve it then I will, but if we don't then elevate it to an issue. Let me know if this is okay. 

Take a look at the comment on the `updateGraph` variable around line of 136 of `Graphsidebar.tsx` for an explanation on how we decide when to call the `sendGraphMessage`. There are alternative approaches we could attempt, see [here](https://stackoverflow.com/questions/54954091/how-to-use-callback-with-usestate-hook-in-react/61842546#61842546). However, this lead to additional buggy behavior that made me take this different route. The issue with the current route is that it does have the potential of race condition bugs. 

When a previous step is checked out, the graphing UI doesn't provide any helpful information to understand why editing the graph doesn't lead to any graph being rendered, but I think that's okay for now. We can address this when we change the UI. 

# Testing

Test creating graphs through the graph taskpane and through the column summary stats. 
Make sure that the steps are properly created and the transpile code works. 

# Documentation

No.


mitosheet/mitosheet/api/api.py
```diff
@@ -29,7 +29,7 @@
 
 # NOTE: BE CAREFUL WITH THIS. When in development mode, you can set it to False
 # so the API calls are handled in the main thread, to make printing easy
-THREADED = True
+THREADED = False
```
aarondr77, line 32: set back to true

mitosheet/mitosheet/api/graph/graph_utils.py
```diff
@@ -1,128 +1,17 @@
...
 import pandas as pd
-import plotly.graph_objects as go
-from mitosheet.api.graph.column_summary_graph import (
-    MAX_UNIQUE_NON_NUMBER_VALUES,
-    filter_df_to_top_unique_values_in_series,
-)
+from mitosheet.api.graph.get_column_summary_graph import MAX_UNIQUE_NON_NUMBER_VALUES, filter_df_to_top_unique_values_in_series
 from mitosheet.sheet_functions.types.utils import is_number_dtype
 from mitosheet.step_performers.bulk_old_rename.deprecated_utils import deprecated
+from mitosheet.step_performers.graph.graph_utils import BAR, BOX, SCATTER
```
aarondr77, line 12: we might not need all of these imports

mitosheet/mitosheet/step_performers/graph/graph.py
```diff
@@ -0,0 +1,201 @@
...
+
+class GraphStepPerformer(StepPerformer):
+    """
+    Creates a graph of the passed parameters and update the graph_data_json
+
+    {
+        preprocessing: {
+            safety_filter_turned_on_by_user: boolean
+        },
+        graph_creation: {
```
aarondr77, line 28: add graph_styling to these events now so that we don't need to do a step upgrade

mitosheet/mitosheet/step_performers/graph/graph.py
```diff
@@ -0,0 +1,201 @@
...
+            else []
+        )
+        y_axis_column_headers = prev_state.column_ids.get_column_headers_by_ids(
+            sheet_index, y_axis_column_ids
+        )
+
+        # Find the height and the width, defaulting to fill whatever container its in
+        graph_rendering_keys = graph_rendering.keys()
+
+        height = (
```
aarondr77, line 112: fix this styling now that I turned off Black formatting

mitosheet/mitosheet/step_performers/graph/graph.py
```diff
@@ -0,0 +1,201 @@
...
+    ) -> str:
+        sheet_index = graph_creation["sheet_index"]
+        graph_type = graph_creation["graph_type"]
+        if df_names is not None:
+            df_name = df_names[sheet_index]
+            return f"Graphed {df_name} as {GRAPH_TITLE_LABELS[graph_type]}"
+        return f"Created {GRAPH_TITLE_LABELS[graph_type]}"
+
+    @classmethod
+    def get_modified_dataframe_indexes(cls, graph_creation, **params) -> Set[int]:  # type: ignore
```
aarondr77, line 199: @naterush what do we want to do here? 

If it returns an empty set, then this step modified every dataframe. If it returned -1, then it modified all new dataframes (on the left side of the dfs array).

Neither of which we want to do

mitosheet/src/api.tsx
```diff
@@ -332,39 +332,70 @@ export default class MitoAPI {
...
+        stepID?: string,
+    ): Promise<string> {
+
+        // If this is overwriting a graph event, then we do not need to
+        // create a new id, as we already have it!
+        if (stepID === undefined || stepID === '') {
+            stepID = getRandomId();
+        }
+
+        console.log(stepID)
```
aarondr77, line 353: remove

mitosheet/mitosheet/api/graph/graph_utils.py
aarondr77: We do, this is for the deprecated function!

mitosheet/mitosheet/api/graph/__init__.py
```diff
@@ -1 +1,6 @@
-from mitosheet.api.graph.get_graph import get_graph
\ No newline at end of file
+#!/usr/bin/env python
+# coding: utf-8
+
+# Copyright (c) Saga Inc.
+# Distributed under the terms of the GPL License.
+from mitosheet.api.graph.get_column_summary_graph import get_column_summary_graph
```
naterush, line 6: Perhaps this can be moved out of `/graph` folder, and condensed into one file?

mitosheet/mitosheet/api/graph/get_column_summary_graph.py
```diff
@@ -1,17 +1,58 @@
+#!/usr/bin/env python
```
naterush, line 1: Cha, since it's in one file, let's move it up out of this folder.

mitosheet/mitosheet/step.py
```diff
@@ -93,6 +99,14 @@ def column_filters(self):
     def column_format_types(self):
         return self.post_state.column_format_types
 
+    @property
+    def graph_data_json(self) -> str:
```
naterush, line 103: This doesn't return a string, does it? This is a more complex type

mitosheet/mitosheet/step_performers/graph/graph.py
```diff
@@ -0,0 +1,203 @@
...
+        )
+        y_axis_column_headers = prev_state.column_ids.get_column_headers_by_ids(
+            sheet_index, y_axis_column_ids
+        )
+
+        # Find the height and the width, defaulting to fill whatever container its in
+        graph_rendering_keys = graph_rendering.keys()
+
+        height = graph_rendering["height"] if "height" in graph_rendering_keys else "100%"
+        width = graph_rendering["width"] if "width" in graph_rendering_keys else "100%"
```
naterush, line 115: Is it easier to save the keys vs just checking graph_rendering?

mitosheet/mitosheet/step_performers/graph/graph.py
```diff
@@ -0,0 +1,203 @@
...
+                x_axis_column_headers,
+                y_axis_column_headers,
+                df_name,
+            )
+
+        post_state.graph_data_json[str(sheet_index)] = {
+            "graphParams": {
+                "graphPreprocessing": graph_preprocessing,
+                "graphCreation": graph_creation,
+                "graphStyling": {},
```
naterush, line 159: Maybe just make this graph_styling? I think this will trip someone up in the future - even if it's always empty now, we should define it.

mitosheet/mitosheet/step_performers/graph/graph.py
```diff
@@ -0,0 +1,203 @@
...
+    @classmethod
+    def transpile(  # type: ignore
+        cls,
+        prev_state: State,
+        post_state: State,
+        execution_data: Optional[Dict[str, Any]],
+        graph_preprocessing: Any,
+        graph_creation: Any,
+        graph_styling: Any,
+        graph_rendering: Any,
```
naterush, line 178: In all functions here, can we type these as Dict[str, Any]?

naterush: Also our standard convention is to put execution_data last, right? I might be wrong about that in which case ignore.

mitosheet/mitosheet/step_performers/graph/graph.py
naterush: I think -1 is fine, since it doesn't create any new data frames we aren't actually modifying anything. 

That being said, we'll have to keep an eye on this this effects other optimization efforts...

mitosheet/src/api.tsx
```diff
@@ -332,39 +332,70 @@ export default class MitoAPI {
         return excelFileString;
     }
 
+
+    async sendGraphMessage(
+        graphType: GraphType,
+        sheet_index: number,
+        safety_filter_turned_on_by_user: boolean,
+        xAxisColumnIDs: ColumnID[] | undefined,
+        yAxisColumnIDs: ColumnID[] | undefined,
```
naterush, line 341: These should never be undefined now, right?

mitosheet/src/api.tsx
```diff
@@ -332,39 +332,70 @@ export default class MitoAPI {
...
-                'sheet_index': sheet_index,
-                'x_axis_column_ids': xAxisColumnIDs,
-                'y_axis_column_ids': yAxisColumnIDs,
-            },
-            'graph_rendering': {
-                'height': height,
-                'width': width
-            }
-
+            'type': 'get_column_summary_graph',
```
naterush, line 394: Nice

mitosheet/src/components/taskpanes/Graph/AxisSection.tsx
```diff
@@ -27,7 +27,7 @@ export enum GraphAxisType {
 */
 const AxisSection = (props: {
     columnIDsMap: ColumnIDsMap;
-    columnDtypesMap: Record<string, string>;
+    columnDtypesMap: Record<string, string> | undefined;
```
naterush, line 30: Why so?

mitosheet/src/types.tsx
```diff
@@ -189,6 +194,45 @@ export type SheetData = {
...
 };
 
+/**
+ * Data about all of the graphs. For each graph, it contains all of the parameters used to construct the graph,
+ * the actual graph html & javascript, and the generated code.
+ * 
+ * @param graphParams - all of the parameters used to construct the graph
+ * @param graphGeneratedCode - the python code to construct the graph
+ * @param graphHTML - the number of rows in the data. Should be equal to data[0].length
+ * @param graphScript - the number of columns in the data. Should be equal to data.length
```
naterush, line 204: These comments are out of date

mitosheet/src/types.tsx
```diff
@@ -189,6 +194,45 @@ export type SheetData = {
...
+ */
+export type GraphData = {
+    graphParams: GraphParams,
+    graphGeneratedCode: string,
+    graphHTML: string | undefined,
```
naterush, line 209: Why would these be undefined?

mitosheet/src/types.tsx
```diff
@@ -189,6 +194,45 @@ export type SheetData = {
...
+    graph_type: GraphType,
+    sheet_index: number,
+    x_axis_column_ids: ColumnID[]
+    y_axis_column_ids: ColumnID[]
+}
+export type GraphStylingParams = undefined
+export type GraphRenderingParams = {
+    width?: number
+    height?: number
+}
```
naterush, line 232: Sweet types! Wanna move these above the GraphParams specification just for clarity?

mitosheet/src/widget.tsx
```diff
@@ -201,9 +199,25 @@ export class ExampleView extends DOMWidgetView {
...
+        // Convert empty string to undefined to adhere to best practices
+        Object.entries(graphDataJSON).forEach(([key, val]) => {
+            if (val["graphHTML"] == '') {
+                graphDataJSON[key]["graphHTML"] = undefined
+            }
+            if (val["graphScript"] == '') {
+                graphDataJSON[key]["graphScript"] = undefined
+            }
+        })
+        parsed["graphDataJSON"] = graphDataJSON
```
naterush, line 219: Ah! I see what you're doing here. Gotcha. So ignore my above comment...

This is an interesting example of this application of "encode failure in the type system" - somewhat different than before. A few thoughts:
1. If we're gonna make them undefined, let's do it on the backend, so what we're passing doesn't need to be patched up. It kinda makes it so you need to think about the graph data json being "created" in multiple places, which is bad news. 
2. Also, not doing it on the backend makes it so we have this grow code here vs. it just being a single json.parse, as it was before - which was sweet!
3. Also, since graphHTML and graphScript are both either defined or empty strings together, does it make sense to nest them in a single entity? What about the following:

```
export type GraphData = {
     graphParams: GraphParams,
     graphOutput?: {
          graphGeneratedCode: string,
          graphHTML: string,
          graphScript: string,
     }
 };
```

Note here that: 
1.  The backend would do this (and it would remove the use of the empty string on the backend).
2. The _entire_ output is either there or not. This makes the types more consistent and less likely to be fucky. 

All-in-all, good work thinking about this empty string thing when I did not (per my comment above). It caused me to think about it a bit more as well!

mitosheet/src/types.tsx
naterush: See my comment below. Nice work putting this failure in the type system!

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -522,7 +541,7 @@ const GraphSidebar = (props: {
                         <TextButton
                             variant='dark'
                             onClick={copyGraphCode}
-                            disabled={loading || graphObj === undefined}
+                            disabled={loading || graphHTML === undefined || graphScript === undefined}
```
naterush, line 544: Yurp, this would be simplified by putting it in one object.

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -412,10 +428,13 @@ const GraphSidebar = (props: {
...
                             <Col>
                                 <Select
-                                    value={props.dfNames[selectedSheetIndex]}
+                                    value={props.dfNames[graphParams.graphCreation.sheet_index]}
                                     onChange={(newDfName: string) => {
                                         const newIndex = props.dfNames.indexOf(newDfName);
-                                        setSelectedSheetIndex(newIndex);
+                                        // Get the new sheet's graph params 
+                                        const newSheetGraphParams = getGraphParams(props.graphDataJSON, newIndex, props.sheetDataArray)
+                                        setStepID(undefined)
```
naterush, line 436: Maybe a comment to explain what's happening here?

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -354,10 +373,7 @@ const GraphSidebar = (props: {
 
         // Log that the user copied the graph code
         void props.mitoAPI.sendLogMessage('copy_graph_code', {
-            'generated_graph': graphParams.xAxisColumnIDs.length !== 0 || graphParams.yAxisColumnIDs.length !== 0, // A helpful variable for the value add event
-            'graph_type': graphParams.graphType,
-            'x_axis_column_ids': graphParams.xAxisColumnIDs,
-            'y_axis_column_ids': graphParams.yAxisColumnIDs
+            'graph_type': graphParams.graphCreation.graph_type
```
naterush, line 376: Nice, good call!

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -305,9 +311,12 @@ const GraphSidebar = (props: {
         // Update the graph type
         setGraphParams({
             ...graphParams,
-            graphType: graphType,
-            xAxisColumnIDs: xAxisColumnIDsCopy,
-            yAxisColumnIDs: yAxisColumnIDsCopy
+            graphCreation: {
+                graph_type: graphType,
+                sheet_index: graphParams.graphCreation.sheet_index,
```
naterush, line 316: Do you need to set the sheet index here?

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -254,26 +243,42 @@ const GraphSidebar = (props: {
...
+    /*
+        Updates the graph params on undo so that the graph configuration is in sync
+        with the graph shown
+    */
+    const refreshParamsAfterUndo = async (): Promise<void> => {        
+
+        // We also set ignoreNextUpdate to true, so that these updates
+        // don't cause a new message to get sent, as then we get trapped
+        // in an infinite loop where an undo will cause a new graph message
+        // to get sent
```
naterush, line 255: Why would this cause an infinite loop? Shouldn't this only be called when there was an undo?

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -119,86 +120,73 @@ const GraphSidebar = (props: {
...
+    const [graphParams, setGraphParams] = useState<GraphParams>(() => getGraphParams(props.graphDataJSON, props.graphSidebarSheet, props.sheetDataArray))
+    const dataSourceSheetIndex = graphParams.graphCreation.sheet_index
+    const graphScript = props.graphDataJSON[dataSourceSheetIndex]?.graphScript
+    const graphHTML = props.graphDataJSON[dataSourceSheetIndex]?.graphHTML
+    const [_copyGraphCode, graphCodeCopied] = useCopyToClipboard(props.graphDataJSON[dataSourceSheetIndex]?.graphGeneratedCode || '');
+    const [stepID, setStepID] = useState<string|undefined>(undefined);
 
-    const [graphObj, setGraphObj] = useState<GraphObject | undefined>(undefined);
     const [loading, setLoading] = useState<boolean>(false)
+    const [changeLoadingGraph] = useDelayedAction(LOAD_GRAPH_TIMEOUT)
```
naterush, line 140: I think we should move this logic to using the `useDebouncedEffect` - which we added after this logic. Thoughts?

naterush (commented):
Ok, this mostly looks really great! Not even added code - and still a functionality bump.


The only major review comment here: I think there is some confusion (at least for me) about when / where the params get refreshed currently, and I think we should do some work to fix that up. I'm gonna clarify my understanding of our _expected_ behavior here:

When the graphing interface is opened, we want:
1. The params to default to whatever the last params for that sheet were (which is fine, we can go get them from the graph data json object). 
2. _NO_ message to get sent to the backend. 

When the graphing params are changed:
1. The new params are set immediately.
2. A message is sent to the backend.

When undo is pressed:
1. The refreshParamsOnUndo is called, which gets the old params, and displays them correctly. No message should be sent here, otherwise we'll enter into an infinite loop. 

When redo is pressed:
1. We want to update the params, but also not send a message to the backend. 

My take aways here: of the 3 cases where we update the graphing params, we don't actually want to send a message to the backend. We really only want to send a message to the backend when the user explicitly changes the params themselves. 

In turn, I think the best bet here is to:
1. Remove the effect that sends a message to the backend. Instead, just manually call this when we change: the graph type, sheet, or axis. 
2. Avoid all the confusing logic with ignoreNextUpdate

I think this leads to potentially more code, but it will be _much simpler_ code - just functions getting called when they should, and not functions getting avoided when they shouldn't be.

There might be some complexity with debouncing, but perhaps ignore that for now and see if this can simplify logic. I might be totally off! I am also not sure debouncing is really super necessary, now that we filter so aggressively by default!

Let me know what you think!

mitosheet/mitosheet/step_performers/graph/graph.py
```diff
@@ -0,0 +1,206 @@
...
+        **params,
+    ) -> Tuple[State, Optional[Dict[str, Any]]]:
+        """
+        Returns the new post state with the updated graph_data
+        """
+
+        # We make a new state to modify it
+        post_state = deepcopy(prev_state)
+
+        # Get graph type
```
aarondr77, line 84: Change to get variables from parameters

mitosheet/mitosheet/step_performers/graph/graph.py
```diff
@@ -0,0 +1,206 @@
...
+        prev_state: State,
+        post_state: State,
+        execution_data: Optional[Dict[str, Any]],
+        graph_preprocessing: Dict[str, Any],
+        graph_creation: Dict[str, Any],
+        graph_styling: Dict[str, Any],
+        graph_rendering: Dict[str, Any],
+        **params,
+    ) -> List[str]:
+        return []
```
aarondr77, line 185: Add a comment here

mitosheet/mitosheet/step_performers/graph/graph_utils.py
```diff
@@ -1,22 +1,21 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# Copyright (c) Saga Inc.
+# Distributed under the terms of the GPL License.
+
+# Graph types should be kept consistent with the GraphType in GraphSidebar.tsx
```
aarondr77, line 7: This comment is out of place or out of date

mitosheet/src/components/taskpanes/Graph/AxisSection.tsx
```diff
@@ -81,7 +81,7 @@ const AxisSection = (props: {
     let selectableColumnIDs: ColumnID[] = []
     if (props.graphType === GraphType.BOX || props.graphType === GraphType.HISTOGRAM) {
         selectableColumnIDs = Object.keys(props.columnIDsMap).filter(columnID => {
-            return isNumberDtype(props.columnDtypesMap[columnID])
+            return props.columnDtypesMap !== undefined && isNumberDtype(props.columnDtypesMap[columnID])
```
aarondr77, line 84: TODO: Figure out if we need this

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -119,104 +121,72 @@ const GraphSidebar = (props: {
...
-        _setSelectedSheetIndex(newSelectedSheetIndex);
-        _setGraphParams(newParams)
-        props.setUIState(prevUIState => {
-            return {
-                ...prevUIState,
-                selectedSheetIndex: newSelectedSheetIndex
-            }
-        });
-    }
+    // When updateGraph is changed from false to true, we send a new getGraphMessage with the current graphParams
```
aarondr77, line 135: Explain that we do this to simulate a callback, but gives us power over what sources of graphParams changing causes the change. Also, note that there is a potential race condition here

aarondr77: Note this in the PR description as well since its important

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -305,10 +289,14 @@ const GraphSidebar = (props: {
         // Update the graph type
         setGraphParams({
```
aarondr77, line 290: Make a prevGraphParams and use that instead of this

mitosheet/src/types.tsx
```diff
@@ -190,6 +191,51 @@ export type SheetData = {
...
+};
+
+/**
+ * Data about all of the graphs. For each graph, it contains all of the parameters used to construct the graph,
+ * the actual graph html & javascript, and the generated code.
+ * 
+ * @param graphParams - all of the parameters used to construct the graph
+ * @param graphGeneratedCode - the python code to construct the graph
+ * @param graphHTML - the html to display the graph
+ * @param graphScript - the script to display the graph
```
naterush, line 223: Comments are out of date

mitosheet/src/types.tsx
```diff
@@ -190,6 +191,51 @@ export type SheetData = {
 };
 
 
+export type GraphPreprocessingParams = {
```
naterush, line 194: These types are great!

mitosheet/src/hooks/useStateCallback.tsx
```diff
@@ -0,0 +1,35 @@
...
+    const setStateCallback = useCallback((state: GraphParams, cb) => {
+        cbRef.current = cb; // store current, passed callback in ref
+        setState(state);
+    }, []); // keep object reference stable, exactly like `useState`
+  
+    useEffect(() => {
+        // cb.current is `null` on initial render, 
+        // so we only invoke callback on state *updates*
+        if (cbRef.current) {
+            console.log('state: ', state)
```
naterush, line 28: console.log

mitosheet/mitosheet/tests/test_utils.py
```diff
@@ -726,6 +726,32 @@ def checkout_step_by_idx(self, index: int) -> bool:
...
 
+    def generate_graph(
+        self, 
+        graph_preprocessing: Dict[str, Any],
+        graph_creation: Dict[str, Any],
+        graph_styling: Dict[str, Any],
+        graph_rendering: Dict[str, Any],
+        step_id: str=None
+    ) -> bool:
+        print(step_id)
```
naterush, line 737: print

mitosheet/mitosheet/tests/test_utils.py
```diff
@@ -726,6 +726,32 @@ def checkout_step_by_idx(self, index: int) -> bool:
         )
 
 
+    def generate_graph(
+        self, 
+        graph_preprocessing: Dict[str, Any],
```
naterush, line 731: Does this have the same interface as the function in the API file? It would be nice if it did - which I don't think it does currently. 

I want to be able to pass nice params to this - like graph type, column headers, etc - and have it do the processing necessary to generate the graph!

mitosheet/mitosheet/tests/steps_performers/test_graph.py
```diff
@@ -0,0 +1,33 @@
...
+        {
+            'height': 400,
+            'width': 400
+        }
+    )
+    
+    assert len(mito.steps) == 2
+    assert mito.curr_step.step_type == 'graph'
+
+
```
naterush, line 33: Now that we have some persistent data, can you test this? I'm looking to check that the column IDs get saved properly, that the javascript, etc are defined, etc. 

Just make sure it actually saves what you expect it to!

If you can create an accessor in the Mito Test Helper thing, so that refactoring internal state makes updating tests easy, that would be great!

mitosheet/mitosheet/step_performers/graph/graph.py
```diff
@@ -0,0 +1,210 @@
...
+            fig.update_layout(
+                margin=dict(
+                    l=0,
+                    r=0,
+                    t=30,
+                    b=30,
+                )
+            )
+
+            html_and_script = get_html_and_script_from_figure(fig, height, width)
```
naterush, line 146: Perhaps make this return a tuple of (html, script) for clarity now!

mitosheet/mitosheet/step_performers/graph/graph.py
```diff
@@ -0,0 +1,210 @@
...
+    def step_version(cls) -> int:
+        return 1
+
+    @classmethod
+    def step_type(cls) -> str:
+        return "graph"
+
+    @classmethod
+    def step_display_name(cls) -> str:
+        return "Graphed a sheet"
```
naterush, line 57: Maybe: Created a Graph?

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -122,101 +124,78 @@ const GraphSidebar = (props: {
     columnDtypesMap: Record<string, string>;
```
naterush, line 124: There's an unused thing called `GraphObject` at the top of this file we can remove, I think

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -122,101 +124,78 @@ const GraphSidebar = (props: {
...
-    model_id: string;
-    lastGraphParams: Record<number, GraphParams | undefined>;
-    setLastGraphParams: (sheetIndex: number, graphParams: GraphParams) => void;
+    graphDataJSON: GraphDataJSON
+    lastStepIndex: number
 }): JSX.Element => {
 
-    const [selectedSheetIndex, _setSelectedSheetIndex] = useState(props.graphSidebarSheet);
+    // We keep track of the graph data separately from the backend state so that 
+    // the UI updates imidietly, even though the backend takes a while to process.
```
naterush, line 132: Typo

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -337,15 +340,29 @@ const GraphSidebar = (props: {
...
+                const copyPrevGraphParams = {...prevGraphParams}
+                return {
+                    ...copyPrevGraphParams,
+                    graphCreation: {
+                        ...copyPrevGraphParams.graphCreation, 
+                        x_axis_column_ids: axisColumnIDsCopy
+                    }
+                }
             })
+            setUpdateGraph(true)
```
naterush, line 353: Can you move this to the end of the function, vs being in both if statements?

mitosheet/src/hooks/useStateCallback.tsx
```diff
@@ -0,0 +1,35 @@
+/*
```
naterush, line 1: Don't think we need this custom hook anymore.

naterush (changes requested):
Overall, the code looks great. Only major comment is about when we decide to send messages to the backend.

Overall, I'm not comfortable with level of complexity in this code. We are currently using the debounced effect AND delayed action hooks to delay things. I also am struggling to understand the issues documented in the code that led to the current solution - which is usually evidence that the current solution is a bit too complex.

Here's a proposal for a simple solution that I think works, but I might be missing something. If I am missing something, feel free to let me know. 

```
const [graphUpdatedNumber, setGraphUpdatedNumber] = useState(0)

useDebouncedEffect(() => {
     setLoading(true);
     await getGraphAsync();
}, [graphUpdatedNumber])

// Toggles the safety filter component of the graph params
    const toggleSafetyFilter = (): void => {
        const newSafetyFilter = !graphParams.graphPreprocessing.safety_filter_turned_on_by_user

  setGraphParams(prevGraphParams => {
      const copyPrevGraphParams = {...prevGraphParams}
      return {
          ...copyPrevGraphParams,
          graphPreprocessing: {
              safety_filter_turned_on_by_user: newSafetyFilter
          }
      }
  })
  setGraphUpdatedNumber((old) => old + 1);
}
```
Either way, if this solution doesn't work, I'm still gonna insist we try and get rid of the usage of both the useDebouncedEffect and the useDelayedAction stuff - can we try and get rid of the useDelayedAction stuff here, or if so why not?

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -122,101 +124,78 @@ const GraphSidebar = (props: {
...
-    const [graphHasNeverBeenConfigured, setGraphHasNeverBeenConfigured] = useState<boolean>(
-        graphParams === getDefaultGraphParams(props.sheetDataArray, selectedSheetIndex)
-    )
+    // Save the last step index, so that we can check if an undo occured
+    const prevLastStepIndex = usePrevious(props.lastStepIndex);
+
+    // When the last step index changes, check if an undo occured so we can refresh the params
+    useEffect(() => {
+        // If there has been an undo, then we refresh the params to this pivot
+        console.log(prevLastStepIndex, props.lastStepIndex)
```
naterush, line 162: console log

naterush (commented):
Minor cleanup!

naterush:
> Using redo updates the displayed graph, but it doesn't update the graph configuration UI.

I think there are no hard rules (or even really soft rules) on when to fix this in the PR vs. not. I'm fine to leave this out for now, as I don't really think there's a good way of doing it - a redo just adds a step like a normal operation!

> When a previous step is checked out, the graphing UI doesn't provide any helpful information ... We can address this when we change the UI.

Yep, sounds good for now!

naterush:
@aarondr77 I've done a bit of manual testing, and things look good! I want to recheck after you simplify the process about when we send update messages. 

Feel free to ping me or give me a call after you attempt the above solution I propose (or write back why it won't work, or have questions). Would be happy to chat [I might be missing something major, but it seems like it works...]

A few hours more and we'll be done here. Good work getting this big boy wrapped up - in some ways in the biggest change we'll be making with the graphing stuff.

mitosheet/mitosheet/tests/test_utils.py
aarondr77: I think I'm confused on how you want to handle these parameters / make it easy for us to pass things straight through. 

This is the same as the API in the api.tsx file
```
  def generate_graph(
        self, 
        graph_preprocessing: Dict[str, Any],
        graph_creation: Dict[str, Any],
        graph_styling: Dict[str, Any],
        graph_rendering: Dict[str, Any],
        step_id: str=None
    ) -> bool:
```

But this is not what you say here: graph type, column headers, etc

Can you clarify which version of this you are looking for. I think I originally had the version where I passed graph type, column headers, etc. but then I switched to this approach after our conversation about the previous PR. So just lmk which one you want and happy to update it to that. 

It looks like the only place where we might be able to pass the params directly through is the graph_styling part which we haven't implemented yet. None of the other places we've done that so far.

aarondr77:
The fact that I had both debounced effect AND delayed action hooks was just bad cleanup work. 

The issue that I was attempting to point out, although its also present in your solution, so maybe its not really an issue, is that there seems to be a race condition between setting the graph params and updating state that has the useEffect on it to update the graph. 

In your code, the race condition would be between `setGraphParams` and `setGraphUpdatedNumber`. In my code, the race condition would be between `setGraphParams` and `setUpdateGraph`.

Is the approach we're taking here okay?

naterush (commented):
Code looks great! This only cleanup left is to remove the delayed action effect, which I can do!

naterush:
This is all working great! The undo + multiple graphs is a little wacky, but I think this will mostly be addressed by the new full-sheet-tab interface we move to (which I think is gonna be a huge win for graphing). 

Good work getting this finished up. Also, the final solution we ended up for when to generate new graph code is actually pretty straight forward and not hard to follow. Nice!

PR #111, opened 2022-02-25
.gitignore: better ignoring

naterush:
# Description

Just ignore some files we don't need.

# Testing

No changes otherwise, so nothing.

# Documentation

No.


naterush (commented):
Only __pycache_ files, so good

PR #113, opened 2022-02-25
Logging changes

naterush:
# Description

NOTE: merge after mitosheet private!

1. Adds logging for how long ops take to run, in 10ths of a second
2. Add an explicit log for errors
3. Add log if DF names are not read in correctly
4. Add location log

# Testing

## How long an op takes to run

Just check log to see processing time

## An explicit log for errors

Generate an invalid formula, and see if an explicit error log is generated

## Testing the df names are not read correctly

Comment out line 142 of Mito.tsx, so the names aren't read in correctly. Then, try and make an edit, and make sure the log gets generated.

## Testing for the location

1. Check in Python terminal (make sure in venv where jupyterlab is run from). Run
```
python3
from mitosheet.user.location import get_location
get_location()
```
Try running this when no labs/notebooks are running, and when they are

6. Test in Jlab

Start a Jlab, and run
```
from mitosheet.user.location import get_location
get_location()
```

7. Test in JNotebook

Kill the Jlab. Run `jupyter notebook`. Run
```
from mitosheet.user.location import get_location
get_location()
```

8. Run both a JLab and Notebook 

In one of them, run
```
from mitosheet.user.location import get_location
get_location()
```
It should print out unknown notebook - we don't have a good way to tell! 

9. VS Code notebook

Open a notebook in VS Code, and run that
```
from mitosheet.user.location import get_location
get_location()
```

10. VS Code python file
Create a Python file `test.py` in VS Code, and from the command line run `python3 test.py`:
```
from mitosheet.user.location import get_location
print(get_location())
```
It should be location unknown.

11. Test in Google cola

Copy in the entire contents of the `mitosheet/user/location.py` file, as well as the `run_command` function from `mitosheet/utils`. Then, run `get_location()` - and make sure it's google collab.

12. Check it's in the logs

Finally, check that this is in the logs

# Documentation

None.

### DIFF
--- a/mitosheet/mitosheet/user/location.py
+++ b/mitosheet/mitosheet/user/location.py
@@ -1,0 +1,79 @@
+import os
+from mitosheet.utils import run_command
+
+
+def is_in_google_colab() -> bool:
+    # From: https://discourse.jupyter.org/t/find-out-if-my-code-runs-inside-a-notebook-or-jupyter-lab/6935/19
+    try:
+        from IPython.core.getipython import get_ipython
+        return 'google.colab' in str(get_ipython())
+    except:
+        return False
+
+def is_in_vs_code() -> bool:
+    # From: https://github.com/microsoft/vscode-jupyter/issues/3364
+    return 'VSCODE_PID' in os.environ
+
+def is_jupyter_lab_running() -> bool:
+    """
+    Format of command output: 
+
+    Currently running servers:
+    http://localhost:8888/?token=wenflwenflqnwdlkqmwldkm :: Path/to/launch
+    """
+    try:
+        return '::' in run_command(['jupyter', 'lab', 'list'])[0]
+    except:
+        return False
+
+def is_jupyter_notebook_running() -> bool:
+    """
+    Format of command output: 
+
+    Currently running servers:
+    http://localhost:8888/?token=wenflwenflqnwdlkqmwldkm :: Path/to/launch
+    """
+    try:
+        return '::' in run_command(['jupyter', 'notebook', 'list'])[0]
+    except:
+        return False
+
+def is_notebook() -> bool:
+    """
+    Returns true if this code is run in some sort of notebook
+    Taken from: https://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook
+    """
+    try:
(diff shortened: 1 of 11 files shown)


mitosheet/mitosheet/user/location.py
```diff
@@ -0,0 +1,79 @@
+import os
```
naterush, line 1: Add doc string

mitosheet/mitosheet/user/utils.py
```diff
@@ -10,15 +10,16 @@
 import getpass
 import os
 from datetime import datetime
+from subprocess import CompletedProcess
```
naterush, line 13: Remove

mitosheet/mitosheet/utils.py
```diff
@@ -233,3 +240,30 @@ def get_new_id() -> str:
     return str(uuid.uuid4())
 
 
+def run_command(command_array: List[str]) -> Tuple[str, str]:
+    """
+    An internal command that should be used to run all commands
+    that run on the command line, so that output from failing
+    commands can be captured.
+
+    Can toggle if this raises an error with fail_on_nonzero_exit_code
```
naterush, line 249: Remove

mitosheet/mitosheet/utils.py
```diff
@@ -233,3 +240,30 @@ def get_new_id() -> str:
...
+    """
+    import subprocess
+    completed_process = subprocess.run(
+        command_array, 
+        # NOTE: we do not use the capture_output variable, as this doesn't work before
+        # python 3.7, and we want users to be able to install before that
```
naterush, line 255: Comments

naterush (commented):
Cleanup!

naterush:
@aarondr77 the other _really cool_ thing I'll note about this work: in the future, we can use these location functions to print / display better error messages if the user is trying to run the mitosheet in other locations. 

As is, I am not super secure in using them for this - in case there is some bug. As of now, it might mislead us - but I want to avoid users sheets not displaying for stupid reasons. So let's just give it 2 weeks or so, see how we feel about these logs, and go from there.

I opened an issue for it here:https://github.com/mito-ds/monorepo/issues/114

aarondr77:
This PR is super super exciting! Its going to really shine light on a few potential major areas of investment. Good work.

PR #132, opened 2022-03-02
Graph fullsheet

aarondr77:
# Description

Makes graphing a full sheet experience. Each graph now is its own tab in Mito. The fact that the graph is an entire sheet and the fact that the graph is persistent makes graphing feel way more powerful, even without the upcoming functionality changes. 

It allows you to delete, duplicate, and rename graphs. It also lets you use generate a graph from a datasheet through the sheet tab actions. 

It also gets rid of the usage triggered feedback so we don't need to maintain that anymore. 

There are a few things that this PR leaves desired:

Number 1:
there is a tradeoff between making redo work properly when the entire graph is deleted, and making the graph refresh when opening the graph. The reason that this tradeoff exists is the same reason that we can't properly handle redo in pivot table (where we can't refresh the params). Namely, we can't tell the difference between a redo event and just a regular edit. See the existing issue here https://github.com/mito-ds/monorepo/issues/120

Specifically, for graphing we can't tell the difference between opening a graph / switching from one graph to another (both of which cause the graphID in the GraphSidebar to change) and a redo event that recreates a graph that was previously completely removed using undo.

Since the graph has existing issues with redo (namely, the parameters aren't kept in sync), and I think its a more common use case that people will create a graph, edit the data, and then want to see the updated graph (this is one of our core design principles of graphing -- respect the relationship between graphing and transformation), I vote we opt for refreshing the graph, causing redo to not work in this specific instance.

Number 2: https://github.com/mito-ds/monorepo/issues/147

Number 3: Since issues documented https://github.com/mito-ds/monorepo/issues/145, when you create a graph, it actually gets created as two different steps. Therefore, using Undo requires pressing it twice to get the graph to actually fully dissapear. 

Number 4: We should make the graph taskpane play nicer with the toolbar. However, I don't want to expand this PR to do so, so lets address it separetly. I added a note to the next spec to specify this in a bit more detail and add it to the plan of attack. 

# Testing

Make sure that it plays nicely with undoing graphs, and that the correct graph is always selected. 

Once we merge into dev, we can test the spacing on windows

# Documentation

Yes, this requires documentation changes, but we should just wait until we are finished this week of changes before updating.


mitosheet/mitosheet/step_performers/graph_steps/graph_rename.py
```diff
@@ -0,0 +1,95 @@
...
+        **params
+    ) -> Tuple[State, Optional[Dict[str, Any]]]:
+        # Bail early, if there is no change or the new name is empty
+        if old_graph_tab_name == new_graph_tab_name or new_graph_tab_name == '' :
+            return prev_state, None
+
+        # Create a new step and save the parameters
+        post_state = copy(prev_state)
+
+        post_state.graph_data[graph_id]["graphTabName"] = new_graph_tab_name
```
naterush, line 59: @aarondr77 I think we should always make the graphTabName be the same as the name of the graph itself; like, the name of the graph is the name of the variable of the graph is the name of the graph itself (with underscores). 

Not that I think we should do that right now. I _do think_ that this is something that we should specify and explore, as I actually think there's a bit of design work that we can do!

naterush: We can add this specification to the other specification if you want, but I think we should lead it out of this PR -- feature creep is real, careful!

mitosheet/mitosheet/state.py
```diff
@@ -116,7 +117,7 @@ def __init__(
             ]
         )
 
-        self.graph_data: Dict[str, Dict[str, Any]] = graph_data if graph_data is not None else dict()
+        self.graph_data: OrderedDict[str, Dict[str, Any]] = graph_data if graph_data is not None else dict()
```
naterush, line 120: A comment here would be nice -- why? Sure you woulda done it, just remindin'

mitosheet/mitosheet/step_performers/graph_steps/graph_delete.py
```diff
@@ -0,0 +1,83 @@
...
+        prev_state: State,
+        graph_id: GraphID,
+        **params
+    ) -> Tuple[State, Optional[Dict[str, Any]]]:
+
+        # Create a new step and save the parameters
+        post_state = copy(prev_state)
+
+        # Execute the graph delete
+        post_state.graph_data.pop(graph_id)
```
naterush, line 52: The more standard way of doing this is `del post_state.graph_data[graph_id]`

mitosheet/mitosheet/step_performers/graph_steps/graph_delete.py
```diff
@@ -0,0 +1,83 @@
...
+    ) -> str:
+        return f'Deleted a graph'
+    
+    @classmethod
+    def get_modified_dataframe_indexes( # type: ignore
+        cls, 
+        graph_id: GraphID,
+        **params
+    ) -> Set[int]:
+        return {-1}
```
naterush, line 83: Comment here would be nice

mitosheet/mitosheet/step_performers/graph_steps/graph_utils.py
```diff
@@ -80,6 +81,25 @@ def get_graph_title(
     return " ".join(graph_title_components)
 
 
+def get_new_graph_tab_name(graph_data: Dict[str, Dict[str, Any]]) -> str:
+    """
+    Creates the name for the new graph tab sheet using the format
+    graph0, graph1, etc.
+    """
+    all_graph_names = []
+    for value in graph_data.values():
```
naterush, line 90: value is vague

mitosheet/src/components/footer/Footer.tsx
```diff
@@ -46,13 +49,28 @@ function Footer(props: FooterProps): JSX.Element {
                 <PlusIcon/>
             </div>
             <div className="footer-tab-bar">
+                {/* First add the data tabs, and then add the graph tabs */}
                 {props.sheetDataArray.map(df => df.dfName).map((dfName, idx) => {
                     return (
                         <SheetTab
```
naterush, line 55: Sheet Tab feels like the wrong name for this thing now

mitosheet/src/components/footer/SheetTabActions.tsx
```diff
@@ -16,58 +17,111 @@ export default function SheetTabActions(props: {
...
-                    selectedSheetIndex: props.sheetIndex - 1
-                }
-            })
-        }
+        // If we are deleting a sheet tab, select the first sheet tab
+        props.setUIState(prevUIState => {
+            return {
+                ...prevUIState,
+                selectedTabType: 'data',
+                selectedSheetIndex: 0,
```
naterush, line 42: Why the first - why can't we select one one before it, whatever if its?

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -137,37 +166,90 @@ const GraphSidebar = (props: {
...
+    // so we can update the selected tab
+    const prevNumberOfGraphs = usePrevious(Object.keys(props.graphDataJSON).length)
 
     // When the last step index changes, check if an undo occured so we can refresh the params
     useEffect(() => {
         // If there has been an undo, then we refresh the params to this pivot
         if (prevLastStepIndex && prevLastStepIndex !== props.lastStepIndex - 1) {
             void refreshParamsAfterUndo()
+
+            const graphIDs = Object.keys(props.graphDataJSON)
```
naterush, line 186: I don't want to do this work in the Undo effect - can we handle this generally in the same place where we handle when a new graph is created?

naterush (changes requested):
A quick pop-in. My main confusion here is how were handling / detecting a new graph. My feeling about the best way to do this is very similar for how we do it with sheets:

1. We store the number of graphs in a ref in the GraphSidebar
2. When the graph data changes, we check if the number of graphs has changed.
3. If it has increased, we select this new graph
4. If it has decreased, we make sure we're not selecting an invalid graph (which means either selecting a different graph, or if there are none of them, selecting a sheet).

Ideally, all of these steps happen in a _single_ use effect. 

Is there motivation behind the current approach? I don't think I really understand what it is currently - so if you can document it somewhere (if you want to keep it) so I can understand, that would be great so we can see if we want to stick w/ that or move to the new one.

aarondr77:
Here is an explanation on my approach to the graph sidebar. I think its quite simple, but let me know if you want me to try something else. 
        
The graphID is the keystone of the graphSidebar. Each graph tab has one graphID that does not switch even if the user changes the data source they are graphing. 
        
In order to properly open a graph in Mito, there are a few things that need to occur:
1. We need to update the uiState's `selectedTabType` to "graph" so that the footer selects the correct tab and the `selectedGraphID`. 
2. We need to set the `currOpenTaskpane` to `{type: TaskpaneType.Graph, graphID: graphID}` so that we display the correct graph. 

Everything else is handled by the graphSidebar.  

With that in mind, lets look at how to do a few things:

### Creating a new graph
The opener of the taskpane is required to create a new graphID and pass that graphID to the `currOpenTaskpane`. They also need to update the `selectedTabType`. 

It looks like this: 
```
setUIState(prevUIState => {
    const newGraphID = getRandomId() // Create a new GraphID
    return {
        ...prevUIState,
        selectedTabType: 'graph',
        selectedGraphID: newGraphID,
        currOpenModal: {type: ModalEnum.None},
        currOpenTaskpane: {
            type: TaskpaneType.GRAPH, 
            graphID: newGraphID
        },
    }
})
```

### Switching sheet tabs
When we are switching sheet tabs, we do the same thing, except instead of creating a new graphID, we use the graphID of the graph sheet tab we clicked on. 


It looks like this: 
```
setUIState(prevUIState => {
      return {
          ...prevUIState,
          selectedTabType: 'graph',
          selectedGraphID: props.tabIDObj.graphID,
          currOpenModal: {type: ModalEnum.None},
          currOpenTaskpane: {
              type: TaskpaneType.GRAPH,
              graphID: props.tabIDObj.graphID
          } 
      }
  })
```

### Updating when an Undo occurs 
When an undo occurs, we can figure out that it happened in the standard way of comparing the `prevLastStepIndex` to the current `lastStepIndex`. If it did occur, we compare the `prevNumberOfGraphs` to `currNumGraphs`. If a graph was removed and there are more graphs in Mito, we display a previous graph . If there are no more graphs, we display the `selectedSheetIndex`. To do this, we just use logic from switching sheet tabs. 

Because switching sheet tabs updates `props.graphID` of the graphSidebar, the `useEffect` that watches for changes in the prop.graphID gets triggered and handles updating all of the necessary state. This useEffect is what gets used when an undo occurs and when switching sheets. And all it does is 

1. reset the stepID to undefined so we don't overwrite previous edits
2. refresh the graphParams so the UI is up to date with the new graphID's configuration.

mitosheet/src/components/Mito.tsx
```diff
@@ -96,6 +96,8 @@ export const Mito = (props: MitoProps): JSX.Element => {
         currOpenTaskpane: {type: TaskpaneType.NONE},
         selectedColumnControlPanelTab: ControlPanelTab.FilterSort,
         selectedSheetIndex: 0,
+        selectedGraphID: Object.keys(props.analysisData.graphDataJSON).length == 0 ? undefined : Object.keys(props.analysisData.graphDataJSON)[0],
```
aarondr77, line 99: Add a comment about why this might be undefined contrary to the sheetIndex

mitosheet/src/components/footer/Footer.tsx
```diff
@@ -27,8 +28,15 @@ type FooterProps = {
...
 
-    const sheetIndex = props.gridState.sheetIndex;
-    const sheetData: SheetData | undefined = props.sheetDataArray[sheetIndex];
+    const selectedSheetIndex = props.uiState.selectedSheetIndex
+    const selectedGraphID = props.uiState.selectedGraphID
+    const selectedTabType = props.uiState.selectedTabType
+
+    // Get the sheet index to display the rows and columns of. 
+    // If the sheet tab is a graph, then display the info from the data being graphed 
+    const sheetIndex =selectedTabType === 'graph' && selectedGraphID !== undefined && props.graphDataJSON[selectedGraphID] !== undefined?
```
aarondr77, line 37: add space between undefined and ?

mitosheet/src/components/footer/SheetTab.tsx
```diff
@@ -28,26 +31,32 @@ type SheetTabProps = {
...
     // We only set this as open if it the currOpenSheetTabActions
     const [displayActions, setDisplayActions] = useState(false);
     const [isRename, setIsRename] = useState<boolean>(false);
-    const [newDataframeName, setNewDataframeName] = useState<string>(props.dfName);
-    const isSelectedTab = props.selectedSheetIndex === props.sheetIndex
-    const selectedClass = isSelectedTab ? 'tab-selected' : '';
+    const [newTabName, setNewTabName] = useState<string>(props.tabName);
 
     // Make sure that if we change the df name that is displayed, we default to 
     // the right new dataframe name as well
```
aarondr77, line 42: Update comment to be more generic

mitosheet/src/components/footer/SheetTabActions.tsx
```diff
@@ -16,58 +17,112 @@ export default function SheetTabActions(props: {
     setUIState: React.Dispatch<React.SetStateAction<UIState>>;
     closeOpenEditingPopups: () => void;
     dfName: string, 
-    sheetIndex: number,
-    selectedSheetIndex: number,
     mitoAPI: MitoAPI
+    tabIDObj: {tabType: 'data', selectedIndex: number} | {tabType: 'graph', graphID: GraphID}
```
aarondr77, line 21: Instead of selectedIndex, just use sheetIndex

mitosheet/src/components/footer/SheetTabActions.tsx
```diff
@@ -16,58 +17,112 @@ export default function SheetTabActions(props: {
...
+        if (props.tabIDObj.tabType === 'data') {
+            props.setUIState(prevUIState => {
+                return {
+                    ...prevUIState,
+                    currOpenTaskpane: {
+                        type: TaskpaneType.GRAPH,
+                        graphID: getRandomId() // Create a new graph
+                    },
+                    // Note: We don't set the selected graph tab because we don't know the graph ID yet. 
+                    // Instead, we let the graphSidebar select the graph tab.
```
aarondr77, line 99: Select it since we now know the graphID

mitosheet/src/components/footer/SheetTabActions.tsx
```diff
@@ -16,58 +17,112 @@ export default function SheetTabActions(props: {
...
 
     return (
         <Dropdown
             closeDropdown={() => props.setDisplayActions(false)}
             width='small'
         >
+            <DropdownItem
+                title='Graph data'
+                onClick={(e) => {
+                    e?.stopPropagation()
```
aarondr77, line 113: Add stop propogation comment from below

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -75,13 +76,22 @@ const getDefaultSafetyFilter = (sheetDataArray: SheetData[], sheetIndex: number)
...
-    sheetIndex: number,
+    graphID: GraphID,
+    selectedSheetIndex: number,
     sheetDataArray: SheetData[],
 ): GraphParams => {
-    const graphParams = graphDataJSON[sheetIndex.toString()]?.graphParams;
+
+    const graphParams = graphDataJSON[graphID]?.graphParams;
+
+    // If the graph exists already exists, get the data source sheet index from the graph params.
```
aarondr77, line 86: double 'exists'

aarondr77: get rid of 'we're' in the line below

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -112,17 +125,40 @@ const GraphSidebar = (props: {
...
     columnDtypesMap: Record<string, string>;
     mitoAPI: MitoAPI;
     setUIState: React.Dispatch<React.SetStateAction<UIState>>;
+    uiState: UIState;
     graphDataJSON: GraphDataJSON
     lastStepIndex: number
 }): JSX.Element => {
 
+    /*
+        The graphID is the keystone of the graphSidebar. Each graph tab has one graphID that does not switch even if the user changes source data sheets.
```
aarondr77, line 138: Take comment from GitHub

mitosheet/src/components/footer/Footer.tsx
aarondr77: To me it reflects the model that makes more sense. We have two different types of sheets -- one for data and one for graphs. But currently, its weird because the graph sheet is hacked to actually just be a large taskpane.

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
aarondr77: A lot of this logic is unique to the undoing. The complex part is figuring out if another graph exists to switch to and if not, switching to a data sheet. So I think it makes sense to not bundle together. 

But, this logic is also needed by the sheet tabs `onDelete` function, so I pulled it out into a helper function.

aarondr77:
If you already have a graph tab open and you click on the graph button in the toolbar, the new graph tab doesn't appear until you configure the new graph

aarondr77:
We need to handle deleting data sheet tabs that are also graphed. Currently there is a sheet crashing bug:

To reproduce: 
1. Import data and graph it
2. Delete the data 
3. Click on the graph sheet tab -- notice that the sheet crashes. 

Some proposals: 
1. If you delete a sheet that is graphed popup a warning that asks if you want to continue -- just like the clear analysis popup. 
2. Don't let the sheet crash and save the graph as a static view where the user cannot edit it. the only thing they can do it delete the graph.

aarondr77:
We need to add some basic tests for the new step types.

mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -139,7 +148,7 @@ def execute(  # type: ignore
                     l=0,
                     r=0,
                     t=30,
-                    b=30,
+                    b=35,
```
naterush, line 151: Let's check these margins on other browsers: specifically, I want to test this thoroughly for formatting on Windows before we merge it.

mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -114,15 +117,21 @@ def execute(  # type: ignore
         df: pd.DataFrame = prev_state.dfs[sheet_index].copy()
         df_name: str = prev_state.df_names[sheet_index]
 
+        # If the graph tab already exists, use its name. Otherwise, create a new graph tab name.
+        graph_name: str = post_state.graph_data[graph_id]["graphTabName"] \
```
naterush, line 121: why not call this variable graph_tab_name?

mitosheet/mitosheet/step_performers/graph_steps/graph_utils.py
naterush, line 90: As before - value is a vague name - can we do better here?

mitosheet/mitosheet/tests/steps_performers/test_graph.py
```diff
@@ -5,7 +5,7 @@
 # Distributed under the terms of the GPL License.
 
 import pandas as pd
-from mitosheet.step_performers.graph.graph_utils import BAR
+from mitosheet.step_performers.graph_steps.graph_utils import BAR
 from mitosheet.tests.test_utils import create_mito_wrapper_dfs
 
 def test_create_graph():
```
naterush, line 11: Yup, more tests would be great, as you say.

Also tests for the other types

mitosheet/src/api.tsx
```diff
@@ -539,6 +541,52 @@ export default class MitoAPI {
...
+            'type': 'graph_delete_edit',
+            step_id: getRandomId(),
+            'params': {
+                'graph_id': graphID
+            }
+        }, {})
+    }
+
+    async editGraphDuplicate(
+        originalGraphID: GraphID,
```
naterush, line 559: in all other places, we use the word `old` to say it's old vs new. Can we change original -> old in all places (including throughout the step)?

mitosheet/src/components/Mito.tsx
naterush, line 99: === 0, for safety, not sure what the diff is here

mitosheet/src/components/footer/Footer.tsx
naterush, line 37: missing space here.

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
naterush: Note to self: look into this logic more.

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -137,37 +173,74 @@ const GraphSidebar = (props: {
...
-    const dataSourceSheetIndex = graphParams.graphCreation.sheet_index
-    const graphOutput = props.graphDataJSON[dataSourceSheetIndex]?.graphOutput
-    const [_copyGraphCode, graphCodeCopied] = useCopyToClipboard(graphOutput?.graphGeneratedCode);
-    const [stepID, setStepID] = useState<string|undefined>(undefined);
-
-    const [loading, setLoading] = useState<boolean>(false)
+    // Log when the graph has been opened
+    useEffect(() => {
+        void props.mitoAPI.log('opened_graph');
+    }, []);
```
naterush, line 179: Don't we get a log when the user creates a new graph as it is a step? Aka, I am not sure that we need to do this additional logging here. 

Additionally: I think we need to make sure logs actually anonymized; given that these params are nested, we need to check that our algorithm correctly loops down.

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -112,17 +125,40 @@ const GraphSidebar = (props: {
...
     graphDataJSON: GraphDataJSON
     lastStepIndex: number
 }): JSX.Element => {
 
+    /*
+        The graphID is the keystone of the graphSidebar. Each graph tab has one graphID that does not switch even if the user changes source data sheets. 
+        
+        In order to properly open a graph in Mito, there are a few things that need to occur:
+            1. We need to update the uiState's `selectedTabType` to "graph" so that the footer selects the correct tab
+            2. We need to set the uiState's `currTaskpaneOpen` to "graph" so that we actually display the graph
```
naterush, line 142: If these two are true, can we potentially simplify and not use `selectedTabType`? We can just check if `currTaskpaneOpen` is `graph` - and otherwise it's a sheet tab selected?

Would this improve things? What do you think?

It's a bit of a red flag that you have to remember to update _two_ pieces of the UI state to pretty much the same variable - and that they always move in unison!

mitosheet/src/components/footer/SheetTabActions.tsx
naterush: Bumping this question - I don't understand! Can you explain to me?

mitosheet/src/components/footer/SheetTabActions.tsx
```diff
@@ -16,58 +17,112 @@ export default function SheetTabActions(props: {
...
+            await props.mitoAPI.editGraphDelete(props.tabIDObj.graphID)
+        }
     }
 
     const onDuplicate = async (): Promise<void> => {
         // Close 
         props.closeOpenEditingPopups();
         
-        await props.mitoAPI.editDataframeDuplicate(props.sheetIndex)
+        if (props.tabIDObj.tabType === 'data') {
```
naterush, line 61: It's kinda interesting each of these functions (and indeed, the sheet tab actions) are shared between the two different tab types. 

IMO, a better way of doing this would be to refactor the sheet tab actions to be totally different for each type - e.g. I think it would be better to have sheet tab actions for the sheet, and graph tab actions for the graph. This would lead to much simpler, much more maintainable code in the long run, as these two things continue to diverge (I think they will continue to diverge, as we're seeing with the Graph Data functionality). 

Can you rip these out into two?

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -75,13 +76,22 @@ const getDefaultSafetyFilter = (sheetDataArray: SheetData[], sheetIndex: number)
...
+    selectedSheetIndex: number,
     sheetDataArray: SheetData[],
 ): GraphParams => {
-    const graphParams = graphDataJSON[sheetIndex.toString()]?.graphParams;
+
+    const graphParams = graphDataJSON[graphID]?.graphParams;
+
+    // If the graph exists already exists, get the data source sheet index from the graph params.
+    // Otherwise we're creating a new graph of the selectedSheetIndex
+    const graphDataSourceSeetIndex = graphParams !== undefined ? graphParams.graphCreation.sheet_index : selectedSheetIndex
```
naterush, line 88: `graphDataSourceSeetIndex` -> `graphDataSourceSheetIndex`

naterush (changes requested):
Ok, a couple minor changes and 2 medium architectural questions (1 of which I am not sure of, and one of which I am reasonably confident is a good idea and would like to do).

Let me know your thoughts on them, and call me sometime if you want to discuss / chat about what you think is worth it, given that you're more in this code than I am. 

Overall, this looks really great and I think is just a few hours from being finished (code-wise). I'll review this again when it's done, and check the functionality more (which I have not done). 

Good big architectural changes! You really crushed it! This change is gonna be awesome for our users!

naterush:
Following up with an additional 2 things to check in on, based on some log review I was doing this morning. See a dashboard here: https://mixpanel.com/s/377ZXC

The main error to figure out is: ValueError: Plotly Express cannot process wide-form data with columns of different type. Going to the events tab and filling down to these events, here's a sample of the error:

["Traceback (most recent call last):","File \"/mitosheet/api/api.py\", line 102, in handle_api_event_thread","handle_api_event(send, event, steps_manager)","File \"/mitosheet/api/api.py\", line 126, in handle_api_event","result = get_graph(event, steps_manager)","File \"/mitosheet/api/graph/get_graph.py\", line 96, in get_graph","fig = get_plotly_express_graph(","File \"/mitosheet/api/graph/plotly_express_graphs.py\", line 233, in get_plotly_express_graph","fig = graph_creation(graph_type, df, x_axis_column_headers, y_axis_column_headers)","File \"/mitosheet/api/graph/plotly_express_graphs.py\", line 106, in graph_creation","return px.bar(df, **all_params)","File \"/plotly/express/_chart_types.py\", line 365, in bar","return make_figure(","File \"/plotly/express/_core.py\", line 1933, in make_figure","args = build_dataframe(args, constructor)","File \"/plotly/express/_core.py\", line 1452, in build_dataframe","raise ValueError(","ValueError: Plotly Express cannot process wide-form data with columns of different type."]

Can you look into this as well and update here about what is going on?

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
aarondr77: As per our discussion, we're not going to separate these

aarondr77:
Re: the plotly error you saw. That is the plotly express error that occurs when trying to graph different type series on the same axis. Its something we're tracking and will decide in the spec if we want to either 1) filter out columns wit a different type 2) give a warning 3) give an error message.

aarondr77:
The only remaining thing to do is make the graphing sheets play nicely with the toolbar buttons. We can either disable some of the buttons or we can make them go to the data source sheet and act on that data. 

We need to make sure that we update the selected sheet tab while doing this.

aarondr77:
Check how it works with fullscreen + toggling fullscreen

mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -139,7 +148,7 @@ def execute(  # type: ignore
                     l=0,
                     r=0,
                     t=30,
-                    b=30,
+                    b=35, # This gives enough space so that the x axis label is not cutoff
```
naterush, line 151: @aarondr77 we should test this stuff on windows before deploying! I'm fine to do this after we deploy on Test PyPi, so that we can avoid your environment setup issues.

mitosheet/mitosheet/step_performers/graph_steps/graph_delete.py
```diff
@@ -0,0 +1,87 @@
...
+    @classmethod
+    def step_event_type(cls) -> str:
+        return 'graph_delete_edit'
+
+    @classmethod
+    def saturate(cls, prev_state: State, params: Dict[str, Any]) -> Dict[str, Any]:
+        graph_id = params['graph_id']
+        old_graph_tab_name = prev_state.graph_data[graph_id]
+        params['old_graph_tab_name'] = old_graph_tab_name
+        return params
```
naterush, line 40: Do we need this? I think this is used in the similar delete step for the data frame, but only because we need the name in the transpile step; since we don't have a transpile step here, this can go.

mitosheet/mitosheet/step_performers/graph_steps/graph_duplicate.py
```diff
@@ -0,0 +1,88 @@
...
+        prev_state: State,
+        old_graph_id: GraphID,
+        new_graph_id: GraphID,
+        **params
+    ) -> Tuple[State, Optional[Dict[str, Any]]]:
+        post_state = deepcopy(prev_state)
+
+        # Execute the step
+        graph_copy = deepcopy(post_state.graph_data[old_graph_id])
+        graph_copy["graphTabName"] = graph_copy["graphTabName"] + '_copy'
```
naterush, line 52: Can you note here that we don't need to insist that graphTabNames are unique? Since we don't transpile them linearly, this is not an issue.

mitosheet/mitosheet/step_performers/graph_steps/graph_utils.py
```diff
@@ -80,6 +81,25 @@ def get_graph_title(
     return " ".join(graph_title_components)
 
 
+def get_new_graph_tab_name(graph_data: Dict[str, Dict[str, Any]]) -> str:
+    """
+    Creates the name for the new graph tab sheet using the format
+    graph0, graph1, etc.
+    """
+    all_graph_names = []
+    for graph_data in graph_data.values():
```
naterush, line 90: This is dangerous! You're redefining an exising variable, which means if you try and use the param later on, it'll be defined as one of the values in the graph data dict. Sorry to be a narc here, but this is even worse than value - this _will_ lead to bugs if we ever refactor this function.

What about renaming the param to the function to `graph_data_dict` or something?

mitosheet/mitosheet/tests/steps_performers/graph_steps/test_graph.py
```diff
@@ -0,0 +1,92 @@
...
+    graph_data = mito.get_graph_data(graph_id)
+
+    assert graph_data["graphParams"]["graphCreation"]["graph_type"] == 'box'
+    assert graph_data["graphParams"]["graphCreation"]["sheet_index"] == 0
+    assert graph_data["graphParams"]["graphCreation"]["x_axis_column_ids"] == ['A']
+    assert graph_data["graphParams"]["graphCreation"]["y_axis_column_ids"] == ['B']
+
+    assert graph_data["graphOutput"]["graphScript"] is not None
+    assert graph_data["graphOutput"]["graphHTML"] is not None
+    assert graph_data["graphOutput"]["graphGeneratedCode"] is not None
```
naterush, line 49: Can you write a little helper function for getting at these values? Specifically: `getGraphType`, `getSheetIndex`, etc. These can live in the test utils file, but don't need to be defined on the class if you don't want - they can just be standalone.

This will allow us to:
1. Change the structure of this object
2. Only update where we change the structure in a single place to make the tests pass

Both are pretty important to me, because they make maintaining tests super easy to maintain! I've learned this lesson the hard way :-)

mitosheet/mitosheet/tests/steps_performers/graph_steps/test_graph.py
```diff
@@ -0,0 +1,92 @@
...
+
+def test_create_graph_histogram():
+    df = pd.DataFrame({'A': ['aaron', 'jake', 'nate'], 'B': [1, 2, 3]})
+    mito = create_mito_wrapper_dfs(df)
+    graph_id = '123'
+    mito.generate_graph(graph_id, HISTOGRAM, 0, False, ['A'], [] , 400, 400)
```
naterush, line 75: Also, it seems like all these tests could be a parameterized version of a _single_ test. Can you make this a `@pytest.mark.parameterize`, so these are similarly easy to maintain going forward?

mitosheet/src/components/footer/GraphSheetTabActions.tsx
```diff
@@ -1,73 +1,86 @@
...
-    mitoAPI: MitoAPI
+    mitoAPI: MitoAPI,
+    graphID: GraphID,
+    graphDataJSON: GraphDataJSON
 }): JSX.Element {
 
     // Log opening the sheet tab actions
     useEffect(() => {
         void props.mitoAPI.log(
             'clicked_sheet_tab_actions',
```
naterush, line 28: Let's change this log to `clicked_graph_tab_actions`

mitosheet/src/components/footer/DataSheetTabActions.tsx
```diff
@@ -0,0 +1,148 @@
...
+    }))
+
+    return Object.entries(filteredGraphDataJSON).map(([graphID, graphData]) => {
+        return {graphTabName: graphData.graphTabName, graphID: graphID}
+    })
+} 
+
+/*
+    Displays a set of actions one can perform on a sheet tab, including
+    deleting, duplicating, or renaming.
```
naterush, line 29: (as well as accessing it's graphs)

mitosheet/src/components/footer/DataSheetTabActions.tsx
```diff
@@ -0,0 +1,148 @@
...
+    that are created from a given sheet index
+*/
+export const getGraphTabNamesAndIDsFromSheetIndex = (sheetIndex: number, graphDataJSON: GraphDataJSON): 
+({graphTabName: string, graphID: GraphID})[] => {
+    // Filter to only grapsh with the sheetIndex, and then get a list of the graph tab names
+    const filteredGraphDataJSON: GraphDataJSON = Object.fromEntries(Object.entries(graphDataJSON).filter(([, graphData]) => {
+        return graphData.graphParams.graphCreation.sheet_index === sheetIndex
+    }))
+
+    return Object.entries(filteredGraphDataJSON).map(([graphID, graphData]) => {
```
naterush, line 22: Nice use of functional functions!

mitosheet/src/components/modals/DeleteGraphsModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+  A modal that confirms with the user that they want to delete the dataframe and 
+  all of the depedant graphs
+*/
+const DeleteGraphsModal = (props: DeleteGraphsModalProps): JSX.Element => {
+    const graphIDs: GraphID[] = props.dependantGraphTabNamesAndIDs.map(graphTabNameAndID => {return graphTabNameAndID.graphID})
+    const graphTabNames: string[] = props.dependantGraphTabNamesAndIDs.map(graphTabNameAndID => {return graphTabNameAndID.graphTabName})
+
+    const clickDelete = async () => {
+
+        // Delete the graphs and
```
naterush, line 31: and

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -137,37 +174,47 @@ const GraphSidebar = (props: {
...
     // When the last step index changes, check if an undo occured so we can refresh the params
     useEffect(() => {
         // If there has been an undo, then we refresh the params to this pivot
         if (prevLastStepIndex && prevLastStepIndex !== props.lastStepIndex - 1) {
             void refreshParamsAfterUndo()
+
+            const currNumGraphs = Object.keys(props.graphDataJSON).length
+            // If the number of graphs changed, update the selected tab
+            if (prevNumberOfGraphs && prevNumberOfGraphs > currNumGraphs) {
+                selectPreviousGraphSheetTab(props.graphDataJSON, props.setUIState)
```
naterush, line 192: Can we rename this function to `selectRightmostGraphSheetTab`?

naterush (changes requested):
Ok, looking great! There are some minor changes here (just cleanup) that shouldn't take more than an hour. 

Can you open an issue for moving the graphing taskpane -> a sheet in it's own right, and mark as medium effort and low priority?

My biggest question, which is not really relevant to this PR but to graphing stuff generally: are we so sure we should treat graph steps differently? Why not just write the code to the code cell below as we normally do? Our explicit code export step adds what? I am not really sure. If we still have an export code and export as png button for graphs, I kinda feel like having it written to the code cell below is... more sensible if anything. Anyways, don't worry about it now. 

I'm going to test the functionality now!

naterush:
Screen Shot 2022-03-08 at 8 47 46 AM

select _an_ axis

naterush:
Screen Shot 2022-03-08 at 8 48 24 AM

Screen Shot 2022-03-08 at 8 48 22 AM

What is this little thing visible under the graphing taskpane? Can we kill it?

naterush:
Screen Shot 2022-03-08 at 8 50 13 AM

I don't think the text in the data sheet tab is centered vertically. I think this might be due to the weird thing I can see beneath the graphing taskpane corner above...

naterush:
1. Create a graph.
2. Rename it.
3. Duplicate it.
4. Undo back to the start.
5. Redo. You can only redo to step (1). Why does this not work further?

I think this is a bug we want to get sorted before deploying - it breaks the redo functionality when working with graphs!

naterush:
Screen Shot 2022-03-08 at 9 00 53 AM

Can we space this better? Specifically, just making things lineup and look nice (and making changing the color of the graph names or something). It looks like old Mito design currently.

naterush:
@aarondr77 I'm gonna pause my review here. I think there are a couple some-what major considerations above that we should dive into - but some of them potentially have pretty complex solutions I want to make sure we're on the same page about before we move into fixing them. 

Can you write a specification for the behavior around: 
1. Creating new graph tabs, how we switch to them (to avoid race conditons where no tab is displayed).
2. Deleting graph tabs. How we a) make sure to close the graph taskpane if we should [e.g. there are no graphs left], and b) make sure we select the right graph tab if there are some left.

I'm looking for the specification to clarify what behavior we want in the first place, and also where this behavior will live (e.g. what file). If you want to explore different options, that would be cool to. This should have been more explicit in the initial specification - so my bad!

aarondr77:
Move to the same system as we use for the sheet indexes, where we check if the graphDataJSON updates in size, and then we go to that one.

aarondr77:
When you delete a graph, we want to move to the new graph in the same index, not the last one that is edited.

mitosheet/src/components/footer/DataSheetTabActions.tsx
```diff
@@ -0,0 +1,148 @@
...
+                sheet_index: props.sheetIndex
+            }
+        )
+    }, [])
+
+    const onDelete = async (): Promise<void> => {
+        const dependantGraphTabNamesAndIDs = getGraphTabNamesAndIDsFromSheetIndex(props.sheetIndex, props.graphDataJSON)
+        
+        if (dependantGraphTabNamesAndIDs.length > 0) {
+            props.setUIState(prevUIState => {
```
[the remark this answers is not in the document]
aarondr77: Yeah, there more work that we need to do if we want to keep them here. We'd need to save the graphScript and graphHTML so it can still be accessed when they switch to that tab. We'd also need to make sure that the graph is not editable so you can't configure a graph that doesn't exists, etc. etc. 

Given I don't think there is really a benefit of this, but it requires more technical complexity, I vote we don't do it.

mitosheet/src/components/footer/DataSheetTabActions.tsx
```diff
@@ -0,0 +1,148 @@
...
+import Dropdown from '../elements/Dropdown';
+import DropdownItem from '../elements/DropdownItem';
+import { ModalEnum } from '../modals/modals';
+import { TaskpaneType } from '../taskpanes/taskpanes';
+
+/*
+    Helper function for finding all of the graph tab names
+    that are created from a given sheet index
+*/
+export const getGraphTabNamesAndIDsFromSheetIndex = (sheetIndex: number, graphDataJSON: GraphDataJSON):
```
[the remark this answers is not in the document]
aarondr77: Btw: This type of comment isn't generally super helpful. most of the time, I just did what I thought was best so if you could be more specific about what you want, that would be a more helpful comment. Happy to change to whatever you want, just have to know what that is. 

In this case, I think it is properly formatted because that is how VSCode told me to format is. But I'll update it to all being in one line.

mitosheet/mitosheet/step_performers/graph_steps/graph_delete.py
aarondr77: I'm using the name in the describe step to make the comment in the steps taskpane informative.

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
aarondr77: Ignoring since we changed the functionality away from selecting the rightmost graph sheet tab.

aarondr77:
Remove the left border

Screen Shot 2022-03-09 at 3 12 59 PM

aarondr77:
Kinda weird situation:

1. Import Data with column A and B
2. Create a graph of column A and B
3. Delete column B from the data
4. Switch back to the graph (notice that column B is no longer in the graph)
5. Undo (notice that the graph displayed now includes column B, but the graph params in the UI don't)

This happens because when you opened the graph, it sent a new `editGraph` message to the backend which created the most recent step. Now, instead of undoing the column deletion, you are undoing the new graph generation. 

A solution would be to actually update the graph in the step that effects the underlying data. For example, in this case, the column delete would trigger an update of the `graph_data_dict` in the column delete step itself. 

To make this nice, since it would be needed for a large number of steps, we would want to handle this in the `mito_widget` for every step.

aarondr77:
The reason the redo is not working is because when the redo occurs and we create a graph that was completely removed, the length of the Object.keys(analysisData.graphDataJSON) changes. When that occurs, and we recognize that a graph was just added, we open the graph sidebar. But since we didn't create a graph in the frontend, we didn't have the opportunity to create the stepID. Therefore the `newGraphStepID` is undefined and so the graphsidebar actually sends an editGraph message because it thinks we're opening a graph that already existed and it needs to refresh the graph view on the new data. Since this message gets sent, we end up adding a new step which clears the `self.undone_step_list_store`

aarondr77:
Bug: 
1. Create a graph
2. Click on the steps taskpane 
3. Notice that the data is now displayed in the sheet, but the selected tab doesn't update.

naterush:
@aarondr77 make sure to merge in dev, when you get the chance (specifically, gonna be some conflicts b/c the new graph types). Sorry for the extra work here!

aarondr77:
Okay, so there is a tradeoff between making redo work properly when the entire graph is deleted, and making the graph refresh when opening the graph. The reason that this tradeoff exists is the same reason that we can't properly handle redo in pivot table (where we can't refresh the params). Namely, we can't tell the difference between a redo event and just a regular edit. See the existing issue here https://github.com/mito-ds/monorepo/issues/120

Specifically, for graphing we can't tell the difference between opening a graph / switching from one graph to another (both of which cause the graphID in the GraphSidebar to change) and a redo event that recreates a graph that was previously completely removed using undo. 

Since the graph has existing issues with redo (namely, the parameters aren't kept in sync), and I think its a more common use case that people will create a graph, edit the data, and then want to see the updated graph (this is one of our core design principles of graphing -- respect the relationship between graphing and transformation), I vote we opt for refreshing the graph, causing redo to not work in this specific instance.

aarondr77:
Screen Shot 2022-03-09 at 5 49 18 PM

mitosheet/mitosheet/steps_manager.py
```diff
@@ -276,7 +276,7 @@ def analysis_data_json(self):
                 "stepSummaryList": self.step_summary_list,
                 "currStepIdx": self.curr_step_idx,
                 "dataTypeInTool": self.data_type_in_mito.value,
-                "graphDataJSON": self.curr_step.graph_data
+                "graphDataJSON": self.curr_step.graph_data_dict
```
naterush, line 279: Should this be named graphDataDict? I think yes; we want to keep these consistent, if we can.

mitosheet/mitosheet/tests/steps_performers/graph_steps/test_graph.py
```diff
@@ -0,0 +1,35 @@
...
+import pandas as pd
+import pytest
+from mitosheet.step_performers.graph_steps.graph_utils import BAR, BOX, HISTOGRAM, SCATTER
+from mitosheet.tests.test_utils import create_mito_wrapper_dfs
+
+GRAPH_CREATION_TESTS = [
+    (BAR),
+    (BOX),
+    (HISTOGRAM),
+    (SCATTER)
```
naterush, line 16: Can we add the other graph types here?

mitosheet/src/api.tsx
```diff
@@ -539,6 +541,52 @@ export default class MitoAPI {
...
+
+    async editGraphDuplicate(
+        oldGraphID: GraphID,
+        newGraphID: GraphID
+    ): Promise<void> {
+        
+        await this.send<string>({
+            'event': 'edit_event',
+            'type': 'graph_duplicate_edit',
+            step_id: getRandomId(),
```
naterush, line 566: put in quotes

mitosheet/src/api.tsx
```diff
@@ -539,6 +541,52 @@ export default class MitoAPI {
...
+
+    async editGraphRename(
+        graphID: GraphID,
+        newGraphTabName: string
+    ): Promise<void> {
+        
+        await this.send<string>({
+            'event': 'edit_event',
+            'type': 'graph_rename_edit',
+            step_id: getRandomId(),
```
naterush, line 582: put in quotes

mitosheet/src/api.tsx
```diff
@@ -539,6 +541,52 @@ export default class MitoAPI {
...
     }
 
+    async editGraphDelete(
+        graphID: GraphID,
+    ): Promise<void> {
+
+        await this.send<string>({
+            'event': 'edit_event',
+            'type': 'graph_delete_edit',
+            step_id: getRandomId(),
```
naterush, line 551: put in quotes

mitosheet/src/api.tsx
```diff
@@ -518,6 +519,7 @@ export default class MitoAPI {
             'type': 'graph_edit',
             step_id: stepID,
```
naterush, line 520: put in quotes

mitosheet/src/components/Mito.tsx
```diff
@@ -219,6 +218,54 @@ export const Mito = (props: MitoProps): JSX.Element => {
         previousNumSheetsRef.current = sheetDataArray.length;
     }, [sheetDataArray])
 
+    const previousNumGraphsRef = useRef<number>(Object.keys(analysisData.graphDataJSON).length)
```
naterush, line 221: These effects are _so_ much more followable and consistent compared to how we were previously thinking about this. This is great!

mitosheet/src/components/Mito.tsx
```diff
@@ -356,12 +391,13 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
-                <FeedbackModal
-                    usageTriggereeFeedbackID={uiState.currOpenModal.feedbackID}
+            case ModalEnum.DeleteGraphs: return (
+                <DeleteGraphsModal
                     setUIState={setUIState}
                     mitoAPI={props.mitoAPI}
-                    numUsages={userProfile.numUsages}
+                    sheetIndex={uiState.currOpenModal.sheetIndex}
+                    dependantGraphTabNamesAndIDs={uiState.currOpenModal.dependantGraphTabNamesAndIDs}
+                    dfName={sheetDataArray[uiState.currOpenModal.sheetIndex].dfName}
```
naterush, line 400: This might be undefined! Never just index into the sheet data array and then get a property of the result without checking if defined. This causes sheet crashing bugs (and should be something we handle in the type system after we have sheet ids!).

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -161,15 +111,23 @@ const GraphSidebar = (props: {
         }
     }, [props.lastStepIndex])
 
+    /*
+        If the props.graphID changes, which happens when opening a graph:
+        1. reset the stepID so we don't overwrite the previous edits.
+        2. refresh the graphParams so the UI is up to date with the new graphID's configuration.
+        3. update the graphUpdateNumber so the graph refreshes
+    */
+    useEffect(() => {
```
naterush, line 120: Clean and clear, baby

mitosheet/src/components/footer/GraphSheetTabActions.tsx
```diff
@@ -0,0 +1,91 @@
...
+            return {
+                ...prevUIState,
+                selectedGraphID: newGraphID,
+                selectedTabType: 'graph',
+                currOpenTaskpane: {
+                    type: TaskpaneType.GRAPH,
+                    graphID: newGraphID,
+                },
+            }
+        })
```
naterush, line 59: I think we can (and should) remove this -- this is now handled by the effect that watches for new graphs, right? Let's not duplicate the behavior here for no reason.

mitosheet/src/components/footer/SheetTab.tsx
```diff
@@ -58,49 +124,85 @@ export default function SheetTab(props: SheetTabProps): JSX.Element {
...
     return (
         <div 
-            className={classNames('tab', selectedClass)} 
+            className={classNames('tab', {'tab-graph': props.tabIDObj.tabType === 'graph'}, {'tab-selected': props.isSelectedTab})} 
             onClick={() => {
                 props.setUIState(prevUIState => {
-                    return {
-                        ...prevUIState,
-                        selectedSheetIndex: props.sheetIndex
+                    if (props.tabIDObj.tabType === 'data') {
```
naterush, line 130: I think it's fine for now, but these if statements make me feel like there's an architectural change we could make that would make this a cleaner file; maybe the onClick or something is a prop? Idk. It's fine for now, just this code smells a bit.

mitosheet/src/components/taskpanes/Graph/GraphUtils.tsx
```diff
@@ -0,0 +1,78 @@
+// Helper function for creating default graph params. Defaults to a Bar chart,
```
naterush, line 1: This file should not be capitalized.

mitosheet/src/components/taskpanes/taskpanes.tsx
```diff
@@ -40,7 +41,8 @@ export type TaskpaneInfo =
     | {type: TaskpaneType.DROP_DUPLICATES}
     | {
         type: TaskpaneType.GRAPH,
-        graphSidebarSheet: number
+        graphID: GraphID,
+        newGraphStepID?: string
```
naterush, line 45: We don't use this anymore - right? Delete if so.

naterush (changes requested):
This code looks awesome. The last round of simplifications really make things much more followable, and way more intelligible and maintainable longterm!

A few tiny code and documentation changes (< 30 minutes). I am gonna go test this code manually now!

naterush:
Number 1: redo not working properly. 

I think this is super fine. Redo doesn't really work anywhere. Thank for digging into this so much and making a good call on which feature was important - I agree with your call!

Number 2: https://github.com/mito-ds/monorepo/issues/147

Super minor, I think this is fine.

Number 3: Undoing a graph creating requires two undos.

This is fine - thanks for the issue!

Number 4: not playing nicely with the toolbar.

I'm mostly fine with what we have currently. Can we just turn some of these buttons (namely, add column and delete column) off when in graphing mode? Literally just disable them in this case - I got tripped up on these two specifically.

naterush:
@aarondr77 I addressed the review, aka this PR is finished. Give it one final glance to peep my changes, and make sure they look good. Then, after we deploy the outstanding work on dev, LGTM!

PR #136, opened 2022-03-07
mitoinstaller: major cleanup, bug fixes with identify

naterush:
# Description

This PR is high priority, as it does a few things:
1. It cleans up our logging in our installer - so we can actually tell what is happening. Things were very messy and inconsistent before, which mostly meant it's pretty much impossible to figure out when/why things are failing.
2. Fixes a bug where many install attempts were not logged (namely, those where users never ran an import mitosheet call), because the identify function was not calling the underlying analytics.identify due to a non-updated guard to stop profile overwriting.

Notably, to do so, I further simplify the architecture of the installer. It's pretty damn simple now, and should be really understandable. As before, I'm continually realizing that simplicity is of the _upmost_ importance here, because debugging it otherwise is pretty much impossible (given the varied places it exists).

# Testing

The command I use:
`rm -rf ~/.mito/user.json; deactivate; rm -rf venv; python3 -m venv venv; source venv/bin/activate; pip install -r requirements.txt; python -m mitoinstaller install`

Test the following:
1. Run the installer, get logs
2. Run the installer with `--pro`, get logs
3. Do Control + C during some random steps, get logs
4. Add `raise Exception()` to random steps in random places, run installer, and make sure it still generates logs that work properly.

NOTE: a major part of this is making sure that the logs make sense. Specifically, our logging strategy is documented in the `run_installer_steps` comment - *make sure that we meet this specification and that you can tell what happened based on the logs, given the test you ran.* 

Being able to debug from the logs is the most important part of this PR!

# Documentation

No.

### DIFF
--- a/mitoinstaller/mitoinstaller/__main__.py
+++ b/mitoinstaller/mitoinstaller/__main__.py
@@ -10,7 +10,7 @@ Long term, we aim to meet:
 from colorama import init
 from termcolor import colored  # type: ignore
 
-from mitoinstaller.install import do_install_or_upgrade
+from mitoinstaller.install import do_install
 
 
 def main() -> None:
@@ -18,7 +18,7 @@ def main() -> None:
     The main function of the Mito installer, this function is responsible
     for installing and upgrading the `mitosheet` package.
 
-    To install Mito (for the first time):
+    To install Mito:
     python -m mitoinstaller install
 
     To upgrade Mito:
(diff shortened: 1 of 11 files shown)


naterush (commented):
Looks good!

naterush:
Note: really, seriously, the most important thing is the logging working! So make sure to check the logs thoroughly.

aarondr77:
In order to get the installer to run, I again had to add `--no-cache-dir` to the install command. I added it to line 101 of commands.py. 

This occured the last round of testing for me as well. Should we add it to the installer?

aarondr77:
Okay, so besides for the comment above, my testing is working as according to the comment in the PR. 

I can certainly tell where everything is failing and I can tell why it is failing, either user interruption or raise exception. 

Your call on the `--no-cache-dir` thing, and then LGTM.

naterush:
@aarondr77 my concern is that `--no-cache-dir` increases the install time, and potentially might do so really dramatically for some user. It's not recommended to turn this off (see here: https://pip.pypa.io/en/latest/topics/caching/#disabling-caching). 

This would be a big concern during upgrading, where it would redownload all packages, I think, and this could result in much much larger upgrade times and so less success. 

If we ever observe this in the wild, we can reevaluate!

PR #140, opened 2022-03-07
mitosheet: add starts with and endswith string filters

### LINKED ISSUE #138: Add startswith and endswith string filter conditions
**Is your feature request related to a problem? Please describe.**
In filter options, it would be very useful to have a "Where" option to find strings that `beginswith` or `endswith` a pattern.  I often need to filter for strings that end with "-Prod" exactly, but I have some rows that contain "-Productivity" in the middle of the string.  A contains filter will include those, so I'd have to add a second filter that does not contain "-Productivity" to get the desired result.  There may be other use cases where beginswith and endswith would be more useful.

naterush:
# Description

Closes #138 

# Testing

See tests. Test in Mito. Note that on Mixed type data, our current string filtering does not perform well when combining different types into multiple filter conditions (namely, `na=None` is not how we combine things in the multi-filter conditions - where we type cast). 

A note on the tests for the "combining filter conditions" for future reference:
1. We no longer test transpired code directly (I think when these were written, we used to do so).
2. The string condition aggregation isn't tested at all, let alone with mixed data types.

Instead of fixing these issues, I opened an issue here: https://github.com/mito-ds/monorepo/issues/139

# Documentation

Nope.

### DIFF
--- a/mitosheet/mitosheet/step_performers/filter.py
+++ b/mitosheet/mitosheet/step_performers/filter.py
@@ -31,6 +31,8 @@ FC_STRING_CONTAINS = "contains"
 FC_STRING_DOES_NOT_CONTAIN = "string_does_not_contain"
 FC_STRING_EXACTLY = "string_exactly"
 FC_STRING_NOT_EXACTLY = "string_not_exactly"
+FC_STRING_STARTS_WITH = "string_starts_with"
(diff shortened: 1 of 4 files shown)


aarondr77:
Beautiful! Amazing getting this turned around so quickly!

PR #146, opened 2022-03-09
Add new logs

naterush:
# Description

1. Docker
2. For performance metric for the backend: the pandas performance time. Note that I tried to strike a balance here for _just_ capturing the pandas operation, while not introducing a ton of new code, and also acknowledging that if statements are _not_ where the performance problems come from. 
3. Not the JS one (See explanation here: https://www.notion.so/trymito/New-Log-Work-48869aa0f3b245ffb2f4728ef5cf9d1c) - we need to talk to an expert if we want this to happen - I think it's a few days of investigation. So not doing it for now (spent a few hours reading about it and trying things).

# Testing

Investigate the two new logs. The docker one just test it runs correctly. - the code looks good otherwise since I didn't come up with it.

# Documentation

No.


mitosheet/mitosheet/step_performers/dataframe_steps/dataframe_delete.py
```diff
@@ -58,7 +58,9 @@ def execute( # type: ignore
         post_state.df_names.pop(sheet_index)
         post_state.df_sources.pop(sheet_index)
 
-        return post_state, None
+        return post_state, {
+            'pandas_processing_time': 0
```
naterush, line 62: Leave a comment

mitosheet/mitosheet/step_performers/dataframe_steps/dataframe_rename.py
```diff
@@ -58,7 +58,9 @@ def execute( # type: ignore
 
         post_state.df_names[sheet_index] = get_valid_dataframe_name(post_state.df_names, new_dataframe_name)
 
-        return post_state, None
+        return post_state, {
+            'pandas_processing_time': 0
```
naterush, line 62: leave a comment here too

naterush (commented):
Some comments

aarondr77:
Can we just add the logging infrastructure for the new graph steps and then LGTM

PR #151, opened 2022-03-10
mitosheet: format summary stats and unique values tab

### LINKED ISSUE #135: Summary stats to reflect number format for selected column
I just noticed that when I changed the number type of a specific column (say to Accounting style), the values displayed in the summary statistics window are still formatted generically.  Not a huge deal, but would be nice if those numbers were displayed using a similar number format that is assigned to the selected column.

naterush:
# Description

Closes #135 

# Testing

Check with all different column types, and different formats. Both the summary stats and the unique values tab.

# Documentation

None.


naterush (commented):
Looks good!

PR #152, opened 2022-03-10
mitosheet: make undo and redo work with shortcuts

### LINKED ISSUE #130: Undo key presses should work on Mito
**Describe the bug**

On Mac, command + z in Mito does not undo the most recent operation.

**Expected behavior**

On Mac: `command + z` in Mito should undo the most recent operation
On Windows: `control + z` should undo the most recent operation.

naterush:
# Description

Closes #130 

# Testing

Test on Mac. @aarondr77 can you also test on windows?

Also make sure to test with multiple mitosheets.

# Documentation

We should add this to the docs somewhere. Where?

### DIFF
--- a/mitosheet/src/components/toolbar/Toolbar.tsx
+++ b/mitosheet/src/components/toolbar/Toolbar.tsx
@@ -55,12 +55,14 @@ const Toolbar = (
         <div className='toolbar-container'>
             <div className='toolbar-left-half'>
                 <ToolbarButton
+                    id='mito-undo-button' // NOTE: this is used to click the undo button in plugin.tsx
                     toolbarButtonType={ToolbarButtonType.UNDO}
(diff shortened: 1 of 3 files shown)


naterush:
Note that these don't work sometimes, e.g. when you have the import taskpane open, because we don't handle focus super great as is. 

I think this is ok for now, given that this is an improvement over the previous. I opened an issue for focus improvement here: https://github.com/mito-ds/monorepo/issues/153

aarondr77:
Since this doesn't work with taskpanes open, it means it also won't work when the graph taskpane is open. I think we should wait to add more keyboard shortcuts until we resolve the the focus issue. Because I think this sort of feature, albeit its really small, just adds confusion. Nobody except us will ever understand when they can use the keyboard shortcuts and when they can't, and the tool will just seem broken.

naterush:
@aarondr77 I vote we merge this, but your call (and whatever you think is best).

PR #155, opened 2022-03-11
mitosheet: don't clear analysis if it doesn't exist

aarondr77:
# Description

Previously, if a user replayed an analysis that they didn't have access to in their .mito folder, it would completely clear all of their generated code. This isn't generally an issue for users, but I spoke with a user that is sharing notebooks back and forth via OneDrive and was concerned about deleting their colleague's work. 

Now, if the analysis doesn't exist, we just give an error that explains what is happening (along with the log you've already created) and we don't clear the analysis. 

# Testing

1. Create an analysis
2. Change the analysis ID
3. Rerun the analysis and notice that an error appears instead of clearing out the generated code. 

# Documentation

### DIFF
--- a/mitosheet/mitosheet/errors.py
+++ b/mitosheet/mitosheet/errors.py
@@ -377,6 +377,25 @@ def make_is_directory_error(file_name: str) -> MitoError:
     )
 
 
+def make_no_analysis_error(analysis_id: str, error_modal: bool=True) -> MitoError:
+    """
+    Helper function for creating a no_analysis_error.
+
+    Occurs when a user tries to replay an analysis that they don't have access to in 
+    their .mito folder
+    """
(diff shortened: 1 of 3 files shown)


naterush (approved):
Code looks good. Testing now.

naterush:
@aarondr77 can you explain to me why this doesn't overwrite the code cell when it did before? Is it because it doesn't change the analysis name on the backend or something? 

If so - can you document this ordering requirements so we don't break them in the future in a refactor? Then, LGTM - this is working great - thanks!

aarondr77:
The reason that this works is because we return before `steps_manager.steps = steps_manager.steps[:1]` which clears all of the steps. I'll document this in the comment.

PR #156, opened 2022-03-12
Make both undo and redo work with graphs

naterush:
# Description

Adds a custom hook that is useful for refreshing params, and uses it on the pivot and graphing task panes - so that both should work with undo and redo now.

Closes: 

# Testing

Undo and redo with pivot and graphing.

# Documentation

No.


naterush (commented):
Looks good!

PR #158, opened 2022-03-13
Graph bug fixes from usability test

naterush:
# Description

- [x] [[Init.py](http://init.py/)](http://Init.py) is missing in deployed version, bc of a merge conflict resolution I did badly, I think.
- [x] If you: import data, create a graph, make a pivot table, and then rerun the top cell, it ends up with the graph selected but the wrong sheet tab selected. We should fix this, obvi.
- [ ] The double message that gets sent out, which causes users to need to click undo twice to get rid of a newly created graph.

I don't have a good way of doing this one. We want:
1. To update the graph if the underlying data has been edited
2. To not update the graph when the user opens it otherwise.

I don't think both of these things are possible, without a backend implementation of "refresh_dependent_graphs" - which IMO would be really horribly annoying and not worth it. Thoughts?

- [ ] Jake felt like the loading indicator for the graph showed up a little bit late. Maybe we can make the loading indicator be a normal effect, and then the actual backend call in a debounced effect. ← I actually don’t think we should fix this, because it will lead to weird race conditions that are not good to debug at *all*.

# Testing

# Documentation

### DIFF
--- a/mitosheet/src/components/Mito.tsx
+++ b/mitosheet/src/components/Mito.tsx
@@ -213,7 +213,6 @@ export const Mito = (props: MitoProps): JSX.Element => {
             return {
                 ...prevUIState,
                 selectedSheetIndex: newSheetIndex,
-                selectedTabType: 'data'
             };
         })
 
--- a/mitosheet/src/components/Mito.tsx
+++ b/mitosheet/src/components/Mito.tsx
@@ -30,7 +30,7 @@ import LoadingIndicator from './LoadingIndicator';
 import ErrorModal from './modals/ErrorModal';
 import MitoAPI from '../api';
 import PivotTaskpane from './taskpanes/PivotTable/PivotTaskpane';
-import { EDITING_TASKPANES, TaskpaneType, FULLSCREEN_TASKPANES } from './taskpanes/taskpanes';
+import { EDITING_TASKPANES, TaskpaneType } from './taskpanes/taskpanes';
 import MergeTaskpane from './taskpanes/Merge/MergeTaskpane';
 import ControlPanelTaskpane, { ControlPanelTab } from './taskpanes/ControlPanel/ControlPanelTaskpane';
 import SignUpModal from './modals/SignupModal';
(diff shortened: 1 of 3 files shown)


mitosheet/src/components/Mito.tsx
```diff
@@ -569,25 +572,28 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
 
     const taskpaneClassNames = classNames({
         'mito-default-taskpane': !taskpaneOpen,
-        'mito-default-fullscreen-taskpane-open': fullscreenTaskpaneOpen,
+        'mito-default-fullscreen-taskpane-open': graphTaskpaneOpen,
         'mito-default-narrow-taskpane-open': narrowTaskpaneOpen,
     })
 
+
+    console.log("MITO", uiState.selectedTabType, taskpaneOpen, graphTaskpaneOpen, narrowTaskpaneOpen)
```
naterush, line 595: Remove

mitosheet/src/components/footer/Footer.tsx
```diff
@@ -32,6 +32,8 @@ function Footer(props: FooterProps): JSX.Element {
     const selectedGraphID = props.uiState.selectedGraphID
     const selectedTabType = props.uiState.selectedTabType
 
+    console.log("FOOTER", selectedTabType)
```
naterush, line 35: Remove

naterush (commented):
Cleanup

mitosheet/src/components/Mito.tsx
```diff
@@ -211,7 +212,8 @@ export const Mito = (props: MitoProps): JSX.Element => {
             
             return {
                 ...prevUIState,
-                selectedSheetIndex: newSheetIndex
+                selectedSheetIndex: newSheetIndex,
+                selectedTabType: 'data'
```
aarondr77, line 216: This is getting triggered when the graph message gets sent and as a result the variable `graphTaskpaneOpen` which is used to set the class on the taskpane is not getting set correctly since it now checks for `selectedTabType`.

aarondr77: https://user-images.githubusercontent.com/18709905/158092319-71f11e7a-d412-4e51-99e0-ef36ab36e624.mov

PR #159, opened 2022-03-14
Dogfood 3-14

naterush:
# Description

## Bugs fixed:

1. If you deleted a sheet that had graph 1 relying on it, while selecting graph 2, then the graphing taskpane would stay open. This is another manifestation of the weirdness around having to set the graphing taskpane to open or closed in two different ways, which is _not nice_. We'll clean this up over time, me thinks. @aarondr77 do we have an open issue for the graphing being a full sheet of it's own?

### Bug 1: double redo to get rid of graph creation

1. Create a new graph.
3. You have to press undo twice to get rid of it. 

As we've talked about, this is because we send a double message, which we do to refresh the graph in case the user changed the underlying data. The only real way I can see to fix it is to add a `refresh_dependent_graphs` to the backend, but I'd rather not do this for now - it's a huge code smell IMO.

### Bug 2:  redo graph creation broken with multiple graphs

This is a manifestation of the same behavior as above.

1. Create a graph 1. Edit it. 
4. Create a graph 2. Edit that. 
5. Undo until you get back to graph 1.
6. Try and press redo. It will not redo the graph 2 creation and edits.

This is because we send out a message when the graphing tab is changed, aka when we do step 3. This new edit event overwrites the steps that we have in the redo buffer.

In terms of fixes, this is the same issue that we have with updating dependent graphs. I'm wondering: do we think the tradeoff of graphs never being out of date is better than all of these bugs with undo / redo? I am not sure. Let me know your thoughts. I think refreshing dependent graphs is a whole thing I don't want to do currently - so let's just pick one of these bugs for now.

## NOTE: I don't think either of these bugs are a huge deal, just minor annoyances. I am down to deploy without fixing either of them, just wanted to document them here (and if we're keeping them in, let's open a low priority issue for them).

# Testing

# Documentation


naterush (commented):
Looks good!

aarondr77:
Here is the issue about moving the graph for a full sheet https://github.com/mito-ds/monorepo/issues/142

aarondr77:
> In terms of fixes, this is the same issue that we have with updating dependent graphs. I'm wondering: do we think the tradeoff of graphs never being out of date is better than all of these bugs with undo / redo? I am not sure. Let me know your thoughts. 

I think since one of our core design principles for graphing is that it should respect the fact that creating a graph is something that is often done in iteration with transforming the data, we should keep the graph refresh. For example, we expect people to create a graph of their data, notice that there is an outlier, apply a filter to the data to attempt to remove the outlier, and then open the graph again to check their work. This is much more natural if the graph automatically refreshes.

aarondr77:
Now that I'm on windows, I can see the little thing hovering here. I believe that its the index column! 

image

PR #161, opened 2022-03-14
Graphing tabs

aarondr77:
# Description

- Creates tabs in the graphing taskpane for setup, style, and export. 
- Adds the color property to the setup tab. 
- Makes graphing work with toggling fullscreen mode. 

Adding the color to graphs can effect the order of the items displayed. I'm not sure what the specific rule is for when the order changes and when it doesn't, however. I also can't find any way of turning off this behavior in Plotly. Instead, we would need to address it by letting the user explicitly pick a column and sort it by ascending or descending order, etc. I think we should usability test it amongst ourselves + look at the graphs that we want to support and see if we want to address it. 

Looking for some design thoughts: Currently, it doesn't do a great job of communicating the that you can only have one color, until you actually have that color applied and the ADD button is disabled. We could switch to just having a select, but I think 1) having lots of different types of input UI's on the screen makes the UI require more thought on how to interact with it. 2) The method of getting rid of the color when its just a select is different from getting rid of other columns (which is not great) + is harder to find, since the user needs to find the "none" button. I think selects are good when you have to have one option, and all of the options are the same type, ie: all column headers. 

To support all of the above, we could do this: 
Where, if None is selected, then the select just says none and there is no X button. But then if you do select one, it shrinks its width and makes room for the X icon. 

Screen Shot 2022-03-16 at 3 17 54 PM

Here is another option: When nothing is selected, it just is faint as to still prompt you what to do + the structure of the UI doesn't have to change when something is the column is added / removed. 

Screen Shot 2022-03-16 at 3 21 27 PM

I vote we just leave it symetric with the X and Y axis ones since the main thing we want to communicate is that all of these are columns that you can add to change the structure of the graph, and I think the symetry accomplishes that best. And I don't think that having the initial confusion about how many you can add is a big deal. **But want to hear your thoughts**

# Testing

# Documentation

### DIFF
--- a/mitosheet/css/elements/Select.css
+++ b/mitosheet/css/elements/Select.css
@@ -83,7 +83,7 @@
 */
 .select-disabled > .select-text {
     color: var(--mito-light-gray) !important;
-    border: none;
+    border-color: var(--mito-light-gray) !important;
 }
 .select-disabled > .select-dropdown-icon-container {
     background-color: var(--mito-light-gray) !important;


aarondr77:
When adding the color parameter to these graphs, it takes the graph out of being sorted in ascending order, and instead sorts them in ascending order within each of the groups.

With color: 
Screen Shot 2022-03-15 at 6 20 24 PM

Without color: 
Screen Shot 2022-03-15 at 6 20 45 PM

This makes the graph at lot less informative. Because it makes it more difficult to compare the different groups. We should try to address this, but it might require adding a 'sort' field.

aarondr77:
Sometimes the graph gets loaded to large and the toolbar gets cut off like this. See the "Export" Tab. 

Screen Shot 2022-03-15 at 6 26 03 PM

aarondr77:
Update the color selection UI to reflect that you can only have one selected at at time.

mitosheet/mitosheet/step_performers/column_steps/delete_column.py
```diff
@@ -82,7 +82,8 @@ def describe( # type: ignore
         df_names=None,
         **params
     ) -> str:
-        formated_column_ids = (', '.join(column_ids))
+        # Remove None from the list, and turn the column_ids into a string
+        formated_column_ids = (', '.join(list(filter(None, column_ids))))
```
aarondr77, line 86: Go back to the original code because we now handle this case in the api call

formated_column_ids = (', '.join(column_ids))

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -46,7 +46,8 @@ def execute_replay_analysis_update(
     # This is particularly important for users who received the notebook from a colleague.
     if analysis is None:
         log('replayed_nonexistant_analysis_failed')
-        raise make_no_analysis_error(analysis_name)
+        return
```
aarondr77, line 49: Ooops this wasn't supposed to be changed in this PR

aarondr77:
This should also add that it deletes all of the graphs

Screen Shot 2022-03-16 at 12 09 52 PM

mitosheet/src/components/taskpanes/Graph/GraphSidebarTabs.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const changeTab = (tab: GraphSidebarTab) => {
+        props.setSelectedGraphSidebarTab(tab)
+
+        // Send a log message saying that this tab has been clicked,
+        // and keeping the tab at the highest level
+        void props.mitoAPI.log('clicked_graph_' + tab + '_tab');
+    }
+
+    return (
+        <div className='graph-sidebar-tab-container'>
```
naterush, line 28: Can you open an issue for making a general Tabs component for all the tabs we have?

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -149,18 +148,22 @@ const GraphSidebar = (props: {
...
                 graphParams.graphCreation.graph_type,
                 graphParams.graphCreation.sheet_index,
+                graphParams.graphCreation.color,
                 graphParams.graphPreprocessing.safety_filter_turned_on_by_user,
                 graphParams.graphCreation.x_axis_column_ids,
                 graphParams.graphCreation.y_axis_column_ids,
-                `${boundingRect?.height - 10}px`, 
-                `${boundingRect?.width - 20}px`, // Subtract pixels from the height & width to account for padding
+                `${boundingRect?.height - 10}px`, // Subtract pixels from the height & width to account for padding
+                `${boundingRect?.width - 20 - 250}px`,
```
naterush, line 166: Where is this 250 coming from. Can you leave a comment?

mitosheet/src/components/taskpanes/Graph/GraphSetupTab.tsx
```diff
@@ -0,0 +1,337 @@
...
+                        A key is required because react does not know that the object x_axis_column_ids changed in all cases. 
+                        Particularly, when the user changes the x_axis_column_ids from [A, B, A] to [B, A] by 
+                        deleting the first A, React does not recognize that the change has occurred and so the Axis Section does 
+                        not update even though the graph updates.
+
+                        We append the indicator xAxis to the front of the list to ensure that both AxisSections have unique keys. 
+                        When the Axis Sections don't have unique keys, its possible for the sections to become duplicated as per 
+                        the React warnings.
+                    */
+                    key={['xAxis'].concat(props.graphParams.graphCreation.x_axis_column_ids).join('')}
```
naterush, line 245: @aarondr77 is this still true? If so, it's because we're not making good copies of things, and I expect that this can die...

Can you a) check if removing the key is working now, and b) figure out where we're updating things that we aren't making a copy of it correctly?

Not a big rush on this, but this is def a code smell, and that comment is nasty nasty. This sort of thing has led to bugs in the past, with things not updating properly (in the column control panel specifically).

mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -32,11 +32,7 @@ class GraphStepPerformer(StepPerformer):
             sheet_index: int
             x_axis_column_ids: ColumnID[],
             y_axis_column_ids: ColumnID[],
-            color: (optional) {type: 'variable', columnID: columnID} | {type: 'constant', color_mapping: Record<ColumnID, string>}
-
-            Note: Color is not in the styles object because it can either be a variable or a discrete color.
-            Color as a variable does not belong in the style object. However, I want to keep the variable
-            and constant color together so we can ensure through the type system that only one can be set at a time.
+            color: columnID: columnID
```
naterush, line 35: @aarondr77 so color is something that goes in graph creation! Cool.

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
```diff
@@ -108,6 +109,8 @@ def graph_creation(
     elif len(y_axis_column_headers) > 1:
         all_params["y"] = y_axis_column_headers
 
+    all_params['color'] = color_column_header
```
naterush, line 112: Wow. This is good design. Look at that change.

mitosheet/src/components/Mito.tsx
```diff
@@ -610,7 +610,7 @@ export const Mito = (props: MitoProps): JSX.Element => {
                     setUIState={setUIState}
                     sheetData={sheetDataArray[uiState.selectedSheetIndex]}
                 />
-                <div className="mito-main-sheet-div"> 
+                <div className="mito-main-sheet-div" id="mito-main-sheet-div">
```
naterush, line 613: Hmm. Why is this the thing we want to get the size of?

mitosheet/src/components/taskpanes/Graph/AxisSection.tsx
```diff
@@ -43,37 +42,20 @@ const AxisSection = (props: {
...
+                columnID={columnID}
+                columnIDsMap={props.columnIDsMap}
+                onChange={(columnID: string) => {
+                    props.updateAxisData(
+                        props.graphAxis,
+                        i,
+                        columnID
+                    )
+                }}
+                onXIconClick={() => props.updateAxisData(props.graphAxis, i)}
```
naterush, line 56: I think `onDelete` is a better name for this, perhaps.

mitosheet/src/components/taskpanes/Graph/ColumnCard.tsx
```diff
@@ -0,0 +1,51 @@
...
+import Select from '../../elements/Select';
+import Row from '../../spacing/Row';
+import Col from '../../spacing/Col';
+
+import '../../../../css/taskpanes/Graph/AxisSection.css'
+import { ColumnID, ColumnIDsMap } from '../../../types';
+import DropdownItem from '../../elements/DropdownItem';
+import { getDisplayColumnHeader } from '../../../utils/columnHeaders';
+
+const ColumnCard = (props: {
```
naterush, line 14: Leave a comment that this is just a select an an X-Icon.

Also, can we refactor this out of graph -> a component it's own right? And then use it for Pivot for the row and column sections?

In fact, I'm wondering if we already have a component for that here.

mitosheet/src/components/taskpanes/Graph/GraphExportTab.tsx
```diff
@@ -0,0 +1,44 @@
...
+import MitoAPI from '../../../api';
+import TextButton from '../../elements/TextButton';
+
+/* 
+    The export tab that lets the user copy the graph code or download as a png
+*/
+function GraphExportTab(
+    props: {
+        graphCodeCopied: boolean;
+        _copyGraphCode: () => void;
```
naterush, line 14: I think these are functions that we can push down to the export tab itself. Aka, we don't need these being defined in the highest level of the graphing taskpane anymore.

mitosheet/src/components/taskpanes/Graph/GraphSetupTab.tsx
```diff
@@ -0,0 +1,337 @@
...
+        if (columnID === undefined) {
+            axisColumnIDsCopy.splice(index, 1)
+        } else {
+            axisColumnIDsCopy[index] = columnID
+        }
+
+        // Update the axis data
+        if (graphAxis === GraphAxisType.X_AXIS) {
+            props.setGraphParams(prevGraphParams => {
+                const copyPrevGraphParams = {...prevGraphParams}
```
naterush, line 91: I think we're doing a double copy here... can we skip this?

mitosheet/src/components/taskpanes/Graph/GraphSetupTab.tsx
```diff
@@ -0,0 +1,337 @@
...
+                    ...copyPrevGraphParams,
+                    graphCreation: {
+                        ...copyPrevGraphParams.graphCreation, 
+                        x_axis_column_ids: axisColumnIDsCopy
+                    }
+                }
+            })
+        } else {
+            props.setGraphParams(prevGraphParams => {
+                const copyPrevGraphParams = {...prevGraphParams}
```
naterush, line 102: A double copy here. But also, it's not a nested copy... so not sure what the deal is here. 

Maybe we want to change both of these to:

```
props.setGraphParams(prevGraphparams => {
    return {
        graphCreation: {
             ... prevGraphparams,
             y_axis_column_ids: axisColumnIDsCopy
        }, 
        graphStyling: {
             ....prevGraphParams.graphStyling
        },
        ....
    }
}
```
Aka, given that I don't think it's making deep copies of the object, I think we actually want to go ahead and make them ourselves in the nested objects. 

The other option here (which is actually better, and what we should do!) is to create a deep copy of this object just with:

```
props.setGraphParams(prevGraphparams => {
    const graphParamsCopy = JSON.parse(JSON.stringify(graphParamsCopy)); 
    return {
        ...graphParamsCopy,
        graphCreation: {
             ... prevGraphparams,
             y_axis_column_ids: axisColumnIDsCopy
        }, 
    }
}
```

This is best for a few reasons:
1. It will probably fix the key issues I commented about already. This failure to make new objects is probably where that issue is occurring. 
2. It will scale as we continue to nest objects (e.g. for styling!). 

Note here: you need to make this change in a few places (namely, any place where we set graph params).

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -100,6 +94,11 @@ const GraphSidebar = (props: {
     */
     const [graphUpdatedNumber, setGraphUpdatedNumber] = useState(0)
 
+    useEffect(() => {
+        fscreen.addEventListener('fullscreenchange', () => {setGraphUpdatedNumber(old => old + 1)});
+        return () => fscreen.removeEventListener('fullscreenchange', () => setGraphUpdatedNumber(old => old + 1))
```
naterush, line 99: I think you need the cleanup function, but not the callback to the cleanup function, right?

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -149,18 +148,22 @@ const GraphSidebar = (props: {
         size.
     */
     const getGraphAsync = async () => {
-        const boundingRect: DOMRect | undefined = document.getElementById('graph-div')?.getBoundingClientRect();
+        // The reason that we use the mito-main-sheet-div instead of the graph-div is because the size of the graph div
```
naterush, line 151: but isn't the `Mito-main-sheet-div` small at this point? Namely, isn't it 0 width (or is that just the grid we shrink down...). 

Ohhh... it's the main container. I see what you mean then. Ok... I am ok with this!

naterush (changes requested):
Some cleanup and changes to the code, nothing major (beyond addressing one copying bug I'd like to ferret out, since it's a real code smell that will make maintaining this code hard long-term [as we get into styling]).

The other thing: I'm gonna take a shot at making an `useEffectOnElementResize`, using a resize observer. This will allow us to handle the graphing thing in a more general way than just the fullscreen mode, and it's also something we use in Endo as well! I'll do this tonight or tomorrow morning.

Gonna push back the release from tomorrow morning to after you can get this addressed. No rush. I don't think the day is super gonna be the make or break for the retention for the March cohort, and we aren't sending out a release for the Feb cohort till next week anyways.

mitosheet/src/components/taskpanes/Graph/GraphSidebarTabs.tsx
aarondr77: Opened: https://github.com/mito-ds/monorepo/issues/171

mitosheet/mitosheet/mito_analytics.py
```diff
@@ -251,8 +251,9 @@ def log(log_event: str, params: Dict[Any, Any]=None, steps_manager: StepsManager
                     private_params[key] = value
             elif 'graph_creation' in key:
                 # Don't log the column ids in the graph, just log the number of series graphed
-                private_params[key]['graph_creation']['x_axis_column_ids'] = len(private_params[key]['graph_creation']['x_axis_column_ids'])
-                private_params[key]['graph_creation']['y_axis_column_ids'] = len(private_params[key]['graph_creation']['y_axis_column_ids'])
+                private_params['params_x_axis_column_ids'] = len(value['x_axis_column_ids'])
+                private_params['params_y_axis_column_ids'] = len(value['y_axis_column_ids'])
+                private_params['params_color'] = 1 if 'color' in value.keys() else 0
```
naterush, line 256: Can we set this to True and False instead of 1 and 0?

aarondr77:
I don't think you can have a date column in the color section. We should disable selecting dates with an error message if so.

aarondr77:
I'm going to leave the Color as the Add thing. Although I think you're right that at some point we'll get feedback about the only being able to set one color, I think its way easier to have a X button than to have to search for the None option, especially because users have 70+ columns and the select is really full. 

I think this highlights that it would be nice to develop a new type of component at some point. I'm not sure what that component is like though. We need to do some research. A component that only lets you have one input at a time and is easily closable.

PR #165, opened 2022-03-15
Installer improvements

naterush:
# Description

Read the retro here: https://www.notion.so/trymito/New-Deployment-Pt-3-8a6df7780d794d57a736c44219f91f86

TLDR:
1. We don't know if installs are terminating or getting stuck (as @aarondr77 does with the cache)
2. Our upgrade success does appear to be trending upwards!

The fixes in this PR:
1. mitoinstaller: add optional log on success, so we can see if steps terminate (and their runtime).
3. mitoinstaller: check for user json more explicitly

# Testing

1. Make sure the new logs are generated for success on the `install_mitosheet_success`, and that gives runtime.
2. Make sure that the user.json is not overwritten when you run the installer.

# Documentation

Nope.

### DIFF
--- a/.github/workflows/test-mitoinstaller.yml
+++ b/.github/workflows/test-mitoinstaller.yml
@@ -16,6 +16,9 @@ jobs:
       matrix:
         python-version: [3.6, 3.8]
         os: [ubuntu-latest, windows-latest, macos-latest]
+        exclude:
+          - python-version: 3.6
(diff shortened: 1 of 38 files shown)


naterush (commented):
Looks good!

naterush:
This contains the stripping path PR, so that tests pass.

PR #170, opened 2022-03-16
Graph styling

aarondr77:
# Description

- Adds all of the style support (edit graph + axis titles, hide legend, hide range slider)
- Adds download as PNG button (however, the Plotly png quality is really poor. So @naterush thoughts on implementing this a different way where we can get a higher quality png?) 

# Testing

Use all of the different styling options. Make sure that the generated code matches!

# Documentation

Yes, stand by for improved documentation.


mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -32,11 +32,7 @@ class GraphStepPerformer(StepPerformer):
...
             x_axis_column_ids: ColumnID[],
             y_axis_column_ids: ColumnID[],
-            color: (optional) {type: 'variable', columnID: columnID} | {type: 'constant', color_mapping: Record<ColumnID, string>}
-
-            Note: Color is not in the styles object because it can either be a variable or a discrete color.
-            Color as a variable does not belong in the style object. However, I want to keep the variable
-            and constant color together so we can ensure through the type system that only one can be set at a time.
+            color: columnID: columnID
         },
         graph_styling: {},
```
naterush, line 37: Assuming there are params here now.

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
```diff
@@ -35,6 +35,13 @@
 # This must be kept in sync with GRAPH_SAFETY_FILTER_CUTOFF in GraphSidebar.tsx
 GRAPH_SAFETY_FILTER_CUTOFF = 1000
 
+RANGE_SLIDER = """dict(
```
naterush, line 38: RANGE_SLIDER_CODE perhaps, for clarity

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
```diff
@@ -189,27 +195,65 @@ def graph_creation_code(
...
+    fig: go.Figure, graph_type: str, column_headers: List[ColumnHeader], filtered: bool, graph_styling_params: Dict[str, Any]
 ) -> go.Figure:
     """
     Styles the Plotly express graph figure
     """
 
     # Create the parameters that we use to style the graph
-    graph_title = get_graph_title(column_headers, [], filtered, graph_type)
+    if graph_styling_params['title']['visible']:
+        graph_title = graph_styling_params['title']['title'] if 'title' in graph_styling_params['title'].keys() else get_graph_title(column_headers, [], filtered, graph_type)
```
naterush, line 206: You can just say `if 'title' in graph_styling_params['title']`, you don't need to do the explicit .keys() call.

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
```diff
@@ -189,27 +195,65 @@ def graph_creation_code(
...
+    else:
+        graph_title = None
+
+    # If the x axis title is visible, then either dispaly the user's custom title or Ploty's default title.
+    # Otherwise, get rid of the title
+    if graph_styling_params['xaxis']['visible']:
+        use_custom_xaxis_title = 'title' in graph_styling_params['xaxis'].keys()
```
naterush, line 213: no need for .keys()

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
```diff
@@ -189,27 +195,65 @@ def graph_creation_code(
...
+        # Only apply the xaxis_title if it is set because if we set it to None, then we don't get ploty's default values
+        if use_custom_xaxis_title:
+            fig.update_layout(
+                xaxis_title = graph_styling_params['xaxis']['title']
+            )
```
naterush, line 218: Can we not update this here, but rather build params as we do elsewhere? Otherwise, it's gonna become spaghetti code.

aarondr77:
We also need to add logging to make sure that we're obscurring the logs for Title and stuff.

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
```diff
@@ -189,27 +195,65 @@ def graph_creation_code(
...
+            xaxis_title = None
+        )
+
+    # If the y axis title is visible, then either dispaly the user's custom title or Ploty's default title.
+    # Otherwise, get rid of the title
+    if graph_styling_params['yaxis']['visible']:
+        use_custom_yaxis_title = 'title' in graph_styling_params['yaxis'].keys()
+        # Only apply the xaxis_title if it is set because if we set it to None, then we don't get ploty's default values
+        if use_custom_yaxis_title:
+            fig.update_layout(
```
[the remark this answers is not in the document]
naterush: Building a parameters dictionary, as you do with graph creation, as clearly the best bet here! I think it will lead to much more maintainable and clean code over time!

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -0,0 +1,261 @@
...
+                        onChange={(e) => {
+                            props.setGraphParams(prevGraphParams => {
+                                const graphParamsCopy: GraphParams = JSON.parse(JSON.stringify(prevGraphParams)); 
+                                return {
+                                    ...graphParamsCopy,
+                                    graphStyling: {
+                                        ...graphParamsCopy.graphStyling,
+                                        title: {
+                                            ...graphParamsCopy.graphStyling.title,
+                                            title: e.target.value !== '' ? e.target.value : undefined
```
naterush, line 46: You cannot use `e.target.value` inside of a callback. We've seen this as a sheet crashing bug a couple of times - generally, you have to a access events _immediately_ in react. This is hard to remember!

At the start of the onChange function, just add `const title = e.target.value` - and then use it below!

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -0,0 +1,261 @@
...
+                        onChange={(e) => {
+                            props.setGraphParams(prevGraphParams => {
+                                const graphParamsCopy: GraphParams = JSON.parse(JSON.stringify(prevGraphParams)); 
+                                return {
+                                    ...graphParamsCopy,
+                                    graphStyling: {
+                                        ...graphParamsCopy.graphStyling,
+                                        yaxis: {
+                                            ...graphParamsCopy.graphStyling.yaxis,
+                                            title: e.target.value !== '' ? e.target.value : undefined
```
naterush, line 190: And here

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
```diff
@@ -218,27 +262,52 @@ def graph_styling_code(
     # Create the params used to style the graph
     all_params: List[Tuple[str, str, bool]] = []
 
-    graph_title = get_graph_title(column_headers, [], filtered, graph_type)
-    all_params.append(("title", graph_title, True))
+    # Create the title param
+    if graph_styling_params['title']['visible']:
+        graph_title =  graph_styling_params['title']['title'] if 'title' in graph_styling_params['title'].keys() else get_graph_title(column_headers, [], filtered, graph_type)
+        all_params.append(("title", graph_title, True))
```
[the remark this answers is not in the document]
aarondr77: Plotly handles titles differently from the the axis labels, we only need to explicitly set the axis labels to None, for some reason. I'll document this.

mitosheet/src/components/taskpanes/Graph/GraphExportTab.tsx
```diff
@@ -0,0 +1,67 @@
...
+                        : "Copied!"
+                    }
+                </TextButton>
+            </div>
+            <div>
+                <TextButton
+                    variant='dark'
+                    onClick={() => {
+                        // Find the Plotly Download plot as png button, and then click it. 
+                        const node: Element | undefined | null = document.querySelector('[data-title="Download plot as a png"]')
```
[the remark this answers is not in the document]
aarondr77: Nice good call on the mitoContainerRef. 

the query selector returns a Element instead of  HTMLElement, which has the .click() function on it. So this approach lets us be type safe and not need to cast.

naterush:
# A bug-ish bug with deleting sheets and graphs

Discovered a real bug in graphing, although it's not present in this PR. TLDR: because we don't use DataframeID and rather use SheetIndex for detecting when we delete a dataframe that has a graph relying on it, you can easily delete data frames in an order that leads to graphs not getting deleted.

Opened a issue here:  https://github.com/mito-ds/monorepo/issues/176

naterush:
> Adds download as PNG button (however, the Plotly png quality is really poor.

Sure. I have a feeling (80% confidence) that there is no easy solution that will lead to a higher-quality PNG (if there was, Plotly probably would have taken it). 

Let's leave this change to the _very last_ thing in this PR. Only after we get all the review addressed, and this PR is finalized, then we can work on that if we want to. I don't want us spending 2 days on that before we've got everything else finished!

mitosheet/mitosheet/step_performers/graph_steps/graph_utils.py
```diff
@@ -52,7 +52,7 @@ def get_barmode(graph_type: str) -> Optional[str]:
         return None
 
 
-def create_parameter(param_key: str, param_value: str, format_as_string: bool) -> str:
+def create_parameter(param_key: str, param_value: Union[str, None], format_as_string: bool) -> str:
```
naterush, line 55: Optional[str]

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
```diff
@@ -192,56 +199,128 @@ def graph_creation_code(
...
+
+    # Create the barmode param
     barmode = get_barmode(graph_type)
+    if barmode is not None:
+        all_params['barmode'] = get_barmode(graph_type)
+
+    # Create the showlegend param
+    all_params['showlegend'] = graph_styling_params['showlegend']
+
+    # Create the range slider param
```
naterush, line 249: Can we move this up to the x-axis title area; just so they are all in the same place? Both in the generate graph code and transpile code, to keep them symmetric!

mitosheet/mitosheet/tests/test_utils.py
```diff
@@ -749,15 +749,32 @@ def generate_graph(
...
+                            'rangeslider': {
+                                'visible': True
+                            }
+                        },
+                        'yaxis': {
+                            'title': None,
+                            'visible': True
+                        },
+                        'showlegend': True
+                    },
```
naterush, line 777: Can we make these optional arguments to the call? And in turn, can you test that they work (e.g. you can set them without error, and that the are in the params of the resulting graph?

mitosheet/src/components/taskpanes/Graph/GraphExportTab.tsx
```diff
@@ -29,16 +30,38 @@ function GraphExportTab(
...
+                    variant='dark'
+                    onClick={() => {
+                        // Find the Plotly Download plot as png button, and then click it. 
+                        const node: Element | undefined | null = props.mitoContainerRef.current?.querySelector('[data-title="Download plot as a png"]')
+                        const clickEvent = new MouseEvent("click", {
+                            "view": window,
+                            "bubbles": true,
+                            "cancelable": false
+                        });
+                        node?.dispatchEvent(clickEvent)
```
naterush, line 57: This doesn't let you do .click for typing reasons. I think the following works:
```
const downloadLink: HTMLLinkElement | undefined | null = props.mitoContainerRef.current?.querySelector<HTMLLinkElement>('[data-title="Download plot as a png"]')
downloadLink?.click()
```

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -1,20 +1,264 @@
...
+                            Title
+                        </p>
+                    </Col>
+                    <Input 
+                        value={graphStylingParams.title.title || ''}
+                        placeholder="Default Graph Title"
+                        onChange={(e) => {
+                            props.setGraphParams(prevGraphParams => {
+                                const graphParamsCopy: GraphParams = JSON.parse(JSON.stringify(prevGraphParams)); 
+                                const newTitle =  e.target.value !== '' ? e.target.value : undefined
```
naterush, line 40: Can you leave a comment here that we leave it undefined so it doesn't think we're trying to set a custom axis title on the backend?

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -1,20 +1,264 @@
...
+                            X axis title
+                        </p>
+                    </Col>
+                    <Input 
+                        value={graphStylingParams.xaxis.title || ''}
+                        placeholder="Default X Axis"
+                        onChange={(e) => {
+                            props.setGraphParams(prevGraphParams => {
+                                const graphParamsCopy: GraphParams = JSON.parse(JSON.stringify(prevGraphParams)); 
+                                const newTitle =  e.target.value !== '' ? e.target.value : undefined
```
naterush, line 99: Same doc here

mitosheet/src/types.tsx
```diff
@@ -206,17 +206,29 @@ export type GraphCreationParams = {
     y_axis_column_ids: ColumnID[]
     color: ColumnID | undefined
 }
-export type GraphStylingParams = undefined
-export type GraphRenderingParams = {
-    width?: number
-    height?: number
+export type GraphStylingParams = {
+    title: {
+        title: string | undefined,
```
naterush, line 211: Comment about using the default when there is an undefined title, for clarity if you please

naterush (approved):
Code looks super great. Gonna test it manually (including upgrading), and then merge it if it looks good!

Will ask my rent to do a usability test first thing tomorrow, and will get the results of said test written up (and hopefully fixed by EOD tomorrow). 

Aka, let's aim for a release on Thursday, the 24th. Sound good?

PR #187, opened 2022-03-22
Background color selector for pro

naterush:
# Description

Adds a background color selector for pro. It works pretty well. Also fixes a bug where upgrading graph steps was broken, and would always reset your params. I'm cool w/ how the color selectors look (I think they look quite cool as is), but feel free to mess about with styling if you want something different.

The one thing here, there's a workaround for making sure the entire background is the right color, which ends up making the edges update before the graph does. I think this is fine (it's kinda like a little preview), and it's better than it staying white anyways.

This is only a pro feature.

# Testing

Test pro and non-pro mode. Test that your styling params get saved when you replay the analysis!

# Documentation

Not yet!


mitosheet/mitosheet/saved_analyses/step_upgraders/graph.py
```diff
@@ -98,3 +98,99 @@ def upgrade_graph_1_to_2(step: Dict[str, Any], later_steps: List[Dict[str, Any]]
...
+            },
+            graph_rendering: {
+                height: int representing the div width
+                width: int representing the div width
+            }
+        }
+    }
+    """
+    params = step['params']
+    print(params)
```
naterush, line 188: Remove

mitosheet/mitosheet/tests/test_utils.py
```diff
@@ -781,7 +782,8 @@ def generate_graph(
                             'title': yaxis_title,
                             'visible': yaxis_visible
                         },
-                        'showlegend': showlegend
+                        'showlegend': showlegend,
+                        'paper_bgcolor': paper_bgcolor
```
naterush, line 786: Add to test!

mitosheet/src/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -186,7 +187,15 @@ const GraphSidebar = (props: {
     } else {
         return (
             <div className='graph-sidebar-div'>
-                <div className='graph-sidebar-graph-div' id='graph-div' >
+                <div 
+                    className='graph-sidebar-graph-div' 
+                    id='graph-div'
+                    // Because we have padding on this div, but we want the graph to appear
+                    // to take up the whole screen, we also style it with the background color
```
naterush, line 194: Note the minor bug here

mitosheet/src/utils/actions.tsx
```diff
@@ -382,6 +382,7 @@ export const createActions = (
 
                 const newGraphID = getRandomId() // Create a new GraphID
                 const graphParams = getDefaultGraphParams(sheetDataArray, sheetIndex)
+                console.log("PARAMS", graphParams)
```
naterush, line 385: Remove

naterush (commented):
Cleanup, test

aarondr77 (commented):
Yay for the start of graph styling! Everything is working well. No bugs found. 

In my use of this feature, the first thing that I did was change the background color of the graph to bright green. I immediately noticed two things 1) I couldn't read the the font anymore 2) There was no way of resetting the color. 

### Can't read the font
I think this feature is going to create the desire for changing the font color of the title + axis labels and gridline color of the graphs as well. Since it seems like that would only be about max 30 minutes of work, I'd be down to make that change! I think it will make the current feature valuable, whereas its pretty limitted right now (especially since the main graph colors we've seen are either default or 'dark mode'. Currently, you can't do a dark mode since the font becomes unreadable which means people will probably not even use this feature. 

Proposal 1: 
- For simplicity of implementation and so that we don't have a bunch of upgrade to pro buttons, we can change the `background` title section to `color` and put all of the colors there. We should reevaluate when we want to add more coloring features at some point in the future. 

Proposal 2:
- Put the colors in the Title, X and Y axis sections for pro users. If they are a free user, then instead of having the the title be `background` switch it to `colors` and say 'Want to set the color of the background, titles, and axis labels? Upgrade to pro."
- But don't have the pro call to action in every section. 
- I feel like this one is the best! 

### Can't get back to original color
I don't think this one is a huge deal since they can just reset it to white. But since people don't know the default plotly color, but probably really want their graphs to match (I def would care about this probably even if it was just a graph I was sharing with you), then they might end up going in and setting the color of every graph. 

Anyways, I don't think we need to worry about it now though.

aarondr77:
I had a weird bug where I couldn't open the background color selector, but I could open the plot color selector. I thought the color selector might've just been already open and hidden behind the browser, but as you can see in the video, it wasn't. I'm not sure what the problem is, but I can't replicate it. I guess something to keep an eye out for. 

https://user-images.githubusercontent.com/18709905/159598493-c1fd5aac-8816-4d2b-bedd-0d2377b066bb.mov

aarondr77:
Last thing, I had no idea what the difference between the `background color` and `plot background color` was initially. Could we maybe change it to `sheet background color` and `plot background color`? Or we could borrow Plotly's language and go with `paper background color`. 

And could we swap the order of the two? I think it makes sense for the plot background color to come first.

mitosheet/css/elements/ColorInput.css
```diff
@@ -0,0 +1,18 @@
+.color-input {
```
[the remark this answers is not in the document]
naterush: I think it's fine. I moved the TS for color selection to the /pro folder though, thanks for the reminder!

naterush:
@aarondr77 the big new cool thing in this PR now: it was getting very annoying to have to update how we created a param dictionary in two places for graph styling. 

So, I wrote a utility that takes a dictionary of params and transpiles it! It was a bit challenging, but I think it will pay off leaps and bounds going forward. I am pretty proud of this utility - I think it's really sweet!

Adding a new styling param is now:
1. Add it to the type on the frontend, display it. (5 minutes, depending if we have input type)
2. Add it to the styling param on the backend, in a single place. (5 minutes. Used to be like 15 when in multiple places)
3. Write an upgrader (3 minutes)
4. Update tests (10 minutes)

It should be around 30 minutes, but notably, if you're adding multiple at once, then these things don't scale linearly. So we should aim to add them in chunks, a few at a time.

aarondr77:
We got to remove this from the open source version

Screen Shot 2022-03-23 at 10 47 58 AM

PR #188, opened 2022-03-22
mitosheet: cleanup and modernize pivot code, thank god

### LINKED ISSUE #175: Cleanup pivot taskpane code
This is just a developer improvement. The code is incredibly old and messy, and needs to be refactored. I tried to refactor it and failed, because there's a whole lot of complexity in there!

naterush:
# Description

Closes #175 

# Testing

Should be identical in behavior. Test with editing, opening edited tables, etc!

# Documentation

No.

### DIFF
--- a/mitosheet/mitosheet/tests/test_ts_types.py
+++ b/mitosheet/mitosheet/tests/test_ts_types.py
@@ -136,7 +136,7 @@ def test_filter_conditions_match():
 
 def test_pivot_aggregation_functions_match():
     aggregation_types = get_enum_from_ts_file(
-        "./src/components/taskpanes/PivotTable/PivotTaskpane.tsx", "AggregationType"
+        "./src/types.tsx", "AggregationType"
     )
     assert set(aggregation_types.values()) == set(PIVOT_AGGREGATION_TYPES)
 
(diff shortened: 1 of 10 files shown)


aarondr77:
I got this sheet crashing bug. 

Steps to reproduce
1. Import data
2. Duplicate that data
3. While having the duplicated dataframe selected, click the pivot button
4. Undo twice

Screen Shot 2022-03-24 at 11 48 19 AM

aarondr77:
In the spirit of tying refactors to improvements to the tool. Do you want to implement this small, but really helpful pivot table improvement? https://github.com/mito-ds/monorepo/issues/121

aarondr77:
Other than that, everything seems to be working great!

aarondr77:
Bug in reading the pivot table params. Steps to reproduce

1. Import data
2. Pivot on that data
3. Close the pivot table taskpane
4. Go back to the original data
5. Click on the pivot table button in the toolbar
6. Notice that the params from the previous pivot table you created are displayed, but they shouldn't be! 

https://user-images.githubusercontent.com/18709905/160303604-b5212f17-e80f-4cfa-a4ca-4a1a58fdffe0.mov

naterush:
@aarondr77 yeppers good bug report. Will get this fixed up. Notably, this bug is something that is currently in the wild. I noted this twice in the video watching from last week. Will be good to get it fixed.

aarondr77:
Nice! I guess this is another example of the fact that we probably want to shake our error squashing activity up a bit since we didn't catch these.

naterush:
@aarondr77 if you have suggestions on how to improve the error reduction process, please do open up an issue and label it ideation. I'd like to see them, and would be happy to chat more there!

One thing I will note quickly: the current "error reduction process" is not the _only_ error reduction process we run. It's meant to target a very specific type of error: those that we have explicit visibility into through our logs. Through usability tests and workflows, we target other things too. 

Given that we're not stopping on either usability tests (or videos), workflows ourselves, and dogfooding, I think the targeting specific error logs we can see from real users is the one big missing piece that the "error reduction process" steps up to fill. 

Feel free to suggest changes to any/all of them, obviously!

PR #200, opened 2022-03-28
mitosheet: fix deployment by moving away from setuptools cli

aarondr77:
# Description

Deployment isn't working right now. This is the error message that we're receiving:

> 'setup.py:45: DeprecatedWarning: ensure_python is deprecated as of 0.7 and will be removed in 1.0. Use `setuptools` `python_requires` instead\n  ensure_python(\'>=3.4\')\nsetup.py:168: DeprecatedWarning: create_cmdclass is deprecated as of 0.8 and will be removed in 1.0. "\nUse `wrap_installers` to handle prebuild steps in cmdclass.\nUse `get_data_files` to handle data files.\nUse `include_package_data=True` and `MANIFEST.in` for package data.\n\n  cmdclass = create_cmdclass("jsdeps",\nsetup.py:174: DeprecatedWarning: install_npm is deprecated as of 0.8 and will be removed in 1.0. Use `npm_builder` and `wrap_installers`\n  install_npm(HERE, build_cmd="build:prod", npm=["jlpm"]),\n/home/runner/work/monorepo/monorepo/mitosheet/venv/lib/python3.8/site-packages/setuptools/dist.py:757: UserWarning: Usage of dash-separated \'description-file\' will not be supported in future versions. Please use the underscore name \'description_file\' instead\n  warnings.warn(\n/home/runner/work/monorepo/monorepo/mitosheet/venv/lib/python3.8/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.\n  warnings.warn(\nUpload failed (400): Invalid value for blake2_256_digest. Error: Use a valid, hex-encoded, BLAKE2 message digest.\nerror: Upload failed (400): Invalid value for blake2_256_digest. Error: Use a valid, hex-encoded, BLAKE2 message digest.\n')

I followed a bunch of github issues and eventually found [this forum post](https://bugs.python.org/issue45590) that explains that "Distutils is deprecated and should no longer be used." It lead me to [this article](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) so I made the changes that they suggested. I'm not sure that they are the correct changes, but I wanted to give it a try!

Feel free to use this as a helpful starting point or to discard it. 

# Testing

# Documentation


naterush:
@aarondr77 thanks for taking a shot at fixing deployment!
 
This command is invoking a Python module named build, but we don't have a build dependency! So these commands are not going to run at all.

We could add the `build` dependency, but we need to use the setup.py file as we're currently setup to handle different package versions within that file. Check out the setup.py to see some details of what it does...

I'll try and fix up this issue. We should just be able to use an old version of setuptools / distutils, and get it to work with that....

PR #208, opened 2022-03-30
mitosheet: only display the first 1500 columns

aarondr77:
# Description

- Only displays the first 1500 column in the pandas dataframe to make Mito snappier! 
- Adds logging for the pandas performance warning that occurs when creating a really large pivot table, so that we can figure out how much guidance is appropriate to provide to users. 
- Dramatically improves the efficiency of building the column evaluation graph by short circuiting if the column has no formula!

# Testing

1. Create a pivot table that has over 1500 columns, but not enough data to crash the sheet -- you can achieve this by adding a column and a value, with no row. See that the pivot table loads quickly and that if you rename, delete, or set the formula of a column, the operation finishes in about 30-60 seconds.  

2. Create a new pivot table with a row and column that each have a large number of unique values. Check your mixpanel logs to see a `pivot_table_performance_warning`. Notice that eventually the kernel will shut down and the log for the pivot table will never get logged. 

3. Make sure the column evaluation graph still works properly

# Documentation

### DIFF
--- a/mitosheet/mitosheet/step_performers/pivot.py
+++ b/mitosheet/mitosheet/step_performers/pivot.py
@@ -4,13 +4,17 @@
 # Copyright (c) Saga Inc.
 # Distributed under the terms of the GPL License.
 from copy import copy
+import json
+import sys
 from time import perf_counter
 from typing import Any, Callable, Dict, Collection, List, Optional, Set, Tuple
 import pandas as pd
+import warnings
 
 from mitosheet.column_headers import flatten_column_header
 from mitosheet.errors import (make_invalid_aggregation_error,
                               make_invalid_pivot_error, make_no_column_error)
+from mitosheet.mito_analytics import log
 from mitosheet.state import DATAFRAME_SOURCE_PIVOTED, State
 from mitosheet.step_performers.step_performer import StepPerformer
 from pandas.core.base import DataError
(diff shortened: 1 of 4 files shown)


aarondr77:
To make setting a column formula, renaming a column and deleting a column faster on datasets with a large number of columns, we need to stop the creation of the column evaluation graph from taking so long. Currently its O(N^2) because for each column, we check if its column formula references any other column. 

### Proposal 1: Use Steps
Use the steps to figure out if there are any column formula on the sheet that we're looking at. If there are not set column formula steps that exist on the sheet index we're looking at, then we know there are no column dependencies! 

However, we don't have access to the steps via the state object. So we don't have enough information to perform this proposal 

### Proposal 2: More efficiently build the column evaluation graph. 
Currently, we parse the formula of every column, no matter if its a data column or a formula column. We can make this much quicker by short circuiting on data columns!

mitosheet/mitosheet/step_performers/pivot.py
```diff
@@ -367,3 +373,13 @@ def get_new_pivot_df_name(post_state: State, sheet_index: int) -> str:
...
+
+
+def log_pivot_table_warnings(message, *args):
+    """
+    Logs warnings that are created by the pandas pivot table function.
+    Using warnings.showwarning passes several additional arguments to the function, 
+    so we must include *args to support them.
+    """
+    params = {'message': str(message)}
+    log('pivot_table_performance_warning', params)
```
naterush, line 385: TODO: test that I can generate this log!

naterush (approved):
Ok, just want to test that log, then LGTM!

These performance gains are pretty awesome, very happy we took it this direction!

PR #210, opened 2022-04-03
mitosheet: make drop duplicates start with no columns selected

aarondr77:
# Description

Removes the confusion caused by drop duplicates automatically deduplicating the data when the taskpane is open by starting the taskpane with no columns selected. This means that no data is dropped until the user configured the taskpane + they have a better idea of how to get back to a state where no duplicates are dropped. 

# Testing

# Documentation

No.

### DIFF
--- a/mitosheet/src/components/taskpanes/DropDuplicates/DropDuplicates.tsx
+++ b/mitosheet/src/components/taskpanes/DropDuplicates/DropDuplicates.tsx
@@ -48,7 +48,7 @@ export const getDefaultParams = (selectedSheetIndex: number, sheetDataArray: She
 
     return {
         sheet_index: selectedSheetIndex,
-        column_ids: sheetDataArray[selectedSheetIndex]?.data?.map(c => c.columnID) || [],
+        column_ids: [],
         keep: 'first',
     }
 }


naterush (changes requested):
The issues I've observed with drop duplicates are not just that when you open it, it drops rows, but also that just opening it performs some operation that you don't understand (and don't even really get that it was applied).

This change fixes the first problem, but not the second -- and in turn I think this might actually be a 10% solution that only fixes 10% of the problem. I vote we take a bit more of a thorough approach to this problem - I am interested in considering the role of live updating vs. non-live updating, generally here. 

Thoughts on this? If you want to merge this, I am fine with it - but I don't think it'll lead to the drop duplicate confusion issues being solved (and I do think they are worth solving!).

aarondr77:
I just made it so that now on initial opening of the drop duplicates no message is sent. I think that addresses your feedback. But I'm not sure. Lmk if I'm missing something. 

Currently, its still live-updating, but unlike Merge, it doesn't make any changes on the initial opening of the Drop duplicates taskpane. The solution is that the user needs to take an actual action to send a message.

naterush:
I retract my comment. I think that your original solution is fine.

Can get rid of the last commit before merging? I don't love that parameter, and every time we go and add a large number of boolean toggles to an interface it ends up getting pretty confusing (see our `DropdownItem` interface for an example of this). 

I'd rather not take on the debt - so let's just undo this last commit and then LGTM!

PR #219, opened 2022-04-05
Communicate Result

aarondr77:
# Description

What are your thoughts on when the original number of rows variable in the filter taskpane should update now that you see it? I'm leaning towards ideally it would update when you either 1) open the taskpane or 2) add/remove a filter condition. 

However, there's a lot more complexity that comes with condition 2 that I'm not sure is worth it. Let me know your thoughts. 

# Testing

Concat files together and make sure that if they have all the same column headers, then the text says they are all included. If a column header is not shared, make sure it shows up in the text. 

Filter and make sure the number of values that it says are removed is correct. 

# Documentation

No.

### DIFF
--- a/mitosheet/src/components/taskpanes/Concat/ConcatTaskpane.tsx
+++ b/mitosheet/src/components/taskpanes/Concat/ConcatTaskpane.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from "react";
 import MitoAPI from "../../../api";
 import useSyncedParams from "../../../hooks/useSyncedParams";
-import { AnalysisData, ConcatParams, SheetData, StepType, UIState } from "../../../types"
+import { AnalysisData, ColumnHeader, ConcatParams, SheetData, StepType, UIState } from "../../../types"
 import DropdownButton from "../../elements/DropdownButton";
 import DropdownItem from "../../elements/DropdownItem";
 import Select from "../../elements/Select";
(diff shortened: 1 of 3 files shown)


mitosheet/src/components/taskpanes/Concat/ConcatTaskpane.tsx
```diff
@@ -44,46 +45,63 @@ const ConcatTaskpane = (props: ConcatTaskpaneProps): JSX.Element => {
     // Make sure the user cannot select the newly created dataframe
     const [selectableSheetIndexes] = useState(props.sheetDataArray.map((sd, index) => index));
 
+    // For each sheet concatonated together, find all of the columns that are not included in the final result
+    const notIncludedColumnsArray = params?.sheet_indexes.map(sheetIndex => {
+        return Object.values(props.sheetDataArray[sheetIndex].columnIDsMap).filter(columnHeader => {
```
naterush, line 50: `props.sheetDataArray[sheetIndex].columnIDsMap` to `props.sheetDataArray[sheetIndex]?.columnIDsMap || {}` for safety.

naterush: Want to make this Object.values call just once and save it in a variable? Right now, it's O(n^3), methinks in the number of columns, at least - given all the copying.

mitosheet/src/components/taskpanes/Concat/ConcatTaskpane.tsx
```diff
@@ -44,46 +45,63 @@ const ConcatTaskpane = (props: ConcatTaskpaneProps): JSX.Element => {
...
+                        })
+                    }}
+                    selectableValues={Object.keys(props.sheetDataArray)} // Note the cast to strings
+                />
+                {notIncludedColumnsArray !== undefined &&
+                    <Row className='text-subtext-1' >
+                        {notIncludedColumnsArray[arrIndex].length === 0 ? `\u2713 All columns are included in the concatenated sheet.` : 
+                            notIncludedColumnsArray[arrIndex].length < 4 ? `Columns ${notIncludedColumnsArray[arrIndex].join(', ')} are not included.` :
+                                `Columns ${notIncludedColumnsArray[arrIndex].slice(0,3).join(', ')} and ${notIncludedColumnsArray[arrIndex].length - 3} others are not included.`
                         }
```
naterush, line 99: Want to break this out into a little utility?

mitosheet/src/components/taskpanes/ControlPanel/ControlPanelTaskpane.tsx
```diff
@@ -58,6 +58,8 @@ export const ControlPanelTaskpane = (props: ControlPanelTaskpaneProps): JSX.Elem
     const [operator, setOperator] = useState(columnFilters !== undefined ? columnFilters.operator : 'And');
     const [stepID, setStepID] = useState('');
 
+    const originalNumRows = useRef(props.sheetData?.numRows || 0);
+    const editedFilter = useRef(false)
```
naterush, line 62: Why are these refs vs. state variables? I think state variables are much safer, as refs don't trigger rerenders automatically.

If you don't need to set the state (e.g. originalNumRows), then you can use the following syntax:
```
const [originalNumRows, ] = useState(props.sheetData?.numRows || 0)
```

naterush:
Screen Shot 2022-04-06 at 3 30 43 PM

Can we make the color of the column headers slightly different? Was hard for me to parse through them at first.

naterush:
In the tooltip, can we say "index information" instead of "indexing information"?

naterush:
Screen Shot 2022-04-06 at 3 32 41 PM

Long column names end up displaying weird. Can we chop them off? Or only show one? 

I think it's perhaps a better strategy to show at most X characters (however many can fit per ur tests), and then chop it there. If you break that code out into a utility, this would actually be a pretty sweet utility to have!

naterush (commented):
Code mostly looking great. Some minor questions and comments. Feel free to merge once these things are fixed up and tested well!

PR #220, opened 2022-04-06
Code chunk

naterush:
# Description

This large PR adds new infrastructure that allows us to do code optimization - and also starts doing code optimization.

In practice, the optimization that we're doing currently is just the adding columns, renaming columns, deleting columns code, setting formulas, and setting cell values code chunks. With the exception of adding columns chunks (which are tough to optimize, although I have some ideas for the future) - things are much improved here! 

Some test cases that used to be like 10 lines of code are literally 0 now. Generally, the experience of working with a single dataframe is much improved!

In doing so, I moved step titles and step descriptions into code chunk comments and descriptions. This is very nice - it leads to a much better ability to actually create informative comments, if we so want!

However, we need to decide what to do with the step list that is displayed in the frontend. Namely - how do we want to handle step checkout? Namely, because we no longer link the steps to specific code chunks, we don't have a notion of what step a code chunk corresponds to. 

Thoughts? We can remove it, and probably avoid 200 lines of code and complexity, or add it back. Let me know what you think.

The other thing to figure out: pro or no?

# Testing

The main beneficial change would be to add test cases! You can manually test functionality, but tests here are the most beneficial thing as they will test the generated code for all of eternity. 

I wrote a bunch more tests, but try and figure out more interesting interactions to test!

# Documentation

We probably want to add documentation on this!


mitosheet/mitosheet/step_performers/graph_steps/graph_delete.py
```diff
@@ -55,26 +53,18 @@ def execute( # type: ignore
...
-        cls,
-        graph_id: GraphID,
-        old_graph_tab_name: str,
-        df_names=None,
-        **params
-    ) -> str:
-        return f'Deleted {old_graph_tab_name}'
+        # Graph steps don't add any generated code to the analysis script. 
+        return [
+            EmptyCodeChunk(prev_state, post_state, params, execution_data)
```
naterush, line 66: TODO: maybe I want to make it so the title and name are part of the empty code chunk, so you can still see it in the description of what occurred?

mitosheet/mitosheet/step_performers/column_steps/set_column_formula.py
```diff
@@ -315,7 +297,7 @@ def transpile_dependant_columns(
         post_state: State, 
         sheet_index: int, 
         column_id: ColumnID
-    ) -> List[str]: 
+    ) -> CodeChunk:
     """
     Use this helper function when making a change to a column and you want to transpile
```
naterush, line 302: TODO: change what this says!

mitosheet/mitosheet/step_performers/column_steps/rename_column.py
```diff
@@ -78,45 +75,31 @@ def execute( # type: ignore
...
+            return [EmptyCodeChunk(prev_state, post_state, {}, {})]
+
+
+        return [
+            RenameColumnsCodeChunk(
+                prev_state, 
+                post_state, 
+                {
+                    'sheet_index': params['sheet_index'],
+                    'column_ids_to_new_column_headers': {
```
naterush, line 96: TODO: comment this

mitosheet/mitosheet/code_chunks/step_performers/sort_code_chunk.py
```diff
@@ -0,0 +1,38 @@
...
+
+        df_name = self.post_state.df_names[sheet_index]
+        column_header = self.post_state.column_ids.get_column_header_by_id(sheet_index, column_id)
+        transpiled_column_header = column_header_to_transpiled_code(column_header)
+        
+        na_position_string = 'first' if sort_direction == ASCENDING else 'last'
+        
+        return [
+            f'{df_name} = {df_name}.sort_values(by={transpiled_column_header}, ascending={sort_direction == ASCENDING}, na_position=\'{na_position_string}\')', 
+        ]
```
naterush, line 38: TODO: add the delete utility here, just cuz we can!

mitosheet/mitosheet/code_chunks/step_performers/merge_code_chunk.py
```diff
@@ -0,0 +1,101 @@
...
+            merge_code.append(
+                f'{df_new_name} = {df_two_to_merge}.copy(deep=True)[~{df_two_to_merge}["{merge_key_two}"].isin({df_one_to_merge}["{merge_key_one}"])]'
+            )
+        else:      
+            merge_code.append(
+                f'{df_new_name} = {df_one_to_merge}.merge({df_two_to_merge}, left_on=[{column_header_to_transpiled_code(merge_key_one)}], right_on=[{column_header_to_transpiled_code(merge_key_two)}], how=\'{how_to_use}\', suffixes=[\'_{suffix_one}\', \'_{suffix_two}\'])'
+            )
+
+        # And then return it
+        return merge_code
```
naterush, line 101: TODO: do we want to do deletes here as well?

mitosheet/mitosheet/code_chunks/step_performers/drop_duplicates_code_chunk.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
```
naterush, line 1: default:

def delete_if_combined_right_with_delete_dataframe_code_chunk_for_sheet_index(sheet_index: int):
      return False
      
Then, steps can overwrite this!

mitosheet/mitosheet/code_chunks/step_performers/dataframe_steps/dataframe_delete_code_chunk.py
```diff
@@ -0,0 +1,24 @@
...
+    
+    def get_description_comment(self) -> str:
+        sheet_index = self.get_param('sheet_index')
+        df_name = self.prev_state.df_names[sheet_index]
+        return f'Deleted {df_name}'
+
+    def get_code(self) -> List[str]:
+        old_dataframe_name = self.get_param('old_dataframe_name')
+
+        return [f'del {old_dataframe_name}']
```
naterush, line 24: It would be nice to have a combine left function as well as a combine right function - as this would make the expression of the `DataframeDeleteCodeChunk` much more natural - we could specify it in a single place, rather than everywhere!

But does this complicate things and make things too hard to reason about? First you do a right optimize path? Then you do a left optimize pass.

mitosheet/mitosheet/code_chunks/step_performers/column_steps/set_column_formula_code_chunk.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+from typing import List
+
+from mitosheet.code_chunks.code_chunk import CodeChunk
+from mitosheet.evaluation_graph_utils import topological_sort_dependent_columns
+from mitosheet.parser import parse_formula
+
+
+class RefreshDependantColumnsCodeChunk(CodeChunk):
```
naterush, line 15: TODO: should we rename this?

mitosheet/mitosheet/code_chunks/step_performers/column_steps/reorder_column_code_chunk.py
```diff
@@ -0,0 +1,45 @@
...
+
+from typing import List
+from mitosheet.code_chunks.code_chunk import CodeChunk
+from mitosheet.transpiler.transpile_utils import \
+    column_header_to_transpiled_code
+
+
+class ReorderColumnCodeChunk(CodeChunk):
+
+    def get_display_name(self) -> str:
```
naterush, line 15: TODO: do we want to specify required params and required execution data?

naterush:
Screen Shot 2022-04-06 at 3 30 43 PM

Can we make the color of the column headers slightly different? Was hard for me to parse through them at first.

mitosheet/mitosheet/code_chunks/add_column_set_formula_code_chunk.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if added_column_id in deleted_column_ids and len(deleted_column_ids) == 1:
+            return NoOpCodeChunk(
+                self.prev_state, 
+                other_code_chunk.post_state, 
+                {},
+                other_code_chunk.execution_data # TODO: this is out of date, but we don't use it!
+            )
+        elif added_column_id in deleted_column_ids:
+            print("added_column_id in deleted_column_ids", added_column_id in deleted_column_ids)
```
aarondr77, line 69: remove

mitosheet/mitosheet/code_chunks/code_chunk.py
```diff
@@ -0,0 +1,91 @@
...
+        execution_data: Optional[Dict[str, Any]]
+    ):
+        self.prev_state = prev_state
+        self.post_state = post_state
+        self.params = params
+        self.execution_data = execution_data
+
+    def get_display_name(self) -> str:
+        """Returns a short name to display for this CodeChunk"""
+        raise NotImplementedError('Implement in subclass')
```
aarondr77, line 40: TODO: Understand where this gets displayed. Does the step history not update too? My assumption was that the step history taskpane was going to still show individual steps.

mitosheet/mitosheet/code_chunks/code_chunk.py
```diff
@@ -0,0 +1,91 @@
...
+        Gets a value from the execution_data passed to this CodeChunk
+        """
+        if self.execution_data and key in self.execution_data:
+            return self.execution_data[key]
+        return None
+
+    def params_match(self, other_code_chunk: "CodeChunk", param_keys: List[str]) -> bool:
+        """
+        Given a different code chunk, and a list of keys to check, returns True if
+        all the give keys match in the params. A useful utility for checking if
```
aarondr77, line 73: give -> given

mitosheet/mitosheet/code_chunks/code_chunk.py
```diff
@@ -0,0 +1,91 @@
...
+    State = Any
+
+class CodeChunk:
+    """
+    A CodeChunk is the a abstract base class that can be inhereited from
+    to create a code generating object. For example, a AddColumnCodeChunk
+    with the correct parameters and states will generate code that adds
+    a column with a specific column header to a specific dataframe.
+
+    When a step is transpiled, it generates a list of code chunks. In turn,
```
aarondr77, line 22: TODO: Understand why it would generate a step would generate a list of code chunks instead of just a single code chunk per step. And then an analysis would generate a list of code chunks

mitosheet/mitosheet/code_chunks/code_chunk_utils.py
```diff
@@ -0,0 +1,142 @@
...
+            optimized = True
+            all_code_chunks_reversed.append(combined_chunk)
+        else:
+            # If we cannot combine the two chunks, take the first chunk, 
+            # and reset the second one to the new first chunk for the next loop
+            code_chunks_list.append(first_code_chunk)
+            all_code_chunks_reversed.append(second_code_chunk)            
+    
+    # Make sure we take the final item in the code chunks list, 
+    # as it has nothing ot combine_right with
```
aarondr77, line 53: ot -> to

mitosheet/mitosheet/code_chunks/code_chunk_utils.py
```diff
@@ -0,0 +1,142 @@
...
+    if len(all_code_chunks_reversed) == 1:
+        code_chunks_list.append(all_code_chunks_reversed[-1])
```
aarondr77, line 55: It seems simpler to me to do all_code_chunks_reversed[0] since what we're doing is checking that only one element exists, we can just take that one element. Its not about the element being the last one in the array, its that its the only one. But nbd to me

mitosheet/mitosheet/code_chunks/code_chunk_utils.py
```diff
@@ -0,0 +1,142 @@
...
+
+    As such, we have a general utility we can apply to apply this optimization
+    to a variety of steps, including: setting cell values, setting column formulas.
+
+    NOTE: If you set a cell value, and there are formulas dependent on this column,
+    then these formulas refresh. Thus, it might seem like optimizing out the setting
+    cell values might lead to invalid code - but notably, you cannot delete columns
+    that have columns that are dependant on them - so this is not a problem!
+
+    NOTE: we cannot use this on sort, as sort potentially changes the indexes of the
```
aarondr77, line 120: Similarly, we can't use it on filter, right? Since you can apply a filter to a column, then delete that column entirely, but the filter is now applied to the data for the rest of the analysis... until you rerun the analysis.

aarondr77: For future proofing, maybe we could add these checks in the code itself and throw an error or warning to ourselves so its easy to catch?

mitosheet/mitosheet/code_chunks/empty_code_chunk.py
```diff
@@ -0,0 +1,27 @@
...
+# Copyright (c) Saga Inc.
+# Distributed under the terms of the GPL License.
+
+from typing import List, Optional
+from mitosheet.code_chunks.code_chunk import CodeChunk
+
+
+class EmptyCodeChunk(CodeChunk):
+    """
+    A code chunk that generates no code, but has a title.
```
aarondr77, line 13: TODO: Understand what this means. What does "has a title" mean? I can't think of an example where we don't want to generate code, but we want to display info about the step someplace else.

aarondr77: Oh! Well the graphing steps are like that, I suppose!

mitosheet/mitosheet/code_chunks/code_chunk.py
aarondr77: Oh, I see. In the case of changing a column dtype for example, we generate the change column dtype code chunk, but then we also need to generate the code chunks for all of the columns that are dependant on it

mitosheet/mitosheet/code_chunks/code_chunk.py
aarondr77: Okay, yeah I think it is the case that we are updating the step history taskpane. That's the only reason that we would need the EmptyCodeCell CodeChunks. But I will see for sure when I test it

mitosheet/mitosheet/tests/test_utils.py
```diff
@@ -83,6 +83,7 @@ def check_final_dataframe(df_name, df):
     )
 
     import mitosheet
+    print("HERE", code)
```
aarondr77, line 86: remove

mitosheet/mitosheet/code_chunks/code_chunk.py
aarondr77: > However, we need to decide what to do with the step list that is displayed in the frontend. Namely - how do we want to handle step checkout? Namely, because we no longer link the steps to specific code chunks, we don't have a notion of what step a code chunk corresponds to.

Would it be difficult to keep the step history as the unoptimized code? That would be my preference as I think it would be the most helpful version of the step history. And it seems like it wouldn't really incur much complexity at all.

aarondr77:
> The other thing to figure out: pro or no? 
Given the retention goal, I'd love to keep this not pro. But lets discuss it with Jake.

mitosheet/mitosheet/code_chunks/code_chunk.py
naterush: ^ this was a good solution - like 10 lines of code! Nice :-) 

I just get the first code chunk for each step, and display that!

naterush: ^ this was a good solution - like 10 lines of code! Nice :-) 

I just get the first code chunk for each step, and display that!

naterush:
@aarondr77 strong agree on fuzz testing. I think it's probably the biggest step up in continued code robustness we could possibly manage - I'm very pro!

Re: pro functionality - just send a message about it in Zulip!

aarondr77:
Is there any logs that would be helpful to collect? I guess since its a pro feature it doesn't matter anyway. But we could still collect the logs from ourselves.

Seems not worth it.

aarondr77:
For a combined code chunk that does Add column, rename column, and set column formula, the comment starts with "Added a column". 

But if its just a Add column, rename column, then the comment only starts with: "Added column"

mitosheet/mitosheet/parser.py
```diff
@@ -357,7 +357,8 @@ def parse_formula(
         column_header: ColumnHeader, 
         column_headers: List[ColumnHeader],
         throw_errors: bool=True,
-        df_name: str='df'
+        df_name: str='df',
+        include_df_set: bool=True,
```
aarondr77, line 361: Can you add a comment about what this is

aarondr77 (commented):
I really like this code architecture. I can tell you put a lot of thought into it! It really fits the Mito step model super well. 

No big comments + I tested it and don't see any problems. LGTM!

mitosheet/mitosheet/code_chunks/step_performers/column_steps/add_column_code_chunk.py
```diff
@@ -0,0 +1,135 @@
...
+        if not self.params_match(other_code_chunk, ['sheet_index']):
+            return None
+        
+        # Check to see if the column ids overlap
+        sheet_index = self.get_param('sheet_index')
+        added_column_header = self.get_param('column_header')
+        added_column_id = self.post_state.column_ids.get_column_id_by_header(sheet_index, added_column_header)
+        column_ids_to_new_column_headers = other_code_chunk.get_param('column_ids_to_new_column_headers')
+
+        if added_column_id in column_ids_to_new_column_headers and len(column_ids_to_new_column_headers) == 1:
```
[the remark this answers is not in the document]
naterush: Yeah. Def having some ability to make these order better would be ideal. I think potentially just having an optimization pass that starts with trying to combine "like" steps together might be a good first step - I added this to the issue about future code optimization!

twelsh37:
**Good work on this Chaps!!** I think it will make a positive improvement on the code generated. 

On the Pro/Non Pro side my view is that this PR should probably be in all versions. As the product matures surely one of the key points is to show how far you have come with the product. Rolling up code into code chunks and thereby pushing optimisations into the generated code would be a very visual way to display how the product is improving. 

From a personal point of view when I look at the generated code, I just see a list of single steps. The genius of rolled up code is the potential for users to learn. Currently, we as users get single actions carried out on a dataframe. If I rename 20 columns in Mito I would get 60 lines in my generated code cell. 20 lines of code and 20 lines of comments separated by 20 blank lines between each. That's a lot of real estate to take up. The optimized code would be a lot shorter and the commands more succinct.

**Code-Chunks and Step history.**
The old adage " You can't please all the people all of the time ,but you can please some of the people some of the time" springs to mind. I think this will be a difficult one to get right. Some users will love the chunked up optimizations and the display in the step history but other users will want to see the granular steps that were  taken to reach that point. If  it could be possible to have the step history displayed, and then have some kind of differentiator to show that chunking had happened (I mean code rollup or whatever) maybe something like single steps have the blue Mito icon next to them and the optimized chunks have an orange Mito Icon.  Im sure its not going to be that easy to implement but kudos if you manage it.

naterush:
@twelsh37 sorry for the late reply here - I didn't check this PR after it was merged! In the future, feel free to open an issue with your thoughts (or ping us on Discord to let us know you posted or something) - don't want to miss your valuable thoughts!

Thanks for the thoughts on pro v. not. We're trying to strike a balance between giving the most value to our users with an open source and free product and making sure we can fund Mito's development in a sustainable way. Give that most of the folks who use Mito's generated code extensively are Mito power users who use the tool frequently, starting this as a pro feature makes the most sense for now. We'll def keep your thoughts in mind if/when we return to this question - the pro feature set is evolving rapidly currently!

Re: code chunks and step history - I think you're right that some people are gonna want to see the specific steps, while some folks will want optimized code. In the future, we can expose a settings interface that allows users to toggle this (and possibly some other cool settings like dark mode and stuff :-)). 

Thanks again for the thoughts... they are so appreciated. You rock @twelsh37 :)

PR #222, opened 2022-04-07
[WIP] Formula Bar

aarondr77:
# Description

This was pretty difficult to implement, in particular setting the location and the size of the formula bar was rough. I don't think the current implementation is the smoothest, as it doesn't use the same approach that we use for in-cell cell editors, but I couldn't really figure out how to do it otherwise. In particular, using the refs was difficult. 

The two known remaining things to do are:
- [ ] Currently, if you use your mouse to select a column while using the formula bar cell editor, the column always gets added to the front of the formula. Not sure why, I have to look into it. 
- [ ] When using the formula bar cell editor, we always want the arrow keys to move inside of the formula, like Excel. This is easy. 

# Testing

# Documentation

### DIFF
--- a/mitosheet/css/endo/CellEditor.css
+++ b/mitosheet/css/endo/CellEditor.css
@@ -4,7 +4,6 @@
     z-index: 2;
     position: absolute;
     box-sizing: border-box;
-    background: white;
 }
 
 .cell-editor-form {
@@ -14,7 +13,7 @@
 }
 
 .cell-editor-input {
-    height: 25px;
+    height: 26px;
     width: 100%;
     border: none;
     box-sizing: border-box;
@@ -26,7 +25,8 @@
     
     display: flex;
     flex-direction: column;
-    overflow-x: hidden;
+    overflow-wrap: break-word;
+    white-space: normal; 
 }
 
 .cell-editor-suggestion {
--- a/mitosheet/css/endo/CellEditor.css
+++ b/mitosheet/css/endo/CellEditor.css
@@ -4,12 +4,26 @@
     z-index: 2;
     position: absolute;
     box-sizing: border-box;
+
+    /* 
+        Set the pointer-events: none allows the user to select elements that are behind the 
+        cell editor. The formula bar cell editor's rectangular div covers the column headers. 
+        However, most of the rectangle is translucent. 
+
+        We set the entire cell-editor to have no pointer events, and we turn on the pointer events
+        for the specific parts of the cell editor that are displayed. This lets the user interact with 
+        all of the visible elements as they would expect. 
+    */
+    pointer-events: none;
 }
 
 .cell-editor-form {
     border: 2px solid var(--mito-purple);
     width: 100%;
     box-sizing: border-box;
+
+    /* Let the user click on the cell editor form */
+    pointer-events: all 
 }
 
 .cell-editor-input {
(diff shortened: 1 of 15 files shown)


mitosheet/css/endo/CellEditor.css
```diff
@@ -14,7 +13,7 @@
 }
 
 .cell-editor-input {
-    height: 25px;
+    height: 26px;
```
aarondr77, line 16: Add a comment that this is 26px so that it lines up with the formula bar

mitosheet/src/components/endo/CellEditor.tsx
```diff
@@ -494,6 +497,7 @@ const CellEditor = (props: {
         }
     }
 
+    console.log(props.containerRef)
```
aarondr77, line 500: remove

mitosheet/src/components/endo/CellEditor.tsx
```diff
@@ -169,6 +170,8 @@ const CellEditor = (props: {
             // Focus the input
             cellEditorInputRef.current?.focus();
 
+            console.log('here')
```
aarondr77, line 173: remove

mitosheet/css/sitewide/flexbox.css
```diff
@@ -18,4 +18,8 @@
 
 .flexbox-justify-end {
     justify-content: end;
+}
+
+.flexbox-align-items-center {
+    align-items: center;
```
aarondr77, line 24: Ended up not using this, remove it

naterush (changes requested):
I think trying to reuse the exact same component for both the cell editor and the formula bar is probably a loosing game - given that we want the former to float about, and the latter to be fixed in place. I think a floating look in the second case is bad news bears.

A cleaner way to approach this probably:
1. Take the `CellEditor` and split it into a `CellEditorInput` and `FloatingCellEditor`. The `CellEditorInput` is the `form` component and all it's children (I think), and all the logic for using the arrow keys to navigate about (and all the focus handling). The `CellEditorInput` fills it's container.
2. The `FloatingCellEditor` wraps the `CellEditorInput` and moves it about - so pretty much just the code of the `CellEditor` currently!
3. Then, create a `FormulaBarCellEditor`. It does not float, but also has the `CellEditorInput` as a child component. 

Thoughts? I might be missing something on why this is bad - but you end up with a floating thing sometimes and not others, without pretty much any duplicated code...

naterush:
@aarondr77 I think we went into the implementation phase of this story a bit too quickly. We've been avoiding this story for a while based on technical complexity - and although we addressed a lot of that with the removal of Endo, the fact that we paused in the past should cause us to take the time to write a little spec. The triggers for writing a spec is: big feature OR small feature + someone thinks it might be scary.

I don't think it had to be much - but if you had written up this proposed approach, I probably would have been able to give some input about floating elements being fragile earlier - which could have saved you/us some time!

No big deal, just a lesson learned for the future :-)

aarondr77:
One issue with your proposal is that we do want parts of the cell editor to be floating -- the dropdown box should float about the grid. So if we break it out, then we will have to add back the floating of just the dropdown box.

aarondr77:
Leaving a note here so I don't forget. We're going takes Nate's proposal + break the cell editor input and the dropdown box into separate components so that we can leave the cell editor input as a placed item of its parent, and make the dropdown box foat

PR #240, opened 2022-04-18
Formula bar attempt 3

aarondr77:
# Description

Makes the formula bar work the same way as the cell editor!

Note: I'm logging the location of the cell editor by passing a param. Works fine, but I don't think we pass uneeded params elsewhere, so lmk if its okay. 

# Testing

I tested this on mac: [firefox, chrome] and windows: [brave, chrome], and it works as expected. Can you pick a different combo?

# Documentation

I added a tiny bit of documentation because there isn't a good place to add it right now. When we redo the formula writing documentation page, we should def highlight this.


mitosheet/src/components/endo/celleditor/CellEditor.tsx
```diff
@@ -1,43 +1,30 @@
...
-    Allowing users to select column headers by using the arrow keys is handeled
-    inside this component, by overwriting the cell navigation logic and updating 
-    the selection here. Clicking on columns is handeled inside the EndoGrid 
-    itself.
+    A CellEditor allows the user to edit the formula or value of a cell.
+    
+    The main complexity is allowing the user to select column headers by 
+    clicking or using the arrow keys. It is handled inside this component, 
+    by overwriting the cell navigation logic and updating the selection here. 
+    Clicking on columns is handeled inside the EndoGrid itself.
```
naterush, line 25: handeled -> handled

mitosheet/src/components/endo/celleditor/CellEditor.tsx
```diff
@@ -49,102 +36,22 @@ const CellEditor = (props: {
...
-        setTimeout(updateCellEditorPosition)
-
-        // We reposition the cell editor when you enter or leave fullscreen mode, to make
-        // sure that it stays visible
-        fscreen.addEventListener('fullscreenchange', updateCellEditorPosition);
-        return () => fscreen.removeEventListener('fullscreenchange', updateCellEditorPosition);
-    }, [])
-
+    // When the CellEditor is created via the FloatingCellEditor, the component tries to set the focus on the input
+    // field while the cellEditorInputRef is still Null. So here, we wait until the cellEditorInputRef is not null
```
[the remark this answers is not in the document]
naterush: Ah, I see what you're doing. You're attempting to watch for changes on the id, and then running this effect when it is set and not undefined.

This only works currently due to luck, I think, and other pre-renders that are happening. The key is that setting a ref or changing a ref generally does not cause effects to rerun, or the react component to re-render generally. 

You're looking for a callback ref. Peep line 251 of the the Dropdown.tsx for an example of how to run a function after the ref is defined. 

Also, sorry if you took this code from somewhere else and I wrote it. I'm also learning about this stuff in real time, so parts of our codebase do this just like you're doing it here - even though it's fragile and pretty much bad!

mitosheet/src/components/endo/celleditor/CellEditor.tsx
```diff
@@ -522,38 +427,40 @@ const CellEditor = (props: {
...
-                        // If we are typing at the end of the formula, and we type a CHARS_TO_REMOVE_SCROLL_IN_FORMULA,
-                        // then we reset the arrowKeysScrollInFormula to false. Furtherrmore, if the formula is empty, 
-                        // we reset the arrow keys to scroll in the sheet. Otherwise, we keep it as is.
-                        // This attempts to match what Excel and Google Sheets do
-                        const atEndOfFormula = (e.target.selectionStart || 0) >= e.target.value.length;
-                        const finalChar = e.target.value.substring(e.target.value.length - 1);
-                        const endsInResetCharacter = atEndOfFormula && CHARS_TO_REMOVE_SCROLL_IN_FORMULA.includes(finalChar)
-                        const isEmpty = e.target.value.length === 0;
-                        const arrowKeysScrollInFormula = props.editorState.arrowKeysScrollInFormula && !endsInResetCharacter && !isEmpty; 
+                        let arrowKeysScrollInFormula: boolean | undefined = true
```
naterush, line 430: Why would it be undefined?

mitosheet/src/components/endo/celleditor/FloatingCellEditor.tsx
```diff
@@ -0,0 +1,152 @@
...
+        // Make it so the setting of the cell editor positon just runs after the
+        // current execution context finishes, to make sure everything is placed
+        // properly.
+        setTimeout(updateCellEditorPosition)
+
+        // We reposition the cell editor when you enter or leave fullscreen mode, to make
+        // sure that it stays visible
+        fscreen.addEventListener('fullscreenchange', updateCellEditorPosition);
+        return () => fscreen.removeEventListener('fullscreenchange', updateCellEditorPosition);
+    }, [])
```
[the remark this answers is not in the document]
naterush: We previously didn't make it do this, which I think is the right call and fine. Let's just update the comment then perhaps.

naterush (changes requested):
Working great. Some minor cleanup, and then ping me and I'll get it merged!

naterush:
Screen Shot 2022-04-19 at 5 02 23 PM

A few styling / polish questions that I think will take this feature up another notch on top of the improvement it already is: 
1. Does the harsh purple feel really out of place up top? I think bumping it to 1px and a higher purple looks a lot better. let me know what you think. 
2. Do we want the suggestion box to get a bit bigger up top? I feel like we can utilize a lot more space up here. I don't know if it needs to get massive, but perhaps like twice as big? Your call here. 
3. There's whitespace between the cell editor and the rest of the formula bar that looks kinda terrible. Like if you start editing, there's the purple border, white space, and then the bottom border of the toolbar. Can we get rid of that white space?

mitosheet/src/components/endo/celleditor/CellEditor.tsx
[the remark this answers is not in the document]
aarondr77: I'm currently using this suggestion, but there still is some race condition with the focus. So I have a really tiny timeout. I can't figure out where the race condition is though. If you want to give it a try feel free to.

mitosheet/src/components/endo/celleditor/CellEditor.tsx
```diff
@@ -49,102 +36,27 @@ const CellEditor = (props: {
...
-                left: left,
-                bottom: bottom,
-                right: right,
-                display: undefined
-            });
+    // When we first render the cell edit input, make sure to save it and focus on it
+    const setRef = useCallback((unsavedInputAnchor: HTMLInputElement) => {
+        if (unsavedInputAnchor !== null) {
+            // Save this node, so that we can update 
+            cellEditorInputRef.current = unsavedInputAnchor;
```
naterush, line 53: I think we can just call unsavedInputAnchor.focus - but this looks much better!

naterush:
@aarondr77 ok update I found one bug in this in final test:
1. Create a new column
2. Click on a column header of the new column in the sheet
3. Double click on the formula bar. 
4. Observe that the formula for for the column (=0) gets submitted as the column header

Pretty much, we don't support header editing in the formula bar. Let's fix this up before merging, either:
1. Making it so you can edit column headers in the formula bar, and it just works.
2. We just turn off formula bar editing for headers (this is worse). 

I think we should do (1), which might push back this PR out of this release. I'm fine with that - but let me know what you think.

aarondr77:
Okay, the bug above is caused by the fact that the column header is a pure function of the state. So when we set the editor state, the column header automatically opened. It's interesting, that architecture is really helpful most of the time, but as things get more complex, it gets harder to track stuff like that down. If we didn't have consistent naming (like if we changed the names of variables when we created new props) this would be a big issue. Yay for clean code. Anyways, just random thoughts. 

To fix this bug, though, I see three options. 

### Cell Editor 
Add a case to the cell editor to submit the column rename. This would be nice because all of the form handling, stying, etc is already there. 

On the other hand, the cell editor is really just for cell editing, and it would be bad to mix it with column editing, especially when we already have the ColumnHeader component. 

### Column Header component abstraction
Just like we did with the cell editor, we could create a FloatingColumnHeader and use that inside of the grid, and then use the ColumnHeader component in the formula bar. Note: We might want to do some special styling here in the formula bar case because currently, the column header rename input doesn't have any styling.

### Don't handle column headers
Very easily we could just not display column headers in the formula bar. This is reasonable for what a formula bar is, but doesn't make editing long column headers any easier for users.

naterush:
Nice bug investigation! So pretty much:
1. Cell editor opened. This opens column header editor. 
2. Formula bar focused on. 
3. The onBlur for the column header submits it. 

I don't think the ColumnHeaderEditor (whatever it is named) floats currently, not do I think it needs to. So My vote here is we do not refactor to a floating boi. 

I am down to take on the technical debt to make the CellEditor support column header editing when it's in formula bar mode. I think we should just finish this off - again, if this doesn't make the release, I think that's a-ok. Let's take the extra half day and polish this - there's really no rush.

mitosheet/src/components/Mito.tsx
```diff
@@ -402,6 +402,7 @@ export const Mito = (props: MitoProps): JSX.Element => {
 
         // Close the cell editor if it is open
         if (editorState !== undefined) {
+            console.log(10)
```
naterush, line 405: Remove

mitosheet/src/components/endo/FormulaBar.tsx
```diff
@@ -5,24 +5,46 @@ import React from 'react';
...
+    const {columnHeader, columnFormula, cellValue} = getCellDataFromCellIndexes(props.sheetData, rowIndex, colIndex);
     const originalFormulaBarValue = '' + (columnFormula !== undefined && columnFormula !== '' ? columnFormula : (cellValue !== undefined ? cellValue : ''));
     const cellEditingCellData = props.editorState === undefined ? undefined : getCellDataFromCellIndexes(props.sheetData, props.editorState.rowIndex, props.editorState.columnIndex);
     const formulaBarColumnHeader = props.editorState === undefined ? columnHeader : cellEditingCellData?.columnHeader;
-    const formulaBarValue = props.editorState === undefined ? originalFormulaBarValue : getFullFormula(props.editorState.formula, formulaBarColumnHeader || '', props.editorState.pendingSelectedColumns);
 
+    // If the formula bar is a cell, display the cell value. If it is a column header, display the column header
+    let formulaBarValue = '' 
+    if (rowIndex == -1 && columnHeader !== undefined) {
+        formulaBarValue = String(columnHeader)
```
naterush, line 42: I don't think this is correct to cast to a string like this. We have a function called `getDisplayColumnHeader` - which I think is what you want...

The fact that we _never_ use the `String(` utility on column headers before (but we do have column headers displayed as text all over the place) is good evidence that this approach is not what we want. 

You've gotten got on type casting before. I think when you find yourself converting between types, try and find a similar example to help guide you!

naterush (changes requested):
Two things:
1. Merge in dev (and keep dev's switch.py - ignore this branches changes).
2. We should remove the suggestion box when you're editing a formula. I think this requires an `isColumnHeader` input to the cell editor or something... make sure to test all the paths here thoroughly.

mitosheet/src/components/endo/celleditor/CellEditor.tsx
```diff
@@ -454,28 +370,40 @@ const CellEditor = (props: {
...
+            } 
         } else {
-            errorMessage = await props.mitoAPI.editSetCellValue(
-                props.sheetIndex,
-                columnID,
-                index,
-                formula
-            )
-        } 
+            // Change of column header
```
naterush, line 402: Can we move this up to the first condition? E.g. check if rowIndex == -1?

It sounds kinda silly, but I would expect the if statements to flow in the same order that the sheet data does - like column header before everything else. My brain took a second to process as it is - I was expecting column headers first!

mitosheet/src/components/endo/FormulaBar.tsx
```diff
@@ -44,7 +66,38 @@ const FormulaBar = (props: {
...
+                        setGridState={props.setGridState}
+                        setUIState={props.setUIState}
+                        scrollAndRenderedContainerRef={props.scrollAndRenderedContainerRef}
+                        containerRef={props.containerRef}
+                        mitoAPI={props.mitoAPI}
+                        currentSheetView={currentSheetView}
+                    />
+                } 
+
+                {(props.editorState === undefined || props.editorState.editorLocation === 'cell') &&
```
naterush, line 85: I think we can do the inverse of the check above? Just `props.editorState?.editorLocation !== 'formula bar'` - rather than this more complicated check?

mitosheet/src/components/endo/FormulaBar.tsx
```diff
@@ -5,24 +5,46 @@ import React from 'react';
...
+
+    const {columnHeader, columnFormula, cellValue} = getCellDataFromCellIndexes(props.sheetData, rowIndex, colIndex);
     const originalFormulaBarValue = '' + (columnFormula !== undefined && columnFormula !== '' ? columnFormula : (cellValue !== undefined ? cellValue : ''));
     const cellEditingCellData = props.editorState === undefined ? undefined : getCellDataFromCellIndexes(props.sheetData, props.editorState.rowIndex, props.editorState.columnIndex);
     const formulaBarColumnHeader = props.editorState === undefined ? columnHeader : cellEditingCellData?.columnHeader;
-    const formulaBarValue = props.editorState === undefined ? originalFormulaBarValue : getFullFormula(props.editorState.formula, formulaBarColumnHeader || '', props.editorState.pendingSelectedColumns);
 
+    // If the formula bar is a cell, display the cell value. If it is a column header, display the column header
+    let formulaBarValue = '' 
+    if (rowIndex == -1 && columnHeader !== undefined) {
```
naterush, line 41: Per my other comment, doing this if (column header) check as the first condition is much more natural!

mitosheet/src/components/endo/ColumnHeader.tsx
```diff
@@ -361,7 +342,7 @@ const ColumnHeader = (props: {
                             }}
                             // We submit the column header if the user focuses outside the input
                             onBlur={() => {
-                                void submitRenameColumnHeader();
+                                void _submitRenameColumnHeader();
```
naterush, line 345: I think you can call submit function directly here, as you never need to pass an e value. Wrapping it might just obfuscate things...

naterush: It's a bit more verbose, but also more clear I think!

naterush (changes requested):
A few more comments, mostly minor

aarondr77:
The only thing I didn't do was remove the suggestion box while editing a formula. Can you help me understand why you think we should do that? When would the suggestion box be helpful if it wasn't for editing a formula?

mitosheet/src/components/endo/celleditor/CellEditor.tsx
```diff
@@ -370,16 +370,21 @@ const CellEditor = (props: {
...
         setLoading(true);
         
         let errorMessage: MitoError | undefined = undefined;
 
+
         // Make sure to send the write type of message, depending on the editor
-        if (rowIndex >= 0) {
+        if (props.editorState.rowIndex == -1) {
+            // Change of column header
+            console.log("hererere")
```
aarondr77, line 384: remove

aarondr77:
The height of the cell editor on windows and mac is a bit different for some reason, so I updated the height to do what looks best on windows, chrome since that is where most of our users are.

naterush:
@aarondr77 sorry, I meant to say: remove the suggestion box when editing a column header? Can you check if this is good?

aarondr77:
Ooo sweet. Yeah that is taken care of. It just says "You are editing the column header" and doesn't have anything else pop up.

PR #251, opened 2022-04-21
Dogfood: 4-21

naterush:
# Description

1. Removed unnecessary print in installer.
2. Focus on search input in file browser once you click on new folder
3. Fixes a massive bug in signup flow where modal was not visible, and buttons were all wacky.

# Testing

# Documentation

### DIFF
--- a/docs/misc/release-notes.md
+++ b/docs/misc/release-notes.md
@@ -2,7 +2,13 @@
 description: Want to see what is new in the Mitosheet? Check it out below.
 ---
 
+
 # Release Notes
+
+## 2022-4-21 <a href="#2022-4-21" id="2022-4-21"></a>
+
+New Features:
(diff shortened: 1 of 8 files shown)


naterush:
Oh weird. Is this also the case on JLab? If so, then it seems like it's potentially from the recent dropdown changes we've made to scroll bars. 

Wanna take a shot at fixing them?

aarondr77:
> 

Yeah, let me check if its on Lab too. And I'll also check if its just edge. And then I can try to fix.

aarondr77:
Okay, yeah so I think the lack of scrollbar is a windows thing and the shakyness is just a [windows edge, notebook thing]. 
I'll fix the lack of scrollbar rn.

naterush (commented):
Looks good!

PR #272, opened 2022-05-04
Add new toolbar

naterush:
# Description

This PR adds the new toolbar, following the specification here: https://www.notion.so/trymito/Megastructure-37506a53542740e9b378a87d73fafc32

# Testing

A few things to pay attention to:
1. All the buttons work / are hooked up correctly. Click em all!
2. Each button is disabled in the correct cases! Check these explicitly. 
3. Things are labeled and grouped consistently. I tried hard to standardize order and grouping, but I am sure I missed something, so thoughts here are great!
4. Spacing. There might be some wackiness as a result of these changes. Keep your eyes peeled.

# Documentation

We'll need to let users know that the search action bar has moved to help (which is legit I think)...


mitosheet/src/utils/actions.tsx
```diff
@@ -84,9 +85,22 @@ export const createActions = (
...
+        [ActionEnum.Catch_Up]: {
+            type: ActionEnum.Catch_Up,
+            shortTitle: 'Catch Up',
+            longTitle: 'Catch up',
+            actionFunction: () => {
+                // Fast forwards to the most recent step, allowing editing
+                void mitoAPI.log('click_catch_up')
+                void mitoAPI.updateCheckoutStepByIndex(-1); // TODO: Check that -1 works! And below
+            },
+            isDisabled: () => {return undefined},
```
naterush, line 97: TODO: update this! and comment above!

mitosheet/src/utils/actions.tsx
```diff
@@ -286,10 +315,29 @@ export const createActions = (
...
+                // We turn off editing mode, if it is on
+                setEditorState(undefined);
+                
+                if (uiState.selectedGraphID) {
+                    const newGraphID = getRandomId()
+                    await mitoAPI.editGraphDuplicate(uiState.selectedGraphID, newGraphID)
+                }
+            },
+            isDisabled: () => {
+                return getGraphIsSelected(uiState) ? undefined : 'No graph is selected. Select a graph tab before duplicating it.'
```
naterush, line 332: TODO: write this function for all the dataframe options!

naterush (commented):
Cleanup and test!

aarondr77:
Can we keep the color of this icon white?

Screen Shot 2022-05-04 at 5 24 17 PM

aarondr77:
In order for the dropdowns in the toolbar to appear, I need to first click on one of them. For example, 

1. Render a mitosheet
2. Click on one of the cells
3. Hover your mouse over any of the dropdown labels in the toolbar and notice that dropdown does not appear. 
4. Click on one of the dropdown labels and notice the dropdown appears
5. Hover over other dropdown labels and notice that they appear 

Can we make it so that if the focus is anywhere in Mito, then hovering over the dropdown labels make the dropdown appear?

aarondr77:
I think these separators are super valuable. But I think we can make them look a little bit better. Could we just make it:
1. a fainter gray
2. not go all the way across, but instead having a little bit of whitespace on either side
3. have rounded edges (now that its not going all the way across)

Screen Shot 2022-05-04 at 5 31 09 PM

aarondr77:
When my mouse hovers over the text, the cursor turns into this boy. Can we keep it the pointer since this isn't an input. 

Screen Shot 2022-05-04 at 5 33 26 PM

aarondr77:
I think the font color used in the menu is darker than the font color below

Screen Shot 2022-05-04 at 5 52 46 PM

aarondr77:
Can we keep this white?

Screen Shot 2022-05-04 at 6 03 37 PM

aarondr77:
There is unnatural behavior when a graph tab is selected. For example, if a graph tab is selected and I select add column then a column is added to the previous sheet I was selected. This is left over from the graphing implementation and we decided not to do it then, but now that we're editing the formula bar it seems like the time to do it.

aarondr77:
> I standardized button highlighting in the toolbar, which in practice meant removing variants of the buttons. They still are very reactive and tbh I kinda think they are more casual and cool now. Let me know thoughts.

I think it looks really great. The softer purple + rounded corners looks dope! 

One small detail would be making the purple vertically centered. I think its probably not because of the top-margin and bottom-padding of the horizontal line between the two sections of the toolbar. 

Screen Shot 2022-05-04 at 6 08 57 PM

aarondr77:
One note on consistent labeling: We sometimes use the word 'selected' and sometimes we don't. If we're going to use it "Filter selected column" then we should have "Rename selected column". I'd be down to get rid of "selected" from these

Screen Shot 2022-05-04 at 6 13 30 PM

aarondr77:
As per our conversation today around `delete row`, here are some proposals (not necessarily to implement now, but just to make sure we're happy with our flexibility going forward):

1. add a new menu item `row` which would contain `add row` `delete row` `promote row to header`
2. use the `Edit` menu for things like `add row` `delete row` `add column` `delete column` and other things in the future like `transpose`. Edit would get all things related to the shape of the data. 
3. have items in multiple places. We could do a combo of 1  + 2 (this one kinda sucks)
4. change the title to `data`, but this is super confusing since I don't know what the difference between that and `dataframes` is.
5. change the title to `Columns/Rows` or `Columns & Rows`. Maybe the first one works. 


I think doing number 1 would be fine for now and would carry us a long way. And we will learn through usability tests what people think!

aarondr77 (commented):
Great work on designing + implementing this. I think you did a great job. I have a few small comments mostly around small design tweaks, but other than that it looks great. 

This is good timing to redo the toolbar also right before we redo a bunch of the docs since this makes all of our docs super outdated!

Issue #249, opened 2022-04-20
Loading message for opening folder


naterush:
**Describe the bug**

Sometimes it takes a while. Let's add a loading message!

Issue #248, opened 2022-04-20
Pivot Table gets renamed


aarondr77:
1. Import data
2. Create a pivot table
3. change the format of one of the columns
4. Reopen the pivot table configuration and add a new column to the value section
5. Notice that the pivot table sheet name now also contains `_1`

naterush:
I addressed this in a release a while ago, this is fixed!

Issue #271, opened 2022-05-03
Plotly sometimes not defined


naterush:
**Describe the bug**
Sometimes plotly does not load, and the graphs do not loading. We have a log now called `plotly_define_failed`, which we should investigate in the coming weeks and see where it occurs, then try to fix it!

**Desktop (please complete the following information):**

We have seen this on Chrome so far. 
<img width="268" alt="Screen Shot 2022-05-03 at 4 06 13 PM" src="https://user-images.githubusercontent.com/21002431/166566233-4349b23d-ba27-48d7-889c-69d345b3b7c8.png">

naterush:
I have a feeling this is a result of the jlab check that we're currently doing. Maybe we should try and define plotly always and then if it doesn't work we can just do the amd approach. Let's see how common this is though!

naterush:
We don't have enough information to work on this yet. We need to deploy the Plotly log first.

PR #278, opened 2022-05-05
Error reduction 5

naterush:
# Description

Does error reduction here: https://www.notion.so/trymito/5-5-22-3676be30e9764ea9a00fa5df0e26bba4

Did you know that we can add multiple columns to the X and Y axis on graphs currently. I thought we blocked this - did we always allow this?

Also see at bottom of document. I think optimizing for this workflow would make me a real Mito user, and aka is something that I think might really be worth doing. Let me know!

# Documentation

Nope.

### DIFF
--- a/mitosheet/src/components/taskpanes/Import/FileBrowser.tsx
+++ b/mitosheet/src/components/taskpanes/Import/FileBrowser.tsx
@@ -163,7 +163,7 @@ function FileBrowser(props: FileBrowserProps): JSX.Element {
                 />
             </div>
             <div className='file-browser-element-list'>
-                {elementsToDisplay?.map((element, i) => {
+                {!props.importState.loadingFolder && elementsToDisplay?.map((element, i) => {
                     return (
                         <FileBrowserElement
                             key={i}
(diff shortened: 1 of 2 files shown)


naterush:
NOTE: for some reason, there is a large number of files being shown as changed (despite that not appearing in the commits). Check out the commits 1-by-1 to see what is going on more easily.

aarondr77:
> Did you know that we can add multiple columns to the X and Y axis on graphs currently. I thought we blocked this - did we always allow this?

Yeah! This is part of that whole thing around letting Plotly do the error handling instead of us trying to capture all of Plotly's intricacies and build better error handling systems. 

In Plotly express the rules are (to my understanding):
1. You can't have multiple series on the x and y axis at the same time. You have to pick one or the other
2. You can't have series of different types on the same axis.

aarondr77:
Btw: There are a few issues tagged with `next-bug-squash` want to attempt those and then I'll review all at once?

naterush:
Thanks for the reminder on those!

Closes #249 
Closes #248 (this was already fixed, so just closed it). 
#271 is not yet fixable!

PR #281, opened 2022-05-07
Sheet tab actions

aarondr77:
# Description

Adds a download button to the sheet tab actions because I saw a user try to do this yesterday.

# Testing

# Documentation


mitosheet/src/components/footer/DataSheetTabActions.tsx
```diff
@@ -124,6 +135,10 @@ export default function SheetTabActions(props: {
                     void graphData()
                 }}
             />
+            <DropdownItem 
+                title='Download'
```
naterush, line 139: Can we call this export to be consistent with the toolbar?

naterush (changes requested):
Three things:
1. Check the commit I pushed. This PR made me realize how it was to add right clicking here (literally as easy as adding a log lol).
1. Can we add this to the graphing tabs also (just have them open the export tab).
2. Can we group the things in the sheet tabs? I think we should do [Graph, Export] | [Duplicate, Rename, Delete]. Mostly, I want delete to be last in both action tabs, so it's easy to click, as this is the most common sheet tab action to click here.

mitosheet/src/components/footer/DataSheetTabActions.tsx
```diff
@@ -109,6 +111,16 @@ export default function SheetTabActions(props: {
         );
     }
 
+    const openDownloadTaskpane = async (): Promise<void> => {
+        props.setUIState(prevUIState => {
+            return {
+                ...prevUIState,
+                currOpenTaskpane: {type: TaskpaneType.DOWNLOAD},
+                currOpenModal: {type: ModalEnum.None},
```
naterush, line 119: I think we have to set the type to data here, unfortunately

mitosheet/src/components/footer/GraphSheetTabActions.tsx
```diff
@@ -50,21 +53,35 @@ export default function GraphSheetTabActions(props: {
...
+
+    const openExportGraphTaskpaneTab = async (): Promise<void> => {
+        props.setUIState(prevUIState => {
+            return {
+                ...prevUIState,
+                currOpenTaskpane: {
+                    type: TaskpaneType.GRAPH, 
+                    graphID: props.graphID, 
+                    graphSidebarTab: GraphSidebarTab.Export
+                },
```
naterush, line 65: Type to graph

naterush (changes requested):
Minor changes, I'll make em!

PR #282, opened 2022-05-08
mitosheet: add right click on column header

naterush:
# Description

This adds a set of action to the column header. Let me know your thoughts!

The only issue with this is that the open column header dropdown is that it's not in global state so that you can open multiple at once. We can refactor this if we want, but I am kinda ok with it. let me know ur thoughts.

# Testing

Try the actions! 

# Documentation

N/A


aarondr77:
Can we add formatting to menu also?

PR #293, opened 2022-05-10
mitosheet: dogfood toolbar

naterush:
# Description

- Update tour
- Fix bug where telemetry wasn't turned off through a rare upgrade path
- Fix formula bar displaying header if editing formula and you clicked on a header
- Make toolbar look the same in the notebook

# Testing

Check tour, upgrading.

# Documentation

N/A


aarondr77:
IMG_6287

aarondr77:
If you double click on these menu items, then you end up selecting the text also, which doesn't go away when you move your mouse. Can we make this text unselectable? 

Screen Shot 2022-05-11 at 11 25 22 AM

You can also do things like this
Screen Shot 2022-05-11 at 11 26 08 AM

PR #298, opened 2022-05-12
add scrollbar to tabs

aarondr77:
# Description

Makes sure that there is always a scrollbar for the user to scroll the sheet tabs even if they don't have a trackpad by using the [scollbar-gutter](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) css property. 

However, this doesn't work on IE or Safari. 

Note: The scrollbar-gutter property only effects computers whose scrollbars are not floating. On Macs, the scrollabrs float, so this has no effect.  Which unfortunately means the footer is different on macs and windows. 

Windows JupyterLab:

footer-scroll-gutter

Windows Notebook: 
footer-scroll-gutter-notebook

If there is no overflow, then it just looks normal:
footer-no-scroll-bar

# Testing

# Documentation

### DIFF
--- a/mitosheet/css/footer.css
+++ b/mitosheet/css/footer.css
@@ -1,5 +1,5 @@
 .footer {
-    height: 32px;
+    height: min-content;
     display: flex;
     justify-content: space-between;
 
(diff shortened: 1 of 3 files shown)


mitosheet/css/footer.css
```diff
@@ -36,10 +39,13 @@
 
 .footer-tab-bar {
     /* Deal with the overflow */
-    overflow-x: auto;
+    overflow: auto;
```
aarondr77, line 42: Add comment explaining that for the scrollbar-gutter property to work, we need to set the overflow property. And since we don't want scrolling on the y axis, we can overwrite that one

PR #306, opened 2022-05-17
Split text to columns

aarondr77:
# Description

Adds Split text to columns on delimiters. It also adds a bit of helpful infrastructure for putting buttons at the bottom of a taskpane. 

Future implementation: 
1. It would be nice if we could press the submit button with Enter. This infrastructure could be used by all of the useEditOnClick taskpanes. I opened an issue for it https://github.com/mito-ds/monorepo/issues/321
2. Its a bit confusing that it doesn't work with formatting. We could take a stab at adding the unformatted view of the data to the preview have that that unformatted view displayed there when the taskpane opens. What do you think? @naterush 

There are a few known bugs in splitting text to columns:
1. Using `.` along with another delimiter behaves weirdly. https://github.com/mito-ds/monorepo/issues/322. 
3. The split function sometimes returns None and other times returns nan, which doesn't get converted to NaN in the `df_to_json_dumpsable` function. https://github.com/mito-ds/monorepo/issues/323

# Testing

Nothing special here. 

# Documentation

Checkout the documentation draft and merge when ready.


aarondr77:
There is weird behavior that if the user sets the `Other` delimiter to `.` it only works if its the only delimiter selected. If an additional delimiter is included, then it just returns a bunch of blank columns. But that is only the case when its the `.`. If for example, you use `3` as the other delimiter than it works fine ....

aarondr77:
It also doesn't respect the formatting of the cells, of course. But one weirder thing is that for dates that have 00:00:00 it doesn't create a column with 00:00:00 if you, for example, split on the space.

naterush:
@aarondr77 that bug is b/c we use the different code to convert from date times to strings when we write for the frontend (180 ish in utils.py), this code: `df[column_header] = df[column_header].dt.strftime('%Y-%m-%d %X')`

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,302 @@
...
+        },
+        StepType.SplitTextToColumns, 
+        props.mitoAPI,
+        props.analysisData,
+    )
+
+    const [preview, setPreview] = useState<(string | number | boolean)[][] | undefined>([])
+    const [loadingPreview, setLoadingPreview] = useState<boolean>(false)
+
+    console.log(loadingPreview)
```
aarondr77, line 93: Got to implement the loading indicator still...

mitosheet/src/components/endo/ColumnHeaderDropdown.tsx
```diff
@@ -18,14 +19,20 @@ export default function ColumnHeaderDropdown(props: {
...
     sheetIndex: number;
     columnID: ColumnID;
+    sheetData: SheetData
     columnDtype: string;
 }): JSX.Element {
 
     // Log opening this dropdown
     useEffect(() => {void props.mitoAPI.log('opened_column_header_dropdown')}, [])
 
+    const rowIndex = 0
```
naterush, line 32: What is rowIndex?

mitosheet/src/components/endo/ColumnHeaderDropdown.tsx
```diff
@@ -96,6 +103,36 @@ export default function ColumnHeaderDropdown(props: {
...
+                        columnIndex: columnIndex,
+                        formula: columnFormula !== undefined ? columnFormula : '',
+                        // As in google sheets, if the starting formula is non empty, we default to the 
+                        // arrow keys scrolling in the editor
+                        arrowKeysScrollInFormula: columnFormula !== undefined && columnFormula.length > 0,
+                        editorLocation: 'cell'
+                    })
+                }}
+                disabled={columnFormula === undefined || columnFormula.length == 0}
+                tooltip={!isNumberDtype(props.columnDtype) ? "Data columns don't support formulas" : undefined}
```
naterush, line 122: I don't think this is the right check! I think you want the disabled check above!

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,302 @@
...
+ 
+    // Get the starting column id
+    let startingColumnID = undefined
+    if (props.startingColumnID !== undefined) {
+        startingColumnID = props.startingColumnID
+    } else if (!isSheetEmpty) {
+        startingColumnID = props.sheetDataArray[props.gridState.sheetIndex].data[startingColumnIndex].columnID
+    } 
+
+    const {params, setParams, loading, edit, editApplied} = useSendEditOnClick<SplitTextToColumnsParams, undefined>(
```
naterush, line 79: Hook should be defined at the stop of the file. If you have to do work to get data to call this, do it in a function for clarity

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,302 @@
...
+}
+
+/* 
+    This taskpane allows users to split a column into multiple columns 
+    by separating on a delimeter
+*/
+const SplitTextToColumnsTaskpane = (props: SplitTextToColumnsTaskpaneProps): JSX.Element => {
+
+    // Use the selected column, unless its the index column, then just use the first column
+    const startingColumnIndex = props.gridState.selections[0].startingColumnIndex >= 0 ? props.gridState.selections[0].startingColumnIndex : 0
```
naterush, line 66: Dont' think you want the first selection here, vs. the last one. But I'd check elsewhere in the codebase, we have a standard practice we do here when pulling a single column out of the selections.

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,302 @@
...
+import DefaultTaskpaneFooter from "../DefaultTaskpane/DefaultTaskpaneFooter";
+
+interface SplitTextToColumnsTaskpaneProps {
+    mitoAPI: MitoAPI;
+    analysisData: AnalysisData;
+    sheetDataArray: SheetData[];
+    gridState: GridState;
+    setUIState: React.Dispatch<React.SetStateAction<UIState>>;
+    dfNames: string[];
+    startingColumnID?: ColumnID
```
naterush, line 28: Just based on how much work you're doing at the start of this file to find the correct column id, why not just make this not optional, ever?

No matter where you're setting the taskpane from, you know the column ID that is selected (even in the actions, this work is already done). 

It would chop out so much code at the start of this file that is just duplicated work anyways.

mitosheet/src/components/taskpanes/DefaultTaskpane/DefaultTaskpaneFooter.tsx
```diff
@@ -0,0 +1,25 @@
...
+const DefaultTaskpaneFooter = (
+    props: {
+        /** 
+            * @param children - The actual content to display in the body
+        */
+        children: ReactNode
+    }): JSX.Element => {
+
+    return (
+        <div>
```
naterush, line 19: No classes on that, is this correct?

mitosheet/src/components/taskpanes/DefaultTaskpane/DefaultTaskpaneFooter.tsx
```diff
@@ -0,0 +1,25 @@
...
+import "../../../../css/taskpanes/DefaultTaskpane.css"
+
+/*
+    A container for the footer content of a taskpane. Helpful for ensuring the footer is always 
+    visible even if the body needs to scroll.
+*/
+const DefaultTaskpaneFooter = (
+    props: {
+        /** 
+            * @param children - The actual content to display in the body
```
naterush, line 13: Comment.

mitosheet/src/components/elements/MultiSelectButtons.tsx
```diff
@@ -0,0 +1,41 @@
...
+    values: string[];
+    selectedValues: string[];
+    onChange: (newValue: string) => void;
+}
+
+/**
+ * The MultiSelectButtons component displays a set of options and allows
+ * the user to select as many as they'd like
+ */ 
+const MultiSelectButtons = (props: MultiSelectButtonsProps): JSX.Element => {
```
naterush, line 16: Ok, a few thoughts on this element.

We have learned over time that making the _value_ you need to display the same as the value you need in the callbacks adds a lot of boilerplate code. You can see this through the way that you naturally ended up storing the delimiter values. 

Instead, we normally allow you to separate an id from the title ( you can see this in the DropdownItems anywhere we use column ids). And note here, creating a MultiSelectButtonItem would be the natural way to do this well.

This leads to a more complex component design, but a lot simpler consumption of these components. Imagine the following code:

```
// NOTE: the delimiters are now stateless. E.g. there is no notion of included within the delimeter
// definition (which is very sensible, we want a definition of just what delimiters we support - not // what we support and if they are included in one variable).
const delimiters = [{',': 'Comma', ' ': 'Space'}]

interface SplitTextToColumnParams {
    sheet_index: number,
    column_id: ColumnID,
    delimiters: string[]
}

const {params, setParams} = useSendEditOnClick(...);


<MultiToggleButtonContainer onToggle={(newDelimiters) => {
     setParams(prevParams => {
         return {
             ...prevParams,
             delimiters: newDelimiters
         }
     })

}}>
     {Object.entries(delimiters).map(([delimiter, delimiterTitle]) => {
          return (
              <MultiToggleButtonItem
                   id={delimiter}
                   title={delimiterTitle}
                   toggled={params.delimiters.includes(delimiter)}
              />
          )
     })}
</MultiToggleButtonContainer>
<Input placeholder="other" onChange={(e) => {
    const newValue = e.target.value;
    setParams(prevParams => {
        const newDelimiters = [...prevParams.delimiters].filter(delimiter => Object.keys(delimiters.includes(delimiter));
        newDelimiters.push(newValue);
        return {
            ...prevParams,
            delimiters: newDelimiters
        }
    })
}}>
```
Of course, this is a sketch of the code I think would work - I'm sure you'll have to adapt to figure out how to do this best. Give me a call if you want to chat implementation.

This has a couple of really important benefits:
1. It removes more state from the taskpane except that defined through our new hook. This makes refactoring like _10x_ easier. E.g. for example look at my drop duplicates refactor, vs. the fact I gave up on the excel import refactor b/c it doesn't use the new hook, and has a bunch of random state!
2. It is much closer to the designs of our other components. This is a very similar interface to what we use elsewhere, which will make using this much more of a breeze.
3. It would lead to _more_ code in the components, and less in the taskpane. Whenever we can do this, it's a good idea!

I think these changes are important to do, for the above reasons. I know they are pretty big, so take it in pieces. Do not crush yourself trying to get these done. We can just push back the release date on this feature - marathon not a sprint!

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,302 @@
...
+    }
+    
+    if (params === undefined || params.column_id === undefined) {
+        return (<DefaultEmptyTaskpane setUIState={props.setUIState}/>)
+    }
+
+    return (
+        <DefaultTaskpane>
+            <DefaultTaskpaneHeader 
+                header="Split Text to Column"
```
naterush, line 148: to Columns

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,302 @@
...
+                        <Select
+                            width='medium'
+                            value={props.dfNames[params.sheet_index]}
+                            // Safe to cast as dfNames are strings
+                            onChange={(newSheet: string) => {
+                                setParams(prevParams => {
+                                    const newSheetIndex = props.dfNames.indexOf(newSheet)
+                                    return {
+                                        ...prevParams,
+                                        sheet_index: newSheetIndex
```
naterush, line 168: I think we should change the column id here, rather than doing all this work at the top. Just update to the default selected column id in the new sheet (which is the first?).

mitosheet/src/jupyter/api.tsx
```diff
@@ -503,6 +503,42 @@ export default class MitoAPI {
...
+                'sheet_index': sheetIndex,
+                'column_id': columnID,
+                'delimiters': delimiters,
+            },
+        }, {})
+
+        if (dfPreviewString !== undefined && dfPreviewString !== '') {
+            const dfPreviewSheetData: SheetData = JSON.parse(dfPreviewString)['dfPreviewSheetData'];
+
+            const dfPreviewArray: (string | number | boolean)[][] = []
```
naterush, line 528: It would be nice if we have a more ergonomic way to get data to the frontend, given that we do this sort of parsing in a few places. 

What would a good type be for this and the unique values? I'm thinking this map:
```
ColumnDataMap: Record<ColumnID => (string | number | boolean)[]>
```

Wanna take a shot at refactoring these two api calls to use this? It would just require writing a new backend and frontend utils (and in return, all this special case code would go away). 

It will be much easier to work with, much cleaner, and much better for maintainability long-term I think!

naterush (changes requested):
Just did a code review for now. There are few somewhat large architectural things I think we would really benefit from changing. 

Give me a call if you want to chat about these. I have some more thoughts generally on process that could move these up in the future, both for you alone and from my feedback, but we can chat about these later!

mitosheet/mitosheet/code_chunks/step_performers/column_steps/split_text_to_columns_code_chunk.py
```diff
@@ -0,0 +1,49 @@
...
+    def get_display_name(self) -> str:
+        return 'Split text to columns'
+    
+    def get_description_comment(self) -> str:
+        sheet_index = self.get_param('sheet_index')
+        column_id = self.get_param('column_id')
+        delimiters = self.get_param('delimiters')
+        column_header = self.post_state.column_ids.get_column_header_by_id(sheet_index, column_id)
+        delimiters_string = (' ,').join(delimiters)
+        return f'Split {column_header} on {delimiters_string}'
```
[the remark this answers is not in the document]
aarondr77: Nah. I mean, weird is subjective, but its the correct delimiter construction.

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
aarondr77: Ooo it actually looks like we don't have a standard. When setting the column formula, for example, we use the first column and when opening the column control panel, we use the second column. 

I'm going to try to unify it to the first column, since that is what Excel does.

aarondr77: We now always default to the top left corner of the last selection.

mitosheet/src/jupyter/api.tsx
aarondr77: Nice. I made a new util. But the format that is most helpful for these cases is row-organized, instead of column organized, since the transformation that we end up performing in api.tsx is converting the data from column-first, to row-first. So check out the new util `get_row_data_array` in `utils.py`

mitosheet/src/components/elements/MultiSelectButtons.tsx
aarondr77: Done!

mitosheet/src/components/elements/MulitSelectButtonItem.tsx
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Mito
+import React from 'react';
+import { classNames } from '../../utils/classNames';
+
+interface MultiSelectButtonItem {
+    /** 
+        * @param id - If you want to display something other than what you actually store 
+        * (e.g. display a columnHeader but store a columnID), then set the ID equal to the value 
+        * that you want to store, and it will be used for the onChange value in the MultiSelectButtons.
```
aarondr77, line 9: Not used in the onChage. Update to new design

mitosheet/mitosheet/utils.py
```diff
@@ -6,7 +6,9 @@
 """
 Contains helpful utility functions
 """
+from ctypes import Union
```
aarondr77, line 9: This is the wrong import

mitosheet/src/jupyter/api.tsx
```diff
@@ -484,17 +484,18 @@ export default class MitoAPI {
...
-                    percentOccurence: (uniqueValueCountsObj.uniqueValueCountsSheetData.data[1].columnData[i] as number) * 100,
-                    countOccurence: (uniqueValueCountsObj.uniqueValueCountsSheetData.data[2].columnData[i] as number),
+                    value: uniqueValueCountsObj.uniqueValueRowDataArray[i][0],
+                    percentOccurence: (uniqueValueCountsObj.uniqueValueRowDataArray[i][1] as number) * 100,
+                    countOccurence: (uniqueValueCountsObj.uniqueValueRowDataArray[i][2] as number),
                     isNotFiltered: true
                 })
             }
 
+            console.log(uniqueValueCounts)
```
aarondr77, line 498: remove

mitosheet/mitosheet/step_performers/column_steps/split_text_to_columns.py
```diff
@@ -58,15 +59,20 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
         # Actually execute the column reordering
         pandas_start_time = perf_counter() 
         # Create the dataframe of new columns. We do this first, so that we know how many columns get created.
-        new_columns_df = final_df[column_id].astype('str').str.split(delimiter_string, -1, expand=True)
+        if is_datetime_dtype(str(post_state.dfs[sheet_index][column_header].dtype)):
+            print('is datetime ')
```
aarondr77, line 63: remove

aarondr77:
@naterush There is weird behavior that occurs with the split function and missing values. 

For starters, depending on the type of column, pandas treat `None` values either as `None` or `NaN`. See here: 
Screen Shot 2022-05-25 at 7 29 31 PM

naterush:
@aarondr77 hmm. nice investigation. That's very strange behavior.

 I think we can leave it for now though. NaN values in string columns do happen, but feel more rare in string columns as far as I have observed. I don't think it's worth the extra line of code...

Let me know when you want a final review!

naterush:
1. Add a column
2. Open the Split Text to Column Taskpane
3. Set a delimiter
4. Press undo (to undo the column creation)
5. The sheet crashes.

Screen Shot 2022-05-26 at 9 04 12 AM

naterush:
Elsewhere, we refer to the `Sheet` as the `Dataframe`. Sometimes even as `Data Source` as in Pivot. 

I think if we can standardize this, let's do it. I think we should either use `Dataframe` - thoughts?

naterush:
Styling change request: the fact that the selects and the delimiter selection isn't lined up made it very hard for me to parse the taskpane when I first saw it. 

I did some manual css messing about to get them to line up -- check out the result. I think it looks so much better. Can we move to this? 

Might require some more manual css, but I think it's worth it. IMO it's much prettier, but more importantly is much easier to read when you first see it.

Screen Shot 2022-05-26 at 9 12 11 AM


Screen Shot 2022-05-26 at 9 15 01 AM

naterush:
Ok some thoughts on the column header names that get created:
1. Can we move the column header to the start of the string so it's easily visible and looks similar to the column header next to it? Think this would 
2. Can we make the prefix a number rather than a string? It is less visually noisy.

Peep these changes. They look so much better IMO!

Screen Shot 2022-05-26 at 9 24 52 AM


Screen Shot 2022-05-26 at 9 25 21 AM

naterush:
The fact that this works on dates is epic!

mitosheet/src/utils/actions.tsx
```diff
@@ -894,6 +884,23 @@ export const createActions = (
...
+                setUIState(prevUIState => {
+                    return {
+                        ...prevUIState,
+                        currOpenModal: {type: ModalEnum.None},
+                        currOpenTaskpane: {type: TaskpaneType.SPLIT_TEXT_TO_COLUMNS, startingColumnID: startingColumnID}
+                    }
+                })
+            },
+            isDisabled: () => {return doesAnySheetExist(sheetDataArray) ? undefined : 'There are no columns to split. Import data.'},
+            searchTerms: ['split', 'extract', 'parse', 'column', 'splice', 'text'],
```
naterush, line 901: Add: delimiter, comma, space, dash

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,287 @@
...
+    )
+
+    const [preview, setPreview] = useState<(string | number | boolean)[][] | undefined>([])
+
+    // When the startingColumnID is updated outside of the taskpane, set it as the column getting split
+    useEffect(() => {
+        setParams(prevParams => {
+            return {
+                ...prevParams,
+                column_id: props.startingColumnID !== undefined ? props.startingColumnID : props.sheetDataArray[props.gridState.sheetIndex].data[0].columnID,
```
naterush, line 63: What if sheetData array is undefined? This will cause a sheet crashing bug. I think you can just steal the logic from the effect like this in fill nan:

if props.startingColumnID is defined, we change the params, and otherwise we don't.

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,287 @@
...
+                            value={props.dfNames[params.sheet_index]}
+                            // Safe to cast as dfNames are strings
+                            onChange={(newSheet: string) => {
+                                setParams(prevParams => {
+                                    const newSheetIndex = props.dfNames.indexOf(newSheet)
+                                    return {
+                                        ...prevParams,
+                                        sheet_index: newSheetIndex,
+                                        // Default to the first column in the new sheet
+                                        column_id: Object.keys(props.sheetDataArray[newSheetIndex].columnIDsMap)[0]
```
naterush, line 123: sheet crashing bugs abound here! Careful. 

```
Object.keys(props.sheetDataArray[newSheetIndex]?.columnIDsMap || {})[0]
```

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,287 @@
...
+    num_cols_created: number;
+}
+
+const delimiters = {',': 'Comma', '-': 'Dash', '\t': 'Tab', ' ': 'Space'}
+
+/* 
+    This taskpane allows users to split a column into multiple columns 
+    by separating on a delimeter
+*/
+const SplitTextToColumnsTaskpane = (props: SplitTextToColumnsTaskpaneProps): JSX.Element => {
```
naterush, line 42: Really happy with how this taskpane looks! It's really stupendously clean - I think that refactor was def the right call. Nice work getting that done!

mitosheet/src/components/taskpanes/ControlPanel/ControlPanelTaskpane.tsx
```diff
@@ -54,7 +54,7 @@ type ControlPanelTaskpaneProps = {
 export const ControlPanelTaskpane = (props: ControlPanelTaskpaneProps): JSX.Element => {
     
     // Get the values for the last cell that was selected
-    const {columnHeader, columnID, columnFormula, columnFilters, columnDtype, columnFormatType} = getCellDataFromCellIndexes(props.sheetData, props.selection.endingRowIndex, props.selection.endingColumnIndex);
+    const {columnHeader, columnID, columnFormula, columnFilters, columnDtype, columnFormatType} = getCellDataFromCellIndexes(props.sheetData, props.selection.startingRowIndex, props.selection.startingColumnIndex);
```
naterush, line 57: Why do you do the starting cell here? The comment makes it seem like we want to do the last?

mitosheet/mitosheet/utils.py
```diff
@@ -237,6 +212,49 @@ def df_to_json_dumpsable(
...
 
 
+def get_row_data_array(df: pd.DataFrame) -> List[Any]:
+    """
+    Returns just the data of a dataframe in the 2d array format of [row idx][col idx]
+    """
+    json_obj = convert_df_to_json(df)
+    return json_obj['data']
+
+def convert_df_to_json(df: pd.DataFrame) -> Dict[str, Any]:
```
naterush, line 222: This is... a fucking great function. What took us so long? Nice work pulling this out.

naterush: Maybe name it `convert_df_to_parsed_json` though?

mitosheet/mitosheet/code_chunks/step_performers/column_steps/split_text_to_columns_code_chunk.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def get_code(self) -> List[str]:
+        sheet_index = self.get_param('sheet_index')
+        column_id = self.get_param('column_id')
+        delimiters = self.get_param('delimiters')
+
+        # The type system says that execution_data can be None, so we need to appease it
+        # TODO: Update the type system so that the execution data is not None when we know it is defined for a step
+        new_column_headers = self.execution_data['new_column_headers'] if self.execution_data is not None else '1'
```
naterush, line 33: There is a self.get_execution_data function I think! Use this rather than indexing in directly.

mitosheet/mitosheet/code_chunks/step_performers/column_steps/split_text_to_columns_code_chunk.py
```diff
@@ -0,0 +1,60 @@
...
+from mitosheet.transpiler.transpile_utils import column_header_list_to_transpiled_code, column_header_to_transpiled_code
+
+
+class SplitTextToColumnsCodeChunk(CodeChunk):
+
+    def get_display_name(self) -> str:
+        return 'Split text to columns'
+    
+    def get_description_comment(self) -> str:
+        sheet_index = self.get_param('sheet_index')
```
naterush, line 19: Everywhere you access params (here and elsewhere), might as well add types to them? If there other steps you see not doing this, go ahead and add them :-)

mitosheet/mitosheet/code_chunks/step_performers/column_steps/split_text_to_columns_code_chunk.py
```diff
@@ -0,0 +1,60 @@
...
+# Copyright (c) Saga Inc.
+# Distributed under the terms of the GPL License.
+
+from typing import List
+from mitosheet.code_chunks.code_chunk import CodeChunk
+from mitosheet.sheet_functions.types.utils import is_datetime_dtype, is_string_dtype
+from mitosheet.transpiler.transpile_utils import column_header_list_to_transpiled_code, column_header_to_transpiled_code
+
+
+class SplitTextToColumnsCodeChunk(CodeChunk):
```
naterush, line 13: Code looks great in this file. Nice work.

mitosheet/mitosheet/api/get_unique_value_counts.py
```diff
@@ -90,16 +90,7 @@ def get_unique_value_counts(params: Dict[str, Any], steps_manager: StepsManagerT
...
-            unique_value_counts_df, 
-            'value counts',
-            'imported',
-            {},
-            {},
-            {'values': 'values', 'percents': 'percents', 'counts': 'counts'},
-            {'values': {'type': FORMAT_DEFAULT}, 'percents': {'type': FORMAT_DEFAULT}, 'counts': {'type': FORMAT_DEFAULT}},
-            max_length=None
-        ),
+        'uniqueValueRowDataArray': get_row_data_array(unique_value_counts_df),
```
naterush, line 93: Very nice!

mitosheet/css/taskpanes/SplitTextToColumns/SplitTextToColumns.css
```diff
@@ -0,0 +1,22 @@
...
+
+td {
+    border: 1px solid #ddd;
+    border-width: 0 1px 0 1px;
+    padding: 4px;
+}
+  
+tr:nth-child(even) {
+    background-color: #f2f2f2;
+}
```
naterush, line 22: Can you make these classes just so they don't accidently go and apply elsewhere?

mitosheet/mitosheet/step_performers/column_steps/split_text_to_columns.py
```diff
@@ -0,0 +1,117 @@
...
+from mitosheet.code_chunks.code_chunk import CodeChunk
+from mitosheet.code_chunks.step_performers.column_steps.split_text_to_columns_code_chunk import SplitTextToColumnsCodeChunk
+from mitosheet.column_headers import try_make_new_header_valid_if_multi_index_headers
+from mitosheet.sheet_functions.types.utils import is_datetime_dtype
+from mitosheet.state import FORMAT_DEFAULT, State
+from mitosheet.step_performers.step_performer import StepPerformer
+from mitosheet.step_performers.utils import get_param
+from mitosheet.types import ColumnHeader, ColumnID
+
+def get_new_colum_header_unique_component() -> str:
```
naterush, line 23: Unused now?

naterush: Or if test utility, perhaps move it there.

mitosheet/mitosheet/step_performers/column_steps/split_text_to_columns.py
```diff
@@ -0,0 +1,117 @@
...
+        delimiters: List[str] = get_param(params, 'delimiters')
+        new_column_header_suffix: str = get_param(params, 'new_column_header_suffix')
+
+        column_header = prev_state.column_ids.get_column_header_by_id(sheet_index, column_id)
+        column_idx = prev_state.dfs[sheet_index].columns.tolist().index(column_header)
+            
+        # Create a new post state
+        post_state = prev_state.copy(deep_sheet_indexes=[sheet_index])
+        final_df = post_state.dfs[sheet_index]
+        delimiter_string = '|'.join(delimiters)
```
naterush, line 58: What happens if we're splitting on a bar? TODO: test this.

naterush (commented):
This code looks pretty fantastic. A few minor comments, but nothing major.

naterush:
Can you add tests for timedetlas (and add a similar special case to date times if you need to).

Screen Shot 2022-05-26 at 9 49 52 AM

naterush:
I'm getting very laggy side-to-side scroll in Endo on Windows. Not sure what the deal is here.

Gonna check on dev.

naterush:
Screen Shot 2022-05-26 at 10 04 06 AM

Scroll bars are messed up windows - overflow on the sides and the bottom scrollbar is cut off.

mitosheet/mitosheet/step_performers/column_steps/split_text_to_columns.py
aarondr77: It seems to have the same issue as `.` do

naterush:
I think we can ignore formatting for now tbh. Splitting on number columns is, as far as I imagine, going to incredibly rare - so I think it's pretty much not worth it!

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,287 @@
...
+    mitoAPI: MitoAPI;
+    analysisData: AnalysisData;
+    sheetDataArray: SheetData[];
+    gridState: GridState;
+    setUIState: React.Dispatch<React.SetStateAction<UIState>>;
+    dfNames: string[];
+    startingColumnID: ColumnID | undefined
+}
+
+interface SplitTextToColumnsResult {
```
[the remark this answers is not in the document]
aarondr77: This is already in the taskpane .tsx file. Is that not what you're looking for?

mitosheet/mitosheet/tests/steps_performers/column_steps/test_split_text_to_columns.py
```diff
@@ -0,0 +1,148 @@
...
+        'A',
+        [' ', '-'],
+        pd.DataFrame({'A': pd.to_datetime(['1/2/22', '4/06/99']), 'A-split-0-1': ['2022', '1999'], 'A-split-1-1': ['01', '04'], 'A-split-2-1': ['02', '06'], 'A-split-3-1': ['00:00:00', '00:00:00']})
+    ),
+    (
+        pd.DataFrame({'A': [pd.to_datetime('3/07/21') - pd.to_datetime('1/12/22')]}),
+        'A',
+        [' '],
+        pd.DataFrame({'A': [pd.to_datetime('3/07/21') - pd.to_datetime('1/12/22')], 'A-split-0-1': ['-311'], 'A-split-1-1': ['days'], 'A-split-2-1': ['+00:00:00']})
+    ),
```
naterush, line 93: Missing tests for timedelta.

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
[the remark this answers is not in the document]
naterush: The Params type could be here too -- but it's fine for now.

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,315 @@
...
+    num_cols_created: number;
+}
+
+const delimiters = {',': 'Comma', '-': 'Dash', '\t': 'Tab', ' ': 'Space'}
+
+const getDefaultParams  = (startingColumnID: ColumnID | undefined, sheetDataArray: SheetData[], sheetIndex: number): SplitTextToColumnsParams | undefined => {
+    if (sheetDataArray.length === 0 || sheetDataArray[sheetIndex] === undefined && sheetIndex > 0) {
```
naterush, line 39: you need to add some parens on this statement, its' very risky as is. The second clause after the or should always be logically grouped (right now you're relying on JS execution order that is fragile).

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,315 @@
...
+            void loadSplitTextToColumnsPreview()
+        } else {
+            setPreview(undefined)
+        }
+    }, [params])
+    
+
+    async function loadSplitTextToColumnsPreview() {
+
+        if (params !== undefined && params.column_id !== undefined) {
```
naterush, line 98: It's weird you do this check in both the effect above and here. Why not do all the checks in the effect, or do all the checks in this function? Splitting them across these two is just confusing.

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,315 @@
...
+            setPreview(undefined)
+        }
+    }
+
+    if (params === undefined || params.column_id === undefined) {
+        if (params !== undefined) {
+            console.log('column_id: ', params.column_id)
+        } else {
+            console.log('params undefined')
+        }
```
naterush, line 117: Remove

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -0,0 +1,315 @@
...
+                            // Prepare for the user splitting again by creating a new column suffix,
+                            // so that the column headers don't overlap
+                            new_column_header_suffix: getNewColumnHeader()
+                        }
+                    })}
+                    disabled={params.delimiters.length === 0}
+                    disabledTooltip="Select at least one delimiter"
+                >
+                    {!editApplied 
+                        ? 'Split on delimiter'
```
naterush, line 295: Can this (and the text below) say "Split on delimiter(s)" - like add an s when there are multiple?

naterush (changes requested):
Very minor code cleanup things.

naterush:
Screen Shot 2022-05-27 at 8 58 35 AM

1. Add a column
2. Set formula
3. Split on it, do the split
4. Undo to get rid of the column being created. 

Still getting a sheet crashing bug here.

naterush:
Btw, the consistent sizes on the delimiter stuff looks... so freaking good. Nice!

mitosheet/mitosheet/tests/steps_performers/column_steps/test_split_text_to_columns.py
aarondr77: This is a timedelta test. I'm just creating a timedelta by subtracting the two datetimes.

aarondr77:
I can't replicate the bug with column editing. Am I recreating it wrong here? 

https://user-images.githubusercontent.com/18709905/170741025-1aa7198b-4f51-49f1-b15d-7715dcfe8ad2.mov

PR #307, opened 2022-05-18
Copy and paste

naterush:
# Description

This PR adds copy and paste to Mito. A few notes about this:
1. One of the major things I had to do was focus refactoring. Now, pretty much no matter where you click in Mito, you'll remain focused on the mito object. The major refactor here was fixing up when / how the dropdowns display.
2. Since we have good focus handling now, check out the shortcuts I added to the menus... they are... kinda sicko mode.

There are two main things to decide:
1. How to handle \t values. 
3. How to handle NaN values.

My thinking on (1) is that a) I don't know how often this occurs and b) I tried and failed to handle it for about 3 hours. I can spend more time here, but my concern is that b/c I don't really understand what I'm doing, I'm as likely to break things with this fix as I am to fix it. Same with 2. Let me know your thoughts.

We'll get these fixed up (or decide not to) before we merge this, so this is kinda an incremental review rather than a final one!

# Testing

Import data, do some pivoting or whatever, and copy and paste the data out! Try this with a few different datasets and all the different datatypes!

# Documentation

Need to write documentation for copy and paste, probably!


mitosheet/src/components/Mito.tsx
```diff
@@ -423,6 +424,38 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
+                return;
+            }
+
+            // Then, we check if the user is actually focused on some input in Mito,
+            // as in this case we don't want to copy from the sheet
+            if (document.activeElement?.tagName.toLowerCase() === 'input') {
+                return;
+            }
+
+            e.stopImmediatePropagation();
```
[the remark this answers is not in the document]
naterush: Also comment the plugin file

mitosheet/src/components/endo/GridData.tsx
```diff
@@ -43,7 +43,7 @@ const GridData = (props: {
                                 return null;
                             }
 
-                            const className = classNames('mito-grid-cell', {
+                            const className = classNames('mito-grid-cell', 'text-unselectable', {
```
naterush, line 46: Make the column headers and the index unelectable.

mitosheet/src/hooks/useDebouncedEffect.tsx
```diff
@@ -3,10 +3,25 @@ import { useEffect } from "react";
...
 */
-export const useDebouncedEffect = (effect: () => void, deps: unknown[], delay: number): void => {
+export const useDebouncedEffect = (
+    effect: () => (void | (() => void)), deps: unknown[], 
+    delay: number,
+): void => {
     useEffect(() => {
-        const handler = setTimeout(() => effect(), delay);
+        // Cleanup defaults to a noop, but you can also return a function 
+        // from the effect to have it run on cleanup
```
naterush, line 12: Note that this is just like the useEffect

mitosheet/src/types.tsx
```diff
@@ -779,6 +781,12 @@ export interface Action {
 
     // Optionally categorize the action, so it can easily be sorted later
     category?: 'spreadsheet formula'
+
+    // Optionally give the action a keyboard shortcut to display with it
```
naterush, line 785: this comment is terrible

mitosheet/src/utils/actions.tsx
```diff
@@ -175,6 +176,42 @@ export const createActions = (
...
+                    gridState.selections
+                );
+
+                if (stringToCopy === undefined) {
+                    return;
+                }
+                
+                navigator.clipboard.writeText(stringToCopy).then(function() {
+                    /* clipboard successfully set */
+                    // TODO: do we want to do some reporting to the user here?
```
naterush, line 196: TODO: look into these!

naterush:
Oh, the other main thing: we don't support Cmd + Z or Cmd + Y on notebooks. I can spend another few hours trying to fix this, but I'd like to do it in another PR, so let's just remove the keyboard shortcut hints from undo and redo if we want to put time into this (I vote yes!). 

Let me know what you think.

aarondr77:
> so let's just remove the keyboard shortcut hints from undo and redo if we want to put time into this (I vote yes!).

That sounds good to me. If the focus is working better, then I'd be super down to invest more in keyboard shortcuts, especially now that items are more clicks away in the toolbar.

naterush:
NOTE: this is the sort of PR that makes a lot of assumptions about the systems users are on. You don't have to do it in testing this PR (although def test on windows), but before releasing, I'll be testing on all major browsers on windows, etc. We'll also do a my browser test too!

aarondr77:
The fact that creating a selection using the Shift key doesn't work well, seems like a much bigger problem now that I vote we fix. 

In order for the selection to work, you need to click outside of the data. 
1. If you don't click, then the selection just follows your cursor around
2. If you click insider of the data, then the selection just gets set as the cell you click on. 

https://user-images.githubusercontent.com/18709905/169867573-7f7183fd-ca2e-43d7-be36-c90454fe7d43.mov

aarondr77:
Selecting everything in the sheet, including the index column doesn't seem to copy at all. (What ended up copying was my previously copied selection)


https://user-images.githubusercontent.com/18709905/169870935-bda7ae8e-90ca-4456-bef8-8f265fea1b14.mov

aarondr77:
The formatting that dates get copied in is different than what we display in the sheet. The date `2022-01-22 00:00:00` in Mito gets copied as `1/22/2022  12:00:00 AM` in the Excel formula bar. Note: Excel displays it differently in a the cell when the cell editor is not open `1/22/2022  0:00`

aarondr77:
The keyboard shorcut overlaps with the scrollbar in the help dropdown. A good fix to this might be using the scollbar-gutter approach that we used to fix the footer scrollbar. Or, for a less robust solution, just changing the margins. 

help-menu

aarondr77:
Weird bug I'm experiencing on windows, JLab. 

Open the values tab on the `Temp` column of the `Forestfires` dataset in `small-datasets`. The loading indicators sit there for a while and the data doesn't load. When I switch to `main` or `dev` though, this doesn't occur. It loads instantly (as there are only about ~500 rows). 

This is repeatable.

mitosheet/src/hooks/useDebouncedEffect.tsx
```diff
@@ -3,10 +3,25 @@ import { useEffect } from "react";
...
 */
-export const useDebouncedEffect = (effect: () => void, deps: unknown[], delay: number): void => {
+export const useDebouncedEffect = (
+    effect: () => (void | (() => void)), deps: unknown[], 
+    delay: number,
+): void => {
     useEffect(() => {
-        const handler = setTimeout(() => effect(), delay);
+        // Just like useEffect, you can return a result function that runs on
+        // cleanup. We just default here to a noop.
```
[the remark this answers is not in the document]
naterush: it's also called noop!

mitosheet/src/components/elements/Dropdown.tsx
```diff
@@ -194,7 +200,20 @@ const Dropdown = (props: DropdownProps): JSX.Element => {
...
+        if (!props.display) {
+            return;
+        }
+
+        // Close the dropdown
+        props.closeDropdown();
+
+        // Refocus on the div that is the parent of the dropdown
+        // so that users are focused where they expect
+        dropdownAnchor.current?.focus();
```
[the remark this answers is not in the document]
naterush: Fixing this is a huge lift for now. We can continue to improve these practices, but I'm comfortable with where this is at so far...

naterush:
@aarondr77 ok, review is addressed to the best of my abilities. The things I want to highlight in this requested rereview:
1. Thoughts on the border for copying? Feel free to change up the styling (it's just defined in a single place, so easy to change if you think something else looks better). 
2. The selection changes you requested are pretty much not doable, to my knowledge. It's a JLab limitation that I spent at least a full day last week trying to work around in another context - but idk what it is. I vote we open an issue and leave it for now.

Let me know if there's anything else you want me to try again on!

aarondr77:
The feedback with the border is probably good enough. Getting it to [dance](https://stackoverflow.com/questions/28365839/dashed-border-animation-in-css3-animation) would be nice, and would be easy if we were applying a class instead of a style, but its a bit of refactoring to get there. So I'm cool with leaving it as is. 

The one thing I think we should change though, is that if I change my selection after copying, the border should probably go away. Right now, the only way to get the copy border to go away is by copying something else. 

https://user-images.githubusercontent.com/18709905/170155111-159b3f17-7c1f-4fe8-8204-ffa7c21df8b1.mov

aarondr77:
Other than that, LGTM

aarondr77:
Other than that, LGTM!

naterush:
The tests don't run when there is merge conflicts - but this is tested as you expect!

PR #310, opened 2022-05-23
Error reducing sheet crashing 5 23

naterush:
# Description

Fixes two bugs:
1. Issues with not letting errors through (making it impossible to dig in on sheet crashing errors). 
2. Reduces a line of code from all pivot steps, as there is a much better way of doing things.
3. Does the same thing as this pr https://github.com/mito-ds/monorepo/pull/304, subsuming it (just wanted it in a variable, otherwise was great).

# Testing

Check the new params (cell_editor_location, error_message, error_name, error_stack, user_agent) are public (this can be hard to text with sheet crashing bugs, but check they are at least in the map), and also the new pivot code!

# Documentation

N/A.


aarondr77:
LGTM, but holding off for your queue since its deploy day

PR #312, opened 2022-05-23
Fill nan usability

naterush:
# Description

Makes changes here: https://www.notion.so/trymito/Jake-DR-Fill-NaN-79e1ddaee6e44a5eb050ed2ce2c38eb3

# Testing

Check the changes listed above!

# Documentation

N/A.

### DIFF
--- a/mitosheet/src/components/taskpanes/FillNa/FillNaTaskpane.tsx
+++ b/mitosheet/src/components/taskpanes/FillNa/FillNaTaskpane.tsx
@@ -20,7 +20,7 @@ import MultiToggleBox from '../../elements/MultiToggleBox';
 import MultiToggleItem from '../../elements/MultiToggleItem';
 import { getDisplayColumnHeader, getFirstCharactersOfColumnHeaders } from '../../../utils/columnHeaders';
 import { getDtypeValue } from '../ControlPanel/FilterAndSortTab/DtypeCard';
-import { addIfAbsent, removeIfPresent } from '../../../utils/arrays';
+import { addIfAbsent, intersection, removeIfPresent } from '../../../utils/arrays';
 import Spacer from '../../spacing/Spacer';
 import Input from '../../elements/Input';
 import { isDatetimeDtype, isNumberDtype, isTimedeltaDtype } from '../../../utils/dtypes';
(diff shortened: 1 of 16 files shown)


mitosheet/src/components/taskpanes/FillNa/FillNaTaskpane.tsx
```diff
@@ -75,9 +80,12 @@ const getDefaultParams = (sheetDataArray: SheetData[], sheetIndex: number, defau
         }
     }
 
+    const columnIDs = startingColumnIDs === undefined ? Object.keys(sheetData.columnIDsMap) : intersection(Object.keys(sheetData.columnIDsMap), startingColumnIDs);
+    console.log(startingColumnIDs)
```
naterush, line 84: Remove

naterush (commented):
Cleanup console

aarondr77:
I like the idea of defaulting to selecting the column that you select through the column header dropdown, but in practice I think its confusing. I bet most people would see this and think that no columns are selected. 

A few ideas:
1. Scroll the multi toggle box to the selected column so you can tell that it is selected
2. Have a message that says which columns are selected. We could reuse the utility we created for concatting dataframes. 
3. Add the column header that was selected to the search bar so its the only one showing. This doesn't seem great because it makes it harder for users to find the other columns. 
4. Just don't do this, and default to having everything selected. 

I think option 1 is the best solution. 

https://user-images.githubusercontent.com/18709905/169877090-4ef58da4-0fad-4e50-aabb-1a58d2bc54f4.mov

aarondr77:
We're incorrectly anonymizing the method that we use to fill. Since we implemented the log anonymization PR, we've done this 2-3 times. Would be nice to add a check for this in the new release process. We could go back to trying to make dashboards that gives us a look into the effectiveness of the PR, which features are being used, etc. 

It could be nice to have a dashboard for each feature that answers these basic questions. And if we include the creation of them in the PR, then it will make sure that our logs are correct. 

Screen Shot 2022-05-23 at 1 49 44 PM

aarondr77:
Can we add a thing right below the multi toggle box that says how many are selected?

Screen Shot 2022-05-23 at 1 54 06 PM

aarondr77:
I was using the Toggle All Matching button and thought it was a bug that the 1 was not changing to 0. I think I would've realized the intent of this message if the language matched. So can we change from `Toggle All Matching` to `Toggle All Displayed` or move both to Matching. 

Screen Shot 2022-05-23 at 1 55 25 PM

aarondr77:
Once those few things are fixed up, LGTM.

PR #325, opened 2022-05-26
mitosheet: save scroll positions in sheets

naterush:
# Description

Saves the scroll positions in sheets. Like every other application every person has ever used.

# Testing

Make sure this works with adding, removing, deleting sheets, etc.

# Documentation

No.


aarondr77:
1. Import the RockArt 600 column datasets in our large dataset.
2. Duplicate it
3. In the first sheet, scroll over to the right a bunch
4. Delete the first sheet
5. Notice that the second sheet (which is now the first sheet) is scrolled to the same position that the first sheet was in.

aarondr77:
Maybe clear analysis should reset the horizontal scrolls too? It feels more resetty

aarondr77:
Other than that LGTM

naterush:
@aarondr77 just like with width data, because we don't have sheet ids, figuring out the saving of scroll with deleting is fucked. 

The same thing is true with column widths, but I've literally never noticed it before (and if anything this is a less big deal with column widths).

naterush:
Clearning an anlaysis is similarly very hard to fix for very little bang for buck. Can we forget about both of these?

aarondr77:
Yeah. forgot!

PR #326, opened 2022-05-26
mitosheet: move excel import to new taskpane hook

naterush:
# Description

Cleans up the old taskpane and moves it to the new hook, which is nice. It also fixes issues were users were confused about not being able to import multiple excel sheets.

Note that while there is more code, it is a) standard code, that is b) much easier to read, and c) much more maintainable!

# Testing

Text Excel import.

# Documentation

No.


naterush:
NOTE: this is branched off of split text to columns - so make sure to merge this into dev properly!

PR #327, opened 2022-05-26
mitosheet: add toggle all to values taskpane, unoptimized

naterush:
# Description

I think that as of now, the the that makes the most sense is an unoptimized toggle all in values. This would personally help me the most - I am dying for this feature almost half of the time I use Mito!

As it is currently, the multi toggle box only shows 1k values, so the limit on the size of generated code is somewhat existent. 

@aarondr77 if you think an optimized version of this makes more sense, then I can write a spec and do that). If we're gonna just end up going with the unoptimized version for now, let's not sit on this - not having it makes Mito annoying to use for me!

# Testing

Use it.

# Documentation

No.


aarondr77:
I created a spec for how I think we should implement this. Check it https://www.notion.so/trymito/Toggle-All-in-Values-bc5211451bbb4ce7844893f9b0d007a4

PR #332, opened 2022-05-29
mitosheet: fix the sheet tab action from closing the rename

aarondr77:
# Description

Fixes https://github.com/mito-ds/monorepo/issues/331


naterush:
1. Right click on column header
2. Click rename
3. Rename opens and closes

Is this related? seems like it might be...

PR #342, opened 2022-06-02
Promote row to header

naterush:
# Description

This PR implements this spec: https://www.notion.so/trymito/Promote-Row-to-header-6bcc410eb9154ea3a13a0d74a33d81a2

With the meta-programming code that had already been written, this took 30 minutes start to finish. Get ready for a whole new world of productivity gain on new step types!

# Testing

Just test it.

# Documentation

Need to write docs for this, and delete row.

### DIFF
--- a/mitosheet/mitosheet/step_performers/promote_row_to_header.py
+++ b/mitosheet/mitosheet/step_performers/promote_row_to_header.py
@@ -9,8 +9,6 @@ from time import perf_counter
 from typing import Any, Dict, List, Optional, Set, Tuple
 from mitosheet.code_chunks.code_chunk import CodeChunk
 from mitosheet.code_chunks.promote_row_to_header_code_chunk import PromoteRowToHeaderCodeChunk
-from mitosheet.errors import make_invalid_promote_row_to_header_error
-from mitosheet.sheet_functions.types.utils import get_float_dt_td_columns
 
 from mitosheet.state import State
 from mitosheet.step_performers.column_steps.rename_column import rename_column_headers_in_state
@@ -39,11 +37,6 @@ class PromoteRowToHeaderStepPerformer(StepPerformer):
         post_state = prev_state.copy(deep_sheet_indexes=[sheet_index])
 
         pandas_processing_time = 0.0
-
-        # Throw a special error if the user is trying to use a row with datetimes or timedeltas
-        _, datetime_column_headers, timedelta_column_headers = get_float_dt_td_columns(post_state.dfs[sheet_index])
-        if len(datetime_column_headers) > 0 or len(timedelta_column_headers) > 0:
-            raise make_invalid_promote_row_to_header_error(datetime_column_headers, timedelta_column_headers)
 
         for old_column_header, new_column_header in zip(post_state.dfs[sheet_index].columns, post_state.dfs[sheet_index].iloc[index]):
 
--- a/mitosheet/mitosheet/step_performers/promote_row_to_header.py
+++ b/mitosheet/mitosheet/step_performers/promote_row_to_header.py
@@ -8,13 +8,14 @@
 from time import perf_counter
 from typing import Any, Dict, List, Optional, Set, Tuple
 from mitosheet.code_chunks.code_chunk import CodeChunk
-from mitosheet.code_chunks.promote_row_to_header import PromoteRowToHeaderCodeChunk
+from mitosheet.code_chunks.promote_row_to_header_code_chunk import PromoteRowToHeaderCodeChunk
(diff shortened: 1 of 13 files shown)


aarondr77:
Try using this on the `forestfires` dataset see this error
Screen Shot 2022-06-03 at 5 01 42 PM

aarondr77:
Can we update the comment, "# Promoted 3 to header in Sheet_name_1" is a bit confusing. Maybe just "# Promoted row 3 to header in Sheet_name_1"

mitosheet/mitosheet/step_performers/promote_row_to_header.py
```diff
@@ -0,0 +1,83 @@
...
+
+        # We make a new state to modify it
+        post_state = prev_state.copy(deep_sheet_indexes=[sheet_index])
+
+        pandas_processing_time = 0.0
+
+        # Throw a special error if the user is trying to use a row with datetimes or timedeltas
+        _, datetime_column_headers, timedelta_column_headers = get_float_dt_td_columns(post_state.dfs[sheet_index])
+        if len(datetime_column_headers) > 0 or len(timedelta_column_headers) > 0:
+            raise make_invalid_promote_row_to_header_error(datetime_column_headers, timedelta_column_headers)
```
aarondr77, line 46: I'd also be down to just convert datetime and timedelta to strings when this occurs and not throw an error

mitosheet/mitosheet/tests/step_performers/test_promote_row_to_header.py
```diff
@@ -0,0 +1,87 @@
...
+        ],
+        0, 
+        0, 
+        [
+            pd.DataFrame({1: [2, 3], False: [False, True]}, index=[1, 2])
+        ]
+    ),
+]
+@pytest.mark.parametrize("input_dfs, sheet_index, index, output_dfs", PROMOTE_ROW_TO_HEADER_TESTS)
+def test_fill_na(input_dfs, sheet_index, index, output_dfs):
```
aarondr77, line 57: update test name

mitosheet/mitosheet/tests/step_performers/test_promote_row_to_header.py
```diff
@@ -0,0 +1,87 @@
...
+    (
+        [
+            pd.DataFrame({'A': [1, 2, 3], 'B': pd.to_timedelta(['1 days', '2 days', '3 days'])})
+        ],
+        0, 
+        0,
+    ),
+]
+@pytest.mark.parametrize("input_dfs, sheet_index, row_index", PROMOTE_ROW_TO_HEADER_FAILING_TESTS)
+def test_fill_na_failing(input_dfs, sheet_index, row_index):
```
aarondr77, line 83: update test name

mitosheet/src/utils/actions.tsx
```diff
@@ -675,6 +695,26 @@ export const createActions = (
...
+                }
+            },
+            isDisabled: () => {
+                const rowsToDelete = getSelectedRowIndexesWithEntireSelectedRow(gridState.selections, sheetData);
+                if (rowsToDelete.length > 0) {
+                    return undefined;
+                }
+                return "There is no selected row to promote to header."
+            },
+            searchTerms: ['make header', 'row to header', 'rename headers', 'column headers'],
```
aarondr77, line 715: other terms: promote, rename, column

aarondr77:
Lets do a bit more testing on it once that bug is fixed up. Lmk when to check it out

naterush:
@aarondr77 this bug is a very interesting one. You can see my previous investigation here: https://github.com/mito-ds/monorepo/issues/267

The TLDR is that this occurs when the user has a column header and a index that are the same (e.g. two zeros). I think untangling this would be a lot of work. How do you want to handle this here?

aarondr77:
While trying to understand that bug report, I attempted to promote the selected row to the header and got this new error. 

Screen Shot 2022-06-05 at 2 23 11 PM

You can follow it here: 

Screen Shot 2022-06-05 at 2 24 23 PM

aarondr77:
I'm okay with releasing this as is, and then checking in to see how many users run into these errors before deciding how to fix it, since you think it'll be tough to fix up. 

We could also add an error message that gives a hint as what the problem might be so that the user has some hope of fixing it up.

naterush:
@aarondr77 this is the same bug, it just gets caught in a different code path sometimes. Pretty much: index and header that match? You might run into errors.

I fixed up column headers so they work as timedeltas and date times now, just like indexes. So this is a good fix. Given how this can manifest in many different code paths, let's ignore this other bug for now and see if it is an issue we run into!

PR #344, opened 2022-06-05
Delete row fixes

naterush:
# Description

1. Makes indexes work with timedelta and datetime
2. Fixes duplicate indexes bug when optimizing

# Testing

Test these things!

# Documentation

N/A. Need to write delete row docs before deploying!


aarondr77:
Nice! This is working much better. The only issue is if you have multiple rows with the same index, when you delete one of them, they both get deleted. This could happen when the user concats two dataframes together and turns off `Ignore Existing Indexes`, and then deletes a row. I'm okay with this, as its somewhat intended + better than the alternative probably, but can we just make sure to note this in the documentation.

PR #346, opened 2022-06-06
Dogfood 6 6

naterush:
# Description

Makes these changes: https://www.notion.so/trymito/Jake-DR-Delete-Row-Promote-Row-ceaccb600f9345dd9b9dce69b5b86c77

# Testing

Test these changes (make sure to test new error).

# Documentation

Check doc for both Promote Row to Header and Delete Rows. Make any changes, then I'll finish dogfooding and we'll get this merged!

### DIFF
--- a/mitosheet/mitosheet/errors.py
+++ b/mitosheet/mitosheet/errors.py
@@ -409,14 +409,14 @@ def make_invalid_promote_row_to_header(error_modal: bool=True) -> MitoError:
     """
     Helper function for creating a invalid_promote_row_to_header.
 
-    Occurs when a user tries to promote a row to header that has row indexes
-    that are the same as this header.
+    Occurs when a user tries to promote a row to header that has duplicated
+    values within it.
     """
-    to_fix = f'This row has values that overlap with row indexes. As such, making this row a header would make column references ambigious, so we cannot make this row the header row' 
+    to_fix = f'This row has duplicated values in it. As such, making this row a header would make column references ambigious, so we cannot make it a header row.' 
     
     return MitoError(
         'no_analysis_error', 
-        "Duplicated Headers and Row Indexes",
+        "Cannot create duplicate headers",
         to_fix,
         error_modal=error_modal
     )
--- a/mitosheet/mitosheet/errors.py
+++ b/mitosheet/mitosheet/errors.py
@@ -405,6 +405,22 @@ def make_no_analysis_error(analysis_id: str, error_modal: bool=True) -> MitoErro
         error_modal=error_modal
     )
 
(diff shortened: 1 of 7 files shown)


mitosheet/mitosheet/errors.py
```diff
@@ -405,6 +405,22 @@ def make_no_analysis_error(analysis_id: str, error_modal: bool=True) -> MitoErro
...
     )
 
+def make_invalid_promote_row_to_header(error_modal: bool=True) -> MitoError:
+    """
+    Helper function for creating a invalid_promote_row_to_header.
+
+    Occurs when a user tries to promote a row to header that has row indexes
+    that are the same as this header.
+    """
+    to_fix = f'This row has values that overlap with row indexes. As such, making this row a header would make column references ambigious, so we cannot make this row the header row'
```
aarondr77, line 415: It also happens if the row has two values that are the same, which would also make column references ambiguous because the column headers would be the same. 

I actually think that might be the issue and it might have nothing to do with indexes. 

For example, take this dataframe `df = pd.DataFrame({'A': [1,2], 'B': [1,2]})` and try to promote the first row to the header. It will error.

Then update A1 -> 0, and try to promote it again. This time it will work. 

If you agree that is the issue, then lets update this error message because I think the explanation we can provide in that case could be way more clear and helpful.

mitosheet/mitosheet/step_performers/promote_row_to_header.py
```diff
@@ -54,6 +56,14 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
         post_state.dfs[sheet_index].drop(labels=[index], inplace=True)
         pandas_processing_time_drop = perf_counter() - pandas_start_time_drop
 
+        # We make sure that this making of headers will work, and not cause issues 
+        # later while trying to convert to json. We throw an error if this causes
+        # errors. See this bug: https://github.com/mito-ds/monorepo/issues/267
+        try:
+            convert_df_to_parsed_json(post_state.dfs[sheet_index].head(MAX_ROWS))
```
aarondr77, line 63: If our updated understanding of this error is correct, then we don't even need to pass in the MAX_ROWS, we can just use the first row to make it even faster.

aarondr77 (commented):
My understanding of this error changed a bit. Let me know if you agree.

mitosheet/mitosheet/errors.py
aarondr77: I added this example to https://github.com/mito-ds/monorepo/issues/267

PR #348, opened 2022-06-06
mitosheet: remove setting on warnings errors

naterush:
# Description

Fixes a warning we were getting with SettingWithCopyWarning in the utils file, which you can read about here: https://www.dataquest.io/blog/settingwithcopywarning/

It was an invalid warning (it wasn't real), but rather than silencing it I just moved where we copy to fix it up. No functionality changes.

Had to do this because warnings on Notebooks appear in the notebook rather than printed elsewhere, which looks awful.

# Testing

Testing nothing broke.

# Documentation

N/A.

### DIFF
--- a/mitosheet/mitosheet/utils.py
+++ b/mitosheet/mitosheet/utils.py
@@ -175,7 +175,7 @@ def df_to_json_dumpsable(
         for row in json_obj['data']:
             # If we're beyond the max columns, we might not have data, and we leave column data empty
             # in this case and don't append anything
-            column_final_data['columnData'].append(row[column_index] if column_index < len(row) else None)
+            column_final_data['columnData'].append(row[column_index] if column_index < MAX_COLUMNS else None)
         
         final_data.append(column_final_data)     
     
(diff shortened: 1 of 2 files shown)


mitosheet/mitosheet/utils.py
```diff
@@ -156,17 +156,9 @@ def df_to_json_dumpsable(
     """
 
     (num_rows, num_columns) = original_df.shape 
+    df = original_df.iloc[: , :max_columns]
```
aarondr77, line 159: I was confused why we needed to do this line here, since we're also doing it in the `convert_df_to_parsed_json` function, and I think I found an existing bug (I think one that I introduced). 

I think we should update line line 192 to use the `original_df` instead of `df` because we want all of the ColumnIDs, not just the first max_columns to be part of the map. 

And then we could also update the construction of `column_final_data` and `column_dtype_map` to use `df` instead of `original_df`. But this doesn't matter because we're looping over the results of `json_obj`.

aarondr77 (commented):
I think I found one existing bug, then LGTM.

PR #350, opened 2022-06-07
[WIP] Network drives

aarondr77:
# Description

# Testing

# Documentation


aarondr77:
This approach is messier than I would like. The key component of this approach was to keep everything in the frontend the same for windows and mac, but it leads to the paths being funky and lots of windows edge cases in the backend. 

I'm going to see if I can simplify this by doing a really small case on the frontend on windows / mac to make the backend much cleaner.

naterush (commented):
I think I might be misunderstanding what the complexity is - specifically, I'm confused where the C vs. C: vs. C:/ distinction needs to enter in on the frontend. 

What doesn't work about the following:
1. On mac, we return paths from the backend as they come.
2. On windows, we return paths from the backend as ['/', 'C:/', 'the', 'rest', 'of', 'it']
3. On the frontend, we display these paths and work with these paths identically. You simply render the `/` as a drive, always. Clicking on it takes you there in both Mac and windows. 
4. Accessing the `/` drive on Mac and windows is different on the backend. On Mac, it does normal things. On windows, it goes and get the drives.

Where does this break down? Where does the C:/ part enter into it? Is it that getting this path doesn't have the slash or something?

aarondr77:
In response to your review comment: 

Yeah that might help simplify some of the code. Returning the windows path as ['/', 'C:/', 'the', 'rest', 'of', 'it'] is an example of messing with the paths in a way that I was trying to avoid. But I can test if that works on my windows computer and then implement that if we're comfortable with it. This is different from the current implementation because we either have that path ['C:/', 'the', 'rest', 'of', 'it'] or [WINDOWS_DRIVE_PATH], we don't ever mess with the paths that we're going to operate on. 

> Where does the C:/ part enter into it? 
The place that C:/ enters into this is that the get_path_parts function takes the path `C:/the/rest/of/it` and returns `[C:, /, the, rest, of, it]` breaking up the C: and the / into separate path parts. This is not what we want for a few reasons the two biggest of which are 1) if the user selects the path part `C:` that is not a valid path and os will not handle it correctly for some reason. It must be formatted like `C:/`. So we then need to do some casing when we receive the path and add the `/` to it. And then doing that, will start the cycle over where the get_path_parts function will break it out into its own path. 

So the solution I'm going to do now is to just filter out the `/` in the return path parts. But the reason I haven't wanted to do that is because it seems fragile and breaks the invariant of trying to not mess with the paths that os gives to us.

naterush:
> This is different from the current implementation because we either have that path ['C:/', 'the', 'rest', 'of', 'it'] or [WINDOWS_DRIVE_PATH], we don't ever mess with the paths that we're going to operate on.

I think that the sort of messing we're doing in either case is fairly similar -- we're just putting at the start vs. in a separate list. So I'm totally fine with adding the / at the start.

> The place that C:/ enters into this is that the get_path_parts function takes the path C:/the/rest/of/it and returns [C:, /, the, rest, of, it]

I think given that reading in a path requires C:/, this is just a demonstration that our splitting code is bugged! I think that casing to a) put the / after the drive name, and b) not including the / in the path is just fixing a bug in our current get_path_parts code -- which notably, is hand rolled and _already is me messing with paths_.

To summarize: we shouldn't mess with paths more than we currently are but the first solution here is pretty much a different flavor of what we're currently doing, and the second solution is fixing a bug in the hand-rolled solution we already have!

aarondr77:
This code looks good to me + is working on both my mac and windows computers. Once we implement the pro feature work this LGTM.

PR #363, opened 2022-06-13
Usability tests

naterush:
# Description

Fixes here: https://www.notion.so/trymito/Usability-Test-In-place-formual-editing-6-13-22-ae65aa7d076147369ae31c5c390ce47f
 
# Testing

Test changes.

# Documentation

No.

### DIFF
--- a/mitosheet/mitosheet/code_chunks/promote_row_to_header_code_chunk.py
+++ b/mitosheet/mitosheet/code_chunks/promote_row_to_header_code_chunk.py
@@ -26,11 +26,13 @@ class PromoteRowToHeaderCodeChunk(CodeChunk):
         sheet_index: int = self.get_param('sheet_index')
         index: int = self.get_param('index')
 
+        print("index", index)
+
         df_name = self.post_state.df_names[sheet_index]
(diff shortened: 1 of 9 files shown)


naterush (commented):
Cleanup.

aarondr77:
I think the log anonymization for fill_na_edit is wrong. We do hide the value of `params_fill_method_type` and we don't hide the value of `params_fill_method_value`.

aarondr77:
Found another bug where if you select Rename Dataframe from the toolbar, the input just opens and closes without letting you type.

aarondr77:
Found a bug where the merge taskpane thinks there is not enough sheets to merge even if there are. 
1. Import a dataframe
2. Duplicate it
3. Create a pivot table of the duplicated sheet, but don't configure it, so that it remains a blank sheet
4. Select the duplicated sheet
5. Attempt to merge and notice that it says you need to import data.

aarondr77:
We still have this disabled check. We should remove it!

Screen Shot 2022-06-13 at 2 37 17 PM

aarondr77:
There are still some toolbar buttons that should close the cell editor that don't. 
1. Copy 
2. Split text to columns

PR #368, opened 2022-06-15
mitosheet: add user install option

naterush:
# Description

This adds the option for users to do a user install with the mitoinstaller, which hopefully will address.

# Testing

This is very hard to test. Here's what you can do:
1. Change `Consider using the `--user` option or check the permissions.` to `KeyboardInterrupt`, and then press control + c during this step (and then see if you enter the condition correctly). 
2. Confirm that you get an error when trying to do a user install inside a venv - this confirms that we are at least trying to do a user install.

This might lead to more bugs with notebook installs or something, but it's very hard to say before we deploy. Given how uncommon this is, I think we should sanity check per above and then see how common this is in the future before doing much more. We know user installs will work in many cases, so this is at least an improvement for many users who cannot even get started as of now!

# Documentation

No.

### DIFF
--- a/mitoinstaller/mitoinstaller/commands.py
+++ b/mitoinstaller/mitoinstaller/commands.py
@@ -88,7 +88,7 @@ def uninstall_pip_packages(*packages: str) -> None:
     run_command(sys_call)
 
 
-def install_pip_packages(*packages: str, test_pypi: bool=False) -> None:
+def install_pip_packages(*packages: str, test_pypi: bool=False, user_install: bool=False) -> None:
     """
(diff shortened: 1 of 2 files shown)


aarondr77:
Okay for test 1 where I change the condition to KeyBoard Interupt I get this. I'm not sure if something is supposed to happen afterwards, but the condition is getting caught correctly. 

Screen Shot 2022-06-18 at 10 06 38 PM

aarondr77:
Code looks great + its cool that you're trying to handle it through the user input. It makes the installer more valuable!

PR #370, opened 2022-06-18
Bulk filter

naterush:
# Description

Implements the final version of the spec here: https://www.notion.so/trymito/Toggle-All-in-Values-bc5211451bbb4ce7844893f9b0d007a4

The one change from the spec is the fact that the refresh button is always there. This is because there are many ways for this data to get out of date (e.g. editing it) - and always being able to refresh was really nice for me when I was testing!

# Testing

Do some bulk filtering!

# Documentation

We need to rewrite filter docs.


aarondr77:
If I have the numbers 1 ... 100 and I apply the filter > 90, then she sheet updates to only include the last 10 numbers, but the unique values tab doesn't refresh. It still displays all of the number 1 ... 100 as toggled on. 

My understanding of the decision we made is that we are going to replicate the behavior that we currently have by refreshing the unique values when the filter updates and removing all of the values that are filtered out via the filter section. So in my understanding, if you applied the >90 filter condition, the unique values tab would update and only display the values 91 ... 100. This would, as you phrased it "replicate our current behavior, where you can't toggle back values that have been filtered out through non-toggle means."

If you then toggled 91 off, it would remain displayed in the unqiue values tab becuase it wasn't filtered out by one of the filters from the first section. 

The other behavior that our current deployed version has around this is: let's say again you have the values from 1 ... 100. You use the unqiue values toggle to remove 1, you then apply a filter for >90. Now you look at the unique values tab again, it will display the numbers 1 and 91 .... 100. 

That is the behavior I thought we were aiming for here, but let me know if you had a different understanding. Or maybe we just haven't gotten to that part of the implementation yet!

aarondr77:
I think there is a bug in the search logic.

1. Upload this dataset. [unicorn.csv](https://github.com/mito-ds/monorepo/files/8944531/unicorn.csv)
2. Convert the valuatin column to an into
3. Open the unique values tab and search for `1`
4. Notice that it says `31 selected of which 2 not displayed`. 

The values that don't get returned in this search are `5` and `9`. But every other value does get returned. That should not be the case. 

I believe the bug is that it searches all of the text in the toggle box. Every other value contains a `1` in it either as the number of occurences, or percent. 

I think we should only search the actual values.

aarondr77:
The filter button in the column header should take you to the filter tab

mitosheet/src/utils/actions.tsx
```diff
@@ -1048,8 +1048,20 @@ export const createActions = (
             searchTerms: ['Melt', 'Unpivot'],
             tooltip: "Unpivot a DataFrame from wide to long format."
         },
+        [ActionEnum.Bulk_Filter]: {
```
naterush, line 1051: Remove this!

mitosheet/src/jupyter/api.tsx
```diff
@@ -737,9 +736,33 @@ export default class MitoAPI {
         }, {})
     }
     
+    
+    async editBulkFilter(
+        sheet_index: number,
+        column_id: ColumnID,
+        toggle_type:
```
naterush, line 743: Update these to be more reasonable

mitosheet/src/components/taskpanes/ControlPanel/FilterTab/UniqueValuesCard.tsx
```diff
@@ -247,47 +243,41 @@ export function ValuesTab(
...
-                                    // Manually change the toggle status so it updates instantaneously
-                                    const uniqueValueCountIndex = getUniqueValueCountIndexFromSortedIndex(index);
-                                    setUniqueValueCounts(oldUniqueValueCounts => {
-                                        const newUniqueValueCounts = oldUniqueValueCounts.slice();
-                                        newUniqueValueCounts[uniqueValueCountIndex].isNotFiltered = !uniqueValueCounts[uniqueValueCountIndex].isNotFiltered
-                                        return newUniqueValueCounts;
-                                    })
-
-                                    toggleExclusiveFilters([uniqueValueCount.value])
+                                    // TODO: handle NaN properly in this function
```
naterush, line 249: TODO: handle NaN values!

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -41,7 +41,10 @@ def execute_replay_analysis_update(
         log('replayed_nonexistant_analysis_failed')
         raise make_no_analysis_error(analysis_name, error_modal=False)
 
-    steps_manager.execute_steps_data(new_steps_data=analysis['steps_data'])
+    try:
```
naterush, line 44: Remove this try except

mitosheet/mitosheet/steps_manager.py
```diff
@@ -367,6 +367,9 @@ def handle_edit_event(self, edit_event: Dict[str, Any]) -> None:
 
         step_performer = EVENT_TYPE_TO_STEP_PERFORMER[edit_event["type"]]
 
+        # We saturate the event with data at the time this event is generated
```
naterush, line 370: TODO: document this better

mitosheet/mitosheet/list_utils.py
```diff
@@ -0,0 +1,21 @@
+
```
naterush, line 1: Add copy right

naterush (commented):
Some cleanup here!

naterush (commented):
TODO:
1. Make the bulk filter action do something
2. Make the filter button click properly

aarondr77:
I think this text is pretty confusing. What does it mean for something to be `selected` in this case if they can't do anything to effect it: They can't change it with the toggle all button or by scrolling to it. 

Screen Shot 2022-06-23 at 11 28 02 AM

The difference between selected and displayed is more relevant when there are so many values that not all of them are displayed. But I think in this case, we already have a message that not all of them are displayed. 

Screen Shot 2022-06-23 at 11 27 06 AM

What if instead, `selected` always referred to the values that match the search term that are also included in the dataset. 
So in the example above, it would say "Toggle all selected (74 selected)" since 74 values match the search term `2`

And similarly, if there were 100k values selected, it would say "Toggle all selected (100,000 selected)" even though only 1000 of them are displayed.

aarondr77:
This is a nitpick, but I think the refresh button is so thin, etc that it doesn't stand out enough. I designed this one. Want to sub it? https://www.figma.com/file/hbh9aMLG3tDOMaQdxHknM4/Mito?node-id=3652%3A86

aarondr77:
Can we update the design for this text? Its super bulky. 

Screen Shot 2022-06-23 at 11 52 29 AM

Maybe we can do something like this? 

Screen Shot 2022-06-23 at 11 56 13 AM

aarondr77:
Even in full screen mode the multi toggle box doesn't fit on the screen. Can we update the formatting on full screen mode so it all fits? I keep scrolling in the wrong scrollbar in this taskpane.

Screen Shot 2022-06-23 at 11 59 39 AM

mitosheet/mitosheet/code_chunks/step_performers/filter_code_chunk.py
```diff
@@ -210,6 +212,9 @@ def combine_filter_strings(
     NOTE: we choose to keep groups together for readibility, and so do not
     split the lines if we are combing a group.
```
aarondr77, line 213: combing -> combining

aarondr77:
Searching for `NaN` when it is included in the unique values doesn't keep it as a result.

aarondr77:
The filter for removing NaN values exclusively does not work. 

Screen Shot 2022-06-23 at 12 17 52 PM

mitosheet/mitosheet/list_utils.py
```diff
@@ -0,0 +1,25 @@
...
+        new_list = []
+        for i in l:
+            if i not in new_list:
+                new_list.append(i)
+        return new_list
+    else:
+        return list(set(l))
+
+
+def get_list_difference(l1: List[Any], l2: List[Any]) -> List[Any]:
```
aarondr77, line 23: Want to just add a note here that its for getting the unique values in l1 and not l2?

mitosheet/mitosheet/step_performers/bulk_filter.py
```diff
@@ -0,0 +1,114 @@
...
+        # This parameter is saturated above
+        bulk_filter: Any = get_param(params, 'bulk_filter')
+
+        post_state = prev_state.copy(deep_sheet_indexes=[sheet_index])
+        if toggle_type['type'] == BULK_FILTER_TOGGLE_SPECIFIC_VALUE:
+            values_to_toggle = set([toggle_type['value']])
+            remove_from_dataframe = toggle_type['remove_from_dataframe'] # if true, filtering out, and if false, adding back
+        elif toggle_type['type'] == BULK_FILTER_TOGGLE_ALL_MATCHING:
+            values_to_toggle = get_matching_values(post_state, sheet_index, column_id, toggle_type['search_string'])
+            remove_from_dataframe = toggle_type['remove_from_dataframe']
```
aarondr77, line 67: In the type comment above, there is no `remove_from_dataframe`.

mitosheet/mitosheet/step_performers/filter.py
```diff
@@ -251,12 +254,23 @@ def _execute_filter(
                 get_applied_filter(df, column_header, filter_or_group)
             )
 
-    
-    if len(applied_filters) > 0:
-        filtered_df = df[combine_filters(operator, applied_filters)]
-    else:
-        filtered_df = df
+    # Get the final filter for the filter list
+    filter_list_filter = combine_filters(operator, applied_filters)
```
aarondr77, line 258: Is there a better name for this? What is the final filter for the filter list mean?

mitosheet/mitosheet/code_chunks/step_performers/set_cell_value_code_chunk.py
```diff
@@ -32,18 +32,21 @@ def get_code(self) -> List[str]:
...
         # If nothings changed, we don't write any code
-        if old_value == new_value:
-            return code
+        if str(self.prev_state.dfs[sheet_index].at[row_index, column_header]) == new_value:
+            return []
 
-        column_header = self.post_state.column_ids.get_column_header_by_id(sheet_index, column_id)
-        transpiled_column_header = column_header_to_transpiled_code(column_header)
+        if new_value == '':
+            new_value = None
```
aarondr77, line 47: Should this come before the previous check so that we can check None values and return early?

aarondr77: We would need to have a special check if they are both `is none`

mitosheet/mitosheet/tests/step_performers/test_bulk_filter.py
```diff
@@ -0,0 +1,249 @@
...
+            df.iloc[1:2]
+        ]
+    ),
+    (
+        [
+            df.copy(deep=True)
+        ],
+        0, 
+        "D", 
+        {'type': BULK_FILTER_TOGGLE_ALL_MATCHING, 'search_string': 's', 'remove_from_dataframe': True},
```
aarondr77, line 125: Should we have some tests with 'remove_from_dataframe': False

mitosheet/src/components/taskpanes/ControlPanel/FilterTab/UniqueValuesCard.tsx
```diff
@@ -122,83 +127,35 @@ export function ValuesTab(
...
-            allDoesNotContainsFilters.forEach(key => {
-                _uniqueValueObjs.push({
-                    value: key,
-                    percentOccurence: 0,
-                    countOccurence: 0,
-                    isNotFiltered: false,
-                })
-            })
-        
+            console.log("Got", _uniqueValueObjs)
```
aarondr77, line 130: remove

mitosheet/src/components/taskpanes/ControlPanel/FilterTab/UniqueValuesCard.tsx
```diff
@@ -79,7 +79,11 @@ export function ValuesTab(
     const [uniqueValueCounts, setUniqueValueCounts] = useState<UniqueValueCount[]>([])
 
     const [searchString, setSearchString] = useState('');
-    const [sort, setSort] = useState<UniqueValueSortType>(UniqueValueSortType.ASCENDING_ALPHABETICAL)
+    const [sort, setSort] = useState<UniqueValueSortType>(UniqueValueSortType.ASCENDING_ALPHABETICAL);
+
+    // We store the toggled values in the frontend so that we can update their toggle
+    // state immediately, for quick feedback to the user
+    const [toggledValues, setToggledValues] = useState<[number | string | boolean, boolean][]>([])
```
aarondr77, line 86: Nice!

mitosheet/src/components/taskpanes/ControlPanel/FilterTab/UniqueValuesCard.tsx
```diff
@@ -122,83 +127,35 @@ export function ValuesTab(
...
-                        Unique Values
-                    </p>
+                    <Row justify='start' align='center' suppressTopBottomMargin>
+                        <Col>
+                            <p className='text-header-2'> 
+                                Unique Values
+                            </p>
+                        </Col>
+                        <Col offset={1} title='Click to refresh unique values, in the case they are out of date.'>
+                            <div onClick={loadUniqueValueCounts}>
```
aarondr77, line 154: Maybe we can make the onClick on the col, so its easier to click? I think this is why Jake struggled to get the ? icons to display the text. ... at least I was able to see the text on Brave.

mitosheet/src/components/taskpanes/ControlPanel/FilterTab/filter/utils.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { isBoolDtype, isDatetimeDtype, isNumberDtype, isStringDtype, isTimedeltaDtype } from '../../../../../utils/dtypes';
+import { CONDITIONS_WITH_NO_INPUT } from './filterConditions';
+
+/*
+    Given that we cannot filter all datatypes well, we have this utility 
+    for getting the message if the filters are disabled for a certain
+    columnDtpye
+*/
+export function getFilterDisabledMessage(columnDtype: string): string | undefined {
+    if (isTimedeltaDtype(columnDtype)) {
```
aarondr77, line 13: Maybe we can log this too?

mitosheet/src/jupyter/api.tsx
```diff
@@ -725,9 +725,32 @@ export default class MitoAPI {
...
+        | {'type': 'toggle_specific_value', 'value': string | number | boolean, 'remove_from_dataframe': boolean} 
+        | {'type': 'toggle_all_matching', 'search_string': string, 'remove_from_dataframe': boolean}
+    ): Promise<void> {
+
+        const stepID = getRandomId();
+        await this.send({
+            'event': 'edit_event',
+            'type': 'bulk_filter_edit',
+            'step_id': stepID,
+            'params': {
```
aarondr77, line 742: I think we are trying to move to having this be 'params': {BulkFilterParams}. But no worries if not.

mitosheet/src/utils/formatColumns.tsx
```diff
@@ -55,17 +55,12 @@ export const formatCellData = (cellData: boolean | string | number, columnDtype:
...
                         return formatCellDataAsStringWithCommas(cellData, 0)
                     } else {
-                        // We show the full number of decimals if it is a float
-                        const cellDataToParse = '' + cellData;
-                        let numDecimals = 0;
-                        if (cellDataToParse.includes('.')) {
-                            numDecimals = cellDataToParse.toString().split(".")[1].length || 0;
-                        }
-                        return formatCellDataAsStringWithCommas(cellData, numDecimals)
+                        // We show the floats with a single decimal by dfault
```
aarondr77, line 62: dfault -> default

aarondr77:
Setting the operator to OR in the UI doesn't effect the filter. It still uses AND.

aarondr77 (commented):
Nice work finishing this up! Its working pretty well and I think having the filters all on one page is going to make more people use the unique values filtering, which is great!

There are a few serious bugs that I found, but I don't think anything is going to be hard to fix. 

Let's make sure to do some more heavy testing of this before we release. I would say I'm not yet 80% sure that there are no bugs in this. So once you clean up this round of bugs, I can do more testing to get to that confidence level.

mitosheet/mitosheet/tests/test_utils.py
```diff
@@ -84,6 +84,7 @@ def check_final_dataframe(df_name, df):
     )
 
     import mitosheet
+    print(code)
```
naterush, line 87: Remove

mitosheet/src/components/taskpanes/ControlPanel/FilterTab/UniqueValuesCard.tsx
```diff
@@ -220,74 +173,78 @@ export function ValuesTab(
...
-                            />)
-                        }
-
+                        const toggledValue = toggledValues.find(([value, ]) => value === uniqueValueCount.value);
+                        const toggle = toggledValue !== undefined ? toggledValue[1] : uniqueValueCount.isNotFiltered
 
                         return((
                             <MultiToggleItem
                                 key={index}
+                                /**
```
naterush, line 220: Move down

naterush (commented):
Test this heavily!

naterush:
What should we do in the case that you:
1. Toggle values
2. Change dtype (changing those values)
3. Look at toggle again?

I think it makes sense to clear the bulk filter (and the filters, tbh) in this case. But we can also just not worry about it (we never really see this being a problem).

naterush:
Ok, I think this is working pretty well, finally. That being said, I've been testing this for hours so am a bit out of steam here... aka if you find a bug of bugs very much sorry and I promise to get them fixed up asap. I'm out of steam on this PR lol. 

There's some weird behavior around how this interacts with filters sometimes where if you:
1. Toggle values out
2. Add filters that filter those same values out
3. Toggle values back in (nothing changes in the actual data)
4. Refresh, these values just disappear

I think it's ok and makes enough sense to be fine!

aarondr77:
Can we make these vertically aligned?

Screen Shot 2022-07-05 at 10 16 48 AM

aarondr77:
When sorting the unique values, it loads and then changes orders.

https://user-images.githubusercontent.com/18709905/177350526-33528d4e-d08a-46cd-860f-ff0872edb19e.mov

aarondr77:
When the unique values tab is open and you scroll, the column headers don't keep up with the data that well. Notice that the lag reduces when the unique values tab is closed. 

Not sure what our options are here though. We could reduce the number of values that we show, but by how much would we need to reduce it by to reduce the lag?

https://user-images.githubusercontent.com/18709905/177351414-0f514946-e732-4aea-9ce8-68be9d2f3f82.mov

aarondr77:
We need to make sure that the X button is clickable on Windows. 

Screen Shot 2022-07-05 at 10 30 59 AM

aarondr77:
Can we leave the "Display first 1000 items..." at the top? Since the multi toggle box doesn't fit on one screen and there is a double scroll bar (toggle box scrollbar + taskpane scrollbar), I end up not scrolling all the way down.

aarondr77:
There is a bug where using the Toggle All button doesn't always remove everything from the dataset. In this video, you see that:
1. Using the Toggle All button on the `followers` column removes everything from the dataset, even though there are over 1000 values and not all of them are displayed. 
2. Using the Toggle All button on the `locations` column does not remove everything from the dataset. The location column also has over 1000 values. 

https://user-images.githubusercontent.com/18709905/177354854-d0f06b66-b125-481c-9974-6e6664639d8b.mov

aarondr77:
When searching for `NaN` values in the unique values section, they appear in the search if the search term is `Na`, but if you make it `Nan` or `NaN` then they do not appear.

aarondr77:
Another interesting bug (?). If you have a column with 1,2,3...10, and you toggle out 4, 5, 6, 7, then you inplace edit the column and you do column = column * 2, then when you look at the unique values tab, it says that there are 0 occurrences of 4. 

And then, if you toggle back on 5, all of the 4s disapear from the dataset.

This is probably edge case enough that it's okay.

aarondr77:
it seems like a few of these issues would be fixed by clearing the filters when changing the dtype, right?

naterush:
This PR is not being merged, and we'll use learnings from this to rearchitect at some point in the future. The main problem with this PR is:
1. How search is implemented differently on the frontend and the backend (one is fuzzy, the other is not). This means that sometimes clicking the button doesn't toggle the specific things you see.
2. The way search is implemented on the frontend generally. It's really confusing logic how the filtering is done currently, which makes it hard to edit.

aarondr77:
The specification process for this was also very messy. There were a few reasons for it:
1. There were a lot of complexities in the filtering mechanics that made it hard to reason upfront about reasonable solutions. We had to actually get into the weeds of the code to figure out what options were available to us. 
2. Nate and Aaron each thought that the other person was super focused on the robustness of replaying analyses. However, we actually thought that the cases where it was going to be an issue were really rare, and it was worth charging forward. 
3. We didn't have precise language to talk about the different mechanics. A few times we were talking / writing past eachother because we wern't on the same page about what we were actually talking about.

PR #384, opened 2022-07-01
mitosheet: add axis type transformations to graph

aarondr77:
# Description

Adds the ability to set [Plotly's axis type ](https://plotly.com/python/axes/) following this [specification](https://www.notion.so/trymito/Log-graph-transformations-d5b4b08a3cf74c25a25c37b9b828b860) 

# Testing

Create some graphs. 

# Documentation

See proposal for doc changes [here](https://app.gitbook.com/o/-MS4px_3crWEYYdbNPU8/s/-MP_U5ZCmiamDOXEOOTC/~/changes/n37dfGIxRPBQd5cYc79g/how-to/graphing). Merge when releasing.

### DIFF
--- a/mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
+++ b/mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
@@ -11,10 +11,12 @@ import Col from '../../spacing/Col';
 import Row from '../../spacing/Row';
 
 export enum AxisType {
+    DEFAULT = 'default',
     LINEAR = 'linear',
     LOG = 'log',
     CATEGORY = 'category',
     DATE = 'date',
+    MULTICATEGORY = 'multicategory',
 }
 /* 
     Contains all of the options for styling graphs,
@@ -197,7 +199,7 @@ function GraphStyleTab(props: {
                 <Row justify='space-between' align='center'>
                     <Col>
                         <p>
-                            Type
+                            Transform
                         </p>
                     </Col>
                     <Select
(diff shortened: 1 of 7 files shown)


mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -3,11 +3,19 @@
...
 import { GraphParams, UserProfile } from '../../../types';
+import DropdownItem from '../../elements/DropdownItem';
 import Input from '../../elements/Input';
+import Select from '../../elements/Select';
 import Toggle from '../../elements/Toggle';
 import Col from '../../spacing/Col';
 import Row from '../../spacing/Row';
 
+export enum AxisType {
+    LINEAR = 'linear',
```
naterush, line 14: Why is default not here?

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -186,6 +194,51 @@ function GraphStyleTab(props: {
                         }}
                     />
                 }
+                <Row justify='space-between' align='center'>
+                    <Col>
+                        <p>
+                            Type
```
naterush, line 200: I think the name "type" is kinda confusing. Can we name it Axis Transform (and then maybe change default to None) or something?

Or I'm misunderstanding what this is, and probably default doesn't mean None. But I think calling this transform instead of type might be helpful, idk...

naterush (changes requested):
Ok, confirming my understanding that we do not add a step upgrader b/c these options are optionally and so don't get saved anyways, correct?

Also, re: smashing this product debt. Let's design a "CollapsableSection" component methinks :0 - but don't worry about this now. 

After addressing my comments in naming (your call), LGTM!

PR #389, opened 2022-07-06
[DO NOT MERGE] mitosheet: add basic streamlit code generation

naterush:
# Description

# Testing

# Documentation


aarondr77:
This is a dope Proof of concept. The relationship between the generated code and getting the streamlit app up and running probably needs some product thinking. I ended up copying it out into a .py file and then running that.

naterush:
Sweet. Closing this PR for now, but this was a good start to the investigation.

PR #392, opened 2022-07-12
Ab test infra

naterush:
# Description

Implements the specification here: https://www.notion.so/trymito/A-B-Testing-Specification-686202c961844c39af2b8fecdb744427

Main differences from spec:
1. Turns out a component for AB testing was in fact helpful, so I added this on the frontend.
2. Logging when the experiment starts is _hard_, so we instead just log the experiment on every event. This makes the resulting dashboard easy to setup still (Any Event with the variant is an easy filter) - without complicated lifecycle reasoning.

# Testing

First, try the signup flow with both variants and make sure you get something different in either cae.

Screen Shot 2022-07-12 at 9 45 04 AM

Try to make a little dashboard.

# Documentation

N/A.


mitosheet/mitosheet/experiments/experiment_utils.py
```diff
@@ -0,0 +1,62 @@
...
+# Distributed under the terms of the GPL License.
+
+"""
+Contains utilities for interacting with experiments. If you want to change the currently running
+experiment, do the following:
+
+1. Change the get_new_experiment() function to return the new experiment you want to run.
+2. Change the user.json version to increase by one.
+3. Write an upgrader function that changes the experiment in the user json to the new experiment.
+
```
naterush, line 14: TODO: leave a note about tracking the ids somewhere, and what the variants mean. So we always know!

mitosheet/src/components/elements/Experiment.tsx
```diff
@@ -0,0 +1,29 @@
...
+    analysisData: AnalysisData, 
+    experimentID: ExperimentID,
+    aElement: React.ReactNode,
+    bElement: React.ReactNode,
+}): JSX.Element => {
+    console.log("Experiment", props.analysisData.experiment);
+
+    // If we are not running this experiment, then we simply default to the aElement
+    if (!isExperimentActive(props.analysisData, props.experimentID)) {
+        console.log("HERE")
```
naterush, line 16: Cleanup these, document this file.

mitosheet/src/types.tsx
```diff
@@ -555,6 +555,13 @@ export type FormatTypeObj =
     }
 
 
+export type ExperimentID = 
+    | 'title_name';
```
naterush, line 559: Add this to the list of steps.

naterush (commented):
Minor cleanup!

aarondr77:
Nice this is working well for me. I was able to get both variants of the experiment!

aarondr77:
When creating the dashboard, I wanted to make a funnel report so I could easily look at the conversion rates between either experiment set. But I'm not able to because the experiment is only logged in the events, not the user. Having the ability to do this would be great:

Screen Shot 2022-07-17 at 12 45 37 PM

naterush:
Screen Shot 2022-07-18 at 10 10 48 AM

Alternatively in this format - is more useful! This looks great :)

Gonna merge

PR #397, opened 2022-07-15
[DO NOT MERGE] Plugin investigation

naterush:
# Description

Search for "custom plugins" in the help bar, and try using one. Also check out how easy it is to specify a custom plugin that doesn't have any params. 

The main complexity with transformation plugins is the params. Check out the README.md for some thoughts on what the looks like, and what the complexities are here...

# Testing

n/a.

# Documentation

n/a.


mitosheet/mitosheet/plugins/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+We should make some components (that we can use in places) for these things. So note that this isn't just the easiest thing in the world.
+
+### How to go from this JSON -> a Frontend
+
+It would be fairly easy to take this JSON, load it up into AnalysisData, and send it to the frontend. Then, we just loop over the params, and turn them into elements. There's nothing particularly complex about this.
+
+## Syntax for Params, V1, attempt 2
+
+We follow bamboolib's example and instead provide Python classes that allow you to build the plugins. I'm less sure how this would work, but you imagine returning a list of params... so imagine something like:
```
aarondr77, line 132: It would be very cool to find some other people who have written Bamboolib plugins (maybe just through GitHub?) and ask them about their experience using that infrastructure.

aarondr77:
I think there is an important open question about the lifecycle of the plugin. How does the developer of the plugin make it accessible to others?

aarondr77:
Currently, in the plugin execute function we have a bunch of standard code and then `new_df = plugin.transform()`. How do we want to handle plugins that have extra work to do? Like updating a column's formula, for example?

aarondr77:
This API based plugin architecture is one route of making it easy for companies to customize Mito. Another approach to the same problem is having users edit the Mito codebase directly, through things like metaprogramming. 

The big tradeoffs being that with the API route, its both simpler for the user to write and its more likely that if we change step performers in the future, we'll be able to continue to support the plugins without requiring anything from the user. 

On the other hand, we add more constraints to the plugins that we support and we're required to write different types of plugins architecture for different types of plugins (transformers, visualizations, etc. etc.)

On initial thought, it seems like the backwards compatibility that we get through the API route is super important and tips the scales. But it might be worth thinking through both approaches.

mitosheet/mitosheet/plugins/custom_plugins/delete_columns_with_a.py
```diff
@@ -0,0 +1,24 @@
...
+    @classmethod
+    def get_plugin_name(cls) -> str:
+        return 'Delete Column with A'
+
+    @classmethod
+    def creates_new_dataframe(cls) -> str:
+        return False
+
+    def transform(self) -> pd.DataFrame:
+        return self.df.drop([column for column in self.df.columns if 'A' in str(column)], axis=1)
```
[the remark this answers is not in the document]
naterush: What invariants do we have currently? I think the only real one would be: if you delete columns, you need to update the state variables to get rid of the metadata we track about them. 

We can def do that automatically in the step performer. I'm not sure what else we would need to do here... but if we introduce invariants in the future, then yes this becomes a problem!

mitosheet/mitosheet/plugins/README.md
naterush: Cool idea -- I can't find anyone through Github though (they seem like they would generally be private) (https://github.com/search?l=&q=TransformationPlugin+language%3APython&type=code)

naterush:
The lifecycle of the plugin is def something we need to investigate! Haven't thought about it yet + not sure how it would work...

naterush:
> How do we want to handle plugins that have extra work to do? Like updating a column's formula, for example?

We wouldn't. The idea here is that we limit what the plugins can do, to make the interface actually usable. At least for a first version of the plugins, depending on what we need, I think this makes a lot of sense.

naterush:
> Another approach to the same problem is having users edit the Mito codebase directly, through things like metaprogramming.

This approach doesn't really work for a few reasons, I think. 
1. Users need much more context on Mito codebase. They go from editing a single file of Python code to editing 10 files, some of which are Python and others are TypeScript (another big dependency!). 
2. Users would need to open source their code - which for some custom plugins or teams, would be a total no-go.
3. We would have no way to give some plugins to some users and not to others! This is huge, but most of the plugins I've seen are so specific they would literally just be bloat for most users.

Working through the direct contribution approach doesn't really require much - it's just asking people to come make contributions to our codebase. But for the reasons I lay out above, I _don't think_ it's even similar to the extension goals. Teams are of course welcome to make contributions to the codebase for features that everyone wants!

naterush:
Since we're going to be doing visualization plugins, I'm gonna pause on this train of thinking now (although check out my comment above, and let's make sure we're on the same page about non-direct edits as the most sensible path forward for transformation plugins). 

For the visualization plugins, the overlap is:
1. We still want users to write just a single file - preferably Python, but perhaps JSON, and hopefully note TypeScript.
2. We still need to think about these lifecycle questions!
3. Different set of considerations for parameters...

We will attack these questions once we have a better understanding of what the visualizations look like! Closing this PR for now :)

PR #398, opened 2022-07-18
Graphing collapsible sections

naterush:
# Description

Implements this: https://www.notion.so/trymito/Make-more-space-for-new-graphing-features-ecebd93a174a42c58cb8c950bdf46e80

This looks and feel much better, and I think the new graphing is pretty intuitive as well! Let's go on a tear and add new sections?

# Testing

Test it.

# Documentation

No.


aarondr77:
Make sure to fix up the docs so that they are in the correct order. Other than that, LGTM.

PR #399, opened 2022-07-18
mitosheet: add one hot encoding of column

naterush:
# Description

This PR adds one-hot encoding as a new step. It's pretty much specification free (there isn't anything to configure), and it was something fun to do one a plane flight. 

It was mentioned as a useful feature by two teams in the past week. Do we want to make this a pro feature? The feedback where it was mentioned isn't really recorded anywhere, but happy to make a mini spec if u want.

# Testing

Check tests, try it out.

# Documentation

Add this (to pro docs depending if it's pro).


mitosheet/mitosheet/code_chunks/one_hot_encoding_code_chunk.py
```diff
@@ -0,0 +1,48 @@
...
+
+        return [
+            'import pandas as pd',
+            encode_line,
+            reorder_columns_line   
+        ]
+
+    
+    def get_edited_sheet_indexes(self) -> List[int]:
+        return [] # TODO: return this here!
```
naterush, line 47: TODO!

mitosheet/mitosheet/code_chunks/one_hot_encoding_code_chunk.py
```diff
@@ -0,0 +1,48 @@
...
+    def get_display_name(self) -> str:
+        return 'One Hot Encoding'
+    
+    def get_description_comment(self) -> str:
+        sheet_index: int = self.get_param('sheet_index')
+        column_id: ColumnID = self.get_param('column_id')
+
+        column_header = self.post_state.column_ids.get_column_header_by_id(sheet_index, column_id)
+
+        return f'One Hot Encoded {column_header}'
```
naterush, line 23: Update

mitosheet/mitosheet/state.py
```diff
@@ -235,6 +235,14 @@ def add_df_to_state(
             # Return the index of this sheet
             return sheet_index
 
+    def add_columns_to_state(self, sheet_index: int, column_headers: List[ColumnHeader]) -> None:
```
naterush, line 238: Document

mitosheet/mitosheet/step_performers/one_hot_encoding.py
```diff
@@ -0,0 +1,81 @@
...
+
+        column_header = prev_state.column_ids.get_column_header_by_id(sheet_index, column_id)
+        df = post_state.dfs[sheet_index]
+        column_index = df.columns.tolist().index(column_header)
+
+        pandas_start_time = perf_counter()
+        new_columns = pd.get_dummies(df[column_header])
+        pandas_processing_time = perf_counter() - pandas_start_time
+
+        # TODO: make sure there isn't overlap here?
```
naterush, line 49: NOTE: right now, there can be bugs if you try to encode on top of something. Is this something we want to address / I should make a spec for?

mitosheet/mitosheet/step_performers/one_hot_encoding.py
```diff
@@ -0,0 +1,81 @@
...
+        params: Dict[str, Any],
+        execution_data: Optional[Dict[str, Any]],
+    ) -> List[CodeChunk]:
+        return [
+            OneHotEncodingCodeChunk(prev_state, post_state, params, execution_data)
+        ]
+
+    @classmethod
+    def get_modified_dataframe_indexes(cls, params: Dict[str, Any]) -> Set[int]:
+        return {get_param(params, 'sheet_index')} # TODO: add the modified indexes here!
```
naterush, line 80: Remove TODO

mitosheet/mitosheet/step_performers/utils.py
```diff
@@ -8,10 +8,25 @@
...
 
 
 def get_param(params: Dict[str, Any], key: str) -> Any:
     if key in params:
         return params[key]
-    return None
\ No newline at end of file
+    return None
+
+def add_columns_to_df(df: pd.DataFrame, new_columns_df: pd.DataFrame, new_column_headers: List[ColumnHeader], column_index: int) -> pd.DataFrame:
```
naterush, line 24: Doc string

mitosheet/mitosheet/tests/test_utils.py
```diff
@@ -84,6 +84,7 @@ def check_final_dataframe(df_name, df):
     )
 
     import mitosheet
+    print(code)
```
naterush, line 87: Remove

naterush (commented):
Cleanup!

aarondr77:
If you one-hot encode one column and then one hot encode a different column that has some of the same values, you get an error.

Can we update this error to say that we're trying to make a column with a name that already exists? 

I wonder how common this is going to be? I think it might be totally possible that the use cases the teams are wanting to use this for would cause this error. Maybe it would be better to create the column name as `{original column name} {value}`? Want to ask the teams that requested this to see what they think?


Screen Shot 2022-07-18 at 12 55 55 AM

mitosheet/mitosheet/code_chunks/one_hot_encoding_code_chunk.py
```diff
@@ -0,0 +1,48 @@
...
+        column_index = self.post_state.dfs[sheet_index].columns.tolist().index(column_header)
+
+        transpiled_column_header = column_header_to_transpiled_code(column_header)
+        new_transpiled_column_headers = column_header_list_to_transpiled_code(new_column_headers)
+
+        encode_line = f'{df_name}[{new_transpiled_column_headers}] = pd.get_dummies({df_name}[{transpiled_column_header}])'
+        reorder_columns_line = f'{df_name} = {df_name}[{df_name}.columns[:{column_index + 1}].tolist() + {new_transpiled_column_headers} + {df_name}.columns[{column_index + 1}:-{len(new_column_headers)}].tolist()]'
+
+        return [
+            'import pandas as pd',
```
aarondr77, line 40: It would be nice to develop a simple way of checking if this already exists and then not importing it if so.

mitosheet/mitosheet/step_performers/one_hot_encoding.py
aarondr77: What do you mean on top of something?

mitosheet/mitosheet/step_performers/one_hot_encoding.py
```diff
@@ -0,0 +1,81 @@
...
+        column_index = df.columns.tolist().index(column_header)
+
+        pandas_start_time = perf_counter()
+        new_columns = pd.get_dummies(df[column_header])
+        pandas_processing_time = perf_counter() - pandas_start_time
+
+        # TODO: make sure there isn't overlap here?
+        new_column_headers = new_columns.columns.tolist()
+
+        print(new_columns, new_column_headers)
```
aarondr77, line 52: remove

aarondr77 (commented):
This is a cool feature. It would be really interesting to learn more about the use cases that the teams want this for. In my limited understanding, this seems like a very ML oriented feature and I bet there are other features that users who are using Mito as a step in ML would need / get a lot of value out of. 

Let's make this a pro feature! 

Then, LGTM.

aarondr77:
Can you make a little icon for it, so it can be displayed in the loading indicator & steps list?

mitosheet/mitosheet/step_performers/one_hot_encoding.py
naterush: Exactly the error you found below! So I'll just add an error for it!

mitosheet/mitosheet/code_chunks/one_hot_encoding_code_chunk.py
naterush: Yes, it def would :-)

naterush:
@aarondr77 I think let's take the easy solution of throwing the error for now. The two teams are not in places where getting feedback from them on this feature really works, I think...

PR #403, opened 2022-07-20
mitosheet: fix up some unrealized race conditions

naterush:
# Description

This PR fixes a bug I discovered while working on the new TTV checklist. Namely, if you switched to a sheet b/c of an open pivot table, you can get a race condition that leads to the wrong param being used as the sheet index.

This bug isn't manifesting in deployed code, but based on the fact adding a checklist made it happen means it might, so let's get this fixed ASAP!

# Testing

Use the pivot table!

# Documentation

N/A.

### DIFF
--- a/mitosheet/src/components/Mito.tsx
+++ b/mitosheet/src/components/Mito.tsx
@@ -391,6 +391,7 @@ export const Mito = (props: MitoProps): JSX.Element => {
                         currOpenModal: {type: ModalEnum.None},
                         currOpenTaskpane: {
                             type: TaskpaneType.PIVOT,
+                            sourceSheetIndex: existingPivotParams.sheet_index,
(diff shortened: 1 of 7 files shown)


aarondr77:
Bug in displaying the name of the source sheet:
1. Import file F
2. Pivot file F
3. Close pivot table
4. Select file F
5. Click pivot table button
6. Notice that the source data is F_pivot, but the columns that it allows you to select from are from F.

https://user-images.githubusercontent.com/18709905/180243733-7655726a-0ba0-4b7e-8029-06384488c031.mov

naterush:
@aarondr77 I am not experiencing this bug. In fact, this looks like exactly the bug that would occur when the old code was being run rather than these new changes.

Wanna check out this branch again and see if you can replicate this? Perhaps it's something about that file F... I am not sure.

PR #404, opened 2022-07-20
Ttv Checklist

naterush:
# Description

Implements this spec: https://www.notion.so/trymito/Time-to-Value-597e7bd8128a4f0c8ccf8972acee4413

See comments in Checklist.tsx with questions about when we should mark the checklist items as done - it's not so clear what is best IMO.

# Testing

## Test Checklist
!rm ~/.mito/user.json
!python -m mitosheet setexperiment installer_communication_and_time_to_value A

Should display the checklist (and check the user json to make sure that it marks the intro tour as completed!)

## Test Tour

!rm ~/.mito/user.json
!python -m mitosheet setexperiment installer_communication_and_time_to_value B

Should display tour (and check user json to make sure that it is the same as the other case as well).

# Documentation

N/A.


mitosheet/dev/create_new_icon.py
```diff
@@ -2,17 +2,27 @@
...
 CLOSE_BRACKET = '}'
 
+
+SPLIT_ON = [
+    'stroke-',
+    'stop-'
+]
+
 def get_tsx_svg_with_props(svg_code: str) -> str:
     """Need to move stroke-width to strokeWidth"""
```
naterush, line 12: Update comment

mitosheet/dev/realinstall.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash -eu
+
+# checktestpypi.sh: this is a script that you can run from anywhere on your
```
naterush, line 3: Comment! And update other one

mitosheet/src/components/icons/SmallDownArrowIcon.tsx
```diff
@@ -0,0 +1,15 @@
+
```
naterush, line 1: Remove

mitosheet/src/components/icons/SmallUpArrowIcon.tsx
```diff
@@ -0,0 +1,15 @@
+
```
naterush, line 1: Remove

mitosheet/src/components/icons/checklist/Checklist1Icon.tsx
```diff
@@ -0,0 +1,25 @@
+
+// Copyright (c) Mito
+
+import React from 'react';
+
+const Checklist1Icon = (): JSX.Element => {
+    return (
+        <svg width="27" height="15" viewBox="0 0 27 15" fill="none" xmlns="http://www.w3.org/2000/svg">
```
naterush, line 8: Lint

naterush (commented):
Minor cleanup, one final test!

naterush:
The spacing is different than the mockups, namely because we have way less space in many cases. Feel free to get in there and mess with things, but reminder that we usually have the biggest screens of any of our users.

naterush:
NOTE: review this after the installer stuff, and pivot stuff is merged. It uses both of them.

aarondr77:
TODO: Open issue for the cell editor not opening with the cursor moving in the sheet. Instead, it is moving in the formula.

mitosheet/mitosheet/tests/user/test_user.py
```diff
@@ -277,4 +278,25 @@ def test_upgrades_version_3_to_version_4():
...
+        UJ_MITOSHEET_TELEMETRY: True,
+        UJ_MITOSHEET_PRO: False,
+        UJ_EXPERIMENT: {
+            'experiment_id': 'an_old_experiment',
+            'variant': 'A',
+        }
+    })
+
+    initialize_user()
+    check_user_json(user_email='', mitosheet_experiment_id=get_new_experiment()['experiment_id'])
```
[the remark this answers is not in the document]
aarondr77: Hmm okay this is my understanding:

- Say I am a user who is onboarded on Jul 1, I will be on user.json version 6. 
- When I upgraded Mito, I will get upgraded to user.json version 7
- In that upgrade process, my user.json will now include a full onboarding_checklist (to ensure that it doesn't pop up), and I will also randomly get added to the experiment in line `206` of `upgrade.py`. 
- Since I've been added to the experiment, all of my future events will include the log parameters `experiment_id`, `experiment_installer_communication_and_time_to_value
`, `experiment_variant`.

This isn't necessarily going to throw off our Mixpanel reports. But it will make them fragile because we will have to be sure to always filter out the users who were not part of the A/B test. 

Alternativly, if we are upgrading a user from user.json 6 to 7 we could either not give them an experiment_id (this would require that the only time we add someone to an experiment is through the install or upgrade process, which seems like it should be the case), or assign them to the experiment_variant called "not included"

What do you think? It seems worth fixing up so we don't make our analyses more fragile in a relm of new analytics work already.

mitosheet/src/components/Mito.tsx
```diff
@@ -725,8 +729,18 @@ export const Mito = (props: MitoProps): JSX.Element => {
 
         const toursToDisplay: TourName[] = []
 
+        // If we are in variant B, then we display the tour to the user. In either case, we 
+        // take special care to set the user json as if we also did the other variant, so
+        // that make sure the user.json for all users is valid at the end of the experiment
```
aarondr77, line 734: If we redo the user.json. Maybe this is a fragility that we want to redesign. This works for onboarding things because its one and done. But if we had something else that was ongoing, we won't want to mark the user as having completed it. 

Instead, when the experiment is done, we'll want to actually want them to have access to the X. 

On second though, this seems like the type of infrastructure design that we should definitely do with a time when we want to use it. So let's hold off on that specific change.

aarondr77 (commented):
Okay sweet!! I just went through the checklist a few times and tried to focus on the logic of the A/B testing too since its a new piece of infrastructure & I see all user.json handling here. 

Everything is mostly working, but I think there might be some tweaks that we want to make to the infrastructure + checklist. 

I'm going to play around with the formatting of the checklist for 20 minutes and then I'll send the rest of this review back to you in Zulip. 

Nice work!! Excited to see if this pays off!

mitosheet/src/components/checklists/Checklist.tsx
```diff
@@ -0,0 +1,268 @@
...
+}): JSX.Element => {
+    return (
+        <ChecklistItem
+            index={props.index}
+            text={getChecklistItemTitle(props.item)}
+            icon={props.completed ? <CheckmarkIcon/> : <EmptyCircleIcon/>}
+            href={getChecklistItemLink(props.item)}
+        />
+    )
+}
```
aarondr77, line 111: It seems like the ChecklistTODOItem isn't that helpful of a component? Want to just remove it and use the ChecklistItem directly?

mitosheet/src/components/checklists/Checklist.tsx
```diff
@@ -0,0 +1,268 @@
...
+    index: number,
+    text: string,
+    icon: JSX.Element,
+    href: string
+}): JSX.Element => {
+    return (
+        <Row 
+            justify='space-between' 
+            align='center' 
+            className='text-body-1 text-color-white-important text-underline-on-hover mt-5px'
```
aarondr77, line 81: It looks like this `mt-5px` is getting overridden by the row's margins.

mitosheet/src/components/checklists/Checklist.tsx
aarondr77: Ahh I guess it is being used for the TODO item and next step item

mitosheet/css/Checklist.css
```diff
@@ -4,4 +4,16 @@
...
+.checklist-item {
+    color: var(--mito-white)
+}
+
+.checklist-item:hover {
+    color: var(--mito-off-white)
+}
+
+.checklist-item-completed {
+    color: #C1A1FF
```
aarondr77, line 18: I could give this a name

mitosheet/src/components/checklists/Checklist.tsx
```diff
@@ -73,20 +74,18 @@ const ChecklistItem = (props: {
...
+    completed?: boolean
 }): JSX.Element => {
     return (
         <Row 
             justify='space-between' 
             align='center' 
-            className='text-body-1 text-color-white-important text-underline-on-hover mt-5px'
-            onClick={() => {
-                window.open(props.href, '_blank');
-            }}
```
aarondr77: I'm pretty sure I have github copilot. And I'm pretty sure it deleted this...

aarondr77:
A bunch of the tests are failing locally for me. 

Screen Shot 2022-07-21 at 2 56 01 PM

mitosheet/mitosheet/user/upgrade.py
```diff
@@ -180,6 +197,14 @@ def try_upgrade_user_json_to_current_version() -> None:
         user_json_object = upgrade_user_json_version_4_to_5(user_json_object)
     if user_json_object[UJ_USER_JSON_VERSION] == 5:
         user_json_object = upgrade_user_json_version_5_to_6(user_json_object)
+    if user_json_object[UJ_USER_JSON_VERSION] == 6:
+        user_json_object = upgrade_user_json_version_6_to_7(user_json_object)
+
+    # We always make sure that the experiment is the most up to date
+    # version of the experiment
+    new_experiment = get_new_experiment()
```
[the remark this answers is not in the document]
naterush: Check out the function. It actually generates a new experiment. The function that reads from the user json is get current experiment.

mitosheet/src/components/checklists/Checklist.tsx
```diff
@@ -0,0 +1,268 @@
...
+const getChecklistIcon = (remainingChecklistItems: string[]): JSX.Element => {
+    if (remainingChecklistItems.length === 1) {
+        return (<Checklist5Icon/>)
+    } else if (remainingChecklistItems.length === 2) {
+        return (<Checklist4Icon/>)
+    } else if (remainingChecklistItems.length === 3) {
+        return (<Checklist3Icon/>)
+    } else if (remainingChecklistItems.length === 4) {
+        return (<Checklist2Icon/>)
+    }
```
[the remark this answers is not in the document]
naterush: Honestly, prob not worth it. This alone took me an hour (per the sizing of icon message I sent in Zulip).

mitosheet/src/components/checklists/Checklist.tsx
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        if (remainingChecklistItems.includes('filter')) {
+            if (props.analysisData.stepSummaryList.filter(stepSummary => stepSummary.step_type === StepType.FilterColumn).length > 0) {
+                void props.mitoAPI.updateChecklist('onboarding_checklist', ['filter']);
+            }
+        }
+
+        if (remainingChecklistItems.includes('pivot')) {
+            // Check that there is a pivot table, and that it has at least one column
```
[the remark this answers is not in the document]
naterush: My concern with this, per my comment, is that the user will -think- they are done with the pivot table but have not in fact configured it. Like they will set rows and columns, and not values, and get confused. 

I guess they can always just close the checklist though!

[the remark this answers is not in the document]
naterush: Updated to check this!

mitosheet/src/components/Mito.tsx
naterush: I agree that we'll def want to change this, and also agree we should do it when we encounter the issue specifically!

mitosheet/mitosheet/tests/user/test_user.py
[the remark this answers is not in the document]
naterush: I think there are ways we want to iterate this infrastructure, but it's hard to say exactly how it's best to do this. I think for now, I want to avoid introducing more complexity into this infrastructure and rather just filter by when the user was first seen or finished signup by date. 

I added fields for just tracking when the experiments start and end to the experiment tracker. I think that this is a good enough solution for now, especially given how few A/B tests we're running and the few analytics we're looking at!

PR #407, opened 2022-07-25
mitosheet: add helper packages, deploy to test-pypi

naterush:
# Description

This PR implements this spec: https://www.notion.so/trymito/Easy-Pro-activation-for-Teams-3a9b4fb75c6642029ef63046299680d3

Also removes the mitosheet-private package (check how we're checking for no display of upgrade screen now), as we can now give different (and easier) instructions for installing Mito Pro.

**One question**: do we want to rename these? So they don't show up so easily in the mitosheet search on pypi. If so, what do we want to rename them too?

# Testing

1. Make sure pro is turned off in user json.
2. Run `python3 -m pip install --index-url https://test.pypi.org/simple/ mitosheet-helper-pro --upgrade`
3. Check that pro is active

Do the same for `python3 -m pip install --index-url https://test.pypi.org/simple/ mitosheet-helper-private --upgrade`

# Documentation

Update the pro docs for installing!


mitosheet-helpers/mitosheet_helper_private/pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+version = "0.0.2"
+authors = [
+  { name="Nate Rush", email="naterush1997@gmail.com" },
+]
+description = "A helper package for mitosheet."
+license = { file="LICENSE" }
+requires-python = ">=3.6"
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "License :: OSI Approved :: MIT License", # TODO: update
```
naterush, line 16: TODO: update this!

mitosheet-helpers/mitosheet_helper_pro/setup.cfg
```diff
@@ -0,0 +1,7 @@
+[metadata]
+name = mitosheet_helper_pro
+version = 0.0.3
+description = A helper
```
naterush, line 4: Update this!

naterush (commented):
Cleanup!

naterush:
Note, once this is merged, I'll do the deploy onto PyPI. Notably, this only ever needs to be done once!

aarondr77:
Hmmm soo:
1. check out this branch
2. startup my jupyter server 
3. Run this command: `python3 -m pip install --index-url https://test.pypi.org/simple/ mitosheet-helper-pro --upgrade`
4. I'm still not on pro

Maybe I need to run the mitosheet.sheet() call too. 

1. Restart my kernel + refresh my browser
2. Run the mitosheet.sheet() call 
3. Get this error.

Screen Shot 2022-07-26 at 11 07 11 AM

mitosheet/mitosheet/user/utils.py
```diff
@@ -66,18 +78,18 @@ def should_upgrade_mitosheet() -> bool:
...
     NOTE: if the user clicks the upgrade button in the app, then we change the upgraded 
     date to this date, so that the user doesn't get a bunch of annoying popups. This just
     pushes back when they are annoyed to upgrade!
     """
     if not is_local_deployment():
         return False
-
-    if package_name == 'mitosheet-private':
+    
+    if MITOSHEET_HELPER_PRO or MITOSHEET_HELPER_PRIVATE:
```
aarondr77, line 90: Are we sure that we want to make this assumption? If we are doing this, then we want to ensure that individual pro users are not upgrading to Pro using this command, because we want them to get upgrade prompts! 

If we want to keep everyone on the same pro approach though, we could add another package called "mitosheet_no_upgrade_prompt" and have specific enterprises use them. This turns into a way of setting config params

aarondr77:
Okay, so to recap this work:

1. There is a circular import in this code that can be resolved by creating MITOSHEET_HELPER_PRIVATE_INDICATOR and MITOSHEET_HELPER_PRO_INDICATOR in the same file. 
2. The two packages are having the intended effect.

aarondr77:
> One question: do we want to rename these? So they don't show up so easily in the mitosheet search on pypi. If so, what do we want to rename them too?

I'm not concerned about it at this point. I think instead of renaming them, we would rather password protect them cause users can find them in our codebase anyways.

mitosheet/mitosheet/user/utils.py
naterush: Yeah. I think this package versioning technique makes a lot of sense, and certainly is something we can do at some point. 

For now, I think this approach is ok. People upgrading isn't so essentially anyways, and I think reminders to do so are probably more annoying than real helpful.

naterush:
Ok, we are deployed! Just need to merge this, and then make sure to test it before we give it to an enterprise. Nice!

PR #409, opened 2022-07-25
More graph functionality

aarondr77:
# Description

Implements this spec: https://www.notion.so/trymito/Graph-Functionality-aba347563c2d47d9a7bcbcc361704d04

Adds a bunch of new functionality: 
- Facet charts 
- Facet Styling (pro)
- Specific graph configurations 
  - barmode
  - points
  - line_shape
  - barnorm
  - histnorm
  - histfunc
  - nbins
- Legend (pro)
- Grid Lines (pro)

It also adds an upgrader script for the new graph step type. We add the non-optional params, which correspond to select inputs that have no None option. 

# Testing

- Test all of the new graph configs
- Test upgrading a graph from an old graph type

# Documentation

Need to create these!


mitosheet/mitosheet/saved_analyses/step_upgraders/graph.py
```diff
@@ -204,4 +204,114 @@ def upgrade_graph_2_to_3(step: Dict[str, Any], later_steps: List[Dict[str, Any]]
...
+
+
+def upgrade_graph_3_to_4(step: Dict[str, Any], later_steps: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    """
+    Adds required graph styling params:
+    - xaxis: showgrid
+    - yaxis: showgrid
+
+    OLD: 
+        'step_version': 2,
```
aarondr77, line 218: update this to 3. Update the next to 4. And make sure the params are correct.

mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -78,6 +80,9 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
         """
         Returns the new post state with the updated graph_data_dict
         """
+
+        print(params)
```
aarondr77, line 84: remove

mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -102,8 +107,21 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
...
+        facet_col_spacing = float(graph_creation['facet_col_spacing']) if 'facet_col_spacing' in graph_creation.keys() else None
+        facet_row_spacing = float(graph_creation['facet_row_spacing']) if 'facet_row_spacing' in graph_creation.keys() else None
+
+        # Validate parameters that are only available for some graph types
+        # Note: We trust the parameters from the frontend, so we don't make sure the params fit the graph type here
+        points = graph_creation['points'] if 'points' in graph_creation.keys() else None
+        line_shape = graph_creation['line_shape'] if 'line_shape' in graph_creation.keys() else None
+        histnorm = graph_creation['histnorm'] if 'histnorm' in graph_creation.keys() else None
+        histfunc = graph_creation['histfunc'] if 'histfunc' in graph_creation.keys() else None
+        nbins = int(graph_creation['nbins']) if 'nbins' in graph_creation.keys() else None
```
aarondr77, line 124: For the ones that don't need to be cast, etc. we could probably just pass them through as params, so we don't need to do all of this validation, because we end up just checking if it is not None, then add it. 

If we just don't do this check, then the param won't be there initially for us to check.

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
```diff
@@ -118,21 +128,76 @@ def get_graph_creation_param_dict(
...
+
+    if nbins is not None: 
+        all_params['nbins'] = nbins
+
+    if histnorm is not None:
+        all_params['histnorm'] = histnorm
+
+    if histfunc is not None:
+        all_params['histfunc'] = histfunc
+
```
aarondr77, line 160: We could get rid of all of this I think

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -291,6 +295,576 @@ function GraphStyleTab(props: {
...
+                                    }
+                                })
+                                props.setGraphUpdatedNumber(old => old + 1)
+                            }}
+                            width='small'
+                        />
+                    </Row>
+                }
+                {GRAPHS_THAT_HAVE_BARMODE.includes(graphCreationParams.graph_type) && 
+                    <Row justify='space-between' align='center'>
```
aarondr77, line 329: Add a title

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -291,6 +295,576 @@ function GraphStyleTab(props: {
...
+                            />
+                            <DropdownItem
+                                title={'percent'}
+                                subtext='fraction multiplied by 100'
+                            />
+                        </Select>
+                    </Row>
+                }
+                {GRAPHS_THAT_HAVE_HISTNORM.includes(graphCreationParams.graph_type) && 
+                    <Row justify='space-between' align='center' title='The type of normalization used for this histogram trace'>
```
aarondr77, line 408: Figure out how to draw a better distinction between barnorm and histnorm.

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -291,6 +295,576 @@ function GraphStyleTab(props: {
...
+                            }}
+                            width='small'
+                            dropdownWidth='medium'
+                        >
+                            <DropdownItem
+                                title={'none'}
+                            />
+                            <DropdownItem
+                                title={'probability'}
+                                subtext='fraction occurence w.r.t total number of sample points'
```
aarondr77, line 437: "occurrences in bin divided by total number of sample points"

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -291,6 +295,576 @@ function GraphStyleTab(props: {
...
+                            width='small'
+                            dropdownWidth='medium'
+                        >
+                            <DropdownItem
+                                title={'count'}
+                                subtext='number of values in each bin'
+                            />
+                            <DropdownItem
+                                title={'sum'}
+                                subtext='sum of the values in each bin'
```
aarondr77, line 485: get rid of "the"

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -291,6 +295,576 @@ function GraphStyleTab(props: {
...
+                            />
+                            <DropdownItem
+                                title={'probability density'}
+                                subtext='probability that a point falls into bin'
+                            />
+                        </Select>
+                    </Row>
+                }
+                {GRAPHS_THAT_HAVE_HISTFUNC.includes(graphCreationParams.graph_type) && 
+                    <Row justify='space-between' align='center' title='The binning function for the historgram'>
```
aarondr77, line 455: "The metric displayed for each bin of data"

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -291,6 +295,576 @@ function GraphStyleTab(props: {
...
+                            />
+                            <DropdownItem
+                                title={'max'}
+                                subtext='max value in each bin'
+                            />
+                        </Select>
+                    </Row>
+                }
+                {GRAPHS_THAT_HAVE_POINTS.includes(graphCreationParams.graph_type) && 
+                    <Row justify='space-between' align='center' title='Decide how to display outlier points'>
```
aarondr77, line 503: "Display outlier points"

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -291,6 +295,576 @@ function GraphStyleTab(props: {
...
+                                id={'suspectedoutliers'}
+                                subtext='display outlier and suspected outlier points'
+                            />
+                            <DropdownItem
+                                title={'all'}
+                                subtext='display all sample points'
+                            />
+                            <DropdownItem
+                                title={'false'}
+                                subtext='display no sample'
```
aarondr77, line 543: "display no individual sample points"

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -291,6 +295,576 @@ function GraphStyleTab(props: {
...
+                            />
+                        </Select>
+                    </Row>
+                }
+                {GRAPHS_THAT_HAVE_LINE_SHAPE.includes(graphCreationParams.graph_type) && 
+                    <Row justify='space-between' align='center' title='Set the shape of the lines in the chart'>
```
aarondr77, line 549: "The shape of the line"

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -291,6 +295,576 @@ function GraphStyleTab(props: {
...
+                {!props.userProfile.isPro &&
+                    <Row justify='space-between' align='center'>
+                        <p className='text-body-1'>
+                            Want to customize the legend? <a href='https://trymito.io/plans' target='_blank' rel="noreferrer"><span className='text-body-1-link'>Upgrade to Mito Pro.</span></a>
+                        </p>  
+                    </Row>
+                }
+                {props.userProfile.isPro &&
+                    <>
+                        <Row justify='space-between' align='center'>
```
aarondr77, line 593: Add a title

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -291,6 +295,576 @@ function GraphStyleTab(props: {
...
+                                                ...graphParamsCopy.graphStyling,
+                                                showlegend: !graphParamsCopy.graphStyling.showlegend
+                                            } 
+                                        }
+                                    })
+                                    props.setGraphUpdatedNumber(old => old + 1)
+                                }}     
+                            />
+                        </Row>
+                        <Row justify='space-between' align='center' title='Set the legend vertical or horizontal'>
```
aarondr77, line 616: Set -> Display

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
```diff
@@ -118,21 +128,76 @@ def get_graph_creation_param_dict(
     if graph_type not in GRAPHS_THAT_DONT_SUPPORT_COLOR and color_column_header is not None:
         all_params['color'] = color_column_header
 
+    if facet_col_column_header is not None:
+        all_params['facet_col'] = facet_col_column_header
```
naterush, line 132: Is the reason that we need to not set them if they are None is because in the transpiled code, we get parameters that are effectivley just set to defaults?

If so, I somehow feel like _filtering_ out in the transpile step might be nice. But this also might suck, because what if we have parameters where None is in fact not the default?

mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -102,8 +105,21 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
         y_axis_column_ids = graph_creation["y_axis_column_ids"] if graph_creation["y_axis_column_ids"] is not None else []
         y_axis_column_headers = prev_state.column_ids.get_column_headers_by_ids(sheet_index, y_axis_column_ids)
 
-        # Validate optional parameters 
-        color_column_header = prev_state.column_ids.get_column_header_by_id(sheet_index, graph_creation["color"]) if 'color' in graph_creation.keys() else None
+        # Validate optional parameters that are available for all graph types
+        color_column_header = prev_state.column_ids.get_column_header_by_id(sheet_index, graph_creation["color"]) if 'color' in graph_creation.keys() and graph_creation["color"] in prev_state.column_ids.column_id_to_column_header[sheet_index].keys() else None
```
naterush, line 109: I think a helper function called `get_column_header_from_optional_column_id_graph_param` might be really nice here. It is just like:
```
def get_column_header_from_optional_column_id_graph_param(state: State, params: Dict[str, Any], param_name: string) -> Optional[ColumnHeader]:
    # TODO
```

mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -102,8 +105,21 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
...
+        facet_col_column_header = prev_state.column_ids.get_column_header_by_id(sheet_index, graph_creation["facet_col_column_id"]) if 'facet_col_column_id' in graph_creation.keys() and graph_creation["facet_col_column_id"] in prev_state.column_ids.column_id_to_column_header[sheet_index].keys() else None
+        facet_row_column_header = prev_state.column_ids.get_column_header_by_id(sheet_index, graph_creation["facet_row_column_id"]) if 'facet_row_column_id' in graph_creation.keys() and graph_creation["facet_row_column_id"] in prev_state.column_ids.column_id_to_column_header[sheet_index].keys() else None
+        facet_col_wrap = int(graph_creation['facet_col_wrap']) if 'facet_col_wrap' in graph_creation.keys() else None
```
naterush, line 112: Here's a helpful bit of syntax:
```
graph_creation.get('facet_col_wrap', None)
```
The second parameter is the default if the key is not there. But it doesn't work because of the cast. 

Is there anyway we can avoid the cast by making this as an integer on the frontend? Or is this hard for some reason?

mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -102,8 +105,21 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
...
+        facet_col_spacing = float(graph_creation['facet_col_spacing']) if 'facet_col_spacing' in graph_creation.keys() else None
+        facet_row_spacing = float(graph_creation['facet_row_spacing']) if 'facet_row_spacing' in graph_creation.keys() else None
+
+        # Validate parameters that are only available for some graph types
+        # Note: We trust the parameters from the frontend, so we don't make sure the params fit the graph type here
+        points = graph_creation['points'] if 'points' in graph_creation.keys() else None
```
naterush, line 118: All of these could be moved to the 
```
points = graph_creation.get('points', None)
```
to great effect I think!

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -291,6 +295,576 @@ function GraphStyleTab(props: {
...
+                                            graphStyling: {
+                                                ...graphParamsCopy.graphStyling,
+                                                legend: {
+                                                    ...graphParamsCopy.graphStyling.legend,
+                                                    y: (newY as number | undefined)
+                                                }
+                                            } 
+                                        }
+                                    })
+                                    props.setGraphUpdatedNumber(old => old + 1)
```
naterush, line 736: I wanna see if I can think of a new helper function that makes this cleaner (the nesting is just brutal). Will investigate when I do a full review!

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
```diff
@@ -399,30 +973,101 @@ function GraphStyleTab(props: {
...
+                                placeholder='num cols'
+                                onChange={(e) => {
+                                    const newNumCols = e.target.value === '' ? undefined : e.target.value
+                                    props.setGraphParams(prevGraphParams => {
+                                        const graphParamsCopy: GraphParams = JSON.parse(JSON.stringify(prevGraphParams)); 
+                                        return {
+                                            ...graphParamsCopy,
+                                            graphCreation: {
+                                                ...graphParamsCopy.graphCreation,
+                                                facet_col_wrap: newNumCols as number | undefined
```
naterush, line 1005: A reminder that there is a difference between _forcing a type_ and _casting the string to a number_. 

Consider the following:
```
const x = "string";
const y: number = x as number;
```
What is the value of `y`? Well, it's "string" still. All you've done is strong-armed the compiler into thinking the that it is a number, when it is in-fact, not. The `as` keyword doesn't change the data that's there, it just changes what type the compiler thinks it is.

Consider the next example.
```
const x = "string";
const y: number = parseFloat(x);
```
What is y equal to? y is NaN! This is because actually parsing the float fails in this case, as it isn't parsable.

# What you want to do here

In this case, we actually want to send a _number_ to the backend. Because you aren't actually parsing the string to a number, this is why you have to do that conversion on the backend. 

That being said, this can get really annoying if the user wants to be able to enter decimals. Check out the FillNaN taskpane for example (line 342). Because we want to allow the user to:
1. Enter a number.
2. That has decimals,
3. So might not be a number while it's being entered (e.g. "." is not a number, but ".123" is)
We need to store the value as a string, and then send it to the backend as a number, only by doing the final conversion _right before_ we send the number to the backend!

To deal with this, the best way I've figured out of generally handling things is:
1. Have a notion of FrontendGraphParams and BackendGraphParams. Frontend params store these values a strings, and backend values store them as numbers. (You could even make a parameteric type to represent this, so you only have to specify this once... but this might be more complexity than it's worth). 
2. Store FrontendGraphParams on the frontend. Convert to backendGraphParams before sending to the backend. This is just just right before the data is sent to the backend. 

Although this might be _net_ more lines of code, it's better because the backend gets data in a format it can directly operate on without casting things, and the frontend is just responsible for sending the correct types. And it accomplishes this while allowing users to enter in whatever number they set. 

# A note on this

This is a bit of work, and I'm happy to do it, but it also could be educational for you. Would love if you want to try it for a bit (an hour?) and then give up if you get stuck. I can finish it off, and we can discuss. 

This is the sort of thing that has popped up exactly _twice_ before, so writing standards before it ahead of time is really hard (it barely was even a distict thing in my head until I saw this casting going on), but I think that this is a great chance to align on these, figure out what a good way to handle them generally is, and learn for the future!

This isn't just me asking you to spin you're wheels for no reasons -- I think you'll learn a lot (about types, casting), and we can have a chance to talk about and figure out better ways to approach this going forward. Let me know if you have more thoughts here.

naterush (commented):
This is just an incremental review. I think to finish this PR:
1. Make a helper for getting column headers on the backend.
2. Move to the `.get(key, default)` syntax where possible.
3. Attempt the frontend refactor to get rid of the casts on the backend (limit 1 hr). 

Let me know wherever you finish up (hard deadline of today, per your budget, right? you're call here) -- and I am tots happy to pick it up from here. Happy to talk about any/all of the above if it would be helpful, just gimme a call.

I am not so sure on how to avoid the `if x is not None:` checks for now, but I think it's not the worst thing... would be happy if we address the above 3!

This is gonna be awesome in making our graphing feel rocking! Good work -- hope you enjoyed it :)

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
```diff
@@ -118,21 +124,60 @@ def get_graph_creation_param_dict(
...
+    if facet_col_spacing is not None:
+        all_params['facet_col_spacing'] = facet_col_spacing
+
+    if facet_row_spacing is not None:
+        all_params['facet_row_spacing'] = facet_row_spacing
+
+    if points is not None:
+        all_params['points'] = points
+
+    if line_shape is not None:
```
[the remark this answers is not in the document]
aarondr77: Yeah that seems nice. Based on your summary comment, doesn't sound like you want to do that now. So resolving this.

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
aarondr77: Yeah the two reasons for this are: 
1) We don't want to include them in the transpiled code
2) I'm trying to keep the transpiled code and the plotly code we execute exactly the same so if there are small intricacies that we don't understand, we don't end up generating a different graph.

mitosheet/mitosheet/step_performers/graph_steps/graph.py
aarondr77: Fixed up so all casting is on the frontend

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
```diff
@@ -235,6 +298,11 @@ def get_graph_styling_param_dict(graph_type: str, column_headers: List[ColumnHea
     if 'type' in graph_styling_params['xaxis']:
         all_params['xaxis']['type'] = graph_styling_params['xaxis']['type']
 
+    all_params['xaxis']['showgrid'] = graph_styling_params['xaxis']['showgrid']
+
+    if 'gridwidth' in graph_styling_params['xaxis']:
+        all_params['xaxis']['gridwidth'] = float(graph_styling_params['xaxis']['gridwidth'])
```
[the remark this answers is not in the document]
aarondr77: Done.

mitosheet/src/components/taskpanes/Graph/GraphStyleTab.tsx
aarondr77: Done! Thanks for this super thorough explanation! Very helpful.

mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -110,17 +110,18 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
         facet_col_column_header = get_column_header_from_optional_column_id_graph_param(prev_state, graph_creation, 'facet_col_column_id')
         facet_row_column_header = get_column_header_from_optional_column_id_graph_param(prev_state, graph_creation, 'facet_row_column_id')
         
-        facet_col_wrap = int(graph_creation['facet_col_wrap']) if 'facet_col_wrap' in graph_creation.keys() else None
-        facet_col_spacing = float(graph_creation['facet_col_spacing']) if 'facet_col_spacing' in graph_creation.keys() else None
-        facet_row_spacing = float(graph_creation['facet_row_spacing']) if 'facet_row_spacing' in graph_creation.keys() else None
+        facet_col_wrap = graph_creation.get('facet_col_wrap', None)
+        facet_col_spacing = graph_creation.get('facet_col_spacing', None)
+        print(facet_col_spacing)
```
aarondr77: remove

mitosheet/mitosheet/step_performers/graph_steps/plotly_express_graphs.py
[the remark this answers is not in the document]
naterush: Can you remove the casting on the backend now? If we don't need it, then having it back here just makes what types we're getting a bit confusing!

There's a couple places we do it in this file, so let's chop it all!

mitosheet/mitosheet/tests/step_performers/graph_steps/test_graph.py
```diff
@@ -83,16 +96,30 @@ def test_all_styling_options():
...
-        yaxis_type=yaxis_type,
         yaxis_visible=yaxis_visible,
+        yaxis_type=yaxis_type,
+        yaxis_showgrid=yaxis_showgrid,
+        yaxis_gridwidth=yaxis_gridwidth,
         showlegend=showlegend,
+        legend_title_text=legend_title_text,
+        legend_orientation=legend_orientation,
+        legend_x=legend_x,
+        legend_y=legend_y,
```
naterush, line 122: It doesn't look like you're testing everything here. Can you check histnorm, for example? And all the rest of them, if they aren't here.

mitosheet/src/components/taskpanes/Graph/GraphSetupTab.tsx
```diff
@@ -26,6 +38,28 @@ const SAFETY_FILTER_ENABLED_MESSAGE = `Turning on Filter to Safe Size only graph
...
+export const GRAPHS_THAT_HAVE_BARMODE = [GraphType.BAR, GraphType.HISTOGRAM]
+export const GRAPHS_THAT_HAVE_BARNORM = [GraphType.BAR, GraphType.HISTOGRAM]
+export const GRAPHS_THAT_HAVE_HISTNORM = [GraphType.HISTOGRAM]
+export const GRAPHS_THAT_HAVE_HISTFUNC = [GraphType.HISTOGRAM]
+export const GRAPHS_THAT_HAVE_POINTS = [GraphType.BOX, GraphType.VIOLIN]
+export const GRAPHS_THAT_HAVE_LINE_SHAPE = [GraphType.LINE]
+
+// This variable is used to figure out which graph types should 
+// havve teh specific graph type configuration section
+export const GRAPHS_WITH_UNIQUE_CONFIG_OPTIONS = [...new Set([
```
naterush, line 53: Nice one liner!

mitosheet/src/components/taskpanes/Graph/GraphSetupTab.tsx
```diff
@@ -26,6 +38,28 @@ const SAFETY_FILTER_ENABLED_MESSAGE = `Turning on Filter to Safe Size only graph
 
 const GRAPHS_THAT_DONT_SUPPORT_COLOR = [GraphType.DENSITY_HEATMAP]
```
naterush, line 39: It's a bit weird that this is a negative thing while the rest of them are positive. But it's cool...

mitosheet/src/components/taskpanes/Graph/GraphSetupTab.tsx
```diff
@@ -271,6 +313,136 @@ function GraphSetupTab(
...
+                                                ...graphParamsCopy,
+                                                graphCreation: {
+                                                    ...graphParamsCopy.graphCreation, 
+                                                    facet_col_column_id: undefined
+                                                }
+                                            }
+                                        })
+                                        props.setGraphUpdatedNumber((old) => old + 1);
+                                    }}
+                                />].concat(
```
naterush, line 353: Nice workaround to the type limitations here!

naterush:
Screen Shot 2022-07-29 at 10 56 17 AM


This is a bit busy to me. Thoughts on putting the facets inside a little toggle able section?

naterush:
Screen Shot 2022-07-29 at 10 57 07 AM

The legend being at the bottom, but not being a Pro feature, might be somewhat annoying. We could move it up if you want.

naterush:
Some thoughts on how we're labeling pro currently: I think if we kept the inputs visible, it would really communicate _how much_ pro gets you so much better.

Thoughts? I don't know an easy way to make this change, but it feels like it would really communicate how much they are missing so much better.

naterush:
Screen Shot 2022-07-29 at 11 34 27 AM

Can we add tooltips for them. I don't know what either of these things do...

naterush:
I kinda feel like the custom plot configuration makes sense to put in graph creation, or at least at the top of graph styling. Thoughts?

naterush:
Screen Shot 2022-07-29 at 11 37 37 AM

This should just be called "Range Slider" now - there's no legend here anymore.

naterush:
Screen Shot 2022-07-29 at 11 38 41 AM

The legend positions being number inputs is really confusing, because it seems they only accept values in certain ranges (e.g. negative numbers don't work). So maybe let's not make them numbers, as the little side toggles just result in invalid values.

naterush:
Screen Shot 2022-07-29 at 11 40 39 AM

What the hell are all these? Can we add descriptions? Do you think we should always just add descriptions to these? I kinda think it would be nice...

naterush:
This looks pretty great. I've tested all the new functionality, and it looks good. I have skimmed the code (not read every line), but it's pretty formulaic, so looks good. 

Like how you broke down the params + graph type relationships with this arrays! Fix up the above, per your call (no need to ping me), and LGTM.

I'll do the deploy Monday morning, so try and have it finished up by then if you can!

mitosheet/mitosheet/tests/step_performers/graph_steps/test_graph.py
aarondr77: I don't test the graph params, like histnorm, that are only for specific graph types in that parameritized function because it will error when trying to apply the histnorm param to other graph types. 

Since we moved to the model of trusting the parameters from the frontend, we don't do a second round of testing on the backend to ensure that the params only get applied to the correct graph types. 

Instead, I created a second set of tests directly below this one that test the specific graph type params.

PR #411, opened 2022-07-26
7 26 usability

naterush:
# Description

1. Made button to reopen onboarding checklist. This only exists during your first usage... that good?
2. For the bulk of the checklist, made it so it opens the feature. This is sweet. It doesn't do the same for the later links, but I think it's fine. Also, what do for those that aren't active (b/c no imported data?) -- I think it could be fine to just leave and usability test.
3. Make the installer not upgrade the user.json if it shouldn't!

# Testing

Test all these things. You'll have to clear your user.json to pretent to be a first time user.

# Documentation

No.

### DIFF
--- a/mitosheet/css/sitewide/margins.css
+++ b/mitosheet/css/sitewide/margins.css
@@ -24,6 +24,19 @@
     margin-right: 2px;
 }
 
+.mt-10px {
+    margin-top: 10px;
+}
+.mb-10px {
+    margin-bottom: 10px;
+}
+.ml-10px {
+    margin-left: 10px;
+}
+.mr-10px {
+    margin-right: 10px;
+}
+
 .mb-2px {
     margin-bottom: 2px;
 }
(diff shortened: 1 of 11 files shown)


naterush (commented):
Looks good!

naterush:
There's another "button click doesn't open import here" - but I think we should let it be, and let the AB test for next week fix it up. We could combine these, see my comments in Zulip!

aarondr77:
In the control variant, the `open checklist` button in the top right corner is still there. It disapears when you click on it.

aarondr77:
When I reopen the checklist, it flashes through all of the previous states before moving to the new screen.

https://user-images.githubusercontent.com/18709905/181091869-3db31f0c-314f-4db4-9d98-d474a356a806.mov

naterush:
@aarondr77 flashing is a very hard fix! We'd need to track different state or something -- it's so minor, I think it's not worth it.

PR #418, opened 2022-08-01
mitosheet: fix user json not existing bug

naterush:
# Description

Fixes https://github.com/mito-ds/monorepo/issues/417

# Testing

Check your previous test case.

# Documentation

No.

### DIFF
--- a/mitosheet/mitosheet/mito_widget.py
+++ b/mitosheet/mitosheet/mito_widget.py
@@ -8,6 +8,7 @@
 Main file containing the mito widget.
 """
 import json
+import os
 import time
 from typing import Any, Dict, List, Optional, Union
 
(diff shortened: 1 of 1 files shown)


aarondr77:
My test case is still not working: 

https://user-images.githubusercontent.com/18709905/182199582-8f891fab-dee8-40c5-901f-2aef33e7c1d6.mov

I also didn't get the signup flow when I followed the commands you listed in the issue: 


https://user-images.githubusercontent.com/18709905/182200154-a43e2cdb-dc5d-49b4-b651-372218a8cb66.mov

aarondr77:
Check if the user.json is making its way to the frontend and if so why is the signup not getting displayed

aarondr77:
Okay, so the reason that the signup was not appearing in this second round of testing is because I had the `mitosheet-helper-private` package on my computer. 

However, even with the `mitosheet-helper-private` removed, when on dev, the signup modal did not appear for me. 

Developer experience improvement iteration ideation: Adding a command to the effect of `python -m mitosheet reset-mito-except-user-id` which will clear out the user.json except for the user_id (so we don't clog up Mixpanel), which removes the user.json, uninstalls the mito packages, etc.

PR #419, opened 2022-08-01
mitoinstaller: make sure experiment is in user.json

aarondr77:
# Description

In trying to understand this experiment. I upgraded an old mitosheet and the upgrade process failed with the error `File "/Users/aarondiamond-reivich/Desktop/mito-upgrade-test/venv/lib/python3.8/site-packages/mitoinstaller/user_install.py", line 73, in try_create_user_json_file
    if new_experiment is not None and updated_user_json['experiment']['experiment_id'] != new_experiment['experiment_id']:
KeyError: 'experiment'`

To recreate it:
1. Delete your user.jsoin
2. Run `pip install mitosheet=0.1.430 --no-cache-dir`
3. Go through the signup flow. 
4. Then kill the server and try to upgrade Mito by running: `python -m pip install mitoinstaller --upgrade` followed by `python -m mitoinstaller upgrade`

On the actual upgrade command the error appears. Also, importantly note that it doesn't show up in the logs at all.

@naterush is thoughts on how we could add logging so we see more upgrade issues? 

# Testing

1. Delete your user.jsoin
2. Run `pip install mitosheet=0.1.430 --no-cache-dir`
3. Go through the signup flow. 
4. Run the dev installer by following the instructions in the `mitoinstaller/readme.md`

# Documentation

None.

### DIFF
--- a/mitoinstaller/mitoinstaller/user_install.py
+++ b/mitoinstaller/mitoinstaller/user_install.py
@@ -67,14 +67,10 @@ def try_create_user_json_file(is_pro: bool=False) -> None:
             updated_user_json['mitosheet_telemetry'] = not is_pro
             updated_user_json['mitosheet_pro'] = is_pro      
 
-        # Make sure that the user.json contains the experiment field
-        if 'experiment' not in updated_user_json.keys():
-            updated_user_json['experiment'] = dict()
-
         # And we also make sure that the experiment is updated, if it needs
         # to be updated
         new_experiment = get_new_experiment()
-        if new_experiment is not None and ('experiment_id' not in updated_user_json['experiment'] or updated_user_json['experiment']['experiment_id'] != new_experiment['experiment_id']):
+        if 'experiment' not in updated_user_json or 'experiment_id' not in updated_user_json['experiment'] or updated_user_json['experiment']['experiment_id'] != new_experiment['experiment_id']:
             updated_user_json['experiment'] = new_experiment
 
         with open(USER_JSON_PATH, 'w') as f:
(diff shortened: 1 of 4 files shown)


naterush:
@aarondr77 I assumed this would be caught by existing logging. Do you know why it isn't?

mitoinstaller/mitoinstaller/user_install.py
```diff
@@ -67,10 +67,14 @@ def try_create_user_json_file(is_pro: bool=False) -> None:
...
 
+        # Make sure that the user.json contains the experiment field
+        if 'experiment' not in updated_user_json.keys():
+            updated_user_json['experiment'] = dict()
+
         # And we also make sure that the experiment is updated, if it needs
         # to be updated
         new_experiment = get_new_experiment()
-        if new_experiment is not None and updated_user_json['experiment']['experiment_id'] != new_experiment['experiment_id']:
+        if new_experiment is not None and ('experiment_id' not in updated_user_json['experiment'] or updated_user_json['experiment']['experiment_id'] != new_experiment['experiment_id']):
```
naterush, line 77: I think this is better expressed as a single if statement:
```
if 'experiment' not in updated_user_json or updated_user_json['experiment']['experiment_id'] != new_experiment['experiment_id']):
     updated_user_json['experiment'] = new_experiment
```

naterush (commented):
A bit of cleanup, then LGTM.

naterush:
Check out the fix for the logging before merging, and let me know if you have questions. I found it by following _where_ this function was called from, and saw that it wasn't contained in any try catch or logging infrastructure, so I added it.

aarondr77:
Since the if statement you suggested got rid of the `is None` check, I updated the type declarations of the `get_experiment()` function, which doesn't seem to ever return None anyways.

naterush:
Ah, sorry - let's keep the inline check def (my bad for leaving it out), and I think that the function can def return None if the user has an older version of the user.json earlier on! Can you change it back?

aarondr77:
Remove the middle check for the experiment_id

PR #425, opened 2022-08-03
Make Plotly work offline

naterush:
# Description

1. Detects if Plotly is not loaded, and loads from the backend. This makes Plotly work when it is offline.
2. Parses the Plotly generated code better.

# Testing

1. Test graphing normally.
2. Open console and `delete window.Plotly`, and make sure graphing still works (and that window.Plotly gets redefined after 1 new graph is made)

# Documentation

N/A.

### DIFF
--- a/mitosheet/mitosheet/step_performers/graph_steps/graph.py
+++ b/mitosheet/mitosheet/step_performers/graph_steps/graph.py
@@ -177,7 +177,6 @@ class GraphStepPerformer(StepPerformer):
                 )
             )
 
-            print("include_plotlyjs:", include_plotlyjs)
             html_and_script = get_html_and_script_from_figure(fig, height, width, include_plotlyjs)
 
             graph_generation_code = get_plotly_express_graph_code(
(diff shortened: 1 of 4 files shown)


mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -176,7 +177,8 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
                 )
             )
 
-            html_and_script = get_html_and_script_from_figure(fig, height, width)
+            print("include_plotlyjs:", include_plotlyjs)
```
naterush, line 180: Remove

naterush (commented):
Almost

mitosheet/mitosheet/step_performers/graph_steps/graph_utils.py
```diff
@@ -163,24 +164,32 @@ def get_html_and_script_from_figure(
...
-    # making sure to remove the script tag (so we can execute it)
-    script_start = '<script type="text/javascript">'
-    script_end = "</script>"
-    split_html = original_html.split(script_start)
-    div = split_html[0]
-    script = split_html[1][: -len(script_end)]
+
+    # Everything in a script tag we want to treat as one big script
+
+    # The get the graph div, which looks like: <div id="9c21d143-3fcd-4958-9295-ac7ada668186" class="plotly-graph-div" style="height:425px; width:970px;"></div>
```
aarondr77, line 176: Can you add some more explanation here of why we are doing this?

aarondr77 (commented):
That testing procedure worked for me and it also worked when I turned off my internet after deleting plotly. 

I don't fully understand what is happening here though. Do you mind just expanding on the comment with a bit more context?

PR #431, opened 2022-08-09
Formatting v2

naterush:
# Description

A few differences to call out:
1. The options in the `Format` dropdown are a bit different, but sensible.
2. The default formatting is different in mito and outside mito (this is fine, I think... unless users want formatting that matches the default, what do?).
3. Also, the default formatting now only displayed 2 decimals on floats, let me know what ya think.

The fact that there are two different dropdowns for formatting is pretty dang weird, but I think we can iterate this once we discover it's confusing (which it is to me!).

# What is left

There are two major things left to do, but I'm pausing here (b/c wrote enough code today...)
- [ ] Write the docs
- [ ] Write some real tests

I will do both before merging, but it's ready to review now!

# Testing

Test all the formats with all the precisions, and make sure the printed dataframes look the same!

# Documentation

Need to write documentation for this!


aarondr77:
> What do we want default formatting to look like?

I vote we keep the formatting as Excel does. I think its much more readable, and quick visibility of data is a key part of why people use Mito! Its also a bit annoying to have to update the formatting everytime you create a new column for example.

naterush:
@aarondr77 if we want to keep the formatting the same as Excel, we need to either add default (I'll probably rename `DEFAULT_MITO` to be clear) to all columns. If we want this to match up with the dataframe that is printed out, then this also means pretty much every analysis will have column formatting in it. 

Or, we can just display things better in the frontend when there is no formatting. And not do that for pandas. I think this is probably the better solution, so I'm gonna go for it!

mitosheet/mitosheet/code_chunks/postprocessing/set_dataframe_format_code_chunk.py
```diff
@@ -0,0 +1,200 @@
...
+        return f"\"{OPEN_BRACKET}:{precision_string}E{CLOSE_BRACKET}\""
+
+    return None
+
+def get_all_columns_format_code(state: State, sheet_index: int) -> Optional[str]:
+    df_format = state.df_formats[sheet_index]
+    df = state.dfs[sheet_index]
+    columns = df_format['columns']
+
+    # TODO: in the future, we probably want to combine equivalent column formats
```
naterush, line 69: TODO: leave this as a note, something we might want to do!

mitosheet/mitosheet/code_chunks/postprocessing/set_dataframe_format_code_chunk.py
```diff
@@ -0,0 +1,200 @@
...
+    
+    return get_transpiled_table_style(
+        'thead', 
+        [
+            ('color', headers.get('color', None)),
+            ('background-color', headers.get('backgroundColor', None))
+        ]
+    )
+
+def get_rows_format_code(state: State, sheet_index: int, even_or_odd: str) -> Optional[str]:
```
naterush, line 104: Update function names to make sense, some doc strings.

mitosheet/mitosheet/code_chunks/postprocessing/set_dataframe_format_code_chunk.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def get_code(self) -> List[str]:
+        dataframes_with_formats = get_dataframes_with_formats(self.post_state)
+        code = []
+        for sheet_index in dataframes_with_formats:
+            dataframe_format_code = get_python_code_for_dataframe_format(self.post_state, sheet_index)
+            if dataframe_format_code is not None:
+                code.append(dataframe_format_code)
+        
+        return code
```
naterush, line 200: TODO: add tests for this code chunk, and more.

mitosheet/mitosheet/pro/download/formatting.py
```diff
@@ -45,28 +48,5 @@ def add_formatting_to_excel_sheet(
...
-                    format = workbook.add_format({'num_format': '#,##0.' + ('0' * format_data['numDecimals'])})
-            elif format_data['type'] == FORMAT_K_M_B:
-                format = workbook.add_format({'num_format': '[<999950]0.0,"K";[<999950000]0.0,,"M";0.0,,,"B"'})
-            elif format_data['type'] == FORMAT_SCIENTIFIC_NOTATION:
-                format = workbook.add_format({'num_format': '0.00E+00'})
-            else:
-                # Default format
-                format = workbook.add_format({'num_format': 'General'})
-        
+            format = workbook.add_format({'num_format': 'General'})
```
naterush, line 51: Leave a TODO.

mitosheet/mitosheet/step_performers/column_steps/change_column_dtype.py
```diff
@@ -194,8 +194,7 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
             pandas_processing_time = perf_counter() - pandas_start_time
 
             # If we're changing between number columns, we keep the formatting on the column. Otherwise, we remove it
-            if not ((is_int_dtype(old_dtype) or is_float_dtype(old_dtype)) and (is_int_dtype(new_dtype) or is_float_dtype(new_dtype))):
-                post_state.column_format_types[sheet_index][column_id] = {'type': FORMAT_DEFAULT}
+            # TODO: fix this up
```
naterush, line 197: TODO: fix this up.

naterush (commented):
- Write the step upgrader!

mitosheet/mitosheet/types.py
```diff
@@ -31,4 +31,7 @@
     StepsManagerType = Any
 
 
+DataframeFormat = Dict[str, Any]
+ColumnFormat = Dict[str, Any] # TODO: strongly type this
```
naterush, line 35: TODO

mitosheet/src/utils/format.tsx
```diff
@@ -0,0 +1,236 @@
+import React from "react"
```
naterush, line 1: TODO: write up some knowledge sharing about this file.

naterush (commented):
- Update trymito.io

mitosheet/src/utils/formatColumns.tsx
```diff
@@ -1,346 +0,0 @@
...
-    return Number(cellData).toLocaleString("en-US", {minimumFractionDigits: decimalPlaces, maximumFractionDigits: decimalPlaces})
-}
-
-/*
-    For a cell to be formatted as a number, the cell must only contain valid number symbols
-    and be either a int or float column. 
-
-    Note: The cellData type of a float is a string.
-*/
-export const displayCellAsNumber = (cellData: boolean | string | number, columnDtype: string): boolean => {
```
naterush, line 22: The reason this function was required was because _floats_ were cast to strings (e.g. "2.0") on the backend, so that they would be displayed with the correct number of decimals. 

Notably, this was a decision that was very very easy to change (I literally had to delete 2 lines of code in this PR, in utils.py). The reason the decision was originally made was because we had no formatting infrastructure on the frontend - but since you were adding it, this would have been a good time to remove it.

Once I remove this random cast to a string, we get to remove _all_ this 'displayCellAsNumber' code. This has other benefits later in the `format.tsx` file now (making a lot of other things way less complex and error prone. 

If you're touching formatting, think about what you're formatting, why it is the way that it is, and what you can change about the data you're consuming that will make your life easier!

mitosheet/src/utils/formatColumns.tsx
```diff
@@ -1,346 +0,0 @@
...
-        return isStringNumeric(cellData)
-    }
-}
-
-/*
-    Determines if a string is a number
-    Adapted from: https://stackoverflow.com/questions/175739/how-can-i-check-if-a-string-is-a-valid-number
-    Note: This won't format numbers that have non . separators correctly, but that's okay for now because we wouldn't want to switch them to decimals
-*/
-export const isStringNumeric = (str: string): boolean => {
```
naterush, line 42: Similar to the comment above, this code can all disappear! We drop a lot of complexity by just making sure the data we're sending the frontend is in a format that is easy for us to consume!

mitosheet/src/utils/formatColumns.tsx
```diff
@@ -1,346 +0,0 @@
...
-           !isNaN(parseFloat(str)) // ...and ensure strings of whitespace fail
-}
-
-/*
-    Returns cellData formatted as a number with decimals if the cell only contains valid number symbols
-    and the columnMitoType is a number_series. Otherwise, returns the unaltered cellData
-*/
-export const formatCellData = (cellData: boolean | string | number, columnDtype: string, columnFormatType: FormatTypeObj): string => {
-    // Wrap in a try, catch because there are lots of type cases and we'd rather be safe than sorry.
-    try {
```
naterush, line 54: One thing I want to call your attention to in this function: there is a _ton_ of rightward drive. The most indented thing in inside a try, inside a if, inside a switch, inside a case, inside an if, inside an if. 

This is really really hard to maintain code. Every line you edit, you have to think about 5 different conditions that must be true. It's also practically just hard to move things about, and simplify cases, as things are so nested.

Check out the new code in format.tsx. Instead of checking the positive versions of things, it checks the opposite of the conditions, and returns early. This allows us to reduce a level or two of the rightward drift. 

Also, because we simplify the datatypes (as mentioned above), we can really just make the most simple checks you could imagine. 

Please take 5 minutes or so to look at the `formatCellData` here, and the new `formatCellData`, and let me know what you think!

mitosheet/src/utils/formatColumns.tsx
```diff
@@ -1,346 +0,0 @@
...
-        return '' + cellData 
-    }
-    
-}
-
-/* 
-    Sends the changeColumnFormat message to the backend so that the format of the selected columns
-    will be updated in the state.
-*/
-export const changeFormatOfSelectedColumns = async (
```
naterush, line 113: This function is extremely thin. It makes one function call, and then calls another really similar function. 

This probably should signify this doesn't make sense as a function at all - just do the getSelectedNumberSeriesColumnIDs in the call location, and remove 30 lines of code!

mitosheet/src/utils/formatColumns.tsx
```diff
@@ -1,346 +0,0 @@
...
-        sheetIndex,
-        numberColumnIDsSelected,
-        newFormatTypeObj
-    )
-}
-
-/* 
-    Change the format of a single columnID
-*/
-export const changeFormatOfColumnID = async (sheetIndex: number, columnID: string, newFormatTypeObj: FormatTypeObj, mitoAPI: MitoAPI): Promise<void> => {
```
naterush, line 133: This shouldn't be a function! It literally just calls another function, with pretty much the same params.

naterush (commented):
I want to use this review to do some knowledge sharing about writing maintainable code. I spent some time working on your older formatting code, and I think there are a couple of concrete lessons to pull out!

Feel free to think about these, debate them, or gimme a call to discuss. The goal of this isn't a callout (the code was reasonably easy to edit!) - but rather to do some knowledge sharing -- so let's make sure the knowledge gets shared :)

aarondr77:
We should update this title from `Format` to `Number Type`?

Screen Shot 2022-08-12 at 10 09 51 AM

aarondr77:
"Remove decimal from number column" made me think it just got rid of the decimal all together. 

Maybe we can change it to "remove last decimal place from number column" or "decrease number of decimal places" 

And then "increase number of decimal places"

Screen Shot 2022-08-12 at 10 12 58 AM

aarondr77:
The decimal place changing / change number type works even if the entire column is not selected. I'm cool with that if you want it, but just calling out that it is different than delete column

aarondr77:
Screen Shot 2022-08-12 at 10 19 44 AM

aarondr77:
Can you update the suggested styles so that it updates the font color too?

Default: Leave as is
Gray: Background color: 9B9B9D... so that the border of the column headers is still visibile
Blue: white font 
Green: white font 
Red: Background color: #8F1B15, Text color: #F7C9CE

aarondr77:
This icon gets pretty hard to see depending on the color...

Screen Shot 2022-08-12 at 10 35 52 AM

aarondr77:
Its possible to get into this whacky edge case where the selection color is the same as the font color.

I think its okay to ignore this. I don't think purple is a common formatted color + its an easy fix for the user

Screen Shot 2022-08-12 at 10 39 24 AM

aarondr77:
Screen Shot 2022-08-12 at 10 42 18 AM

Screen Shot 2022-08-12 at 10 42 01 AM

aarondr77:
Bug: 

1. Import 2 dataframes
2. Open the formatting taskpane
3. Apply a default formatting to the first dataframe
4. Use the dataframe selector to move to the second dataframe
5. Notice that the first dataframe remains styled in mito
6. Notice that the styler is no longer included in the generated code

I think we want the dataframe to remain styled + the styler to remain!

aarondr77:
Because of the above bug, its hard to tell what the intended behavior is: but can we allow users to open the color formatting taskpane, apply formatting to one sheet, swtich dataframes, and apply formatting to that other sheet while preserving the formatting changes to the other dataframe. 

Since we somewhat expect the formatting to be a step that the user does towards the end of their analysis, I think it is likely they might go through and format multiple dfs at the same time

aarondr77:
When the user switches dfs to format in the taskpane, can we update the selected sheet to that new df?

aarondr77:
> The options in the Format dropdown are a bit different, but sensible.

Looks good to me!

aarondr77:
> The default formatting is different in mito and outside mito (this is fine, I think... unless users want formatting that matches the default, what do?).

Yeah I think this is fine!

aarondr77:
> Also, the default formatting now only displayed 2 decimals on floats, let me know what ya think.

I think this is in some ways ideal, but its also not what excel does, so slightly concerned that people will be confused. I think its good though. Its best for most users and they can always change the precision.

aarondr77:
The on hover over the cells doesn't look that great ...

mitosheet/src/utils/formatColumns.tsx
aarondr77: Yeah that makes a lot of sense. I think the fact that the code had is a great sign that checking the opposite condition first would've simplified things

```
        } else {
            return '' + cellData
        }
    } catch {
        return '' + cellData 
    }

```

aarondr77 (commented):
WOoo! Big PR! It looks great. Very excited to get this released to people!

Lots of small things here, but nothing major at all. + this is a lot of "this looks cool" comments.

mitosheet/src/components/endo/selectionUtils.tsx
```diff
@@ -747,6 +747,19 @@ export const getSelectedRowLabelsWithEntireSelectedRow = (selections: MitoSelect
...
 
+export const getNumberColumnIDs = (sheetData: SheetData | undefined, columnIDs: (ColumnID | undefined)[]): ColumnID[] => {
+    const columnIDsAndDtypes: [ColumnID, string][] = columnIDs
+        .filter(colId => colId !== undefined)
+        .map(columnID => [columnID as ColumnID, sheetData?.columnDtypeMap[columnID as ColumnID] || ''])
+
+    // Filter out any columns that are not number series
+    return columnIDsAndDtypes
+        .filter(([, columnDtype]) => {return columnDtype !== undefined && isNumberDtype(columnDtype)})
+        .filter(([columnID, ]) => {return columnID !== undefined})
```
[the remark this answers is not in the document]
naterush: Just in case

mitosheet/src/components/taskpanes/SetDataframeFormat/SetDataframeFormatTaskpane.tsx
```diff
@@ -0,0 +1,207 @@
...
+                            return updateDataframeFormatParams({rows: {even: {backgroundColor: newColor}}});
+                        }}
+                    />
+                    <LabelAndColor
+                        label='Even Row: Text Color'
+                        color={params.df_format.rows.even.color || EVEN_ROW_TEXT_COLOR_DEFAULT}
+                        onChange={(newColor) => {
+                            return updateDataframeFormatParams({rows: {even: {color: newColor}}});
+                        }}
+                    />
```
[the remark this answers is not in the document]
naterush: Yep! Shout out the infrastructure we've been slowly building up...

mitosheet/mitosheet/transpiler/transpile.py
```diff
@@ -44,7 +45,12 @@ def transpile(
 
     # We only transpile up to the currently checked out step
     all_code_chunks: List[CodeChunk] = get_code_chunks(steps_manager.steps_including_skipped[:steps_manager.curr_step_idx + 1], optimize=optimize)
-    
+
+    # We also make sure to include all the post_processing code chunks, which are those
+    # code chunks that are always at the end of the dataframe
+    for postprocessing_code_chunk in POSTPROCESSING_CODE_CHUNKS:
+        all_code_chunks.append(postprocessing_code_chunk(steps_manager.curr_step.initial_defined_state, steps_manager.curr_step.final_defined_state, {}, {}))
```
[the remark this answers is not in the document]
naterush: Just a requirement of the CodeChunks (although I'm not sure any of them use the prev_state...) so maybe we can chop it?

mitosheet/mitosheet/code_chunks/postprocessing/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+# at the end. Note that these should be transpiled with with the latest state in the
+# analysis
+
+
+from mitosheet.code_chunks.postprocessing.set_dataframe_format_code_chunk import SetDataframeFormatCodeChunk
+
+
+POSTPROCESSING_CODE_CHUNKS = [
+    SetDataframeFormatCodeChunk
+]
```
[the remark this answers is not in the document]
naterush: I think it's a pretty great solution! We'll def want to iterate it, but it's really simple and feels pretty robust!

mitosheet/mitosheet/step_performers/column_steps/change_column_dtype.py
```diff
@@ -193,9 +193,9 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
             post_state.dfs[sheet_index][column_header] = new_column
             pandas_processing_time = perf_counter() - pandas_start_time
 
-            # If we're changing between number columns, we keep the formatting on the column. Otherwise, we remove it
-            if not ((is_int_dtype(old_dtype) or is_float_dtype(old_dtype)) and (is_int_dtype(new_dtype) or is_float_dtype(new_dtype))):
-                post_state.column_format_types[sheet_index][column_id] = {'type': FORMAT_DEFAULT}
+            # If we're changing away from a number column, then we remove the formatting on the column if it exists
+            if not is_number_dtype(new_dtype) and column_id in post_state.df_formats[sheet_index]['columns']:
+                del post_state.df_formats[sheet_index]['columns'][column_id]
```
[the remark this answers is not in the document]
naterush: We don't. Currently, if you have a number format applied to a number column, and change it with a formula to a different type, it might still try and format it. 

I am fine with this - it's just a bit of extra generated code for a super rare case.

naterush:
@aarondr77 the only thing left on this PR is finishing up what selected/hovered items look like. I think we can move to what it was before (and just have to work not great for custom color users) - but your call. 

If you could finish this up and get it merged this week, would appreciate it!

PR #434, opened 2022-08-11
mitosheet: add back mitosheet-private package

naterush:
# Description

Check that it undoes the changes here: https://github.com/mito-ds/monorepo/pull/407/files
And that it activates pro. 

Test this by running the switch, and actually setting up the `mitosheet-private` package. **I would reccomend editing the macsetup.sh script to switch to the mitosheet-private package, and then just running that. It makes life much easier.**
# Documentation

No.

### DIFF
--- a/.github/workflows/deploy-mitosheet-private.yml
+++ b/.github/workflows/deploy-mitosheet-private.yml
@@ -1,0 +1,56 @@
+name: Deploy mitosheet-private
+on: 
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Package version to deploy to. Go to pypi and add 1 to the previous version of mitosheet-private"
+        required: true
+
+jobs:
+  deploy-mitosheet-private:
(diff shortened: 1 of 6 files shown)


naterush:
Two things:
1. I tried another approach (moving to a thin wrapper), but for some reason this was not working at all. So ditched this for now. 
2. I will test upgrading from the old version fully once we deploy, before we let them know this is good to go!


Also, **reset your user.json and env, to make sure you don't have pro for other reasons.**

aarondr77:
We should fix this: If the user doesn't go through signup because they are on pro (here, from mitosheet-private) they don't have the setup step. 

If they are on pro, can we just fill in that part of the checklist? 

Screen Shot 2022-08-11 at 2 49 40 PM

aarondr77:
The only thing that I don't see added back is the `.github/workflow/deploy-mitosheet-private.yml` file.

PR #447, opened 2022-08-25
mitosheet: pin to stable ipywidgets mito version

aarondr77:
# Description

Pins ipywidgets and jupyter-widgets packages to stable version with Mito. 

See errors caused [here](https://github.com/mito-ds/monorepo/issues/446)

Note [here](https://ipywidgets.readthedocs.io/en/8.0.1/migration_guides.html) works that we need to do to migrate to ipywidgets 8.0


dcnadler:
This is capping my ipywidgets at 7.0.5

Can you update this to
```
'ipywidgets~=7.0',
```
or 
```
'ipywidgets==7.*',
```

I'm assuming it was ipywdigets>=8.0 that was breaking?

Thanks!

PR #457, opened 2022-09-04
Metaprogram 2.0

naterush:
# Description

This PR does a few things:
1. Adds conditional formatting.
2. Adds df import into the sheet. I think the implementation is pretty simple + sweet (from a UI perspective) - and will extend well to different types of import. Let me know thoughts.
3. Overhauls the metaprogramming package to do a lot more useful things. We're going to keep investing in this, as it literally pays huge dividends!

### Duplicated Imports

There's one thing to figure out. If the user imports the same dataframe twice (either in the mitosheet.sheet call, or through the new interface), the generated code gets out of sync of the sheet. Thoughts? 

We could just auto-copy the dataframe the case you import two of them - I'm fine w/ that. Let me know.

### Empty Sheet

I'm thinking we could change the empty sheet so that it has two things:
1. A button that lets you import files.
2. Something that let's you easily click on a dataframe name to get it in the sheet immediately!

Thoughts? I think it could be really cool.

# Testing

Test df import. Test some interesting conditional formats. Some specific things to test with conditional formats is:
1. Weird indexes in df
2. Invalid filters
4. Changing types after making the filters
5. Etc. Put it in weird positions.

# Documentation

We will need to write docs before deploying!


naterush:
1. Fix highlighting (move styling up or down to the correct div).
2. Fix bug where column ids are in generated code
3. Test the generated code

aarondr77:
> How should we do highlighting on top of conditional formatting? 

Okay, I messed around with this for a little bit. Check out this approach: https://github.com/mito-ds/monorepo/pull/457/commits/a05bec70f2f11c484796de504eb2ce764d0b31d8

If this continues to give us problems, another proposal is to move the selection styling from the cell div itself to an additional div on top of each cell that can either be selected or not.

aarondr77:
I'm sure you're going to clean this up, but if you have the conditional format card open and the purple button is off of the screen, then you can't scroll down in the taskpane far enough to click the button.

aarondr77:
Don't forget to add the close buttons for the cards + make it so that only one of them is open at a time

aarondr77:
Note: the values in the conditionalFormattingResults.results are `null`, not `undefined` which the type system expects.

mitosheet/src/components/taskpanes/ConditionalFormatting/ConditionalFormattingCard.tsx
```diff
@@ -0,0 +1,243 @@
...
+            <XIcon 
+                onClick={(e) => {
+                    e.stopPropagation();
+                    onDelete();
+                }}
+            ></XIcon>
+        </Col>
+    );
+    
+    // TODO: should we sort these
```
naterush, line 92: TODO

mitosheet/src/components/taskpanes/ConditionalFormatting/ConditionalFormattingCard.tsx
```diff
@@ -0,0 +1,243 @@
...
+                />
+                <LabelAndColor 
+                    label="Background Color"
+                    color={backgroundColor}
+                    onChange={(newColor) => {
+                        const newConditionalFormats = [...props.df_format.conditional_formats];
+                        newConditionalFormats[props.index].backgroundColor = newColor;
+                        props.updateDataframeFormatParams({...props.df_format, conditional_formats: newConditionalFormats});
+                    }}              
+                />
```
naterush, line 235: TODO: cleanup this full file!

mitosheet/src/components/taskpanes/ConditionalFormatting/ConditionalFormattingTaskpane.tsx
```diff
@@ -0,0 +1,197 @@
...
+                        const newConditionalFormats = [...params.df_format.conditional_formats];
+                        newConditionalFormats.push(getDefaultEmptyConditionalFormat());
+                        console.log(newConditionalFormats);
+                        return updateDataframeFormatParams({conditional_formats: newConditionalFormats});
+                    }}
+                >
+                    Add Conditional Formatting Rule
+                </TextButton>
+            </DefaultTaskpaneBody>
+        </DefaultTaskpane>
```
naterush, line 193: Clean this up as well.

mitosheet/src/components/taskpanes/DataframeImport/DataframeImportTaskpane.tsx
```diff
@@ -0,0 +1,170 @@
...
+                    disabled={(params?.df_names.length || 0) === 0}
+                >
+                    Import Dataframes
+                </TextButton>
+            </DefaultTaskpaneFooter>
+        </DefaultTaskpane>
+    )
+}
+
+export default DataframeImportTaskpane;
```
naterush, line 170: TODO: cleanup this taskpane

mitosheet/src/components/taskpanes/taskpanes.tsx
```diff
@@ -76,7 +78,13 @@ export type TaskpaneInfo =
     | {type: TaskpaneType.SPLIT_TEXT_TO_COLUMNS, startingColumnID: ColumnID | undefined}
     | {type: TaskpaneType.MELT}
     | {type: TaskpaneType.SET_DATAFRAME_FORMAT}
-// AUTOGENERATED LINE: TASKPANEINFO (DO NOT DELETE)
+    | {type: TaskpaneType.CONDITIONALFORMATTING}
```
naterush, line 81: No

aarondr77:
Instead of generating a separate line of code for the color and background color, you can combine them like this

Screen Shot 2022-09-19 at 10 42 49 AM

aarondr77:
Note: these steps are not duplicating the bug that I thought they would. Noting them here until I can find the bug again. 

1. Import a dataframe
2. Apply 2 conditional formatting rules to it
3. Duplicate the dataframe
4. Go to the conditional formatting taskpane for the duplicated dataframe
5. Delete the rules from top top bottom
6. Notice that the rules change when getting deleted.

aarondr77:
The dataframe import taskpane shows any dataframe that is in memory, including those that are not used in the notebook anymore. This seems like a bad UX that we don't want to support. For example:

1. create a dataframe named `aaron`
3. Notice that the taskpane included `aaron`
4. Rename the dataframe `aaron1` 
5. Reload the mitosheet, open the taskpane
6. Notice the taskpane includes `aaron` and `aaron1`

If a user imports a dataframe that no longer gets created in the notebook, they will be able to import it, but as soon as they restart their kernel, the script will be broken without recourse via Mito. 

Proposal: Filter out any dataframes that are no longer created in the notebook or were created through the mitosheet that they are using. If a user is going to import a dataframe that was created through the mito sheet they are currently using, the correct behavior is to duplicate the dataframe, not import it.

aarondr77:
Screen Shot 2022-09-19 at 11 13 44 AM

aarondr77:
> I'm thinking we could change the empty sheet so that it has two things:

    A button that lets you import files.
    Something that let's you easily click on a dataframe name to get it in the sheet immediately!

I think having something like this would be really cool! Need to figure out the specifics. 

1. Should this be a multiselect so users can import multiple files? Should this just be a separate button that says "Import dataframes" and opens the taskpane?
2. Should we get rid of the mitosheet.sheet(df) text?

Screen Shot 2022-09-19 at 11 17 37 AM

aarondr77:
Screen Shot 2022-09-19 at 11 21 14 AM

I think the subtext would be more clear as: "Import an existing dataframe as a new sheet tab in Mito"

aarondr77:
Screen Shot 2022-09-19 at 11 23 31 AM

This text is cut off

aarondr77:
There are few missing features from the conditional formatting spec:
- [ ] Being able to reorder the conditional formatting rules to apply them in a different order. <- I think this is a good UX improvement, but I'm okay to cut it for now. 
- [ ] The four new filtering conditions (highest/lowest X values, most/least frequent X values). <- I think this is a big use case for conditional formatting and should be included.

aarondr77:
I'm not clear on when the we consider filters to be invalid for specific columns. For example, I expected that the `=` condition was going to be considered invalid if applied to a string column. Was that a purposeful decision to not have that be the case?

mitosheet/mitosheet/utils.py
```diff
@@ -82,7 +82,61 @@ def is_default_df_names(df_names: List[str]) -> bool:
...
+    formatted_result: ConditionalFormattingCellResults = dict()
+
+    for conditional_format in conditional_formatting_rules:
+        format_uuid = conditional_format["format_uuid"]
+        column_ids = conditional_format["columnIDs"]
+
+        for column_id in column_ids:
+            # Skip columns this is not applied to
+            if column_id not in column_ids:
+                continue
```
aarondr77, line 106: I don't think we need this check. We're iterating over colum_id in column_ids already

mitosheet/mitosheet/utils.py
```diff
@@ -82,7 +82,61 @@ def is_default_df_names(df_names: List[str]) -> bool:
...
+            backgroundColor = conditional_format.get("backgroundColor", None)
+            color = conditional_format.get("color", None)
+
+            column_header = state.column_ids.get_column_header_by_id(sheet_index, column_id)
+
+            # Use the get_applied_filter function from our filtering infrastructure
+            from mitosheet.step_performers.filter import get_full_applied_filter
+            try:
+                full_applied_filter, _ = get_full_applied_filter(df, column_header, 'And', filters)
+                applied_indexes = full_applied_filter[full_applied_filter].index.tolist()
```
aarondr77, line 121: This might be more clear as `df[full_applied_filter`] since we're applying the boolean index array to the dataframe to get the true indexes, right?

mitosheet/mitosheet/utils.py
```diff
@@ -82,7 +82,61 @@ def is_default_df_names(df_names: List[str]) -> bool:
...
+
+                for index in applied_indexes:
+                    # We need to make this index valid json, and do so in a way that is consistent with how indexes
+                    # are sent to the frontend
+                    json_index = json.dumps(index, cls=NpEncoder)
+                    formatted_result[column_id][json_index] = {'backgroundColor': backgroundColor, 'color': color}
```
aarondr77, line 127: I think instead of doing these by dataframe indexes, it would be better to do it as mito row indexes. We already talked about this in the spec, so up to you. But note that there is an existing bug with doing it this way.

mitosheet/src/components/Mito.tsx
```diff
@@ -699,6 +702,26 @@ export const Mito = (props: MitoProps): JSX.Element => {
                     selectedSheetIndex={uiState.selectedSheetIndex}
                 />
             )
+            case TaskpaneType.CONDITIONALFORMATTING: return (
```
aarondr77, line 705: Our convention is to do CONDITIONAL_FORMATTING

mitosheet/src/components/Mito.tsx
```diff
@@ -699,6 +702,26 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
+                <ConditionalFormattingTaskpane
+                    userProfile={userProfile}
+                    analysisData={analysisData}
+                    sheetDataArray={sheetDataArray}
+                    setUIState={setUIState}
+                    mitoAPI={props.mitoAPI}
+                    selectedSheetIndex={uiState.selectedSheetIndex}
+                />
+            )
+            case TaskpaneType.DATAFRAMEIMPORT: return (
```
aarondr77, line 715: convention should be IMPORT_DATAFRAME

mitosheet/src/jupyter/api.tsx
```diff
@@ -747,7 +765,24 @@ export default class MitoAPI {
         }, {})
     }
     
-    // AUTOGENERATED LINE: API (DO NOT DELETE)
+    
+    async editDataframeImport(
+        df_name: string,
+    ): Promise<void> {
```
aarondr77, line 771: We could use the `_edit` api call instead of creating a new one

mitosheet/src/utils/actions.tsx
```diff
@@ -576,8 +576,30 @@ export const createActions = (
...
+
+                setUIState(prevUIState => {
+                    return {
+                        ...prevUIState,
+                        displayImportToolbarDropdown: true
+                    }
+                })
+            },
+            isDisabled: () => {return undefined},
+            searchTerms: ['import', 'upload', 'new', 'excel', 'csv', 'add'],
```
aarondr77, line 598: Instead of having an action to open the import dropdown, we'd be better off just having one to to open each import taskpane

mitosheet/src/utils/actions.tsx
```diff
@@ -1139,14 +1161,58 @@ export const createActions = (
                     }
                 })
             },
-            isDisabled: () => {return undefined}, // TODO
+            isDisabled: () => {return undefined},
```
aarondr77, line 1164: This should be disabled if there are no dataframes.

mitosheet/src/utils/actions.tsx
```diff
@@ -1139,14 +1161,58 @@ export const createActions = (
...
+                        currOpenTaskpane: {type: TaskpaneType.CONDITIONALFORMATTING},
+                        selectedTabType: 'data'
+                    }
+                })
+            },
+            isDisabled: () => {
+                return doesAnySheetExist(sheetDataArray) ? undefined : 'There is no data to format. Import data.';
+            },
+            searchTerms: ['formatting', 'conditional', 'color', 'background', 'highlight'],
+            tooltip: "Format"
```
aarondr77, line 1188: "Set the background color and text color of the cell based on a condition"

aarondr77:
Screen Shot 2022-09-19 at 12 38 34 PM

Nice touch! Can we clean it up slightly by 1) adding a space between the column header and (invalid), 2) Making (invalid) mito-light-gray so that it stands out

mitosheet/mitosheet/code_chunks/empty_code_chunk.py
```diff
@@ -28,4 +28,11 @@ def get_code(self) -> List[str]:
         return []
 
     def combine_right(self, other_code_chunk: CodeChunk) -> Optional[CodeChunk]:
-        return other_code_chunk
\ No newline at end of file
+        # We just return the other code chunk, while also updating the prev_state. To avoid
+        # causing issues by modifying data, we make a copy of this object
+        return type(other_code_chunk)(
```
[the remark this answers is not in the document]
naterush: Yep! This returns the class, so it creates a new instance of this object (rather than editing it, which will 100% lead to bugs).

mitosheet/mitosheet/step_performers/dataframe_import.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def get_variable_with_name_from_caller(variable_name: str) -> Optional[Any]:
+    """
+    This helper function tries to find a variable with in the calling context. 
+    It will first check ipython to see if it is defined, and get the variable
+    from this context. 
+
+    If ipython is not defined, it will assume we are in some other python enviornment,
+    and attempt to get the variable from the callers stack frames.
```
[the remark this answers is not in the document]
naterush: Added a comment!

mitosheet/mitosheet/utils.py
naterush: I'm gonna leave this as is for now -- I think it's fine and reasonable to reference rows in this way. We can adjust this approach if it doesn't work going forward, but I think it's ok!

mitosheet/mitosheet/user/upgrade.py
```diff
@@ -164,14 +165,19 @@ def upgrade_user_json_version_6_to_7(user_json_version_6: Dict[str, Any]) -> Dic
...
+    long-term solution going forward.
     """
     # First, bump the version number
     user_json_version_6[UJ_USER_JSON_VERSION] = 7
 
     # Then, set new field
     if UJ_RECEIVED_CHECKLISTS not in user_json_version_6:
         current_experiment = user_json_version_6[UJ_EXPERIMENT]
-        if current_experiment['experiment_id'] != 'installer_communication_and_time_to_value':
+        if current_experiment['experiment_id'] != 'installer_communication_and_time_to_value' or package_name == 'mitosheet-private':
```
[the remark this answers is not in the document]
naterush: No, there shouldn't be. A random commit or two got pulled in, unclear why!

naterush:
@aarondr77 we can pick up the final part of the df import when we make a spec for it -- no rush I think!

aarondr77:
There are some issues with the flexibility in being flexible with which filter types are applied to which columns: if you apply the `=` filter to a string column with the filter value `sep`, then the generated code just contains `sep` instead of `sep` which throws an error when executing it. 

Keeping the flexibility is fine, but then we got to fix these up.

aarondr77:
Move the string filter conditions up in the dropdown

mitosheet/mitosheet/api/get_defined_df_names.py
```diff
@@ -30,7 +30,7 @@ def __exit__(self, *args):
     with Capturing() as output:
         ipython.run_line_magic("who",  "DataFrame")
         
-    output = [line.strip() for line in output]
+    output = [df.strip() for line in output for df in line.split("\t") if df.strip() != '']
```
aarondr77, line 33: It looks like you can instead do `who_ls` to get a list of the dataframe names. This would save us from this parsing. See here: https://stackoverflow.com/a/34950500/14993381

mitosheet/mitosheet/tests/step_performers/test_filter.py
```diff
@@ -271,6 +323,7 @@ def test_filter(df, condition, value, filtered_df):
     if filtered_df.empty and mito.dfs[0].empty:
         return True
 
+    print(mito.dfs[0], filtered_df)
```
aarondr77, line 326: remove

mitosheet/mitosheet/tests/test_utils.py
```diff
@@ -74,6 +61,8 @@ def check_dataframes_equal(test_wrapper):
     # Then, construct code that is just the code we expect, except at the end
     # it compares the dataframe to the final dataframe we expect
     def check_final_dataframe(df_name, df):
+        print(final_dfs[df_name])
+        print(df)
```
aarondr77, line 65: remove

mitosheet/mitosheet/tests/test_utils.py
```diff
@@ -85,6 +74,7 @@ def check_final_dataframe(df_name, df):
     )
 
     import mitosheet
+    print(code)
```
aarondr77, line 77: remove

aarondr77:
I think we should delete the `import all dataframes` from the import dataframes taskpane. I think too often users are going to have dataframes in that last that no longer exist in the notebook and its going to brick their analysis

aarondr77:
Mixpanel is not refreshing rn, but let's make sure the logs are good, especially for the import dataframes errors so we can track if that is an issue.

mitosheet/mitosheet/api/get_defined_df_names.py
naterush: For some reason, this isn't working in the context of this magic call! 

```
def get_df_names():
    from IPython import get_ipython
    from io import StringIO 
    import sys

    ipython = get_ipython()

    class Capturing(list):
        def __enter__(self):
            self._stdout = sys.stdout
            sys.stdout = self._stringio = StringIO()
            return self
        def __exit__(self, *args):
            self.extend(self._stringio.getvalue().splitlines())
            del self._stringio    # free up some memory
            sys.stdout = self._stdout

    with Capturing() as output:
        ipython.run_line_magic("who_ls",  "DataFrame")
                
    output = [line.strip() for line in output]
    return output

get_df_names()
```

Feel free to test, but gonna stick with the working approach for now.

naterush:
Import logs look good (and work on failure too). This is ready to merge once tests pass!

PR #460, opened 2022-09-06
mitosheet: bump ipywidgets and jupyterlab-widgets versions

aarondr77:
# Description

- Bumps ipywidgets on the mitosheet package to version 8
- Bumps jupyterlab-widgets on the mitosheet package to version 3, which is required by ipywidgets 8

### DIFF
--- a/mitosheet/setup.py
+++ b/mitosheet/setup.py
@@ -209,10 +209,10 @@ elif name == 'mitosheet' or name == 'mitosheet3' or name == 'mitosheet-private':
         packages                 = setuptools.find_packages(exclude=['deployment']),
         install_requires=[        
             "jupyterlab~=3.0",
-            'ipywidgets>=8,<9',
+            'ipywidgets~=8',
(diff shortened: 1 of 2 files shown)


mitosheet/setup.py
```diff
@@ -209,10 +209,10 @@
         packages                 = setuptools.find_packages(exclude=['deployment']),
         install_requires=[        
             "jupyterlab~=3.0",
-            'ipywidgets>=7,<9',
+            'ipywidgets>=8,<9',
             # In JLab 3, we move to needing to install the jupyterlab-widgets package, which
             # is equivlaent to the @jupyter-widgets/jupyterlab-manager extension for jlab 2
-            'jupyterlab-widgets>=1.0.0',
+            'jupyterlab-widgets>=3.0.0',
```
naterush, line 215: I think we just want >= 3. Otherwise, I think the resolution looks for a version with 3.0.x.

naterush (changes requested):

PR #462, opened 2022-09-07
mitosheet: bump to 2.0

naterush:
# Description

# Testing

# Documentation

### DIFF
--- a/mitosheet/setup.py
+++ b/mitosheet/setup.py
@@ -212,7 +212,7 @@ elif name == 'mitosheet' or name == 'mitosheet3' or name == 'mitosheet-private':
             'ipywidgets>=7',
             # In JLab 3, we move to needing to install the jupyterlab-widgets package, which
             # is equivlaent to the @jupyter-widgets/jupyterlab-manager extension for jlab 2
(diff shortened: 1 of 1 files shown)


mitosheet/setup.py
```diff
@@ -209,10 +209,10 @@
         packages                 = setuptools.find_packages(exclude=['deployment']),
         install_requires=[        
             "jupyterlab~=3.0",
-            'ipywidgets>=7,<9',
+            'ipywidgets>=7',
             # In JLab 3, we move to needing to install the jupyterlab-widgets package, which
             # is equivlaent to the @jupyter-widgets/jupyterlab-manager extension for jlab 2
-            'jupyterlab-widgets>=1.0.0',
+            'jupyterlab-widgets>=2.0',
```
aarondr77, line 215: Remember that jupyterlab-widgets 2.0 doesn't work with ipywidgets 8 though https://www.notion.so/trymito/Python-3-10-Error-cbec7bf27bae42e58083ece3bc052082#ba40eaa43d4d4b0d8c88c7db30290fcf

naterush: yes, this is just required for the tests to pass

PR #474, opened 2022-09-20
Error reduction 9/20

naterush:
# Description

Investigation here: https://www.notion.so/trymito/9-20-22-99326e7707f7431281cf925f848fc4e3

Closes https://github.com/mito-ds/monorepo/issues/466
Closes https://github.com/mito-ds/monorepo/issues/416
Need more info on: https://github.com/mito-ds/monorepo/issues/450 (@aarondr77 do you remember what you meant?)
This is one too hard: https://github.com/mito-ds/monorepo/pull/425

# Testing

Test all these changes.

# Documentation

N/A.

### DIFF
--- a/mitoinstaller/mitoinstaller/print_utils.py
+++ b/mitoinstaller/mitoinstaller/print_utils.py
@@ -1,6 +1,7 @@
 
 
 
+import sys
 import time
 import os
 from typing import List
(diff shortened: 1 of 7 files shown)


aarondr77:
The none parameter is still giving me errors when I switch the barnorm to fraction and then back to barnom.

PR #475, opened 2022-09-20
mitosheet: make experiment always go to B

naterush:
# Description

This is a much better easier hack than the other hack we used to remove the experiment -- I'm super happy with this simplification!

We def need to iterate the AB test infrastructure, but given that we're using it, pretty low priority! I also noted that the experiment ended today in the tracker, so we don't get confused when it stops!

# Testing

Test you get B!

# Documentation

No.

### DIFF
--- a/mitoinstaller/mitoinstaller/experiments/experiment_utils.py
+++ b/mitoinstaller/mitoinstaller/experiments/experiment_utils.py
@@ -3,8 +3,10 @@ from typing import Dict, Optional
 
 
 def get_random_variant() -> str:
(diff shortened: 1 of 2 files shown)


aarondr77:
I tested this and if I start w/ no user.json than I do go to variant B, but this doesn't stop all of the people who are currently on variant A from continuing to get that functionality. 

Do we want to take on the technical debt of all that additional code that we don't really want? You're call

naterush:
I removed a lot of this functionality from the installer in the installer fix PR. the frontend stuff I'll remove at some point, but it's ok to keep for now (it's really just the tour).

PR #476, opened 2022-09-21
Installer fix 9/21

naterush:
# Description

See here: https://www.notion.so/trymito/9-21-22-4cc50797022742c499d0c809bdfaad34

Check out the dashboard. The new, prettier installer is approx 0% better at successful installs, but it adds lots of complexity and makes it hard to prompt the user to do --user installs (which is the most common install error). 

# Testing

Test installer. Testing user install is tricky - but do your best to fudge the error (in the install_pip_packages function).

# Documentation

N/A.

### DIFF
--- a/mitosheet/mitosheet/preprocessing/preprocess_check_args_type.py
+++ b/mitosheet/mitosheet/preprocessing/preprocess_check_args_type.py
@@ -3,8 +3,10 @@
 
 # Copyright (c) Mito.
 # Distributed under the terms of the Modified BSD License.
+import collections
 from typing import Any, Dict, Collection, List, Optional, Tuple
 import pandas as pd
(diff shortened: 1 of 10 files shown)


mitosheet/mitosheet/preprocessing/preprocess_check_args_type.py
```diff
@@ -3,8 +3,10 @@
 
 # Copyright (c) Mito.
 # Distributed under the terms of the Modified BSD License.
+import collections
```
aarondr77, line 6: Make sure this works on Python 3.10

aarondr77: Tested, it works.

aarondr77:
LGTM

PR #477, opened 2022-09-23
Change import

aarondr77:
# Description

This PR adds the update imports taskpane. This is the first step in the plan of action. The next steps are to add the error messaging, correct popups, and undo/redo/clear. 

The two main things to focus on in this review are:
1. Because I very much wanted to reuse the existing import taskpanes so we can get error handling in them for free, we need to pass the updated import params around since we can't store state in the update import taskpane while the other taskpanes are open. This isn't so bad, but isn't ideal. It makes it hard, for example, to detect which imports have changed and which have not. 
2. The existing_import.py file. This is the method described in the spec, but there are more edge cases that I had to handle than initially expected (which is expected)
3. And of course, whatever else stands out. I'm going to do more polishing, but want to get your review of the main architecture before I move forward. 

# Testing

# Documentation


mitosheet/mitosheet/api/api.py
```diff
@@ -119,8 +120,10 @@ def handle_api_event(
     API must return the same ID that the incoming message contains,
     so that the frontend knows how to match the responses.
     """
+    print('handle api event')
```
aarondr77, line 123: remove

mitosheet/mitosheet/api/api.py
```diff
@@ -119,8 +120,10 @@ def handle_api_event(
...
     result: Union[str, List[str]] = ''
     params = event['params']
+    print('params: ', params)
```
aarondr77, line 126: remove

mitosheet/mitosheet/api/get_imported_files_and_dataframes.py
```diff
@@ -0,0 +1,88 @@
...
+                imported_files_and_dataframes.append({
+                    'step_id': step.step_id,
+                    'type': 'df',
+                    'import_params': {
+                        'df_names': [df_name],
+                        **step_params_copy
+                    }
+                })
+
+    print(imported_files_and_dataframes)
```
aarondr77, line 87: remove

mitosheet/mitosheet/mito_widget.py
```diff
@@ -149,6 +149,7 @@ def handle_update_event(self, event: Dict[str, Any]) -> None:
 
         try:
             self.steps_manager.handle_update_event(event)
+            print('got back to original mito_widget')
```
aarondr77, line 152: remove

mitosheet/mitosheet/updates/existing_imports.py
```diff
@@ -0,0 +1,65 @@
...
+    
+    # Refresh the anlaysis starting from the first step
+    steps_manager.execute_and_update_steps(steps_manager.steps_including_skipped, 0)
+
+
+
+    # Write the analysis
+    
+    #analysis['steps_data'])
+
```
aarondr77, line 58: remove

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -7,6 +7,7 @@
 Replays an existing analysis
 """
 
+from typing import Optional
```
aarondr77, line 10: remove

mitosheet/src/components/elements/RadioButtons.tsx
```diff
@@ -21,6 +21,8 @@ const RadioButtons = (props: RadioButtonsProps): JSX.Element => {
 
     const [possibleAnswers, setPossibleAnswers] = useState<string[]>([])
 
+    console.log('radio button values: ', props.values)
```
aarondr77, line 24: remove

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportWithDataframeTaskpane.tsx
```diff
@@ -0,0 +1,109 @@
...
+                        const newUpdatedImports: UpdatedImport[] = JSON.parse(JSON.stringify(props.updatedImports))
+                        newUpdatedImports[props.importIndex] = {
+                            ...newUpdatedImports[props.importIndex],
+                            type: 'df',
+                            import_params: {
+                                df_names: [selectedDFName]
+                            }                            
+                        }
+
+                        console.log('new udpated imports: ', newUpdatedImports)
```
aarondr77, line 92: remove

aarondr77 (commented):
cleanup workspace

mitosheet/src/components/elements/RadioButtonBox.tsx
```diff
@@ -0,0 +1,67 @@
...
+    */
+    width?: Width;
+    /** 
+        * @param className - Optional class name to add to the multi toggle box
+    */
+    className?: string
+}
+
+/**
+ * The RadioButtonBox component lets the user select on option from a set of options,
```
aarondr77, line 37: on -> one

mitosheet/src/hooks/useSendEditOnClick.tsx
```diff
@@ -93,7 +95,15 @@ function useSendEditOnClick<ParamType, ResultType>(
         setLoading(true);
         const newStepID = getRandomId(); // always use a new step id
 
-        const possibleError = await mitoAPI._edit<ParamType>(editEvent, finalParams, newStepID);
+        let possibleError = undefined
+        console.log(1)
```
aarondr77, line 99: remove

mitosheet/css/elements/MultiToggleBox.css
```diff
@@ -1,10 +1,10 @@
-.mutli-toggle-box-container {
+.multi-toggle-box-container {
```
naterush, line 1: I prefer mutli. It sounds like a healthy brand of cereal.

mitosheet/mitosheet/api/get_imported_files_and_dataframes.py
```diff
@@ -0,0 +1,89 @@
...
+            step_id: str
+            type: 'df'
+            df_names: List[str] <- Note: This should only have one entry
+        }
+	]
+	
+    """
+    imported_files_and_dataframes = []
+    for step in steps_manager.steps_including_skipped:
+        step_params_copy = step.params.copy()
```
naterush, line 51: This looks a bit scary to me, especially since we're deleting from it later. I wonder if making it a copy before returning it is safer, just so if things get moved around this line doesn't get lost.

mitosheet/mitosheet/tests/test_utils.py
```diff
@@ -1176,6 +1176,18 @@ def rename_graph(self, graph_id: GraphID, new_graph_tab_name: str) -> bool:
...
 
+    def update_existing_imports(self, updated_import_objs: List[Dict[str, Any]]) -> bool:
+        return self.mito_widget.receive_message(
+            self.mito_widget,
+            {
+                'event': 'update_event',
+                'type': 'existing_import_update',
+                'id': get_new_id(),
+                'params': {
+                    'updated_import_objs': updated_import_objs
```
naterush, line 1187: Something that ends in objs isn't a great name - can we delete that?

mitosheet/mitosheet/tests/updates/test_existing_import.py
```diff
@@ -0,0 +1,121 @@
...
+from mitosheet.tests.test_utils import create_mito_wrapper, create_mito_wrapper_dfs
+from mitosheet.tests.decorators import pandas_post_1_only, python_post_3_6_only
+
+TEST_EXCEL_FILE = 'excel_file.xlsx'
+TEST_CSV_FILE = 'csv_file.csv'
+TEST_CSV_FILE_TWO = 'csv_file_two.csv'
+
+
+@pandas_post_1_only
+@python_post_3_6_only
```
naterush, line 20: Why these?

mitosheet/mitosheet/tests/updates/test_existing_import.py
```diff
@@ -0,0 +1,121 @@
...
+    new_csv_import_step = mito.steps_including_skipped[1]
+    assert new_csv_import_step.step_type == 'simple_import'
+
+    new_excel_import_step = mito.steps_including_skipped[2]
+    assert new_excel_import_step.step_type == 'excel_import'
+
+    new_df_import_step = mito.steps_including_skipped[3]
+    assert new_df_import_step.step_type == 'dataframe_import'
+
+    assert mito.curr_step.df_names == ['csv_file', 'Sheet3', 'df2']
```
naterush, line 83: I feel like checking the dfs in this test would be a good call too!

mitosheet/mitosheet/tests/updates/test_existing_import.py
```diff
@@ -0,0 +1,121 @@
...
+            'type': 'excel',
+            'import_params': {
+                'file_name': TEST_EXCEL_FILE,
+                'sheet_names': ['Sheet3'],
+                'has_headers': True,
+                'skiprows': 0,
+            }
+        },  
+        {
+            'step_id': step_id,
```
naterush, line 63: Why do all of these have the same step id? Shouldn't they have the specific step id of the step they are trying to overwrite?

mitosheet/mitosheet/updates/existing_imports.py
```diff
@@ -0,0 +1,91 @@
...
+            step_performer = EVENT_TYPE_TO_STEP_PERFORMER[step_type + '_edit']
+
+            # Create the new step from the import_params
+            new_step = Step(
+                step_performer.step_type(), updated_import['step_id'], updated_import["import_params"]
+            )
+
+            if number_of_times_this_step_updated > 0:
+                # Insert the new step without overwriting any existing steps
+                steps_manager.steps_including_skipped.insert(i + number_of_times_this_step_updated, new_step)
```
naterush, line 72: There's an interface for manipulating the steps. Generally, you can see the process in the steps_manager.undo function, but what you want to do is:
1. Make a copy of all the steps
2. Edit this list
3. Then, use the function execute_and_update_steps

You don't want to edit the steps_including_skipped directly. This could be more clear in the interface certainly, but leave that state variable alone and only update once you're done making all the changes to the list.

mitosheet/mitosheet/updates/existing_imports.py
```diff
@@ -0,0 +1,91 @@
...
+    Updates the step list with new import steps
+
+    We must preserve the order of the sheets within the import. Consider this example:
+    In original step (step id: abc): imports sheet_1 and sheet_2 from .xlsx file
+    In updated imports, replaces those files with file_1.csv and file_2.csv
+    The updated_import obj for file_1 and file_2 will both have step id: abc, but we must ensure
+    that file_1.csv gets imported before file_2.csv so it replaces the correct file. 
+
+    Assumptions:
+    1. If we add multiple sheets in one step, the sheets get added to state in the same order as the params
```
naterush, line 32: This is a safe assumption and will remain one methinks!

mitosheet/mitosheet/tests/updates/test_existing_import.py
naterush: Oh. I see. These are updating the same step that imports three import. This is then what leads to the complexity about needing to check how many times a specific step has been updated. 

Architecturally, I think you're right we want a single object that corresponds to each step. So it should look like:
```
{
             'step_id': step_id,
             'import_updates': [
                 {
                    'type': 'excel',
                     'import_params': {
                          'file_name': TEST_EXCEL_FILE,
                          'sheet_names': ['Sheet3'],
                          'has_headers': True,
                          'skiprows': 0,
                      },
                  },
                   {
                    'type': 'excel',
                     'import_params': {
                          'file_name': TEST_EXCEL_FILE,
                          'sheet_names': ['Sheet3'],
                          'has_headers': True,
                          'skiprows': 0,
                      },
                  },
                  ...
             ]
         }, 
```

Then, we insist on the variant that the number of import_updates is equal to the number of imports in that original step. 

This would simplify a ton of update logic, without adding much work at all!

mitosheet/mitosheet/updates/existing_imports.py
```diff
@@ -0,0 +1,91 @@
...
+
+EXISTING_IMPORTS_UPDATE_EVENT = 'existing_import_update'
+EXISTING_IMPORTS_PARAMS = ['updated_import_objs']
+
+
+def execute_existing_imports_update(steps_manager: StepsManagerType, updated_import_objs: List) -> None:
+    """
+    Updates the step list with new import steps
+
+    We must preserve the order of the sheets within the import. Consider this example:
```
naterush, line 25: Per comment above, we def should refactor with new types to avoid having to do all this nasty counting of how many times we update a step!

mitosheet/src/components/elements/RadioButtonBox.tsx
```diff
@@ -0,0 +1,67 @@
...
+import '../../../css/elements/RadioButtonBox.css'
+import { Height, Width } from './sizes.d';
+import RadioButtonItem from './RadioButtonItem';
+
+
+interface RadioButtonsProps {
+    /**
+        * @param values - The options for the user to pick between 
+    */
+    values: string[];
```
naterush, line 13: I think we want this to have the same interface as the MultiToggleBox. Generally, making the creator of a radio button responsible for creating the RadioButtonItems is a much better call - as in this current architecture, it gets very complicated to have the displayed text different than the id that is displayed.

So, this should take children, and no selected value. The onChange can either be defined on the RadioButton (and then copied to the child - check out the MultiToggleBox `React.cloneElement`) - or it could be on each RadioButtonItem.

mitosheet/src/components/taskpanes/DataframeImport/DataframeImportTaskpane.tsx
```diff
@@ -35,7 +58,13 @@ const DataframeImportTaskpane = (props: DataframeImportTaskpaneProps): JSX.Eleme
...
 
     const {params, setParams, edit} = useSendEditOnClick<{df_names: string[]}, {df_names: string[]}>(
-        {df_names: []}, StepType.DataframeImport, props.mitoAPI, props.analysisData, {allowSameParamsToReapplyTwice: true}
+        {df_names: []}, 
+        StepType.DataframeImport, 
+        props.mitoAPI, 
+        props.analysisData, 
+        {allowSameParamsToReapplyTwice: true}, 
+        // If we're updating already imported data, then pass the override edit function
+        props.updateImportedData === undefined ? undefined : (newImportedData) => updateImportedDataWithDataframe(props.updateImportedData, newImportedData, props.setUIState)
```
naterush, line 67: Oh. I see what you did here. Hm. Let me think on this one.

naterush: There's another way to factor this, that at first blush seems much more maintainable to me long-term:
1. Abstract the DataframeImportTaskpane into a DataframeImportScreen and DataframeImportTaskpane. 
2. DataframeImportScreen is just pretty much everything except `useSendEditOnClick`. It takes params, setParams, and edit() function as props.
3. DataframeImportTaskpane is a thin wrapped on top of this, that has useSendEditOnClick and passes these directly to the DataframeImportScreen
4. You consume the DataframeImportScreen directly in the UpdateImports taskpane, without needing to worry about this function.

The benefits of this is that you don't have to change the `useSendEditOnClick` function, which is already quite complex, and probably should only be responsible for sending edits to the backend. This also doesn't really require any code duplication, as well, so it's not a messy refactor. It also ends with different components being responsible for different things in a way that is very sensible and not hard to reason about.

mitosheet/src/components/taskpanes/Import/CSVImport.tsx
```diff
@@ -184,10 +187,11 @@ function CSVImport(props: CSVImportProps): JSX.Element {
     // and not for importing the file
     const [fileMetadata, setFileMetadata] = useState<CSVFileMetadata | undefined>(undefined);
     const {params, setParams, loading, edit, editApplied, error} = useSendEditOnClick(
-        () => getDefaultParams(props.fileName),
+        () => getDefaultCSVParams(props.fileName),
         StepType.SimpleImport,
         props.mitoAPI, props.analysisData,
-        {allowSameParamsToReapplyTwice: true}
+        {allowSameParamsToReapplyTwice: true},
+        props.updateImportEdit
```
naterush, line 194: Probably the same architectural points here. I know it's a more code, but in practice it feels like we'll end up with a much better breakdown that doesn't really complicate logic as much.

naterush: I think I'd want to try it at least and see how it goes. It might not be the right move, but I I'd want to see.

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsTaskpane.tsx
```diff
@@ -0,0 +1,100 @@
...
+    {
+        step_id: string
+        type: 'df'
+        import_params: DataframeImportParams
+    }
+
+/* 
+    This is the updateImports taskpane.
+*/
+const updateImportsTaskpane = (props: updateImportsTaskpaneProps): JSX.Element => {
```
naterush, line 45: Should be caps. I thought this was a function at first.

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsUtils.tsx
```diff
@@ -0,0 +1,74 @@
...
+    newUpdatedImportObjs[importIndex] = {
+        ...newUpdatedImportObjs[importIndex],
+        type: 'df', 
+        import_params: newImportParams
+    }
+
+    setUIState(prevUIState => {
+        return {
+            ...prevUIState,
+            currOpenTaskpane: {type: TaskpaneType.UPDATEIMPORTS, updatedImportObjs: newUpdatedImportObjs}
```
naterush, line 35: Oh. I see what's happening here. Different taskpanes get opened, which then requires you to reopen this taskpane, so you're passing through the global state. That's messy.

I think the DataframeImportScreen approach would generally avoid all of this, and allow you to just get the params back directly from the screen. It would make the UpdateImportsTaskpane a simple tree, rather than a taskpane that is repeatedly opened and closed -- which is much better and long-term maintainable.

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsUtils.tsx
```diff
@@ -0,0 +1,74 @@
...
+        }
+    })
+}
+
+export const updateImportedDataWithFile = (
```
naterush, line 40: It's not immediately obvious to me that this couldn't be a single function with the one above -- is there a reason?

mitosheet/src/jupyter/api.tsx
```diff
@@ -552,6 +553,20 @@ export default class MitoAPI {
...
 
+    async getImportedFilesAndDataframes(): Promise<UpdatedImportObj[] | undefined> {
+        const resultString = await this.send<string>({
+            'event': 'api_call',
+            'type': 'get_imported_files_and_dataframes',
+            'params': {}
+        }, {})
+
+        if (resultString !== undefined && resultString !== '') {
+            return(JSON.parse(resultString).imported_files_and_dataframes);
```
naterush, line 564: We should just be able to return the the results of JSON.parse here - so change what we're returning from the backend so we can just make the types the same on both sides!

naterush (commented):
Good work getting this up, and thanks for requesting an intermediate review!

There are two large architectural changes I think we should make here - that I think might result in a bit more code but something that is much simpler and works more like the rest of our codebase. There are comments that explain each of them (and motivations) in more detail, but I can summarize them.
1. Change the types to group all updates to imports for a single step id together. This is fairly easy and also a no-brainer I think!
2. Refactor how we're reusing the other ImportTaskpanes. Generally, making it so we don't need to use the same `useSendEditOnClick` infrastructure at all, and can effectively keep the UpdateImportTaskpane open at all times. 

The latter is a fairly big refactor, I think. But it has the benefit of making the UpdateImport taskpane contain all it's own state, and not have to reopen and reclose. It also has the benefit of not complicating the useSendEditOnClick functionality, which I think long-term is really important. 

It likely will result in more code, but also code that is much simpler and works in a way that is very easy to understand - the UpdateImport taskpane stays open, and displays different import screens (that are also displayed by the other, standard import taskpanes) -- it just does different things with the params when it gets them!

Once you read through this review and figure out what you think about it (would love to hear any thoughts / comments / questions / pushback!), shoot me a message in Zulip and we can zoom and chat :-) 

Talk soon much love :-}

mitosheet/css/elements/MultiToggleBox.css
aarondr77: lol

mitosheet/mitosheet/tests/updates/test_existing_import.py
aarondr77: Cause we're using the excel import which requires this I believe. I took it becuase the Excel import tests were marked with it.

mitosheet/mitosheet/updates/existing_imports.py
aarondr77: How come?

mitosheet/mitosheet/tests/updates/test_existing_import.py
aarondr77: Yeah there is a tradeoff between having simple logic on the backend and having simpler logic on the frontend. If we move to this, then we'll have to keep track of more information about which specific sheet within which step is getting updated. 

We'll also still need to keep track of how many times we've updated that step on the backend so we can insert the steps in the correct location. 

I think there are other ways we could write the algorithm that would make it so we don't need to do that, but I'm not sure that this data type is actually a requirement for that. Although maybe we want it anyway

mitosheet/src/components/elements/RadioButtonBox.tsx
aarondr77: I just did it this way because I didn't want to over architect the radio button for the current use case. We don't need anything fancy now.

aarondr77:
Priority:
1. Do the frontend refactor, create a spec first. 
2. Do the type refactor

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsTaskpane.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+    const [originalStepImportData] = useStateFromAPIAsync(
+        undefined,
+        () => {return props.mitoAPI.getImportedFilesAndDataframes()},
+        (loadedData) => {
+            // On load, update the updated import data
+            setUpdatedStepImportData(loadedData || [])
+        }
+    )
+
```
aarondr77, line 79: I don't see a distinction between the `updatedStepImportData` and `originalStepImportData`. Currently, they are the same thing as we are setting the updatedStepImportData during the `useStateFromAPIAysnc` call. 

After I update a file import, the `originalStepImportData` is also getting updated somewhere

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsTaskpane.tsx
```diff
@@ -0,0 +1,236 @@
...
+                        disabled={undefined } // TODO, disable this if there is an error
+                    >
+                        <p>
+                            Update Imports
+                        </p>
+                    </TextButton>
+                </DefaultTaskpaneFooter>
+            </DefaultTaskpane>
+        )
+    } else if (replacingDataframeState.screen === 'file_browser') {
```
aarondr77, line 169: Instead of having one case for file_browser, csv_import, and xlsx_import, I could see having just one case for `file` which handles all of this. It would be structured just like the `FileImportTaskpane` but for updating a file. 

So the structure would be:
- UpdateImportsTaskpane
   - ExistingImports
   - UpdateWithFileTaskpane
      - UpdateFileBrowserTaskpane
      - UpdateCSVImportTaskpane
      - UpdateXLSXImportTaskpane
   - UpdateWithDataframeTaskpane

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsTaskpane.tsx
```diff
@@ -0,0 +1,236 @@
...
+        return (
+            <UpdateFileBrowserTaskpane
```
aarondr77, line 171: Can we stick to the convention of only adding `Taskpane` to the end of components that exist on the `Taskpane` type for clarity?

mitosheet/src/components/taskpanes/FileImport/FileImportTaskpane.tsx
```diff
@@ -0,0 +1,100 @@
...
+                currPathParts={props.currPathParts}
+                setCurrPathParts={props.setCurrPathParts}
+            
+                selectedFile={selectedFile}
+                setSelectedFile={setSelectedFile}
+
+                setScreen={setScreen}
+            />
+        )
+    } else if (screen == 'csv_import') {
```
aarondr77, line 70: While we're in this refactor, can we change the name of `csv_import` and `CSVImportTaskpane` to make it clear that this is the csv import params configuration screen? As is, it feels like it is how you actually import the csv, but most of the time I just double click on the file browser element to import it.

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsTaskpane.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+export interface StepImportData {
+    step_id: string,
+    imports: (CSVImportData | ExcelImportData | DataframeImportData)[];
+}
+
+export interface DataframeCreationIndex {step_id: string, index: number}
+export interface ReplacingDataframeState {
+    dataframeCreationIndex: DataframeCreationIndex, 
+    screen: ImportScreen | 'dataframe_import',
```
aarondr77, line 56: Since there is now a difference between the import screen and the update screens, maybe we should separate them here too.

mitosheet/mitosheet/updates/existing_imports.py
```diff
@@ -0,0 +1,51 @@
...
+        for _import in imports:
+            import_steps_to_replace_with.append(
+                Step(_import['step_type'], create_step_id(), _import["params"])
+            )
+
+        # Then, replace the single old step with all the new steps
+        new_steps = new_steps[:original_step_index] + import_steps_to_replace_with + new_steps[(original_step_index + 1):]
+
+    # Refresh the anlaysis starting from the first step
+    steps_manager.execute_and_update_steps(steps_manager.steps_including_skipped, 0)
```
aarondr77, line 44: I think we're missing actually updating the `steps_including_skipped`. By the time we `execute_and_update_steps`, we've only updated the copy of the steps.

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsTaskpane.tsx
aarondr77: This would allow us to keep track of less state in the top level that I think will make this UpdateImportsTaskpane simpler moving forward. There's a lot of state here that is a bit confusing.

naterush:
1. Fix up the Import taskpane totally. Test all the import taskpanes a lot (with errors, etc). 
2. Finish up styling on the replacement stuff. 
3. Cleanup everything. 
4. Request a review. 

Optionally, we can get the rest of the spec working. I'm not sure exactly what is included, so have to check.

mitosheet/mitosheet/api/get_imported_files_and_dataframes.py
```diff
@@ -0,0 +1,98 @@
...
+from typing import Any, Dict, List, Optional
+
+import pandas as pd
+from mitosheet.types import StepsManagerType
+from mitosheet.step_performers.import_steps.simple_import import SimpleImportStepPerformer
+from mitosheet.step_performers.import_steps.excel_import import ExcelImportStepPerformer
+from mitosheet.step_performers.import_steps.dataframe_import import DataframeImportStepPerformer
+from mitosheet.step_performers.import_steps import is_import_step_type
+
+def get_index_from_possible_null_list(l: Optional[List[Any]], index: int) -> Optional[Any]:
```
naterush, line 18: TODO: make this function more useful, cuz as of now it's always wrapped by the same call

mitosheet/mitosheet/api/get_imported_files_and_dataframes.py
```diff
@@ -0,0 +1,98 @@
...
+    if l is None:
+        return None
+        
+    return l[index]
+
+def get_import_data_with_single_import_list(step_type: str, params: Dict[str, Any]) -> List[Dict[str, Any]]:
```
naterush, line 24: Comment and explain this!

mitosheet/mitosheet/api/get_imported_files_and_dataframes_from_analysis_name.py
```diff
@@ -0,0 +1,67 @@
...
+from typing import Any, Dict, List
+
+import pandas as pd
+from mitosheet.types import StepsManagerType
+from mitosheet.step_performers.import_steps import is_import_step_type
+from mitosheet.saved_analyses import read_and_upgrade_analysis
+from mitosheet.api.get_imported_files_and_dataframes import get_import_data_with_single_import_list
+
+
+def get_step_import_data_from_saved_analysis(analysis_name: str) -> List[Dict[str, Any]]:
```
naterush, line 18: Doc string

mitosheet/mitosheet/api/get_imported_files_and_dataframes_from_analysis_name.py
```diff
@@ -0,0 +1,67 @@
...
+            if is_import_step_type(step_type):
+                step_import_data_list.append({
+                    'step_id': 'fake_id', 
+                    'imports': get_import_data_with_single_import_list(step_type, params)
+                })
+        
+        return step_import_data_list
+
+
+def get_imported_files_and_dataframes_from_analysis_name(params: Dict[str, Any], steps_manager: StepsManagerType):
```
naterush, line 38: TODO: explain the relationship to the other API call

mitosheet/mitosheet/api/get_test_imports.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import Any, Dict, List
+
+from mitosheet.errors import MitoError
+from mitosheet.saved_analyses.save_utils import read_and_upgrade_analysis
+from mitosheet.step import Step
+from mitosheet.step_performers.import_steps import is_import_step_type
+from mitosheet.types import StepsManagerType
+from mitosheet.api.get_imported_files_and_dataframes import get_import_data_with_single_import_list
+
+GENERIC_DATA_ERROR = 'There was an error importing this data. Please select a different file or dataframe and try again.'
```
naterush, line 17: TODO: can we do better on the errors here, depending on the step type?

mitosheet/mitosheet/errors.py
```diff
@@ -429,6 +429,20 @@ def make_no_analysis_error(analysis_id: str, error_modal: bool=True) -> MitoErro
         error_modal=error_modal
     )
 
+def make_invalid_update_imports_error() -> MitoError:
+    """
+    Helper function for creating a invalid_update_imports_error.
+
+    Occurs when the user tries to update imports to an invalid
```
naterush, line 436: When?

naterush: TODO: test it occurs here in both update cases?

mitosheet/mitosheet/step.py
```diff
@@ -129,7 +129,7 @@ def final_defined_state(self) -> State:
         return self.post_state if self.post_state is not None else \
             (self.prev_state if self.prev_state is not None else State([]))
 
-    def set_prev_state_and_execute(self, new_prev_state: State) -> None:
+    def set_prev_state_and_execute(self, new_prev_state: State) -> bool:
```
naterush, line 132: Do we need this still?

mitosheet/mitosheet/tests/updates/test_update_existing_import.py
```diff
@@ -0,0 +1,207 @@
...
+                }
+           ]
+        }
+    ]
+
+    mito.update_existing_imports(updated_import_obj)
+    mito.redo()
+
+    # Make sure the updates occured correctly 
+    assert mito.dfs[0].equals(pd.DataFrame({'Unnamed: 0': [0, 1, 2], 'A': [1, 2, 3], 'B': [1, 2, 3]}, index=[0, 1, 2]))
```
naterush, line 207: More test:
1. A later import fails, reverts earlier change (aka, atomicity)
2. Test import properly tests success
3. Test the error conditions are properly sent back -- can we do this?

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -6,21 +6,57 @@
...
     'analysis_name',
+    'import_data_to_overwrite'
 ]
 
+def overwrite_import_data(analysis: Dict[str, Any], import_data_to_overwrite: List[Dict[str, Any]]) -> Dict[str, Any]:
+    """
+    If there is import data to overwrite, we go through and overwrite it. We notably have to do this
+    by keeping track of the index of the import, as we do not have step ids to link these things. 
+    
+    TODO: we could optimize and combine these together, but it's not the biggest deal :-)
```
naterush, line 30: TODO

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -6,21 +6,57 @@
...
+    """
+
+    all_imports = [_import for import_data in import_data_to_overwrite for _import in import_data['imports']]
```
naterush, line 33: Cleanup and comment

mitosheet/mitosheet/updates/update_existing_imports.py
```diff
@@ -0,0 +1,71 @@
...
+        arr = imports_for_step_id.get(step_id, [])
+        arr.extend(imports)
+
+        imports_for_step_id[step_id] = arr
+
+    for step_id, imports in imports_for_step_id.items():
+        original_step_index = [index for index, step in enumerate(new_steps) if step.step_id == step_id][0]
+
+        # Build all the new steps
+        # TODO: we could combine them, but this is done in code optimization otherwise. It might lead to strange undo/redo
```
naterush, line 44: TODO: cleanup and comment this entire file! Doc string as well.

mitosheet/mitosheet/utils.py
```diff
@@ -321,6 +322,9 @@ def get_random_id() -> str:
 def get_new_id() -> str:
     return str(uuid.uuid4())
 
+def create_step_id() -> str:
```
naterush, line 325: remove this func

mitosheet/src/components/elements/RadioButtonBox.tsx
naterush: Fair enough!

mitosheet/src/components/icons/RightPointerIcon.tsx
```diff
@@ -0,0 +1,15 @@
+
+// Copyright (c) Mito
+
+import React from 'react';
+
+const RightPointerIcon = (): JSX.Element => {
```
naterush, line 6: TODO: finish up the styling in the card, so they look like the spec!

mitosheet/src/components/import/CSVImportScreen.tsx
```diff
@@ -180,76 +182,59 @@ function getSuccessMessage(fileElement: FileElement): string {
...
-    const currentDelimeter = params.delimeters[0];
-    const currentEncoding = params.encodings[0] === 'default' ? 'utf-8' : params.encodings[0];
-    const currentErrorBadLines = params.error_bad_lines[0];
+    const delimeters = props.params.delimeters;
+    const encodings = props.params.encodings;
+    const error_bad_lines = props.params.error_bad_lines
+
+    const currentDelimeter = delimeters !== undefined ? delimeters[0] : DEFAULT_DELIMETER;
+    const currentEncoding = ((encodings !== undefined ? encodings[0] : DEFAULT_ENCODING) === 'default') ? 'utf-8' : (encodings !== undefined ? encodings[0] : 'utf-8');
+    const currentErrorBadLines = error_bad_lines !== undefined ? error_bad_lines[0] : DEFAULT_ERROR_BAD_LINES;
```
naterush, line 227: Can I cleanup all this nonsense?

naterush (changes requested):
Major cleanup!

naterush:
There are a few things left to do in this PR:
1. Extensive testing.
2. Minor styling fixes, potentially.
3. Adding a search to the RadioButtonBox.

That being said, as 95% of this is implemented, I'd love to get a review. The architecture changes I ended up making ended up being not so massive, and mostly result in an extremely clean interface for importing things that I am super happy with.

naterush:
I'm getting an extremely annoying bug where the intro checklist keeps popping up. Why the hell would this happen? Unclear.

aarondr77:
Missing space here

Screen Shot 2022-10-09 at 9 57 38 PM

aarondr77:
The update with dataframe taskpane still has the "Add" button 

Screen Shot 2022-10-09 at 9 59 28 PM

aarondr77:
Let's add a tooltip "Resolve all errors in the taskpane before updating imports" 

Screen Shot 2022-10-09 at 10 09 19 PM

aarondr77:
There's a weird case we don't seem to handle where:
1. You replay an analysis and it doesn't succeed, opening the update import taskpane because a file could not be imported. 
2. Instead of using the update import taskpane to resolve the error, you just import a new file 
3. The sheet updates and you can use it as you normally would, but no generated code appears and the analysis ID does not change. 

We should not let the user do anything until they fix the imports. They should instead of have to start a new analysis if they want to start over.

aarondr77:
This error is getting printed to my python kernel console, but no error message is appearing in Mito, when I update the imported data to a file that doesn't contain columns that are needed. 

```
mitosheet.errors.MitoError: ('invalid_update_imports_error', 'Cannot update imports', 'The newly imported data does not have the same structure as the original data, and as such the analysis failed to apply. Please update the imports to preserve the structure of the data and try again.')
```

aarondr77:
Prior to executing the update, maybe this should say `Update to` instead of `Updated to`

Screen Shot 2022-10-09 at 10 20 39 PM

aarondr77:
"The newly imported data does not have the same structure as the original data, and as such the analysis failed to apply. Please update the imports to preserve the structure of the data and try again." - "The analysis failed to refresh because the  newly imported data does not have the same structure as the original data. Please update the imports to preserve the structure of the data and try again."

Screen Shot 2022-10-09 at 10 21 44 PM

aarondr77:
I should only be able to press the update button once all of errors are resolved. Currently, once I resolved the first error, the button became enabled. 

Screen Shot 2022-10-09 at 10 25 16 PM

aarondr77:
Because the update button allowed me to press it even when not all of the files were updated, I got in this state where I had double error messages. If the error messages are duplicated, let's just show one of them. 

Screen Shot 2022-10-09 at 10 27 13 PM

aarondr77:
If the analysis fails to replay and I update the file imports, then when I use Undo, it undoes the steps in the analysis. 
If the analysis does not fail to replay, and I update the imports, then when I use Undo, it undoes the changed imports. 

Can we standardize it to always undoing the changed imports?

aarondr77:
- Import 3 files 
- Change the name of the first file so that the analysis will fail on replay
- Replay the analysis
- Update all of the imports so that the three imports are [df1, df2, df1]
- Update the imports
- Notice that the sheet tabs are named df1, df1, df1

aarondr77:
"delemeter" -> "delimiter"
"delimeter" -> "delimiter"

We should run a find and replace on this
Screen Shot 2022-10-09 at 10 42 41 PM

aarondr77:
If I try to update a file to the `skiplines.txt` file, it doesn't display an error until I hit the final `Update` button. In the spec, we say that we want to validate that the import will at least be successful when we are still in the import configuration phase so they get the relevant info when they are trying to import. Can we push that check up?

aarondr77:
Here are some other things from the spec that are not yet implemented: 
1. This popup / button Screen Shot 2022-10-09 at 10 54 32 PM

aarondr77:
Different behavior depending on the size of the analysis getting replayed. 

- Import single-sheet-excel from the large datasets folder
- Import df1
- Refresh the kernel 
- Run the mitosheet.sheet() call before creating df1 (notice the replay analysis loading indicator appears) 
- Notice that the error taskpane never appears, but the error does get generated in the Python kernel. 

If instead, you just: 
- Import df1
- Refresh the kernel 
- Run the mitosheet.sheet() call before creating df1 (notice no loading indicator appears) 
- Notice that the error taskpane does appear.

mitosheet/src/components/taskpanes/FileImport/CSVImportConfigTaskpane.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    const {params, setParams, loading, edit, editApplied, error} = useSendEditOnClick(
+        () => getDefaultCSVParams(props.filePath),
+        StepType.SimpleImport,
+        props.mitoAPI, props.analysisData,
+        {allowSameParamsToReapplyTwice: true},
+    )
+
+    return (
+        <CSVImportScreen
```
aarondr77, line 46: Oh can we update the name of this component. It looks like it should be called `CSVImportConfigScreen`

aarondr77 (commented):
Okay! Architecture looks great. My understanding is that we've turned the taskpanes into componenents and then we use them by just defining the state, setState, edit functions for that component depending on if we are using it in an import or update context. 

In general, I think we need to do a large amount of usability testing this, starting with ourselves. I think there are probably some very rough edges that we want to catch before releasing. I started that process and found a few in this report below!

aarondr77:
If I import some data through the mitosheet.sheet() call and some data through the import files taskpane, then if the replay analysis fails because the file no longer exists, the import screen appears, but the imported dataframe is also displayed in the spreadsheet, which is a bit confusing. Replaying the analysis should either succeed or fail completey

aarondr77:
- Import a file `file.csv`
- Take the file `skiplines.txt` and name is `file.csv`
- Rerun the mitosheet.sheet() call
- You'll see this error message. 

Screen Shot 2022-10-09 at 11 34 58 PM

Instead of displaying this message, I think it should acknowledge that the the import failed and tell them it failed because the delimter or encoding was incorrect. Or something like that. And then tell them to try changing the delimiter. 

Mostly, there should just be part of this taskpane that says the replay analysis failed.

aarondr77:
- Import df1
- Import all 3 files from the multi-sheet.xlsx
- restart browser
- rerun mitosheet.sheet() without first defining df1
- see in the error taskpane that only the first sheet of the multi-sheet.xlsx is showed as imported.

mitosheet/mitosheet/api/get_imported_files_and_dataframes.py
```diff
@@ -0,0 +1,108 @@
...
+                'step_type': step_type,
+                'params': {
+                    'df_names': [df_name],
+                }
+            }]
+
+    return []
+
+
+def get_imported_files_and_dataframes(params: Dict[str, Any], steps_manager: StepsManagerType) -> str:
```
aarondr77, line 71: Maybe we can call this `get_imported_files_and_dataframes_from_steps_manager` so that its contrasted with `get_imported_files_and_dataframes_from_analysis_name`. I was confused about them in the beginning.

mitosheet/mitosheet/api/get_imported_files_and_dataframes.py
```diff
@@ -0,0 +1,108 @@
...
+	
+    """
+
+    # Then, we turn import steps into a version of the step that just creates a single dataframe
+    # as this is what is easiest to work with on the frontend
+    import_steps_with_just_one_dataframe = []
+    for step in steps_manager.steps_including_skipped:
+        if is_import_step_type(step.step_type):
+            import_steps_with_just_one_dataframe.append({
+                'step_id': step.step_id,
```
aarondr77, line 103: I don't understand why we use the actual step_id here and we use `fake_id` in `get_imported_files_and_dataframes_from_analysis_name`

It seems like we need the id so we can match the step_id in the function `execute_update_existing_imports_update` at line 42: 

`original_step_index = [index for index, step in enumerate(new_steps) if step.step_id == step_id][0]
`

mitosheet/mitosheet/step_performers/import_steps/__init__.py
```diff
@@ -2,4 +2,13 @@
...
+# Distributed under the terms of the GPL License.
+
+from mitosheet.step_performers.import_steps.simple_import import SimpleImportStepPerformer
+from mitosheet.step_performers.import_steps.excel_import import ExcelImportStepPerformer
+from mitosheet.step_performers.import_steps.dataframe_import import DataframeImportStepPerformer
+
+def is_import_step_type(step_type: str) -> bool:
+    return step_type == SimpleImportStepPerformer.step_type() \
+        or step_type == ExcelImportStepPerformer.step_type() \
+        or step_type == DataframeImportStepPerformer.step_type()
```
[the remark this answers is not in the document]
naterush: I think it's kewl

mitosheet/mitosheet/tests/updates/test_update_existing_import.py
```diff
@@ -0,0 +1,327 @@
...
+                }
+            }]
+        }
+    ])
+
+    assert mito.dfs[0].equals(df)
+    assert mito.dfs[1].equals(df1)
+
+
+def test_test_import_returns_good_data():
```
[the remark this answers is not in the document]
naterush: Actually, it's testing test_import -- so all g.

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsPostReplayTaskpane.tsx
```diff
@@ -0,0 +1,135 @@
...
+        undefined,
+        () => {return props.mitoAPI.getImportedFilesAndDataframes()},
+        (loadedData) => {
+            // We default the prevUpdatedStepImportData to be the the original import data, if it's undefined
+            if (loadedData !== undefined && props.updatedStepImportData === undefined) {
+                props.setUpdatedStepImportData(loadedData)
+            }
+        },
+        []
+    )
```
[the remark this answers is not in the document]
naterush: They actually aren't the same. The structurally look very similar, but there are differences in when the buttons are disabled (one is in the case

They also call different functions when you actually click the button - one replays the analysis, the other is overwriting imports. 

I don't think we'd save a ton by factoring them into a shared component... I tried a bit and it wasn't so helpful.

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsPreReplayTaskpane.tsx
```diff
@@ -0,0 +1,137 @@
...
+    useEffect(() => {
+        props.setUpdatedStepImportData(prevUpdatedStepImportData => {
+            if (prevUpdatedStepImportData === undefined) {
+                return JSON.parse(JSON.stringify(props.failedReplayAnalysisOnImports?.importData));
+            }
+            return prevUpdatedStepImportData;
+        })
+    }, [])
+
+    const [invalidReplayError, setInvalidReplayError] = useState<string>('Mito failed to replay the analysis because it could not import the files or used in the analysis. Correct the errors below.');
```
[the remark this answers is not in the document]
naterush: Actually, the default in this case is always that there was an import error. So the og message is correct.

mitosheet/mitosheet/updates/update_existing_imports.py
```diff
@@ -0,0 +1,72 @@
...
+    contain a single dataframe creation, as this is what is created by the
+    frontend for convenience.
+    """
+    old_steps = copy(steps_manager.steps_including_skipped)
+    new_steps = copy(steps_manager.steps_including_skipped)
+
+    # We group up all the steps by their ids
+    imports_for_step_id: Dict[str, List[Dict[str, Any]]] = dict()
+    for step_import_data in updated_step_import_data_list:
+        step_id = step_import_data['step_id']
```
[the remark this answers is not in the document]
naterush: `fake_id` is only used when you're replaying the full analysis!

mitosheet/mitosheet/updates/update_existing_imports.py
```diff
@@ -0,0 +1,72 @@
...
+# Copyright (c) Saga Inc.
+# Distributed under the terms of the GPL License.
+from copy import copy
+from typing import Any, Dict, List
+from mitosheet.step import Step
+from mitosheet.types import StepsManagerType
+from mitosheet.utils import create_step_id
+
+
+EXISTING_IMPORTS_UPDATE_EVENT = 'update_existing_import_update'
```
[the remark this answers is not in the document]
naterush: Yeah, forgot to update this. The thing is that this update event is actually called `update_existing_imports`. So the name of the event string is actually correct with both update at the start and end!

mitosheet/mitosheet/api/get_imported_files_and_dataframes_from_analysis_name.py
```diff
@@ -0,0 +1,51 @@
...
+        return []
+    else:
+        # Find all the import steps, and turn them into imports
+        step_import_data_list = []
+        for step_data in saved_analysis['steps_data']:
+            step_type = step_data['step_type']
+            params = step_data['params']
+            if is_import_step_type(step_type):
+                step_import_data_list.append({
+                    'step_id': 'fake_id',
```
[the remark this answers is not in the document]
naterush: Added a helpful comment!

mitosheet/mitosheet/tests/updates/test_update_existing_import.py
```diff
@@ -0,0 +1,327 @@
...
+                'params': {
+                    'df_names': ['df100000fake']
+                }
+            }]
+        }
+    ]}, mito.mito_widget.steps_manager))
+
+    assert result["1"] == CSV_IMPORT_ERROR
+    assert result["3"] == EXCEL_IMPORT_ERROR
+    assert result["5"] == DATAFRAME_IMPORT_ERROR
```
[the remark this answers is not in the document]
naterush: Which other indexes?

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -6,27 +6,76 @@
...
+        params = step_data['params']
+
+        if not is_import_step_type(step_type):
+            # If it's not an import step, we don't need to change anything
+            final_analysis['steps_data'].append(step_data)
+        else:
+            # If it is an import step, they we replace each dataframe creation with a single
+            # step that is created with the new updated dataframe creation data 
+            num_imports_in_step = len(get_import_data_with_single_import_list(step_type, params))
+            for i in range(num_imports_in_step):
```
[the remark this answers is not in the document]
naterush: Potentially, but there's not an easy way around it. I also think this is a niche case, likely.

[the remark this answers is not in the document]
naterush: We can also get around this with better code optimization alone!

mitosheet/mitosheet/updates/update_existing_imports.py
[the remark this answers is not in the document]
naterush: Added a helpful comment about this.

mitosheet/mitosheet/api/get_imported_files_and_dataframes.py
naterush: Yeah. I added a comment in the the `get_imported_files_and_dataframes_from_current_steps` and `get_imported_files_and_dataframes_from_analysis_name` functions. See them.

Note that when we're updating import pre replay, we use the replay_analysis api call. When we're updating imports post replay, we use the update_imports call (where we use the step ID).

naterush:
Ok... things have gotten much better, a bunch of really serious bugs have been squashed, and the error reporting is actually much improved. The error messages are likely to give you good information most of the time -- and all we have to do is add better error messages to the rest of Mito to get them here!

There are a few things left to do:
1. Add the popup. I'm happy to add this -- you're call -- but I'd also be down to roll out after we get a bit more feedback in the wild on this feature (e.g. after it's deployed for a bit and we see if users have issue with it). This feature is tricky, and I'm concerned about confusing literally everybody.
2. Make it so you can't click other buttons in the toolbar if an analysis fails to replay. I am not sure of a good way to do this -- what do you think? We could deactivate all these buttons, but that seems like a... ton. Any other ideas?
3. Add the search the the RadioButtonBox.

A 1 hour review of functionality and the code changes would be greatly appreciated - to see if you like the changes I ended up making here. Then, I'll do one final cleanup, a final review, and we'll be done. Sorry this is such a marathon -- it's massive...

aarondr77:
Can we try to center this?

Screen Shot 2022-10-12 at 9 51 14 AM

aarondr77:
- Import file 
- Change imports 
- Undo change imports
- Add a column to the sheet 
- Reference the first column in that sheet
- Notice this error

Screen Shot 2022-10-12 at 9 53 27 AM

aarondr77:
Can we:
1. Use a display version of the step name instead of filter_step
2. Say: "Failed to change import files because .... "

Screen Shot 2022-10-12 at 9 57 24 AM

aarondr77:
I think "Start new analysis" is more clear than "restart analysis"

Screen Shot 2022-10-12 at 10 01 27 AM

aarondr77:
I got this error and then selected the new file that I wanted to import. When I'm brought back to this screen, I think users might take the error message to mean that the new file still has that error. 

A few ideations:
1. Change the message to say something to the effect of: "Press the change imports button at the bottom to finalize"
2. Add text near the bottom of the taskpane that says "Use the change imports button to try replaying analysis on the new data"

I don't love either of those options, do you have any ideas? Otherwise, this is something we can certainly test in usability tests. 

Screen Shot 2022-10-12 at 10 07 58 AM

aarondr77:
To make this taskpane easier to navigate can we:
-  add a ? tooltip in the tittle that explains how this taskpane works: "Click on the imported data to swap it for a different dataset. Then press the Change Imports button at the bottom to replay the analysis on the new data"
- Add a tooltip to the TextButton at the bottom that explains what it does.

aarondr77:
When I get the error message in the csv configuration screen, I should not be able to press the `Update to skiplines.txt` button until I pass the import validation. 

https://user-images.githubusercontent.com/18709905/195367553-861322e8-8440-4f37-b0f1-c66d350857cd.mov

aarondr77:
This error message is confusing. Is it actually displaying two different error messages? Does it want me to select a delimiter and encoding or select a new file?

Screen Shot 2022-10-12 at 10 19 07 AM

aarondr77:
I bet a common way that users will try to resolve this replay analysis error is by finding the df1 definition in their notebook and recreating the variable. If you do that though, you still can't press the "Retry with Updated Imports" button because Mito still has the old state of the dataframe not existing. 

I think its fine -- users can either rerun the mitosheet.sheet() call or use the change import taskpane to reselect df1 

Screen Shot 2022-10-12 at 10 25 49 AM

aarondr77:
Can we add a little bit more space between these rows? Its a bit small and crammed together

Screen Shot 2022-10-12 at 10 56 15 AM

aarondr77:
Only the file name in the second row has a tooltip when you hover over it. Can we add it to all of them. 

Screen Shot 2022-10-12 at 10 58 24 AM

aarondr77:
Can we add a message here. If there is no data in the sheet it should tell them to go import data. If there is data in the sheet, but no imports, then maybe say you can change the contents of the dataframe that you're passing in

Screen Shot 2022-10-12 at 11 01 45 AM

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -53,10 +53,14 @@ def overwrite_import_data(analysis: Dict[str, Any], step_import_data_list_to_ove
             num_imports_in_step = len(get_import_data_with_single_import_list(step_type, params))
             for i in range(num_imports_in_step):
                 import_data_index = import_number + i
+                print(import_data_index, all_imports[import_data_index])
```
aarondr77, line 56: remove

mitosheet/src/components/taskpanes/UpdateImports/UpdateDataframeImportTaskpane.tsx
```diff
@@ -0,0 +1,119 @@
...
+                }
+            })}
+            loading={loading}
+        />
+    ) 
+
+    if (props.params === undefined) {
+        return (
+            <div className='text-body-1'>
+                There has been an error loading dataframes to import. Please try again, or contact support.
```
aarondr77, line 74: Want to make this `contact support` a hyperlink? You can use the Slack link in the codebase

aarondr77: Here: https://join.slack.com/t/trymito/shared_invite/zt-1h6t163v7-xLPudO7pjQNKccXz7h7GSg

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsTaskpane.tsx
```diff
@@ -152,14 +153,41 @@ const UpdateImportsTaskpane = (props: UpdateImportsTaskpaneProps): JSX.Element =
...
+                        }
+                    }
+
+                    // First, check that this is a valid import, and if it is, then update this
+                    const indexToErrorMap = await props.mitoAPI.getTestImports([{
+                        'step_id': 'fake_id',
+                        'imports': [dataframeCreationData]
+                    }])
+
+                    console.log("Checkling", indexToErrorMap)
```
aarondr77: remove

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsTaskpane.tsx
```diff
@@ -152,14 +153,41 @@ const UpdateImportsTaskpane = (props: UpdateImportsTaskpaneProps): JSX.Element =
...
+
+                    // if it's not a valid import, then we send the user to the CSV config screen
+                    if (indexToErrorMap === undefined || Object.keys(indexToErrorMap).length > 0) {
+                        console.log("Error", indexToErrorMap !== undefined ? indexToErrorMap[0] : undefined)
```
aarondr77: remove

mitosheet/src/components/elements/TextButton.tsx
```diff
@@ -62,6 +62,10 @@ interface TextButtonProps {
     */
     className?: string
 
+    /**
+        * @param [tooltip] -- Messaage to display as tooltip on button. Has lower priority than the tooltip
```
aarondr77, line 66: Is this supposed to say "has lower priority than the disabled tooltip"?

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsPreReplayTaskpane.tsx
```diff
@@ -63,10 +67,12 @@ const getErrorMessage = (
 
     let errorText: string | undefined = undefined;
 
-    if (numImportsInvalidPreReplayThatAreNotUpdated > 0 || numImportsInvalidPostReplay > 0) {
-        errorText = 'Please fix the failed imports below before updating imports';
-    } else if (invalidReplayError !== undefined) {
-        errorText = invalidReplayError;
+    if (!loadingImportDataAndErrors) {
+        if (numImportsInvalidPreReplayThatAreNotUpdated > 0 || numImportsInvalidPostReplay > 0) {
```
aarondr77: `numImportsInvalidPreReplayThatAreNotUpdated` would this just be `numImportsInvalidPreReplay`?
What is the note supported part?

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsPreReplayTaskpane.tsx
```diff
@@ -63,10 +67,12 @@ const getErrorMessage = (
...
+            errorText = 'Please fix the failed imports below before updating imports';
```
aarondr77: Maybe this can say "Please fix the errors below before updating imports". Otherwise, its a bit confusing with fixing imports before updating imports..

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsPreReplayTaskpane.tsx
```diff
@@ -114,63 +121,90 @@ const UpdateImportsPreReplayTaskpane = (props: UpdateImportPreReplayTaskpaneProp
...
-                    }}
-                    // TODO: move this to pre-replay as well!
-                    disabled={!allErrorsUpdated || invalidPostUpdate}
-                    disabledTooltip={(!allErrorsUpdated || invalidPostUpdate) ? "Please resolve all errors with above imports." : undefined}
-                >
-                    <p>
-                        Change Imports
-                    </p>
-                </TextButton>
+                            tooltip={"This will start a new, blank mitosheet."} // TODO: this text is awful
```
aarondr77, line 129: Ooo hmm this text isn't showing up for me...

aarondr77: "This will create a new, empty mitosheet"

aarondr77:
> Add the popup 

I'm down to wait on this

aarondr77:
> Make it so you can't click other buttons in the toolbar if an analysis fails to replay. I am not sure of a good way to do this -- what do you think? We could deactivate all these buttons, but that seems like a... ton. Any other ideas?

I think this is something we should fix. Currently, if you fail to replay the analysis, you can easily get into a super weird state if instead of resolving the analysis, you import new data (which isn't a super unreasonable thing for someone to do). 

Could we add something to the mito_widget that just stores its currently in an invalid replay analysis state. And then if the user tries to do anything except update the imports, we generate an error modal that says you got to update the imports or start a new analysis.

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsPreReplayTaskpane.tsx
naterush: No - this is the number of imports that were invalid before the replay occurred, that are still not updated!

aarondr77:
If I change an import and a column is missing for the filter step, it displays the correct column. 
If I change an import and the column is missing for a sort step, it just tells me "likely a column is missing"

mitosheet/mitosheet/mito_widget.py
```diff
@@ -152,12 +152,18 @@ def handle_update_event(self, event: Dict[str, Any]) -> None:
 
             # Update all state variables
             self.update_shared_state_variables()
-        except:
+        except Exception as e:
+            from mitosheet.errors import get_recent_traceback
+            print(get_recent_traceback())
```
naterush, line 157: Remove

mitosheet/mitosheet/mito_widget.py
```diff
@@ -152,12 +152,18 @@ def handle_update_event(self, event: Dict[str, Any]) -> None:
...
             # We handle the case of replaying the analysis specially, because we don't
             # want to display the error modal - we want to display something specific
             # in this case. Note that we include the updating of shared state variables
             # in the try catch, as this is sometimes where errors occur
             if event["type"] == REPLAY_ANALYSIS_UPDATE['event_type']:
+                # Propagate the Mito error if we can
```
naterush, line 163: Note that we never want this to go to the error modal!

mitosheet/mitosheet/tests/updates/test_update_existing_import.py
```diff
@@ -0,0 +1,327 @@
...
+            }]
+        }
+    ])
+
+    assert mito.dfs[0].equals(df)
+    assert mito.dfs[1].equals(df1)
+
+
+def test_test_import_returns_good_data():
+    df = pd.DataFrame(data={'A': [1, 2, 3], 'B': [2, 3, 4]})
```
naterush, line 249: TODO: Add tests for:
1. Multiple imports in the middle, still gets the correct indexes on the errors
2. Updating imports with multiple from the same step id, also works

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -6,27 +6,79 @@
...
+    If there is import data to overwrite, we go through and overwrite it. We notably have to do this
+    by keeping track of the index of the import, as we do not have step ids to link these things. 
+
+    As this import data is split so that only one dataframe is created during any specific import,
+    this means we make a bit more steps than we might want. But it's ok for now!
+    """
+    
+    # Flatten all the imports into a single list, so that we can index into them. We notably
+    # don't need the step ids for these
+    all_imports = [_import for import_data in step_import_data_list_to_overwrite for _import in import_data['imports']]
```
naterush, line 36: TODO: let's take this code, refactor into a utility, and use it anywhere we need to (e.g. anywhere we manually track and index).

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -41,7 +93,16 @@ def execute_replay_analysis_update(
...
-    steps_manager.execute_steps_data(new_steps_data=analysis['steps_data'])
+    # If there is import data to overwrite, we do that
+    if len(step_import_data_list_to_overwrite) > 0:
+        analysis = overwrite_import_data(analysis, step_import_data_list_to_overwrite)
+
+    try:
+        steps_manager.execute_steps_data(new_steps_data=analysis['steps_data'])
+    except:
+        from mitosheet.errors import get_recent_traceback
+        print(get_recent_traceback())
```
naterush, line 104: No

mitosheet/mitosheet/updates/update_existing_imports.py
```diff
@@ -0,0 +1,71 @@
...
+        steps_manager.execute_and_update_steps(new_steps, 0)
+    except Exception as e:
+        # We also make sure to catch any error, and turn it into an error that
+        # does not go directly to an error modal.
+        from mitosheet.errors import MitoError, make_invalid_update_imports_error
+        if isinstance(e, MitoError):
+            e.error_modal = False
+            raise e
+        
+        from mitosheet.errors import make_invalid_update_imports_error
```
naterush, line 61: Remove

mitosheet/src/components/Mito.tsx
```diff
@@ -204,25 +205,30 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
-                     * report this to the user by displaying a modal that tells them as much
+                     * If an analysis fails to replay, we open the update import pre replay 
+                     * taskpane with the error. There are two cases where the analysis failed
+                     * to replay:
+                     * 
+                     * 1. One of the import steps failed
+                     * 2. A step other than an import step failed
+                     * 
+                     * In the first case, we can give very specific error messages. In the case 
+                     * a different step has failed, we do our best to help the user!
```
naterush, line 221: Do better here!

mitosheet/src/components/taskpanes/FileImport/CSVImportConfigTaskpane.tsx
```diff
@@ -0,0 +1,72 @@
...
+            editApplied={editApplied}
+            loading={loading}
+            error={error !== undefined ? error : props.error}
+        
+            backCallback={() => {
+                props.setScreen({screen: 'file_browser'})
+            }}
+        />
+    )
+
```
naterush, line 67: Extra space

mitosheet/src/components/taskpanes/FileImport/XLSXImportConfigTaskpane.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+/* 
+    Allows a user to import an XLSX file with the given name, and
+    in turn allows them to configure how to import sheets in this
+    file.
+*/
+function XLSXImportConfigTaskpane(props: XLSXImportConfigTaskpaneProps): JSX.Element {
+
+
+    // NOTE: this loading state is just for getting the metadata about the sheets
```
naterush, line 41: Remove this comment

mitosheet/src/utils/actions.tsx
```diff
@@ -41,6 +41,9 @@ export const createActions = (
     const lastStepSummary = analysisData.stepSummaryList[analysisData.stepSummaryList.length - 1];
 
 
+    const disabledDueToReplayAnalysis = uiState.currOpenTaskpane.type === TaskpaneType.UPDATEIMPORTS && uiState.currOpenTaskpane.failedReplayData !== undefined;
```
naterush, line 44: Document this!

mitosheet/src/components/toolbar/PlanButton.tsx
```diff
@@ -7,17 +7,24 @@ import { TaskpaneType } from '../taskpanes/taskpanes';
...
     userProfile: UserProfile,
+    uiState: UIState;
     setUIState: React.Dispatch<React.SetStateAction<UIState>>,
     mitoAPI: MitoAPI
 }
 
 // The component in the toolbar that either tells you are pro or asks you to upgrade
 const PlanButton = (props: PlanButtonProps): JSX.Element => {
 
+    const disabledDueToReplayAnalysis = props.uiState.currOpenTaskpane.type === TaskpaneType.UPDATEIMPORTS && props.uiState.currOpenTaskpane.failedReplayData !== undefined;;
```
naterush, line 18: Double!

mitosheet/src/components/taskpanes/UpdateImports/updateImportsUtils.tsx
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export const updateDataframeCreation = (
+    dataframeCreationIndex: number,
+    dataframeCreationData: DataframeCreationData,
+    setUpdatedStepImportData: React.Dispatch<React.SetStateAction<StepImportData[] | undefined>>,
+    setUpdatedIndexes: React.Dispatch<React.SetStateAction<number[]>>,
+    setInvalidImportMessages: React.Dispatch<React.SetStateAction<Record<number, string | undefined>>>,
+    setReplacingDataframeState: React.Dispatch<React.SetStateAction<ReplacingDataframeState | undefined>>
+) => {
```
naterush, line 64: Return type

naterush (changes requested):
Cleanup!

aarondr77:
This error message still appears even if its data that is imported through the mitosheet.sheet() call. We could try to correct this by checking if the error occurs in a sheet index that was created through the args and throwing a more specific error for this case. But since we expect it to be pretty rare, I'm okay to leave this error as is for now, since its still pretty helpful in resolving! 

Screen Shot 2022-10-13 at 3 46 51 PM

aarondr77:
If you click "Start New Analysis" in the above taskpane, it still passes any args that were passed. In my mental model of the args and mito, starting a new analysis, should start completely blank

aarondr77:
- Import dataframe with column A using taskpane
- Create new column and reference A
- In creation of dataframe, change it from column A to column B
- Rerun code
- Notice that the error doesn't tell you what sheet the error occurred in. 

Screen Shot 2022-10-13 at 3 56 30 PM

Unlike a filter step that causes this error: 
Screen Shot 2022-10-13 at 3 57 25 PM

aarondr77:
If an analysis fails because it couldn't read the file, maybe the error message should say ".... update the import and specify the encoding and delimiter". The current error message made me think that I would get taken to the screen to set the encoding and delimiter right away. 

Screen Shot 2022-10-13 at 4 01 31 PM

aarondr77:
Can we add a "Change import" to the sheet tab menu?

naterush:
LGTM once tests pass!

PR #480, opened 2022-09-28
Conditional format usability fix

naterush:
# Description

- Adds better error messages
- Makes new filter conditions more clear (this was the best I could think of, feel free to change if you have something better).
- Makes error handling a bit more defensive to try and avoid crashes

# Testing

Test new conditional format error messages!

# Documentation

Before release, I have to write these docs!


aarondr77:
LGTM once tests pass

PR #481, opened 2022-09-28
mitosheet: cleanup per jakes usability test

naterush:
# Description

1. Fixes bug with some filter conditions leading to incorrectly conditional formats (and gives comment if this is the case in generated code). Feel free to rephrase comment.
2. Changes `filter` -> `condition` in the conditional format, as this is much more clear.

# Testing

# Documentation

### DIFF
--- a/mitosheet/mitosheet/code_chunks/postprocessing/set_dataframe_format_code_chunk.py
+++ b/mitosheet/mitosheet/code_chunks/postprocessing/set_dataframe_format_code_chunk.py
@@ -232,6 +232,22 @@ def get_conditional_format_code_list(state: State, sheet_index: int) -> Optional
 
     return None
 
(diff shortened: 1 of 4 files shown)


mitosheet/mitosheet/code_chunks/postprocessing/set_dataframe_format_code_chunk.py
```diff
@@ -251,6 +273,8 @@ def get_dataframe_format_code(state: State, sheet_index: int) -> Optional[str]:
     if conditional_format_code:
         format_code += conditional_format_code
 
+    # If the conditional format code has a .head call
```
naterush, line 276: Remove

naterush (commented):
cleanup!

PR #482, opened 2022-09-28
mitosheet: add different export options for graphs

naterush:
# Description

Per spec: https://www.notion.so/trymito/Download-graph-as-HTML-d43470a091e7465d98c8b705e0885364

Takes a slightly simpler approach of just appending the export code on the frontend, which is nice methinks!

Can you think of better text for these buttons?

# Testing

Test both export options.

# Documentation

Minor updates necessary.


aarondr77:
Ready to merge after release

PR #492, opened 2022-10-10
trymito: switch github for install docs

aarondr77:
# Description

Moves the install buttons back to our installation docs instead of GitHub. 

# Testing

# Documentation


naterush:
LGTM

PR #494, opened 2022-10-11
Get Support Button

aarondr77:
# Description

This PR: 
1. Builds infrastructure for interacting with the `mitosheet-helper-config` package 
2. Implements the first configuration feature -- the get support button. It adds a get support button to the toolbar and updates all of the existing get support buttons to use the configured email, if it exists.
3. Removes the OpenOnboardingChecklist button from the toolbar.

Following this spec: https://www.notion.so/trymito/Support-Button-04e88c3c065a491d82dfb7a7609933b3

# Testing

1. Before installing the `mitoshseet-helper-config` package, make sure the get support button and all other support buttons take you to the Mito Slack. 
4. Follow the instructions [here](https://github.com/mito-ds/mitosheet_helper_config) to configure the `mitosheet-helper-config` package and make sure the get support button updates. 

# Documentation

We should add this to our docs when we release!


mitosheet/mitosheet/mito_config.py
```diff
@@ -0,0 +1,62 @@
...
+
+# Copyright (c) Saga Inc.
+# Distributed under the terms of the GPL License.
+
+from typing import Any, Dict, Optional
+
+# These variables must match the variables defined in 
+# mito_config.py of the mitosheet_helper_config package
+MEC_VERSION = 'mec_version'
+SUPPORT_EMAIL = 'support_email'
```
naterush, line 12: Can we prefix both of these to `MEC_CONFIG_KEY_VERSION`, and `MEC_CONFIG_KEY_SUPPORT_EMAIL`, so that it's clear they are keys?

mitosheet/mitosheet/mito_config.py
```diff
@@ -0,0 +1,62 @@
...
+# }
+mec_upgrade_functions: Dict[int, Any] = {}	
+
+def upgrade_mito_enterprise_configuration(mec: Optional[Dict[str, Any]]) -> Optional[Dict[str, Any]]:
+    if mec is None:
+        return None
+
+    # So mypy tests recognize that mec is not None
+    _mec = mec
+    while _mec[MEC_VERSION] < len(mec_upgrade_functions) + 1:
```
naterush, line 32: `while _mec[MEC_VERSION] in mec_upgrade_functions:` actually checks the condition you use to index in below!

mitosheet/mitosheet/mito_config.py
```diff
@@ -0,0 +1,62 @@
...
+    used by the rest of the app. 
+
+    If the mito_helper_configuration package does not exist or does not set every configuration option, 
+    the MitoConfig class sets defaults. 
+    """
+    def __init__(self, mito_enterprise_configuration: Optional[Dict[str, Any]]):
+        self.mec = upgrade_mito_enterprise_configuration(mito_enterprise_configuration)
+
+    @property
+    def support_email(self) -> str:
```
naterush, line 50: I learned this week we use `@property` wrong. In general, you want to use a property if you can set it. 

We'll get around to updating these at some point, but for now let's just make this a function!

mitosheet/mitosheet/mito_config.py
```diff
@@ -0,0 +1,62 @@
...
+        if self.mec is None or self.mec[SUPPORT_EMAIL] is None:
+            return DEFAULT_SUPPORT_EMAIL
+        return self.mec[SUPPORT_EMAIL]
+
+    # Add new mito configuration options here ...
+
+    @property 
+    def mito_config(self) -> Dict[str, Any]:
```
naterush, line 58: Same here, should be a function

mitosheet/mitosheet/mito_config.py
```diff
@@ -0,0 +1,62 @@
...
+        return {
+            SUPPORT_EMAIL: self.support_email
```
naterush, line 60: Let's expose the version here as well

mitosheet/mitosheet/tests/test_ts_types.py
```diff
@@ -194,6 +195,13 @@ def test_graph_safety_filter_cutoff_matches():
 
     assert int(graph_safety_filter_cutoff) == GRAPH_SAFETY_FILTER_CUTOFF
 
+def test_user_profile_defaults_matches():
+    user_profile_support_email = get_constant_from_ts_file(
```
naterush, line 199: Can we make the keys of the Mito enterprise config a `MitoEnterpriseConfigKey` enum in the `types.ts` file, and update the tests accordingly? 

I want a test that will fail when we add something in on the frontend or something on the backend without updating the either!

mitosheet/src/components/elements/GetSupportButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+    If the mitoConfig support email is set to a custom email, it opens an email, otherwise, it opens Slack. 
+*/
+const GetSupportButton = (props: GetSupportButtonProps): JSX.Element => {
+
+    return (
+        <TextButton 
+            className={props.className}
+            variant='dark'
+            width={props.width || 'medium'}
+            href={props.userProfile.mitoConfig['support_email'] === DEFAULT_SUPPORT_EMAIL ? SLACK_INVITE_LINK : `mailto:${props.userProfile.mitoConfig['support_email']}`}
```
naterush, line 30: Can you use the new keys enum requested above here?

mitosheet/src/utils/actions.tsx
```diff
@@ -572,7 +574,12 @@ export const createActions = (
                 setEditorState(undefined);
 
                 // Open slack
-                window.open(SLACK_INVITE_LINK, '_blank')
+                if (userProfile.mitoConfig['support_email'] === DEFAULT_SUPPORT_EMAIL) {
```
naterush, line 577: Def move to enum everywhere. I want to be able to find and replace w/ 100% accuracy :-)

mitosheet/mitosheet/mito_widget.py
```diff
@@ -34,6 +35,11 @@
                                     UJ_RECEIVED_TOURS, UJ_USER_EMAIL)
 from mitosheet.user.utils import is_excel_import_enabled, is_pro, is_running_test
 
+try:
+    from mitosheet_helper_config import MITO_ENTERPRISE_CONFIGURATION 
+except ImportError:
+    MITO_ENTERPRISE_CONFIGURATION = None
```
naterush, line 41: Another potentially sensible option here is to have a default config ourself. 

FWIW, I think we should be _really thoughtful_ about what we add to this config. We're gonna get an explosion of code paths if we're casing on random stuff throughout the codebase, which is not gonna be fun to manage and test.

naterush (commented):
Can you add some tests? Specifically, I'd love to test the following three conditions:

1. If no helper package, works the same as current.
2. If helper package, gets the right variables. 
3. If helper package with an _invalid_ defined config, throws a good error!

This is user-facing code, in some ways, so let's get 3-5 good tests. 

Ping me when so, I'll do a final review, and then merge!

mitosheet/mitosheet/tests/test_ts_types.py
```diff
@@ -194,6 +195,22 @@ def test_graph_safety_filter_cutoff_matches():
...
     assert int(graph_safety_filter_cutoff) == GRAPH_SAFETY_FILTER_CUTOFF
 
+def test_mito_enterprise_keys_match():
+    mito_enterprise_config_keys = get_enum_from_ts_file("./src/types.tsx", "MitoEnterpriseConfigKey")
+    assert set(mito_enterprise_config_keys.values()) == set(
+        [
+            MEC_CONFIG_KEY_VERSION,
+            MEC_CONFIG_KEY_SUPPORT_EMAIL,
+        ]
+    )
```
naterush, line 205: Let's actually check this against an initialized object!

naterush (commented):
Tiny cleanup, I'll test.

naterush:
This looks great! Because it is optional, this is good to merge -- LGTM!

PR #501, opened 2022-10-21
Change imports usability test fixes

aarondr77:
# Description

- [x] Displays a success message when the an analysis is replayed on new data. This was the major area of confusion during usability test
- [x] When an original analysis has less than 1000 rows and the user graphs, we turn off the graph limit. But if the user then replays the analysis on data with more than 1000 rows, we want to not crash the browser. This updates the default safety filter to True instead of False. When there are less than 1000 rows there is no change to the graph. When there are more, the user is default safe to changing data sizes. 
- [x] Adds anonymous logging

# Testing

# Documentation


mitosheet/src/components/elements/BottomLeftPopup.tsx
```diff
@@ -35,31 +36,22 @@ const BottomLeftPopup = (props: {
...
-
     return (
         <>
             {displayLoadingIndicator && 
                 <div className='bottom-left-popup-container'>
                     <LoadingIndicator loading={props.loading}/>
                 </div>
             }
-            {!displayLoadingIndicator && displayChecklist &&
+            {!displayLoadingIndicator && props.ephemeralMessage &&
```
naterush, line 46: So we're not doing a checklist on onboarding anymore? I don't know if I think that's a great idea. 

We AB tested tour vs checklist, but never checklist v nothing - and IMO a tour checklist is kinda nice for getting people to do something!

mitosheet/src/components/EphemeralMessage.tsx
```diff
@@ -0,0 +1,41 @@
...
+                    ...prevUIState,
+                    ephemeralMessage: undefined
+                }
+            })
+        }, 5000)
+        return () => {clearInterval(interval)};
+    }, [props.message])
+
+    return (
+        <p className={classNames('ephemeral-message-container', 'text-body-1')}>
```
naterush, line 35: There should be some way to close this before the 5 seconds are up. I think this is infrastructure that should be handled by the popup container!

naterush (changes requested):
I think we def should have a success message -- but can we treat this like the rest of our taskpanes and put this in the taskpane?

I didn't see the popup at first when I used it since it was pretty far away (and small) from where I clicked the button. 

A little popup above the button is a) inline with what we do elsewhere, and b) probably easier to see, while c) requiring less code to implement. 

If you move to this (I vote yes), can you make sure to undo all these changes?

naterush:
The other thing to do from these usability tests: can you check logging and compare them to the videos, and see how we can make the logs better? Thanks!

mitosheet/src/components/elements/BottomLeftPopup.tsx
aarondr77: We might do it at some point. But we are not currently doing it

aarondr77:
The popup infrastructure is maybe a bit overkill for just this one popup, but its the infrastructure that we previously discussed wanting to have for future popup functionality.

mitosheet/src/components/taskpanes/Graph/GraphSetupTab.tsx
```diff
@@ -37,7 +37,7 @@ export const GRAPH_SAFETY_FILTER_CUTOFF = 1000;
 
 // Tooltips used to explain the Safety filter toggle
 const SAFETY_FILTER_DISABLED_MESSAGE = `Because you’re graphing less than ${GRAPH_SAFETY_FILTER_CUTOFF} rows of data, you can safely graph your data without applying a filter first.`
-const SAFETY_FILTER_ENABLED_MESSAGE = `Turning on Filter to Safe Size only graphs the first ${GRAPH_SAFETY_FILTER_CUTOFF} rows of your dataframe, ensuring that your browser tab won’t crash. Turning off Filter to Safe Size graphs the entire dataframe and may slow or crash your browser tab.`
+const SAFETY_FILTER_ENABLED_MESSAGE = `Turning on Limit ${GRAPH_SAFETY_FILTER_CUTOFF} only graphs the first ${GRAPH_SAFETY_FILTER_CUTOFF} rows of your dataframe, ensuring that your browser tab won’t crash. Turning it off graphs the entire dataframe and may slow or crash your browser tab.`
```
naterush, line 40: Think this is missing `rows` after it

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsPostReplayTaskpane.tsx
```diff
@@ -117,8 +117,14 @@ const UpdateImportsPostReplayTaskpane = (props: UpdateImportPostReplayTaskpanePr
                             if (_invalidImportIndexes === undefined) {
                                 return;
                             }
+
                             props.setInvalidImportMessages(_invalidImportIndexes);
 
+                            void props.mitoAPI.log('get_test_import_results', {
+                                'open_due_to_replay_error': false,
```
naterush, line 124: I think a more clear log param is `update_imports_pre_replay` and then it's true or false

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsPostReplayTaskpane.tsx
```diff
@@ -117,8 +117,14 @@ const UpdateImportsPostReplayTaskpane = (props: UpdateImportPostReplayTaskpanePr
                             if (_invalidImportIndexes === undefined) {
                                 return;
                             }
+
                             props.setInvalidImportMessages(_invalidImportIndexes);
 
+                            void props.mitoAPI.log('get_test_import_results', {
```
naterush, line 123: Do we want a `clicked_update_from_post_replay_update_imports` log here as well?

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsTaskpane.tsx
```diff
@@ -97,6 +97,12 @@ const UpdateImportsTaskpane = (props: UpdateImportsTaskpaneProps): JSX.Element =
                 invalidImportIndexes = {};
             }
 
+            void props.mitoAPI.log('opened_update_imports_taskpane', {
+                'open_due_to_replay_error': updatePreReplay,
+                'num_invalid_imports': invalidImportIndexes === undefined ? 0 : Object.keys(invalidImportIndexes).length,
+                'num_total_imports': importData === undefined ? 0 : importData.length
+            });
```
naterush, line 104: Can we move this to an
```
useEffect(() => {
    /// TODO: logging here
})
```

Just to separate it from the logic for cleanliness.

naterush (changes requested):
A few minor thoughts on logging, and a minor placement thing below, and then LGTM!

Screen Shot 2022-10-27 at 3 09 47 PM

This placement is wacky as is - feels like it's floating in the middle. I actually think it would look best at bottom right -- want to move it there? Just mirrored from the bottom left.

Also, thoughts on moving the X to the other side? When I first saw it, I thought it was a bullet point - I think X's are almost always on the right side, even if the the popup is on the right side as well.

naterush:
Can you also resolve merge conflicts so that tests run?

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsPostReplayTaskpane.tsx
aarondr77: I think `open_due_to_replay_error` because you can understand what its telling you even if you don't have the context that `update_imports_pre_replay` only occurs when there is an error. 

If you have the context, then both tell you the same thing. If you don't have the context, then `open_due_to_replay_error` gives it to you!

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsPostReplayTaskpane.tsx
aarondr77: Sure

aarondr77:
Okay, I left the placement of the ephemeral message where it is. In my mind this is a notification and very naturally goes in the top right hand corner. I tried moving it up a little bit, but over top of the column headers looks messy. 

To make it more like a notification, I added some animation so it now slides in and out smoothly -- I think its pretty nice! 

Feel free to move it somewhere else if you feel strongly though.

PR #504, opened 2022-10-24
Change imports usability fix 2

aarondr77:
# Description

# Testing

# Documentation


aarondr77:
Leaving the taskpane opens up a can of worms:
1. Since the step_ids change, we need to refresh the step import data to get the correct step ids in the `UpdateImportsPostReplayAnalysis`.
2. In the `UpdateImportsPostReplayAnalysis` we have now successfully replayed the analysis, we technically want to be in the post replay analysis state. 

This all gets simplified if we just close the taskpane.

naterush:
Gotcha. This is good complexity that I didn't realize existed that make these proposed changes much harder than they should be!

I think moving this back to the popup makes sense - you should just be able to configure where the popup goes easily perhaps?

For my knowledge, did you know about this complexity on the previous PR? Or is this something that you discovered in trying to make this change.

aarondr77:
I knew about the `UpdateImportsPostReplayAnalysis` one, but then discovered those other two as I did this PR.

aarondr77:
Closed in favor of https://github.com/mito-ds/monorepo/pull/501/files

PR #505, opened 2022-10-25
Sharing analyses

naterush:
# Description

This PR implements this specification: https://www.notion.so/trymito/Sharing-Analyses-Specification-5acb830b2345454da6de8e71e15d25fd

major differences:
1. We store the author hash rather than the static_user id
2. Write the cell metadata rather than notebook metadata
3. Allow users to expand the steps to see what they look like before trusting them

# Testing

There are a few things to test:
1. Create an analysis, delete it off your computer in ~/.mito/saved_analyses, it should still replay
2. Create an analysis, delete the cell below, it should still replay
3. Create an analysis, delete from both locations, it should still display an error
4. Create an analysis, change your static_user_id in user.json, and it should ask you to trust it before replaying (make sure to refresh kernel)
5. Create multiple analyses in notebook
6. Create analyses in Lab, move to Notebook, and vice versa
7. Create analyses, change user.json and move one of the imported files. Should first ask you to trust, then open the PreReplayUpdateImports taskpane (and this should work). 

Any more tests you can think of as well!

# Documentation


mitosheet/mitosheet/saved_analyses/save_utils.py
```diff
@@ -123,14 +125,45 @@ def rename_saved_analysis(old_analysis_name, new_analysis_name):
...
-        }
+        f.write(saved_analysis_json)
+
+static_user_id = None
+
+
+TEST_AUTHOR_HASH = 'test-hash'
+def get_author_hash(steps_data:  List[Dict[str, Any]]) -> str:
+    """
+    For security reasons, we store the hash... to explain this
```
naterush, line 142: TODO: fill this out!

mitosheet/mitosheet/tests/saved_analyses/test_save_utils.py
```diff
@@ -333,4 +333,18 @@ def test_save_replays_overwrite_by_ids_propererly():
 
     assert new_mito.dfs[1].equals(
         pd.DataFrame({'A': [1, 2, 3], 'A count': [1, 1, 1]})
+    )
+
+def test_replay_analysis_pass_analysis_upgrades_analysis():
```
naterush, line 338: Add another test here?

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -85,8 +87,16 @@ def execute_replay_analysis_update(
     if analysis_name == steps_manager.analysis_name:
         return
 
-    # If we're getting an event telling us to update, we read in the steps from the file
-    analysis = read_and_upgrade_analysis(analysis_name)
+    if analysis is None:
+        print("READING IN", analysis_name)
```
naterush, line 91: Remove

naterush: Add some documentation at the top of this file!

mitosheet/src/components/Mito.tsx
```diff
@@ -184,9 +184,26 @@ export const Mito = (props: MitoProps): JSX.Element => {
             if (analysisData.analysisToReplay) {
                 const analysisToReplayName = analysisData.analysisToReplay?.analysisName;
 
-                // First, if the analysis to replay does not exist at all, we just open an error modal
-                // and tell users that this does not exist on their computer
-                if (!analysisData.analysisToReplay.existsOnDisk) {
+                /**
+                 * We look for this analysis in two locations: in the notebook metadata, and on disk
```
naterush, line 188: TODO: clean this up!

mitosheet/src/components/Mito.tsx
```diff
@@ -225,7 +242,7 @@ export const Mito = (props: MitoProps): JSX.Element => {
                                 type: TaskpaneType.UPDATEIMPORTS,
                                 failedReplayData: {
                                     analysisName: analysisToReplayName,
```
naterush, line 244: We need to pass through the analysis to the failedReplayData!

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsPreReplayTaskpane.tsx
```diff
@@ -155,7 +155,7 @@ const UpdateImportsPreReplayTaskpane = (props: UpdateImportPreReplayTaskpaneProp
 
                                         props.setInvalidReplayError(undefined) // Clear the error
 
-                                        const replayAnalysisError = await props.mitoAPI.updateReplayAnalysis(props.failedReplayData.analysisName, props.updatedStepImportData);
+                                        const replayAnalysisError = await props.mitoAPI.updateReplayAnalysis(props.failedReplayData.analysisName, undefined, props.updatedStepImportData);
```
naterush, line 158: This changes!

mitosheet/src/jupyter/jupyterUtils.tsx
```diff
@@ -96,4 +96,36 @@ export const getArgs = (analysisToReplayName: string | undefined): Promise<strin
         }
         return resolve([]);
     })
+}
+
+export const writeToNotebookMetadata = (key: string, value: string): void => {
+    console.log("Writing", value, "to", key)
```
naterush, line 102: Cleanup and document these

mitosheet/src/plugin.tsx
```diff
@@ -258,6 +258,32 @@ function activateWidgetExtension(
         }
     });
 
+    app.commands.addCommand('set-metadata', {
```
naterush, line 261: Cleanup and document

mitosheet/src/jupyter/notebook/pluginUtils.tsx
```diff
@@ -268,4 +268,26 @@ export const writeEmptyMitosheetCell = (): void => {
         writeToCell(activeCell, 'import mitosheet\nmitosheet.sheet()');
         (window as any).Jupyter?.notebook?.execute_cell_and_insert_below();
     }
+}
+
+export const setMetadataInNotebook = (key: string, value: string): void => {
```
naterush, line 273: Cleanup and document

mitosheet/src/types.tsx
```diff
@@ -645,6 +645,8 @@ interface Experiment {
  * @param lastResult - This is the result of the last step that was applied. This might be undefined if the 
  *        step does not return a result
  * @param experiment - The experiment that this user is currently running, which may not be defined
+ * @param savedAnalysisJSON - The string that is saved in the analysis file that is stored on disk and in this
```
naterush, line 648: Clean this up

naterush (commented):
Minor cleanup. Nothing huge. Then do some testing!

mitosheet/mitosheet/Untitled.ipynb
```diff
@@ -0,0 +1,24 @@
+{
```
naterush, line 1: remove

mitosheet/mitosheet/errors.py
```diff
@@ -532,6 +532,19 @@ def make_dataframe_not_found_error(df_name: str) -> MitoError:
         f'{df_name} is not defined.'
     )
 
+def make_replay_analysis_permissions_error() -> MitoError:
```
naterush, line 535: Assert this is a constant, so it works on the frontend

naterush: Aka add tests

mitosheet/mitosheet/mito_widget.py
```diff
@@ -94,6 +94,7 @@ def update_shared_state_variables(self) -> None:
         self.analysis_data_json = self.steps_manager.analysis_data_json
         self.user_profile_json = json.dumps({
             # Dynamic, update each time
+            'static_user_id': get_user_field(UJ_STATIC_USER_ID),
```
naterush, line 97: No longer needed!

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -59,19 +66,24 @@ def overwrite_import_data(analysis: Dict[str, Any], step_import_data_list_to_ove
...
         analysis_name: str,
-        step_import_data_list_to_overwrite: List[Dict[str, Any]]
+        analysis: Optional[Dict[str, str]],
+        step_import_data_list_to_overwrite: List[Dict[str, Any]],
+        ignore_author_hash: Optional[bool]
     ) -> None:
     """
-    This function reapplies the analysis saved at analysis_name. Notably, if
-    step_import_data_list_to_overwrite are passed, this it overwrites all the
+    This function either:
```
naterush, line 81: TODO: document much better!

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -85,8 +97,23 @@ def execute_replay_analysis_update(
...
+        # go and get these steps from the disk
+        analysis = read_and_upgrade_analysis(analysis_name)
+    else:
+        # Otherwise, if there is a saved analysis provided from the frontend,
+        # we verify it's author_hash to make sure to the user is replaying
+        # their own edits. If not, we throw an error
+        if get_author_hash(analysis['steps_data'], get_user_field(UJ_STATIC_USER_ID)) != analysis['author_hash']:
+            if ignore_author_hash != True:
+                raise make_replay_analysis_permissions_error()
+        print("Checked author hash!")
```
naterush, line 111: Remove

mitosheet/src/components/Mito.tsx
```diff
@@ -184,9 +185,18 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
+                 * We look for this analysis in two locations: in the notebook metadata, and on disk
+                 * as well. When replaying, on the backend, if there is an analysisSavedInNotebook, then
+                 * this will be used to replay. Otherwise, the analysis on disk will be used
+                 */
+
+                const analysisSavedInNotebookJSON = await getNotebookMetadata(analysisToReplayName, `saved_analyses:${analysisToReplayName}`);
+                const analysisSavedInNotebook: SavedAnalysis | undefined = analysisSavedInNotebookJSON !== undefined ? JSON.parse(analysisSavedInNotebookJSON) : undefined;
+                
+                // If there is no analysis anywhere, then tell the user that
+                // TODO: update this error message
```
naterush, line 198: TODO: cleanup all of this

mitosheet/src/components/Mito.tsx
```diff
@@ -255,10 +281,19 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
          * means that we have read in and replayed the updated analysis, etc. 
+         * 
+         * We also write the analysis to the metadata of the notebook in this case.
          */
         if (analysisData.renderCount >= 1) {
             // Finially, we can go and write the code!
             writeGeneratedCodeToCell(analysisData.analysisName, analysisData.code, userProfile.telemetryEnabled);
+            
+            // And we write the analysis to the metadata
+            writeToNotebookMetadata(
```
naterush, line 292: Rename and cleanup functions

mitosheet/src/components/modals/ReplayAnalysisPermissionsModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+import Row from '../layout/Row';
+import Col from '../layout/Col';
+import { isMitoError } from '../../utils/errors';
+import { TaskpaneType } from '../taskpanes/taskpanes';
+
+
+/**
+ * This function returns a summary view of the 
+ */
+const getAnalysisSummary = (analysis: SavedAnalysis): JSX.Element => {
```
naterush, line 19: TODO: style this, look at mockups!

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsPreReplayTaskpane.tsx
```diff
@@ -155,7 +155,7 @@ const UpdateImportsPreReplayTaskpane = (props: UpdateImportPreReplayTaskpaneProp
 
                                         props.setInvalidReplayError(undefined) // Clear the error
 
-                                        const replayAnalysisError = await props.mitoAPI.updateReplayAnalysis(props.failedReplayData.analysisName, props.updatedStepImportData);
+                                        const replayAnalysisError = await props.mitoAPI.updateReplayAnalysis(props.failedReplayData.analysisName, props.failedReplayData.analysis, props.updatedStepImportData);
```
naterush, line 158: TODO: make sure that we update the hash before passing to this?

naterush: Or pass True to the ignore?

mitosheet/src/jupyter/jupyterUtils.tsx
```diff
@@ -96,4 +96,35 @@ export const getArgs = (analysisToReplayName: string | undefined): Promise<strin
         }
         return resolve([]);
     })
+}
+
+export const writeToNotebookMetadata = (analysisName: string, key: string, value: string): void => {
```
naterush, line 101: rename functions

mitosheet/src/jupyter/lab/pluginUtils.tsx
```diff
@@ -90,6 +90,27 @@ export function getCellCallingMitoshetWithAnalysis(tracker: INotebookTracker, an
...
+
+    const notebook = tracker.currentWidget?.content;
+    const cells = notebook?.model?.cells;
+
+    if (notebook === undefined || cells === undefined) {
+        return;
+    }
+
+    const codeCell = getCellAtIndex(cells, mitosheetCallIndex + 1);
+    console.log("code cell", codeCell)
```
naterush, line 110: Remove console logs here

mitosheet/src/plugin.tsx
```diff
@@ -258,6 +266,35 @@ function activateWidgetExtension(
...
+        label: 'Get Metadata',
+        execute: (args: any): string | undefined => {
+            const key = args.key as string;
+            const analysisName = args.analysisName as string;
+
+            // TODO: check that this cast is legit
+            const codeCellData = getCellWithCodeForAnalysis(tracker, analysisName);
+            console.log("returned", codeCellData)
+            const currentMetadata = codeCellData?.metadata.get('mitosheet') as Record<string, string>;
+            console.log('current metadata', currentMetadata)
```
naterush, line 293: Remove console logs

mitosheet/src/types.tsx
```diff
@@ -677,6 +684,7 @@ export interface AnalysisData {
  * @param usageTriggeredFeedbackID - the id of the usage triggered feedback id to display to the user
  */
 export interface UserProfile {
+    staticUserID: string,
```
naterush, line 687: Remove

mitosheet/src/utils/code.tsx
```diff
@@ -29,13 +29,17 @@ export function getCodeString(
         }
     }
 
+    const notEditableComment = '# This cell is not editable. To edit it, copy it to a new cell\n';
```
naterush, line 32: Clean this up

mitosheet/src/jupyter/notebook/pluginUtils.tsx
```diff
@@ -250,12 +261,13 @@ export const notebookWriteGeneratedCodeToCell = (analysisName: string, codeLines
...
         const activeCell = (window as any).Jupyter?.notebook?.get_cell((window as any).Jupyter?.notebook?.get_anchor_index());
         writeToCell(activeCell, code);
+        writtenCell = activeCell;
     }
-}
-
-
-
 
+    if (writtenCell !== undefined) {
```
naterush, line 267: Comment

mitosheet/src/components/modals/ReplayAnalysisPermissionsModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+                    </Row>
+                )
+            })}   
+        </>
+    )
+}
+
+
+/*
+    This modal displays to the user when:
```
naterush, line 41: Update

naterush (commented):
Cleanup!

naterush:
**NOTE: do not merge this until we get the current things deployed. We have enough on our plate without adding this one. We'll do a deploy next Monday, and merge this after that!**

aarondr77:
When testing 3) "Create an analysis, delete from both locations, it should still display an error", I noticed that if I click the Get Support button, and go back to Mito, the error message is gone. It starts me at a blank analysis, but with the same analysis id. We should probably require that users end up with a new analysis id.

aarondr77:
3 updates here:
1. Can you add a space between the end of the sentence and "View all steps". 
2. Can you clean up the scrollbars so that I can see the text that I'm scrolling to view
3. Can we display more intelligible steps than the raw json. Maybe use comments we generate for each step? I think as is, people aren't going to make sense of what they are looking at and we aren't doing an effective job of protecting them.

aarondr77:
Context: I have dataframes passed into the mitosheet.sheet() call and I'm prompted to start a new analysis or trust the analysis

When I press start new analysis, it resets the analysis id, but doesn't get rid of the passed dataframes. That is not what I expect because it wouldn't keep other imported datasets and it is not a new, blank mitosheet.

aarondr77:
In Notebooks, I'm not able to delete the generated code cells. 

https://user-images.githubusercontent.com/18709905/198912231-3054d83c-e913-4d01-817c-0a9c2e9f99b8.mov

aarondr77:
Whenever a mito generated code cell is blank, we should unlock it. Having random blank code cells that end up in my notebook because I delete some mitosheets is really confusing.

aarondr77:
If the user gets the Trust Modal and they ignore it, by for example, opening a taskpane, the modal closes and they update the sheet, but we don't generate additional code. Users shouldn't be able to avoid the Trust modal.

aarondr77:
Context: Create analyses, change user.json and move one of the imported files. Call mitosheet.sheet(), press the trust analysis button. 

Before updating the import, rerun the mitosheet.sheet() call and notice that it prompts you to retrust the analysis. Only until I correctly update the imports does it stop asking me to trust the analysis. 

I expect that once I press the Trust button, it will not prompt me to trust the analysis until the next time someone else edits it.

aarondr77:
1. ```
import pandas as pd
df = pd.DataFrame({'A': [1,2,3]}).to_csv('test.csv')
```
2. Create a new mitosheet and import test
3. Delete test.csv
4. Rerun the mitosheet call and update the import in the PreReplayUpdateImportTaskpane to forestfires.csv
5. Launch Jupyter Notebook
6. Rerun the mitosheets.sheet() call and notice that the PreReplayUpdateImportTaskpane opens with mixed messaging:
<img width="993" alt="Screen Shot 2022-10-30 at 9 35 03 PM" src="https://user-images.githubusercontent.com/18709905/198913941-10aa4ff9-8f6a-437e-99c7-269ccf1405fe.png">

**Comment by aarondr77** [comment]
Btw: Jupyter has a notion of Trusting the notebook too.. 

Screen Shot 2022-10-30 at 9 50 14 PM

**Comment by aarondr77** [comment]
1. Create new mitosheet
2. Create a mitosheet and import data
3. Copy the analysis id, but don't save the notebook.
4. Close the notebook and reopen it
5. Notice that its a blank notebook
6. Create a new mitosheet.sheet() call and pass in the analaysis id
7. Notice that the analysis replays.

File: mitosheet/src/components/modals/ReplayAnalysisPermissionsModal.tsx [hunk 27]
```diff
@@ -0,0 +1,149 @@
...
+import { ModalEnum } from './modals';
+import { overwriteAnalysisToReplayToMitosheetCall } from '../../jupyter/jupyterUtils';
+import { isMitoError } from '../../utils/errors';
+import { TaskpaneType } from '../taskpanes/taskpanes';
+
+
+/**
+ * This function returns a summary view of the analysis. Note that we don't do anything
+ * super fancy here, as this will likely be a very rare flow as users will most often
+ * trust where they are getting their notebooks from.
```
aarondr77, line 17: I think its worth doing the cleanup that I proposed. I think polish and around security practices is really important for user trust.

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsTaskpane.tsx
```diff
@@ -53,7 +53,9 @@ export interface ReplacingDataframeState {
 
 export interface FailedReplayData {
     analysisName: string,
-    error: MitoError
+    analysis: SavedAnalysis | undefined,
+    error: MitoError,
+    ignoreAuthorHash?: boolean
```
aarondr77, line 58: I thought that this was only going to be used for testing. Why would we want to ignore the author hash?

mitosheet/src/jupyter/api.tsx
```diff
@@ -1279,15 +1279,19 @@ export default class MitoAPI {
...
 
         const result: MitoError | undefined = await this.send({
             'event': 'update_event',
             'type': 'replay_analysis_update',
             'params': {
                 'analysis_name': analysisName,
-                'step_import_data_list_to_overwrite': stepImportDataListToOverwrite === undefined ? [] : stepImportDataListToOverwrite
+                'analysis': analysis === undefined ? null : analysis,
+                'step_import_data_list_to_overwrite': stepImportDataListToOverwrite === undefined ? [] : stepImportDataListToOverwrite,
+                'ignore_author_hash': ignoreAuthorHash === undefined ? false : true
```
aarondr77, line 1294: Hmm instead of having this one function do so many different things with different configurations, I think it might be better if we had the user press `Trust Analysis`, then we called an update function like `trust_analysis_update` which updated the author hash. Then if the analysis failed to update because of the imports / invalid steps, then we open up the PreReplayUpdateImportTaskpane which handles things as we normally do. 

Wrapping the security model into these different cases feels more fragile to me

mitosheet/src/jupyter/jupyterUtils.tsx
```diff
@@ -96,4 +96,34 @@ export const getArgs = (analysisToReplayName: string | undefined): Promise<strin
...
+            analysisName: analysisName
+        });
+    } else if (isInJupyterNotebook()) {
+        setGeneratedCodeCellMetadataInNotebook(analysisName, key, value);
+    } else {
+        console.error("Not detected as in Jupyter Notebook or JupyterLab")
+    }
+}
+
+export const getNotebookMetadata = (analysisName: string, key: string): Promise<void | string> => {
```
aarondr77, line 115: Maybe we could be more specific that this isn't just getting the notebook metadata, but is getting an analysis from the notebook metatadata: `getAnalysisFromNotebookMetadata`

mitosheet/src/jupyter/jupyterUtils.tsx
```diff
@@ -96,4 +96,34 @@ export const getArgs = (analysisToReplayName: string | undefined): Promise<strin
         }
         return resolve([]);
     })
+}
+
+export const writeToNotebookMetadata = (analysisName: string, key: string, value: string): void => {
```
aarondr77, line 101: I'm not sure based on this function api what I'm supposed to pass to it. Can we:
1. change the name of the function to `writeAnalysisToNotebookMetadata` 
2. update the argument names of `key` and `value` to whatever these things actually are for our use case of writing the analysis to the metadata?

aarondr77:
1. Can we move the comment to the top of the code cell? (I def vote yes on this)
2. Maybe change the text to: "This code cell is generated by Mito and is not editable. Copy the code to a new cell to edit it" (don't care as much, but it might be nice to say its created by Mito!)

Screen Shot 2022-10-30 at 11 24 17 PM

aarondr77:
I think its fine since the security model stays the same, but just calling out then when a you use the Jupyter functionality to copy and then paste the cells, the metadata gets copied too (and the cell stays locked)

Screen Shot 2022-10-30 at 11 30 50 PM

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsTaskpane.tsx
naterush: I'll rename this to be more clear. This should be called `explicitlyTrustAnalysisByIgnoringAuthorHash`

mitosheet/src/jupyter/jupyterUtils.tsx
naterush: A more clear option probably is renaming to `writeMetadataToGeneratedCodeCell`

mitosheet/src/jupyter/api.tsx
naterush: This was the first approach I took. The problem with this approach is that it leads to an extra roundtrip anytime you're replaying anything, which feels slow. There's also a lot of extra code you have to add that was pretty much just replicating what already exists. 

Let's see how this evolves. As of now, the backend has a single if statement, so it's not too messy currently. But I agree this is something to keep an eye on!

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -85,8 +108,22 @@ def execute_replay_analysis_update(
...
+    if analysis is None:
+        # If we aren't provided steps from the frontend, then we have to 
+        # go and get these steps from the disk
+        analysis = read_and_upgrade_analysis(analysis_name)
+    else:
+        # Otherwise, if there is a saved analysis provided from the frontend,
+        # we verify it's author_hash to make sure to the user is replaying
+        # their own edits. If not, we throw an error
+        if get_author_hash(analysis['steps_data'], get_user_field(UJ_STATIC_USER_ID)) != analysis['author_hash']:
+            if ignore_author_hash != True:
```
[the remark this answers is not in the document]
naterush: so much better

mitosheet/mitosheet/saved_analyses/save_utils.py
```diff
@@ -123,14 +125,56 @@ def rename_saved_analysis(old_analysis_name, new_analysis_name):
...
+        import hashlib
+        return hashlib.sha256(f'{json.dumps(steps_data)}-{user_id}'.encode('utf-8')).hexdigest()
+
+def create_saved_analysis_json(steps_manager: StepsManagerType, version: str=__version__) -> str:
+    
+    steps_data = make_steps_json_obj(steps_manager.steps_including_skipped)
+
+    saved_analysis = {
+        'version': version,
+        'author_hash': get_author_hash(steps_data),
```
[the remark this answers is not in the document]
naterush: I think it's ok, given that we'll likely change what we hash over time!

mitosheet/mitosheet/updates/replay_analysis.py
```diff
@@ -59,19 +66,35 @@ def overwrite_import_data(analysis: Dict[str, Any], step_import_data_list_to_ove
...
+        step_import_data_list_to_overwrite: List[Dict[str, Any]],
+        ignore_author_hash: Optional[bool]
     ) -> None:
     """
-    This function reapplies the analysis saved at analysis_name. Notably, if
-    step_import_data_list_to_overwrite are passed, this it overwrites all the
+    This update function replays an analysis. There are two ways to replay
+    an analysis
+
+    1. By providing an analysis_name, but no analysis. In this case, the analysis
```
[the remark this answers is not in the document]
naterush: Ok, so one option here is to just not save data on disk. And just gradually move everything over time into notebook. I think... let's wait. 

Let's wait and see how this roll out goes - it's a nice gradual transition for now. If the notebook stuff works well, we can stick with that. If not, then we can do something different.

naterush:
# On Notebook cells not being editable

We need to make a call on what we want to do here. There are only a few places to store metadata in the notebook, and there are only two locations that are "visible" to the user: on cells, and on outputs.

If we just keep the the mitosheet.sheet call cell and the generated code cell, we have two inputs and two outputs. We can't put the metadata in the output really of either, because if the user clears outputs (very standard), then we lose the analysis. So we need to put in the metadata of the input cell. 

If we put it in the input metadata of the mitosheet call, then we have the same "the analysis is invisible" issue. If we put it in the metadata of the generate code cell, then it's visible that there is private data here, unless the user clears the cell. We can only stop them from clearing the cell by making it not editable, but that's how we got to the current issue of it not being deletable in notebooks.

One option is to add a new cell that just stores analyses - potentially not in metadata, just in a comment or something - but this sounds really gross and ugly and like not a great solution.

In other words, I am not sure we have any good options. Let's def call and ideate + make a call on this one.

# On the View All Steps

I cleaned this up lots, and it looks pretty pleasant currently. Let me know if you want more styling here, happy to oblige. I don't think JSON is a problem as is though!

aarondr77:
1. Open file A.ipynb open in Jupyter Lab and Jupyter Notebooks at the same time
2. Edit the file in Jupyter Notebook
3. Save the file
4. Go to Jupyter Lab and refresh the browser
5. Rerun the mitosheet.sheet() call and notice that the file updates with your most recent changes
6. Edit the file in Jupyter Lab
7. Save the file
8. Go to Jupyter Notebook and refresh the browser
9. Rerun the mitosheet.sheet() call and notice that the file **does not** update with your most recent changes

aarondr77:
Cool, so a couple interesting things:
1. You can have a file open in both Jupyter Lab and Jupyter Notebook and switch off between which one you edit the file in so long as you save the file before switching and refresh the browser before starting to edit. 
2. When you download a .ipynb file, it downloads starting from the last saved version of the notebook so unless you save your work, the metadata you created reverts back to the last checkpoint.

aarondr77:
Can we add some logging to this, so we can see things like:
1. How often are users working on analysis that are shared
2. How often users trust the analysis
3. How often users start the analysis over

aarondr77:
A scenario that we haven't really accounted for is what happens if users are on different version of Mito. For example:
- Alice is on version 1, Bob is on version 2
- In version 2 we changed a param from columnID to columnIDs since we want to support bulk dtype changes.
- Alice creates an analysis and sends it to Bob
- Bob's mitosheet updates the analysis and everything works fine
- Bob sends it back to Alice 
- Alice reruns the analysis and now it is broken. 

A couple of ideas:
1. Maybe let's not worry about it and see how popular this is to begin with. In enterprises where we expect sharing to occur more commonly, users will all be on the same version
2. Question: If an analysis has an additional parameter that is not used, this shouldn't cause an issue right? Cause we just won't read that param. This handles most step upgrade scenarios. 
3. Since the analysis stores the mitosheet version, if it errors, we could check the version of the analysis and the user's version. If they are different, we can give the user an error that explains the problem and tells them they can either 1) upgrade, 2) start a new analysis 3) read the old version of the analysis of their disk

aarondr77:
Note: Read this all the way through before trying to follow along

A use case that we want to support so we can help users and colleagues can help eachother: 

- You have an error in your analysis 
- You send it to me to help resolve it
- I have to update the file path so I can import the file
- I fix the error and send you back the file
- You run the analysis and get the replay error because the file path changed, but the error should be corrected. 

See testing steps here: 
1. Create this file `pd.DataFrame({'A': [1,2,3]}).to_csv('test.csv')`
2. Import it 
3. Add a column and reference A
4. Run code `pd.DataFrame({'B': [1,2,3]}).to_csv('test.csv')`
5. Rerun mito and see that you get an error
6. Change the static-user-id & restart your kernel, but write down the original static-user-id so you can switch back
7. Run this code `pd.DataFrame({'A': [1,2,3], 'B': [1,2,3]}).to_csv('test.csv')`
8. Rerun mito, trust the analysis, and don't get an error
9. Update formula in the column you wrote to reference B
10. Save notebook
11. Return back to original static-user-id & restart your kernel
12. Delete test.csv
13. Run code `pd.DataFrame({'B': [1,2,3]}).to_csv('test2.csv')`
14. Rerun mito and update the import from test.csv to test2.csv


Note: That this works. However, I'm not sure if this will actually work multiplayer since the analysis on disk will be different.

naterush:
Good catch on the versioning stuff. For now, it actually doesn't really matter, because users will only be able to successfully share analyses if they are both on the most up-to-date version of Mito. 

So yes, let's see how popular this is for now, and then figure out how to handle it. One more logging thing I'm gonna add is which version of Mito the shared analysis was created with, vs. what version the shared analysis is being replayed on.

This will give us lots of good context, methinks!

naterush:
Re: your complex workflow above lol. Because we _always_ read in the analysis from the notebook before reading it off disk (as described in the specification), this exact scenario should work as we want it to. It defaults to the notebook and only goes on disk if it is not there!

naterush:
Sharing analyses dashboard: https://mixpanel.com/s/padWh

naterush:
Frontend tests as passing!

PR #509, opened 2022-10-26
Import config improvements

aarondr77:
# Description

- Adds the `decimal` configuration option to excel and csv imports
- Adds the `skiprows`configuration options to csv imports
- Refactors some of the simple import and excel import code to make adding params a bit easier in the future
- Logs user's pandas version as `version_pandas`
- Moves the package checking infrastructure to the frontend so we can just check it where we need it instead of having to continually add to the UserProfile.

Note: `decimal` was only added to read_excel in version 1.4 of pandas, so I only display that configuration option in the XLSX config taskpane when the user is post 1.4. 

# Testing

- See the tests
- Import files using `dev` and then switch to this branch and make sure imports still work
- Make sure that the `decimal` config is only available in the Excel import taskpane on or after pandas 1.4

Here is a csv and excel file with `,` decimals for testing:
[test_comma_decimal.xlsx](https://github.com/mito-ds/monorepo/files/9881195/test_comma_decimal.xlsx)
[test_comma_decimal.csv](https://github.com/mito-ds/monorepo/files/9881196/test_comma_decimal.csv)

# Documentation

### DIFF
--- a/mitosheet/mitosheet/api/get_imported_files_and_dataframes_from_current_steps.py
+++ b/mitosheet/mitosheet/api/get_imported_files_and_dataframes_from_current_steps.py
@@ -40,6 +40,7 @@ def get_import_data_with_single_import_list(step_type: str, params: Dict[str, An
                 'delimeters': get_sublist_at_index_from_optional_list(params.get('delimeters', None), index),
                 'encodings': get_sublist_at_index_from_optional_list(params.get('encodings', None), index),
                 'decimals': get_sublist_at_index_from_optional_list(params.get('decimals', None), index),
+                'skiprows': get_sublist_at_index_from_optional_list(params.get('skiprows', None), index),
                 'error_bad_lines': get_sublist_at_index_from_optional_list(params.get('error_bad_lines', None), index)
             }
         } for index, file_name in enumerate(params['file_names'])]
--- a/mitosheet/mitosheet/api/get_imported_files_and_dataframes_from_current_steps.py
+++ b/mitosheet/mitosheet/api/get_imported_files_and_dataframes_from_current_steps.py
@@ -39,6 +39,7 @@ def get_import_data_with_single_import_list(step_type: str, params: Dict[str, An
                 'file_names': [file_name],
                 'delimeters': get_sublist_at_index_from_optional_list(params.get('delimeters', None), index),
                 'encodings': get_sublist_at_index_from_optional_list(params.get('encodings', None), index),
+                'decimals': get_sublist_at_index_from_optional_list(params.get('decimals', None), index),
                 'error_bad_lines': get_sublist_at_index_from_optional_list(params.get('error_bad_lines', None), index)
             }
         } for index, file_name in enumerate(params['file_names'])]
@@ -50,7 +51,8 @@ def get_import_data_with_single_import_list(step_type: str, params: Dict[str, An
                 'file_name': params['file_name'],
                 'sheet_names': [sheet_name],
                 'has_headers': params['has_headers'],
-                'skiprows': params['skiprows']
+                'skiprows': params['skiprows'],
+                'decimal': params['decimal']
             }
         } for sheet_name in params['sheet_names']]
 
--- a/mitosheet/mitosheet/step_performers/import_steps/excel_import.py
+++ b/mitosheet/mitosheet/step_performers/import_steps/excel_import.py
@@ -39,12 +39,14 @@ class ExcelImportStepPerformer(StepPerformer):
         sheet_names: List[str] = get_param(params, 'sheet_names')
         has_headers: bool = get_param(params, 'has_headers')
         skiprows: int = get_param(params, 'skiprows')
+        decimal: bool = get_param(params, 'decimal')
 
         post_state = prev_state.copy()
 
         read_excel_params = {
             'sheet_name': sheet_names,
-            'skiprows': skiprows
+            'skiprows': skiprows,
+            'decimal': decimal
         }
 
         # Get rid of the headers if it doesn't have them
--- a/mitosheet/mitosheet/step_performers/import_steps/excel_import.py
+++ b/mitosheet/mitosheet/step_performers/import_steps/excel_import.py
@@ -11,7 +11,7 @@ from typing import Any, Dict, List, Optional, Set, Tuple
 import pandas as pd
 from mitosheet.code_chunks.code_chunk import CodeChunk
 from mitosheet.code_chunks.step_performers.import_steps.excel_import_code_chunk import \
-    ExcelImportCodeChunk
+    ExcelImportCodeChunk, build_read_excel_params
 from mitosheet.errors import make_file_not_found_error
 from mitosheet.state import DATAFRAME_SOURCE_IMPORTED, State
 from mitosheet.step_performers.step_performer import StepPerformer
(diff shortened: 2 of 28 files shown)


aarondr77:
Currently, we handle params in the csv import and excel import differently. CSV import params are optional whereas excel imports are mandatory. As a result, we will need to add a step ugprader to the excel import. 

I think a better option would be to have defaults for these additional params and make them optional.

mitosheet/src/components/import/CSVImportConfigScreen.tsx
```diff
@@ -206,7 +216,7 @@ function CSVImportConfigScreen(props: CSVImportConfigScreenProps): JSX.Element {
                 ...prevParams,
                 delimeters: fileMetadata?.delimeters || [DEFAULT_DELIMETER],
                 encodings: fileMetadata?.encodings || [DEFAULT_ENCODING],
-                error_bad_lines: [DEFAULT_ERROR_BAD_LINES]
+                error_bad_lines: [DEFAULT_ERROR_BAD_LINES],
```
naterush, line 219: Do you want to add decimals here as well?

mitosheet/mitosheet/api/get_imported_files_and_dataframes_from_current_steps.py
```diff
@@ -50,7 +52,8 @@ def get_import_data_with_single_import_list(step_type: str, params: Dict[str, An
                 'file_name': params['file_name'],
                 'sheet_names': [sheet_name],
                 'has_headers': params['has_headers'],
-                'skiprows': params['skiprows']
+                'skiprows': params['skiprows'],
+                'decimal': params['decimal']
```
naterush, line 56: I am guessing that you forgot to add these when you initially did this PR. I am wondering if there is some way to test we are making changes here effectively -- even if it's just adding a comment somewhere!

mitosheet/mitosheet/step_performers/import_steps/excel_import.py
```diff
@@ -36,21 +36,11 @@ def step_type(cls) -> str:
...
-
-        read_excel_params = {
-            'sheet_name': sheet_names,
-            'skiprows': skiprows
-        }
-
-        # Get rid of the headers if it doesn't have them
-        if not has_headers:
-            read_excel_params['header'] = None
+        read_excel_params = build_read_excel_params(params)
```
naterush, line 40: This is a good refactor!

mitosheet/mitosheet/telemetry/telemetry_utils.py
```diff
@@ -191,6 +191,11 @@ def _get_processing_time_log_params(steps_manager: StepsManagerType=None, start_
     from notebook import __version__ as notebook_version
 except:
     notebook_version = 'No notebook'
+try:
+    from pandas import __version__ as pandas_version
+except:
+    pandas_version = 'unkown'
```
naterush, line 197: `No pandas` might be more accurate!

naterush: While you're here, can you add a log for the version of `ipywidgets`?

mitosheet/mitosheet/tests/decorators.py
```diff
@@ -35,6 +35,11 @@
     reason='This test only runs on later versions of Pandas. API inconsistencies make it fail on earlier versions'
 )
 
+pandas_post_1_5_only = pytest.mark.skipif(
```
naterush, line 38: This was released like last week, so if we have things failing pre 1.5 we should def spend the time to make them work elsewhere!

mitosheet/mitosheet/tests/test_ts_types.py
```diff
@@ -214,3 +215,17 @@ def test_update_events_enum_defined():
     for UPDATE in UPDATES:
         assert UPDATE['event_type'] in update_types.values()
 
+
+def test_update_events_default_import_decimal():
```
naterush, line 219: Great work!

mitosheet/mitosheet/user/utils.py
```diff
@@ -105,19 +105,18 @@ def should_upgrade_mitosheet() -> bool:
...
+    Returns the pandas version
     """
-    from mitosheet.saved_analyses.schema_utils import is_prev_version
+    return pd.__version__
 
-    python_version_valid = sys.version_info.minor > 6
-    pandas_version_valid = not is_prev_version(pd.__version__, '0.25.0')
 
-    return python_version_valid and pandas_version_valid
+def get_python_version() -> int:
```
naterush, line 115: I think this function is wrong - we want it to return the entire version string, no?

naterush: If not def rename it!

mitosheet/src/utils/packageVersion.tsx
```diff
@@ -0,0 +1,43 @@
...
+    // Make sure that the current version is of the format x.y.z
+    if (currentVersionParts.length == 1) {
+        currentVersionParts[1] = 0
+    }
+    if (currentVersionParts.length == 2) {
+        currentVersionParts[2] = 0
+    }
+
+    let i = 0
+    for (i; i < currentVersionParts.length; i ++) {
```
naterush, line 21: for let i = 0 in the for loop!

mitosheet/src/components/import/CSVImportConfigScreen.tsx
```diff
@@ -120,15 +120,20 @@ const ENCODINGS = [
...
 const DELIMETER_TOOLTIP = 'The text that seperates one column from another.'
 const ENCODING_TOOLTIP = 'Set the encoding used to save this file.' // I can't think of anything better lol
+const DECIMAL_TOOLTIP = 'Set the character used to separate the decimal places.'
+const SKIP_ROWS_TOOLTIP = 'The number of rows at the top of the file to skip when reading data into the dataframe'
 const ERROR_BAD_LINES_TOOLTIP = 'Turn on to skip any lines that are missing fields.'
 
-export const DEFAULT_DELIMETER = ',';
-export const DEFAULT_ENCODING = 'default';
+export const DEFAULT_DELIMETER = ",";
+export const DEFAULT_ENCODING = "utf-8";
```
naterush, line 132: Cool, this is a good cleanup!

naterush (changes requested):
Code looking very clean, there's some great cleanups in here that are going to make future extensions much easier. Good work on this PR and making good calls on when / what to refactor, you nailed it!

Very minor changes requested to the code.

# Functionality

## The decimal doesn't seem to work

Screen Shot 2022-10-27 at 3 39 23 PM

See here. I cannot get the decimals read in correctly. Am I doing something wrong? 

Once you fix this up (or tell me what I'm doing wrong), I can do a re-review.

## The decimal splitter select

Screen Shot 2022-10-27 at 3 33 22 PM

I almost didn't see this in the dropdown. Want to just add a title to these that says ". (period)" and ", (comma)"?

## Minor cuttoff on ?

Screen Shot 2022-10-27 at 3 34 29 PM

## Tooltip on CSV import but not Excel import

Self explanatory, want to add tooltips to Excel import as well!

mitosheet/mitosheet/api/get_imported_files_and_dataframes_from_current_steps.py
aarondr77: Check out these tests, it catches this! https://github.com/mito-ds/monorepo/pull/509/commits/da3d93c74b98fc4e8a12e8070732a6d9b0a28767

You're right! I did forget at first so I added this test.

mitosheet/mitosheet/tests/decorators.py
aarondr77: Just added this since we had one for all of the other pandas versions -- we're not using anything that recent though.

naterush:
This is ready to merge. Waiting until the release is fully wrapped.

naterush:
Ok there's one refactor that I actually think we would be really nice to make sure that we're not introducing new bugs in the future.

Currently, on line 74 of the simple import, it says that if you specify one of the import params, you need to specify them all. The net result is that if we add a new param, and forget to set it in the frontend by default (which would be easy, given that it's optional), we'd end up not generating the rest of the params.

Do you think it's worth it to  refactor the simple_import file to actually check which of the parameters are optional, and use any of them that aren't? This might an hour of refactoring, but it's seems pretty sensible to me....

mitosheet/mitosheet/step_performers/import_steps/simple_import.py
```diff
@@ -46,6 +46,8 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
         skiprows: Optional[List[int]] = get_param(params, 'skiprows')
         error_bad_lines: Optional[List[bool]] = get_param(params, 'error_bad_lines')
 
+        print(params)
```
aarondr77, line 49: remove

naterush:
This change looks great. LGTM whenever you're ready.

PR #513, opened 2022-11-08
Filter in pivot backend

naterush:
# Description

Part 1 of this spec: https://www.notion.so/trymito/Filter-in-pivot-table-a9f9c63f470244948830e4b22510ae6f

The only complexity not in the spec was handling different pandas versions, but it's not bad + minimal impact.

**Backend only PRs are epic. Really nice development process, and I think the frontend process is going to massively fast as a result of having this. Loving this smaller PR breakdown!**

# Testing

Write any more tests you can think of. If you can't think of any more, all good.

# Documentation

No.


mitosheet/mitosheet/tests/step_performers/test_pivot.py
```diff
@@ -343,4 +349,243 @@ def test_edit_pivot_table_then_delete_optimizes():
...
+def test_pivot_filter(original_df, pivot_rows, pivot_columns, values, pivot_filters, pivoted_df):
+    mito = create_mito_wrapper_dfs(original_df)
+    mito.pivot_sheet(0, pivot_rows, pivot_columns, values, pivot_filters=pivot_filters)
+
+    assert mito.dfs[0].equals(original_df)
+    # For some reason, we need to check if dataframes are equal differently if
+    # they are empty, due to bugs in pandas .equals
+    if len(pivoted_df) > 0:
+        print(mito.dfs[1])
+        print(pivoted_df)
```
aarondr77: remove

PR #514, opened 2022-11-08
Mito config 2

aarondr77:
# Description

- Moves the mito-configuration approach to using environment variables instead of the mitosheet-helper-config package. 
- Also spells the word `environment` correctly all over the codebase, lol. 

Note that if you create the environment variables in the notebook and then refresh the kernel the environment variables are no longer set. As per the investigation below: The order of operations matter -- the environment variables need to be added, then the jupyter server needs to be launched.

# Testing

Set environment variables in the notebook and make sure Mito updates correctly: 
```
os.environ['MITO_CONFIG_VERSION_KEY'] = "1"
os.environ['MITO_CONFIG_SUPPORT_EMAIL_KEY'] = "aaron@sagacollab.com"
```

Then delete the environment variables and make sure Mito goes back to our defaults. 
```
del os.environ['MITO_CONFIG_VERSION_KEY']
del os.environ['MITO_CONFIG_SUPPORT_EMAIL_KEY'] = "aaron@sagacollab.com"
```

# Documentation

### DIFF
--- a/mitosheet/mitosheet/mito_config.py
+++ b/mitosheet/mitosheet/mito_config.py
@@ -7,6 +7,8 @@
 from typing import Any, Dict, Optional
 import os
 
+from mitosheet.telemetry.telemetry_utils import log
+
 # Note: Do not change these keys, we need them for looking up 
 # the environment variables from previous mito_config_versions.
 MITO_CONFIG_VERSION_KEY = 'MITO_CONFIG_VERSION'
@@ -72,6 +74,7 @@ class MitoConfig:
 
     def get_version(self) -> str:
         if self.mec is None or self.mec[MITO_CONFIG_VERSION_KEY] is None:
+            log('loaded_mito_enterprise_config')
             return '1' # NOTE: update this to be the most recent version, when we bump the version
         return self.mec[MITO_CONFIG_VERSION_KEY]
 
--- a/mitosheet/mitosheet/mito_config.py
+++ b/mitosheet/mitosheet/mito_config.py
@@ -13,7 +13,7 @@ MITO_CONFIG_VERSION_KEY = 'MITO_CONFIG_VERSION'
 MITO_CONFIG_SUPPORT_EMAIL_KEY = 'MITO_CONFIG_SUPPORT_EMAIL'
 
 # The default values to use if the mec does not define them
-DEFAULT_SUPPORT_EMAIL = 'founders@sagacollab.com'
+DEFAULT_MITO_CONFIG_SUPPORT_EMAIL = 'founders@sagacollab.com'
 
 # When updating the MEC_VERSION, add a function here 
 # to update the previous mec to the new version. For example, 
@@ -36,8 +36,8 @@ def upgrade_mito_enterprise_configuration(mec: Optional[Dict[str, Any]]) -> Opti
     return _mec
 
 # Since Mito needs to look up indivdual environment variables, we need to 
-# know the names of the variables associated with each mito congif version,
-# so we store them as a list here. 
+# know the names of the variables associated with each mito config version. 
+# To do so we store them as a list here. 
 mec_version_keys = {
     '1': [MITO_CONFIG_VERSION_KEY, MITO_CONFIG_SUPPORT_EMAIL_KEY]
 }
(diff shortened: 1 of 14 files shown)


mitosheet/mitosheet/mito_config.py
```diff
@@ -5,65 +5,86 @@
...
+
 class MitoConfig:
     """
     The MitoConfig class is repsonsible for 1) reading the settings from the 
-    mito_helper_config package if it is installed and 2) returning the configuration variables
+    environment variables 2) returning them as a mito_enterprise_configuration object
     used by the rest of the app. 
 
-    If the mito_helper_configuration package does not exist or does not set every configuration option, 
+    If the enviornment variables does not exist or does not set every configuration option,
```
aarondr77, line 66: environment

mitosheet/mitosheet/mito_config.py
```diff
@@ -5,65 +5,86 @@
...
+    def __init__(self):
+        mito_enterprise_configuration: Optional[Dict[str, Any]] = create_mec_from_environment_variables()
         self.mec = upgrade_mito_enterprise_configuration(mito_enterprise_configuration)
 
-    def get_version(self) -> Optional[int] :
+    def get_version(self) -> Optional[int]:
         if self.mec is None:
-            return 1 # NOTE: update this to be the most recent version, when we bump the version
-        return self.mec[MEC_CONFIG_KEY_VERSION]
+            return '1' # NOTE: update this to be the most recent version, when we bump the version
```
aarondr77, line 75: I could add a test to make sure that this is a string?

aarondr77:
Okay, the things I want to understand better is what environment variables the Mito Config file has access to. 

A quick test:
- In the initialization of MitoConfig, I am printing out `os.environ` and in my virtual environment, I am running the bash command `printenv`. 
- The vast majority of the environmnent variables printed out are in both, but each has some unique variables. For example, `GIT_PAGER` is only included in the `os.environ` result, and `OLDPWD` is only included in `printenv`.

Okay, so now I am going to try adding an eviornment variable to via a bash script and checking if Mito can access it. 
- In my bash shell, I ran the command:
```
export MITO_CONFIG_VERSION="1"; export MITO_CONFIG_SUPPORT_EMAIL="aaron@sagacollab.com"
```
- And confirmed the variables were created by rerunning `printenv` in my bash shell.

mitosheet/mitosheet/mito_config.py
```diff
@@ -5,65 +5,89 @@
...
-        _mec = mec_upgrade_functions[_mec[MEC_CONFIG_KEY_VERSION]](_mec)
+    while _mec[MITO_CONFIG_VERSION_KEY] in mec_upgrade_functions:
+        _mec = mec_upgrade_functions[_mec[MITO_CONFIG_VERSION_KEY]](_mec)
 
     return _mec
 
+# Since Mito needs to look up indivdual environment variables, we need to 
+# know the names of the variables associated with each mito config version. 
+# To do so we store them as a list here. 
+mec_version_keys = {
```
naterush, line 43: Caps, b/c it's a constant

mitosheet/mitosheet/mito_config.py
```diff
@@ -5,65 +5,89 @@
...
+        mec_potentially_outdated = create_mec_from_environment_variables()
+        self.mec = upgrade_mito_enterprise_configuration(mec_potentially_outdated)
 
-    def get_version(self) -> Optional[int] :
-        if self.mec is None:
-            return 1 # NOTE: update this to be the most recent version, when we bump the version
-        return self.mec[MEC_CONFIG_KEY_VERSION]
+    def get_version(self) -> str:
+        if self.mec is None or self.mec[MITO_CONFIG_VERSION_KEY] is None:
+            log('loaded_mito_enterprise_config')
```
naterush, line 77: I think that this log should be in the init, if the `self.mec is not None`

That way, we can make sure we always log when we create it, even if we somehow move to not using this get_version thing.

mitosheet/mitosheet/mito_config.py
```diff
@@ -5,65 +5,89 @@
...
+            return '1' # NOTE: update this to be the most recent version, when we bump the version
+        return self.mec[MITO_CONFIG_VERSION_KEY]
 
     def get_support_email(self) -> str:
```
naterush, line 81: get_version and get_email are internal only functions. I think we probably want to keep it that way, so we don't rely on it elsewhere outside of the get_mito_config function. 

Want to put a double underscore before them to make it so we can't access them from the outside?

naterush (changes requested):
Working great, and I honestly like this solution lots. For non-code configs, this is great. Minor cleanup, then LGTM.

mitosheet/mitosheet/mito_config.py
aarondr77: According to [this](https://stackoverflow.com/questions/551038/private-implementation-class-in-python) convention is to use a single underscore, so I'm going to do that.

PR #516, opened 2022-11-08
NaN column Headers

aarondr77:
# Description

Makes `column_header_to_transpiled_code` work with NaN and NaT column headers. 

# Testing

See pytest and rename these columns:
```
df = pd.DataFrame({pd.NA: [1], pd.NaT: [1], np.NaN: [1]})
```

# Documentation

No.


aarondr77:
Okay here is a very related bug:

1. Create this dataframe
```
df = pd.DataFrame({'Unnamed: 0': [None, '1/2/22', '1/3/22']})
```
2. Convert the column to a date
3. Promote the first row to header
4. Rename the header 
5. See the invalid generated code
```
df.rename(columns={NaT: 'new'}, inplace=True)
```

naterush (changes requested):
Screen Shot 2022-11-09 at 11 37 02 AM

This handles pd.NA, but not np.NaN. Can you add more checks for np.NaN, specifically with `np.isnan`?

Also, a few tests for this column header function on NaN would be great. Then, LGTM!

naterush:
Also fix tests plz!

naterush (changes requested):
Once you fix the failing tests for earlier versions of pandas, LGTM. 

The main error I'm seeing is `AttributeError: module 'pandas' has no attribute 'NA'`. It looks like this was added in 1.0 -- so let's do this check for `pd.NA` behind a version check of pandas.

Then, this should be good to go. Sad that this generates a warning, but it's fine for now.

PR #522, opened 2022-11-14
Small dx improvements

aarondr77:
# Description

Makes some small developer experience improvements:
1. Changes the print for steps so that it displays helpful information instead of the object address
2. Makes the variable names of `is_prev_version` function easier to understand. 

**Note branches off nan-column-headers to make it easy to see differences.**

# Testing

# Documentation


naterush:
Do you want to be merging this into `nan-column-headers`?

naterush:
This looks great, nice work improving this situation for us. LGTM (make sure to update the branch depending on access timing).

PR #530, opened 2022-11-17
mitosheet: fix search string in unique values tab

aarondr77:
# Description

Fixes bug with search in unique values tab where if the value was `6000` as int, so that Mito formatted it to `6,000`, searching for `6000` did not return the correct result. 

# Testing

Try testing above scenario

# Documentation

No.


naterush (commented):
This is an ok partial fix, and I'm fine to merge it. I'll note it doesn't really solve the full problem - b/c if you search with commas you don't get the right values still. Pretty sure this is because the backend doesn't search on the formatted values... 
Screen Shot 2022-11-17 at 11 25 29 AM

PR #533, opened 2022-11-17
Optimize code chunks more

naterush:
# Description

- Optimizes pivot table code chunks, if you edit them immediately
- Optimizes our reorder if you delete column
- Optimizes our rename if you delete these columns

# Testing

The above. Add a test or two if you can think of any good ones!

# Documentation

No.

### DIFF
--- a/mitosheet/mitosheet/code_chunks/step_performers/column_steps/delete_column_code_chunk.py
+++ b/mitosheet/mitosheet/code_chunks/step_performers/column_steps/delete_column_code_chunk.py
@@ -10,6 +10,7 @@ from typing import List, Optional
 from mitosheet.code_chunks.code_chunk import CodeChunk
 from mitosheet.transpiler.transpile_utils import \
     column_header_list_to_transpiled_code
+from mitosheet.code_chunks.step_performers.column_steps.reorder_column_code_chunk import ReorderColumnCodeChunk
 
 
 class DeleteColumnsCodeChunk(CodeChunk):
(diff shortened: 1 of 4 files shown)


naterush:
I tried some other optimizations, like optimizing out filters and sorts if the column was then deleted. But this actually doesn't make sense. If you do a sort and delete the column, you still want the sort to be applied (or at least, it does in Mito currently). Not having any code makes the sheet not match with the dataframes the code generates.

naterush:
@aarondr77 can you think of any more _extremly_ low hanging fruit. Specifically, looking for optimizations that:
1. Don't take changing the format of any steps or code chunks.
2. Are extremely easy to reason about. 

I'd love to just do a few more. They are extremely fun, and have the very-nice property that as we do more, they combine with each other in un-expected ways. It feels very non-linear to work on them....

mitosheet/mitosheet/code_chunks/step_performers/column_steps/delete_column_code_chunk.py
```diff
@@ -65,4 +74,101 @@ def combine_right(self, other_code_chunk: CodeChunk) -> Optional[CodeChunk]:
...
+        if len(to_remove_from_rename) == len(column_ids_to_new_column_headers):    
+            return DeleteColumnsCodeChunk(
+                other_code_chunk.prev_state,
+                self.post_state,
+                self.params,
+                self.execution_data
+            )
+        
+        # Otherwise, we don't optimize
+        return None
```
aarondr77, line 160: I wonder in practice how helpful this optimization is going to be without filling this in. But we certainly cover the case where we can do the most optimization, so that's nice

PR #536, opened 2022-11-20
Full Width Columns

aarondr77:
# Description

Allows you to double click on the column resizer to set the column at the min(column header, longest data cell). Performs operations for all selected columns and the column you double click on. 

# Testing

Resize columns using the double click and make sure it works. 

# Documentation

Yeah, we should probably note this somewhere. Maybe we could create a spreadsheet navigation page?

### DIFF
--- a/mitosheet/src/components/endo/ColumnHeader.tsx
+++ b/mitosheet/src/components/endo/ColumnHeader.tsx
@@ -14,6 +14,7 @@ import { DEFAULT_HEIGHT } from './EndoGrid';
 import { ControlPanelTab } from '../taskpanes/ControlPanel/ControlPanelTaskpane'
 import { submitRenameColumnHeader } from './columnHeaderUtils';
 import ColumnHeaderDropdown from './ColumnHeaderDropdown';
+import { changeColumnWidthDataArray, guessFullWidth } from './widthUtils';
 
 
 export const HEADER_BACKGROUND_COLOR_DEFAULT = '#E8EBF8' // This is var(--mito-light-blue) - update this if we change this variable
@@ -46,6 +47,7 @@ export const HEADER_TEXT_COLOR_DEFAULT = '#494650' // This is var(--mito-gray) -
 */
 const ColumnHeader = (props: {
     gridState: GridState,
+    setGridState: React.Dispatch<React.SetStateAction<GridState>>
     sheetData: SheetData,
     editorState: EditorState | undefined;
     setEditorState: React.Dispatch<React.SetStateAction<EditorState | undefined>>;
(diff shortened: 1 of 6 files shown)


mitosheet/src/components/endo/widthUtils.tsx
```diff
@@ -116,4 +116,21 @@ export const reconciliateWidthData = (prevWidthData: WidthData | undefined, oldC
...
     return getWidthData(sheetData, oldWidths);
+}
+
+/*
+    Guess the minimum column width such that the entire column header and each cell is displayed
+    completely withour wrapping.
+
+    Returns a number of pixels 
+*/
+export const guessFullWidth = (sheetData: SheetData, columnIndex: number, displayColumnHeader: string): number => {
```
naterush, line 127: `guessFullWidthOfColumnHeaderOrContent` is a more informative name I won't wonder about in future, I think

naterush: E.g. much clearer, given we have functions like `getWidthData`

mitosheet/src/components/endo/ColumnHeader.tsx
```diff
@@ -94,6 +96,24 @@ const ColumnHeader = (props: {
         setTimeout(() => focusGrid(props.containerRef.current), 100);
     }
 
+    const getColumnIndexesFullWidthArray = (columnIndexes: number[]): WidthData[]  => {
```
naterush, line 99: can we rename to `getWidthArrayAtFullWidthForColumnIndexes`

naterush: Also, I think this can go in `widthUtils.tsx` -- it doesn't need to be defined in ColumnHeader for any reason.

mitosheet/src/components/endo/ColumnHeader.tsx
```diff
@@ -94,6 +96,24 @@ const ColumnHeader = (props: {
...
+        let widthDataArray = props.gridState.widthDataArray
+        columnIndexes.forEach(columnIndex => {
+            const columnHeader = getCellDataFromCellIndexes(props.sheetData, -1, columnIndex).columnHeader;
+
+            if (columnHeader === undefined) {
+                return
+            }
+
+            const finalColumnHeader  = getColumnHeaderParts(columnHeader).finalColumnHeader
+            const displayColumnHeader = getDisplayColumnHeader(finalColumnHeader)
```
naterush, line 109: I think you can just call `getDisplayColumnHeader` on this, no?

naterush (changes requested):
Maybe bump the spacing on the column header guess -- mine inconsistently makes enough size on safari:

Screen Shot 2022-11-21 at 9 26 25 AM

The other this is that I can't shift click on the column header resizes currently, to select a range of columns. This is not ideal. I think this is because the onClick, onMouseDown, etc stops propagation now.

So that shift+click still works on the resize for selecting a range, can we just case on if shiftKey is pressed in the onClick later in this file? E.g. if you shift + click on the column header anything, it shouldn't open anything -- this is most def just a selection change!

aarondr77:
> So that shift+click still works on the resize for selecting a range, can we just case on if shiftKey is pressed in the onClick later in this file?

I didn't do this because I don't think we want you to be able to select columns on the resizer when using shift+click, but not just using click. Its inconsistent. So instead, I made the resizer smaller so that there is still space on the right hand side of the icons to click on the column. 

This is what Excel does.

PR #545, opened 2022-11-28
mitosheet: make cell editor textarea

aarondr77:
# Description

This is getting a bit complex. In particular, the textArea, unlike the Input does not default submit the form it is in when the user presses Enter. As a result, we need to handle the Enter on our own. 

To mirror the behavior of Excel and Google sheets, we want: 
- [x] Enter => Submits the formula
- [x] CMD + Enter => Adds new line to the cell editor
- [ ] Shift + Enter => Submits the formula

TODO:
- [x] If the user presses Enter to open the cell editor, this event gets fired and it immedietly closes the cell editor..
- [x] It doesn't handle adding the newline in the correct location of the formula. It should be added wherever the user's cursor is. 
- [x] It doesn't keep the cursor in the correct location. It jumps to the end of the text area
- [ ] Make the CellEditor only a TextArea when its the formula bar. Otherwise, keep using the input
- [ ] Some more design questions below

# Testing

# Documentation


aarondr77:
Implementation Notes:

### Google Sheets
- Enter => Submits the formula
- CMD + Enter => Adds a new line to the cell editor
- Shift + Enter => Submits the formula

### Excel
- Enter => Submits the formula
- CMD + Enter => Submits the formula
- Shift + Enter => Submits the formula 

### Slack 
- Enter => Send message
- CMD + Enter => Sends message 
- Shift + Enter => Adds new line to message

### Mito Implementation
- Enter => Submits the formula
- CMD + Enter => Adds new line to the cell editor
- Shift + Enter => Submits the formula

aarondr77:
In Excel, using the tab key does not indent the code. Instead, if there is a suggestion selected in the cell editor, it will take it. Right now that is the same for us. But not having the tab key for multi line formula is horrible. My proposal is that when the user presses tab: 
- If there is no suggestion selected (meaning the user has not used the arrow keys to select it), then we add the tab to the formula. 
- If there is a suggestion selected, then we select it. (what we currently do)

aarondr77:
We also might want to automatically open the text area to multiple lines if the formula has multiple lines in it. For example, if the user wrote a formula: 

```
IF(
    loan_amount > 5000,
    1, 
    0
)
```

Then, when they open the formula in the formula bar, we pop it open to show 5 lines. We should probably pick a default, other 1 line or 5 lines, so that it's constant for users.

mitosheet/src/components/endo/celleditor/CellEditor.tsx
```diff
@@ -193,15 +209,82 @@ const CellEditor = (props: {
...
 
         // Make sure we jump to the end of the input, as we took the suggestion
-        cellEditorInputRef.current?.setSelectionRange(
+        cellEditorTextAreaRef.current?.setSelectionRange(
             fullFormula.length, fullFormula.length
         )
     }
 
-    const onKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    const onSubmit = async (e?: React.FormEvent<HTMLFormElement> | undefined) => {
```
aarondr77, line 217: None of this is new. This is just moved into a function.

naterush:
@aarondr77 appreciate the work you've put in here. However, this feels to me like it's clearly one of those things that actually needs a specification of proper behavior before we continue -- there are very many complexities to understand -- and we probably want to consider just matching sheets or excel exactly at the end of the day.

I think there were some great learnings in this PR about what is hard / complex / unexpected - want to capture them in a shaping document so that we can shape something around "better formula editor" in the future?

Then, I vote we retire this PR for now. I think it blew up more than either of us expected -- but maybe should have known -- last time we tried to touch the cell editor, it blew up a bit as well.

aarondr77:
Some notes for the the future:

By default, the Textarea adds a new line when `Enter` is pressed. This is is contrast to the input field, which submits the form that it is in on `Enter`. Because we want to replicate Excel behavior, where `Enter` submits the formula, this is the primary source of complexity that is increasingly complex due to the following: 

1. OnKeyDown cannot detect Shift+Enter, but onKeyUp can. 
2. When we use ctrl+Enter to add a new line to our cell editor, the text area defaults to moving the cursor to the end of the cell editor because the cell editor value is updated programatically such that the value we are setting it to does not match the value that the cell editor expects. We're not able to set the cursor position in the onKeyDown event because the updating of the textArea value happens afterwards. Instead, we can use a useEffect to update the cursor position after the text area's value has been updated.
3. Getting `Enter` in the text area to submit the form is easiest if we just case on when Enter is pressed and call the `onSubmit` function directly. Alternatively, we can try to create a submit event and dispatch it, but I was not able to get that to work. 
4. In Excel, using the tab key does not indent the code. Instead, if there is a suggestion selected in the cell editor, it will take it. Right now that is the same for Mito. But not having the tab key for multi line formula is horrible. My proposal is that when the user presses tab:
-- If there is no suggestion selected (meaning the user has not used the arrow keys to select it), then we add the tab to the formula.
-- If there is a suggestion selected, then we select it. (what we currently do)

PR #565, opened 2022-12-12
mitosheet: add basic code snippet backend

naterush:
# Description

Implements first sprint: https://www.notion.so/trymito/Code-Snippets-581c52cf00854476bac257a78e17e37c

BTW: I think there's too much detail in this spec. You effectively wrote all the code, which makes sense b/c we wanted to be able to test the integration, but at that point -- let's just write it in the codebase. 

I think if we  find ourselves writing 80% of the code for a sprint in a spec, then let's just either leave it out, or do it in the code.

# Testing

See test.

# Documentation

No.


naterush (commented):
Looks good.

PR #568, opened 2022-12-13
mitosheet: update copying to allow users to select non endo text

aarondr77:
# Description

Fixes our copy logic by relying on the active element so that users can copy things like:
- Summary stats data
- Values from unique values tab
- Text from the modal
- Text from the formula bar

# Testing

Copy text from Endo as well as different text from Mito and make sure it copies correctly. 

# Documentation

No

### DIFF
--- a/mitosheet/mitosheet/mito_frontend.js
+++ b/mitosheet/mitosheet/mito_frontend.js
@@ -22977,6 +22977,7 @@ For more info, visit https://reactjs.org/link/mock-scheduler`);
         if (!Object.keys(KEYBOARD_SHORTCUTS).includes(e.key) || !e.ctrlKey && !e.metaKey) {
           return;
         }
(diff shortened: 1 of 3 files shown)


aarondr77:
Fixes our selection code so that users can copy things like:

Summary stats data
Values from unique values tab
Text from the modal
Text from the formula bar
There is one bug where if the user has a taskpane open > selects a column header > copies: the copy remains in the taskpane because the Selection does not update when the user clicks a column header. The selection does update, however, when the user selects a cell.

This implementation, I believe, is a bit of a hack (see comments in code).

aarondr77:
This is working on Safari, but way less well on FF. I think this characterizes our problem: 

This is only 

Screenshot 2022-12-13 at 5 43 48 PM

When the user clicks on the grid, they aren't actually selecting text, they are just changing like... what is active?

PR #573, opened 2022-12-16
mitosheet: add id, wrap text on code snippets

naterush:
# Description

- When the `MITO_CONFIG_CODE_SNIPPETS_VERSION` and `MITO_CONFIG_CODE_SNIPPETS_URL` are set, uses them to  retrieve the code snippets and displays them in the taskpane
- Adds a Get Support option to each code snippet dropdown that can be configured through the `MITO_CONFIG_CODE_SNIPPETS_SUPPORT` environment variable
- In order to type check the requests code, I installed `types-requests`. 
- Fixes race condition in the code snippets taskpane where clicking on another code snippet while a code snippet was already open resulted in no dropdowns being open. 

# Testing

Set the environment variables and test them. Here's some starter code
```
import os
os.environ['MITO_CONFIG_VERSION'] = '2'
os.environ['MITO_CONFIG_SUPPORT_EMAIL'] = 'aaron@sagacollab.com'
os.environ['MITO_CONFIG_CODE_SNIPPETS_SUPPORT_EMAIL'] = 'naterush1997@gmail.com'
os.environ['MITO_CONFIG_CODE_SNIPPETS_VERSION'] = '1'
os.environ['MITO_CONFIG_CODE_SNIPPETS_URL'] = 'json_test'
```

- make sure that the get_code_snippets taskpane uses the environment variables to get the code snippets instead of the default code snippets.
- make sure that clicking the get support button generates an email instead of sending users to slack. 

# Documentation

Yes, we should add code snippets to the docs


mitosheet/mitosheet/api/get_code_snippets.py
```diff
@@ -63,4 +73,21 @@
...
+        mito_config_code_snippets_url = os.environ.get(MITO_CONFIG_KEY_CODE_SNIPPETS_URL)
+        if mito_config_code_snippets_version == '1' and mito_config_code_snippets_url is not None:
+                
+                # Request the code snippets from the url
+                response = req.get(mito_config_code_snippets_url, verify=False)
+
+                if response.status_code == 200:
+                        # Parse the respone body into JSON 
+                        code_snippets = response.json()
+                        return json.dumps(code_snippets)
```
naterush, line 87: We aren't doing the caching we said we're going to do in the spec. 

Let's pull out all this logic into a `get_custom_code_snippets` function and cache w/ a global variable the first time this function is run.

mitosheet/mitosheet/enterprise/mito_config.py
```diff
@@ -13,9 +13,33 @@
 # the environment variables from previous mito_config_versions.
 MITO_CONFIG_KEY_VERSION = 'MITO_CONFIG_VERSION'
 MITO_CONFIG_KEY_SUPPORT_EMAIL = 'MITO_CONFIG_SUPPORT_EMAIL'
+MITO_CONFIG_KEY_CODE_SNIPPETS_SUPPORT_EMAIL = 'MITO_CONFIG_CODE_SNIPPETS_SUPPORT_EMAIL'
```
naterush, line 16: It feels wrong we don't include the other keys we just added here. I think we probably should include these keys here (and access them through here) - we want all the configuration to be managed through a single location.

Since the code snippets URL and version should always be either both there or not both there, I could see the accessor for them returning them both together or neither (aka, we can enforce that in the type system). 

I think you'll have to change up how you then access the config, where you pass it to the steps manager, so you can pass it through to the API. That seems very sensible to me.

mitosheet/mitosheet/tests/test_mito_config.py
```diff
@@ -14,48 +14,72 @@ def delete_env_var_if_exists(env_var: str) -> None:
...
+        MITO_CONFIG_KEY_VERSION: '2',
+        MITO_CONFIG_KEY_SUPPORT_EMAIL: 'aaron@sagacollab.com',
+        MITO_CONFIG_KEY_CODE_SNIPPETS_SUPPORT_EMAIL: DEFAULT_MITO_CONFIG_CODE_SNIPPETS_SUPPORT_EMAIL
     }    
 
     # Delete the environmnet variables for the next test
-    delete_env_var_if_exists(MITO_CONFIG_KEY_SUPPORT_EMAIL)
-    delete_env_var_if_exists(MITO_CONFIG_KEY_VERSION)
\ No newline at end of file
+    delete_all_mito_config_environment_variables()
```
naterush, line 85: And add some tests for the above changes to the config so all is accessed here.

mitosheet/src/components/taskpanes/CodeSnippets/CodeSnippetsTaskpane.tsx
```diff
@@ -129,6 +130,14 @@ const CodeSnippetsTaskpane = (props: CodeSnippetsTaskpaneProps): JSX.Element =>
                                         title="Write to Notebook" 
                                         onClick={writeToCell}
                                     />
+                                    <DropdownItem
+                                        title='Get Support'
+                                        onClick={() => {
+                                            const openLocation = props.userProfile.mitoConfig[MitoEnterpriseConfigKey.CODE_SNIPPETS_SUPPORT_EMAIL] === DEFAULT_SUPPORT_EMAIL ? SLACK_INVITE_LINK : `mailto:${props.userProfile.mitoConfig[MitoEnterpriseConfigKey.CODE_SNIPPETS_SUPPORT_EMAIL]}?subject=Mito Code Snippet Support: ID ${codeSnippet.Id}`
```
naterush, line 136: Let's only show this if code snippet email is set!

mitosheet/src/components/elements/Dropdown.tsx
```diff
@@ -60,9 +60,11 @@ interface DropdownProps {
     */
     display: boolean;
     /** 
-        * @param closeDropdown - The function to used to close the dropdown when the user clicks
+        * @param closeDropdown - The function to used to close the dropdown when the user clicks. For a single dropdown, you 
+        * probably want to provide a closeDropdown function. But when there are multiple dropdowns and the you want to ensure that 
+        * only one is open at a time, the parent component of the dropdowns can handle closing the dropdown instead to avoid race conditions.
     */
-    closeDropdown: () => void;
+    closeDropdown?: () => void;
```
naterush, line 67: I don't really understand this. What is the parent component in this case?

mitosheet/setup.py
```diff
@@ -121,7 +121,8 @@
                 'pytest',
                 'flake8',
                 'types-chardet',
-                'mypy'
+                'types-requests',
```
naterush, line 124: Do we use this? Do we need this in the deployed package (I think no!) or just the development version?

mitosheet/mitosheet/api/get_code_snippets.py
```diff
@@ -63,4 +73,21 @@
...
+        mito_config_code_snippets_version = os.environ.get(MITO_CONFIG_KEY_CODE_SNIPPETS_VERSION)
+        mito_config_code_snippets_url = os.environ.get(MITO_CONFIG_KEY_CODE_SNIPPETS_URL)
+        if mito_config_code_snippets_version == '1' and mito_config_code_snippets_url is not None:
+                
+                # Request the code snippets from the url
+                response = req.get(mito_config_code_snippets_url, verify=False)
+
+                if response.status_code == 200:
+                        # Parse the respone body into JSON 
+                        code_snippets = response.json()
```
naterush, line 86: I think we should check that these have the correct format, no? Otherwise the frontend will likely crash. 

If they don't, we can raise a ValueError with that specific value that says they are undefined.

mitosheet/mitosheet/api/get_code_snippets.py
```diff
@@ -63,4 +73,21 @@
 ]
 
 def get_code_snippets(params: Dict[str, Any], steps_manager: StepsManagerType) -> str:
-    return json.dumps(DEFAULT_CODE_SNIPPETS)
+        # If code snippet environment variables are set, then use them to get the code snippets. 
+        mito_config_code_snippets_version = os.environ.get(MITO_CONFIG_KEY_CODE_SNIPPETS_VERSION)
```
naterush, line 77: Can we write a test for this? In mitosheet/tests/api -- I think we can start writing tests for all these (I already wrote one for code snippets).
1. one that additional tests if the URL is provided that is legit
2. one where the Url gives code snippets that aren't the exact format

Both can be done with the JSON echo service you used in the spec, methinks!

naterush (commented):
Some code cleanup, and some testing I'd love to see added. Then, I'll rereview the functionality again - since there's big changes I'm requesting, I'll wait on that for now!

mitosheet/src/components/elements/Dropdown.tsx
aarondr77: We have the CodeSnippetsTaskpane keeping track of which dropdowns are open via the `const [openDropdownIndex, setOpenDropdownIndex]`. And then in the dropdown component we display it if `display={codeSnippetIndex === openDropdownIndex} `.

aarondr77: But actually, maybe I'll update this so we do have the closeDropdown because that handles the click outside of this taskpane and I'll just make it so that the closeDropdown ignore any click within the rows of the taskpane because that is handled by the taskpane itself using the logic above.

mitosheet/setup.py
aarondr77: Right now it's just in the `test` section. I think that section is just used in development and not the deployed package, right? That's where we have mypy, pytest, and types-charset, none of which are needed in the deployed package.

aarondr77:
We should also add a loading indicator to the taskpane incase the URL takes a while to load

mitosheet/mitosheet/api/get_code_snippets.py
aarondr77: I can't get the JSON echo service to return a list of json, so I can't use it to test that it works properly, but I can use it to test that it errors when it should

mitosheet/mitosheet/api/get_code_snippets.py
```diff
@@ -62,5 +70,117 @@
         } 
 ]
 
+def get_code_snippets_format_error(code_snippets: Any) -> str:
```
aarondr77, line 73: Update the comment to actually say what this does .

mitosheet/mitosheet/api/get_code_snippets.py
```diff
@@ -62,5 +70,117 @@
...
+                if error:
+                        return f"Custom code snippets should have the format: {correct_code_snippets_type} {NEWLINE}{NEWLINE}But this code snippet was included: {code_snippet}"
+
+        return ''
+
+
+# Global variable used to cache the custom code snippets so that when users
+# open the code snippets taskpane multiple times in the same mito instantiation, 
+# we can display the custom code snippets without querying the url each time.
+CACHED_CUSTOM_CODE_SNIPPETS: Optional[List[CodeSnippet]] = None
```
aarondr77, line 130: Add separating new line

mitosheet/mitosheet/api/get_code_snippets.py
```diff
@@ -62,5 +70,117 @@
...
+                else:
+                        return json.dumps({
+                                'status': 'error',
+                                'error_message': code_snippet_format_error
+                        })
+
+        else:
+                return json.dumps({
+                        'status': 'error',
+                        'error_message': f"Error accessing the code snippets data from the URL. Response status code: {response.status_code}" ,
```
aarondr77, line 169: I can save code by creating an error response object and then just filling in the error message and returning json.dumps object.

mitosheet/mitosheet/api/get_code_snippets.py
```diff
@@ -62,5 +70,117 @@
...
+                code_snippets = response.json()
+                code_snippet_format_error = get_code_snippets_format_error(code_snippets)
+
+                if code_snippet_format_error == '':
+                        # Cache the code snippets so we don't need to request them from the url next time
+                        CACHED_CUSTOM_CODE_SNIPPETS = code_snippets
+                        return json.dumps({
+                                'status': 'success',
+                                'code_snippets': code_snippets
+                        })
```
aarondr77, line 159: Make this an object too so that we can make sure to update them all properly at once

mitosheet/mitosheet/tests/test_mito_config.py
```diff
@@ -14,48 +14,89 @@ def delete_env_var_if_exists(env_var: str) -> None:
...
+
+def test_mito_config_update_version_1_to_2():
     # Set environment variables
     os.environ[MITO_CONFIG_KEY_VERSION] = "1"
     os.environ[MITO_CONFIG_KEY_SUPPORT_EMAIL] = "aaron@sagacollab.com"
     
     # Test reading environment variables works properly
     mito_config = MitoConfig()
+    print(mito_config.get_mito_config())
```
aarondr77, line 90: remove

mitosheet/src/components/elements/Dropdown.tsx
```diff
@@ -60,9 +60,13 @@ interface DropdownProps {
...
     display: boolean;
     /** 
-        * @param closeDropdown - The function to used to close the dropdown when the user clicks
+        * @param closeDropdown - The function to used to close the dropdown when the user clicks.
     */
     closeDropdown: () => void;
+    /**
+        * @param [classToIgnoreCloseDropdown] - Optional class that if clicked, does not trigger the closeDropdown function
+    */
+    classToIgnoreCloseDropdown?: string;
```
aarondr77, line 69: Don't need this anymore

mitosheet/src/components/taskpanes/CodeSnippets/CodeSnippetsTaskpane.tsx
```diff
@@ -129,11 +146,24 @@ const CodeSnippetsTaskpane = (props: CodeSnippetsTaskpaneProps): JSX.Element =>
...
+                                </Dropdown>
                             </Col>
                         </Row>
                     )
                 })}
+                {codeSnippetAPIResult === undefined && 
+                    <p>
+                        Loading code snippets <LoadingDots />
+                    </p>
+                }
```
aarondr77, line 166: Make this text prettier. Maybe move down to middle of taskpane

mitosheet/src/jupyter/api.tsx
```diff
@@ -687,8 +687,10 @@ export default class MitoAPI {
         }, {})
 
         if (resultString !== undefined && resultString !== '') {
+            console.log(resultString)
```
aarondr77, line 690: remove

mitosheet/src/jupyter/api.tsx
```diff
@@ -687,8 +687,10 @@ export default class MitoAPI {
...
             return JSON.parse(resultString);
         }
+        console.log(undefined)
```
aarondr77, line 693: remove

mitosheet/mitosheet/tests/test_mito_config.py
```diff
@@ -14,48 +14,88 @@ def delete_env_var_if_exists(env_var: str) -> None:
...
 
+def delete_all_mito_config_environment_variables() -> None:
+    """
+    Deletes all of the MITO CONFIG environment variables that are used in the MEC
+    """
+    delete_env_var_if_exists(MITO_CONFIG_KEY_VERSION)
+    delete_env_var_if_exists(MITO_CONFIG_KEY_SUPPORT_EMAIL)
+    delete_env_var_if_exists(MITO_CONFIG_KEY_CODE_SNIPPETS_SUPPORT_EMAIL)
+    delete_env_var_if_exists(MITO_CONFIG_KEY_CODE_SNIPPETS_URL)
+    delete_env_var_if_exists(MITO_CONFIG_KEY_CODE_SNIPPETS_VERSION)
```
naterush, line 25: Let's write a loop over the list in the config class, so we can stop updating this

naterush (commented):
Happy hollydays! Sorry for the large-ish rearview -- I think we can improve our types in a few places here (nothing major). Also, I'm confused on how to test the success case - let me know if I'm missing something.

mitosheet/mitosheet/enterprise/mito_config.py
```diff
@@ -77,19 +122,42 @@ def __init__(self):
...
         return self.mec[MITO_CONFIG_KEY_VERSION]
 
     def _get_support_email(self) -> str:
         if self.mec is None or self.mec[MITO_CONFIG_KEY_SUPPORT_EMAIL] is None:
             return DEFAULT_MITO_CONFIG_SUPPORT_EMAIL
         return self.mec[MITO_CONFIG_KEY_SUPPORT_EMAIL]
 
+    def _get_code_snippets_version(self) -> str:
+        if self.mec is None or self.mec[MITO_CONFIG_KEY_CODE_SNIPPETS_VERSION] is None:
+            return '1'
```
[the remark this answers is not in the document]
aarondr77: Hmmm I think either is fine. We default `MITO_CONFIG_VERSION` to 2 even if the variable is not set, so I'm just following the same standard here.

mitosheet/mitosheet/tests/api/test_get_code_snippets.py
```diff
@@ -1,9 +1,37 @@
...
+def test_get_code_snippets_incorrectly_formatted_code_snippets():
+    os.environ[MITO_CONFIG_KEY_VERSION] = "2"
+    os.environ[MITO_CONFIG_KEY_CODE_SNIPPETS_URL] = "http://echo.jsontest.com/key/value/one/two"
+    os.environ[MITO_CONFIG_KEY_CODE_SNIPPETS_VERSION] = "1"
+
+    mito = create_mito_wrapper_dfs()
+    code_snippet_response = get_code_snippets({}, mito.mito_backend.steps_manager)
+    assert json.loads(code_snippet_response)['status'] == 'error'
+
+    delete_all_mito_config_environment_variables()
```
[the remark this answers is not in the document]
aarondr77: I couldn't get this approach to work, but I did get an approach with `pytest_httpserver` to work. It's obviously not ideal, but I tried for a few hours. Feel free to give it a go.

mitosheet/src/types.tsx
```diff
@@ -723,6 +727,11 @@ export interface UserProfile {
     mitoConfig: {
         [MitoEnterpriseConfigKey.MEC_VERSION]: number | undefined
         [MitoEnterpriseConfigKey.SUPPORT_EMAIL]: string
+        [MitoEnterpriseConfigKey.CODE_SNIPPETS]: {
+            [MitoEnterpriseConfigKey.CODE_SNIPPETS_VERSION]: string | undefined,
+            [MitoEnterpriseConfigKey.CODE_SNIPPETS_URL]: string | undefined
+            [MitoEnterpriseConfigKey.CODE_SNIPPETS_SUPPORT_EMAIL]: string
```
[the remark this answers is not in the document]
aarondr77: I see that point for sure. But I didn't take this update because I wanted to give users helpful error messages when they try to open the code snippets taskpane. If they have only defined some of the variables, then we give them helpful guidance on how to correct it. 

And since we don't use these variables in the frontend except for CODE_SNIPPETS_SUPPORT_EMAIL, I think it's fine for now. Another option would be:

Moving all of the environment variable error handling into the MITO_CONFIG file itself so the errors are reported right away instead of when the user opens the code snippets taskpane and then having all of the config variables in the step performer being validated and ready to go. 

But potentially that is worse for end users who would see those errors right away even if they don't use the features that are not configured properly. + there is not a lot that the end user can actually do about it besides go ask their IT team to fix.

mitosheet/mitosheet/enterprise/api/code_snippets_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    code_snippet_format_error = get_code_snippets_format_error(code_snippets)
+
+    if code_snippet_format_error is not None:
+        # Note: We do not log the code_snippet_format_error because it contains user-defined code snippets
+        log('get_code_snippet_error', {'get_code_snippet_error_reason': 'code_snippet_format_error'})
+        return create_error_return_obj(code_snippet_format_error)
+    
+    # Cache the code snippets so we don't need to request them from the url next time
+    cached_custom_code_snippets = code_snippets
+    log('using custom code snippets')
```
naterush, line 122: This is a strange log message! Underscores.

mitosheet/mitosheet/enterprise/mito_config.py
```diff
@@ -75,21 +120,44 @@ def __init__(self):
...
+        if self.mec is None or self.mec[MITO_CONFIG_KEY_CODE_SNIPPETS_SUPPORT_EMAIL] is None:
+            return DEFAULT_MITO_CONFIG_CODE_SNIPPETS_SUPPORT_EMAIL
+        return self.mec[MITO_CONFIG_KEY_CODE_SNIPPETS_SUPPORT_EMAIL]
+
+    def get_code_snippets(self) -> Dict[str, Optional[str]]:
+        return {
+            MITO_CONFIG_KEY_CODE_SNIPPETS_VERSION: self.get_code_snippets_version(),
+            MITO_CONFIG_KEY_CODE_SNIPPETS_URL: self.get_code_snippets_url(),
+            MITO_CONFIG_KEY_CODE_SNIPPETS_SUPPORT_EMAIL: self.get_code_snippets_support_email(),
+        }
```
naterush, line 153: Let me try and be really specific here -- I think we've got back and forth a few times on this request. 

The type of this function should be Optional[Dict[str, str]]. That is, if all of the 3 variable are defined, then we get all of the objects back. If any of them are not defined, then we get None back. 

This allows us to enforce, in the type system, that the user is providing _all three_ required variables for the code snippets - and makes consumption of the config easier throughout the codebase!

mitosheet/mitosheet/tests/api/test_get_code_snippets.py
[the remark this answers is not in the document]
naterush: I'm fine with this, as long as it's a test dependency only!

naterush (commented):
Minor review, and a type cleanup. Then, LGTM.

mitosheet/mitosheet/enterprise/mito_config.py
naterush: I think this would be further improved by:
1. Removing the accessors for getting the version or URL individually (or at least underscoring them to make them private). Again, relying more on the type system here. 
2. Using a TypedDict (see types.py) -- and actually being clear what is in the dictonary. 

These two things together will result in a _very nice_ config API. I think having a nice config API is pretty important, as we will have a combinatorial explosion of configuration options eventually. Protecting ourselves from ourselves is going to be really important here!

mitosheet/mitosheet/enterprise/mito_config.py
```diff
@@ -166,18 +166,21 @@ def get_code_snippets(self) -> Optional[CodeSnippetEnvVarsDefined]:
                 "The code snippet environment variables are configured improperly. The MITO_CONFIG_CODE_SNIPPETS_VERSION environment variable is set, but the MITO_CONFIG_CODE_SNIPPETS_URL environment variable is not set."
             )  
 
-        return {
-            MITO_CONFIG_KEY_CODE_SNIPPETS_VERSION: code_snippets_version,
-            MITO_CONFIG_KEY_CODE_SNIPPETS_URL: code_snippets_url, 
-            MITO_CONFIG_KEY_CODE_SNIPPETS_SUPPORT_EMAIL: code_snippets_support_email
-        }
+        # Note: Since typed dicts aren't smart enough to recognize constants are the correct key, 
+        # instead of using constants, we rely on type help from the TypedDict for the code snippets
```
aarondr77, line 170: Remove

PR #575, opened 2022-12-19
mitosheet: fix bug with quotes

naterush:
# Description

- This fixes a bug where data with ` in it would result in no JS running properly.
- In testing, found a segmentation fault in df.to_json, which was patched here: https://github.com/pandas-dev/pandas/issues/50307

# Testing

See new test.

# Documentation

No.

### DIFF
--- a/mitosheet/mitosheet/mito_backend.py
+++ b/mitosheet/mitosheet/mito_backend.py
@@ -114,7 +114,7 @@ class MitoBackend():
             'shouldUpgradeMitosheet': self.should_upgrade_mitosheet,
             'numUsages': self.num_usages,
             'mitoConfig': self.mito_config.get_mito_config()
-        })
+        }).replace('\\', '\\\\')
 
 
     def handle_edit_event(self, event: Dict[str, Any]) -> None:
@@ -323,9 +323,9 @@ def get_mito_frontend_code(kernel_id: str, comm_target_id: str, div_id: str, mit
     js_code = js_code.replace('REPLACE_THIS_WITH_COMM_TARGET_ID', comm_target_id)
     # NOTE: we need to turn \ into \\ anywhere it exists, as otherwise we might get characters
     # that are invalid JS, and we get a SyntaxError when trying to render the mitosheet
-    js_code = js_code.replace('REPLACE_THIS_WITH_SHEET_DATA_ARRAY', mito_backend.steps_manager.sheet_data_json.replace('\\', '\\\\'))
-    js_code = js_code.replace('REPLACE_THIS_WITH_ANALYSIS_DATA', mito_backend.steps_manager.analysis_data_json.replace('\\', '\\\\'))
-    js_code = js_code.replace('REPLACE_THIS_WITH_USER_PROFILE', mito_backend.get_user_profile_json().replace('\\', '\\\\'))
+    js_code = js_code.replace('REPLACE_THIS_WITH_SHEET_DATA_ARRAY', mito_backend.steps_manager.sheet_data_json)
+    js_code = js_code.replace('REPLACE_THIS_WITH_ANALYSIS_DATA', mito_backend.steps_manager.analysis_data_json)
+    js_code = js_code.replace('REPLACE_THIS_WITH_USER_PROFILE', mito_backend.get_user_profile_json())
     js_code = js_code.replace('REPLACE_THIS_WITH_CSS', css_code_from_file)
     return js_code
 
--- a/mitosheet/mitosheet/mito_backend.py
+++ b/mitosheet/mitosheet/mito_backend.py
@@ -14,6 +14,7 @@ import time
 from sysconfig import get_python_version
 from typing import Any, Dict, List, Optional, Union
 
+import numpy as np
 import pandas as pd
 from ipykernel import get_connection_file
 from ipykernel.comm import Comm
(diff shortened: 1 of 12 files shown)


aarondr77:
I am able to import data with "`" in it, but about 100 pytests are failing when I run this locally. Once those tests are passing then LGTM.

naterush:
# Mini Spec

There are characters that are unsupported in Javascript JSON - specifically \t, \r, etc. This leads to parsing errors. We can try and find them in the string and replace them, but consider the following examples:
```
import pandas as pd
df = pd.DataFrame({'Should Escape': ['\t'], 'Should Not Escape': ['\\this']})
```
You want to additional escape some of the `\t`'s you get, but not others. Figuring out which is a hassle. We'd like to avoid all of this hackery and just pass the actual values to the frontend.

To do this: we're going to encode the string into bytes. I did this above.

To understand what encodings/decodings are, please read this article: https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/

It's epic. And made it clear to me that encoding a string to a byte array is actually not at all complicated, and is pretty easy!

mitosheet/mitosheet/mito_backend.py
```diff
@@ -321,12 +322,19 @@ def get_mito_frontend_code(kernel_id: str, comm_target_id: str, div_id: str, mit
     js_code = js_code_from_file.replace('REPLACE_THIS_WITH_DIV_ID', div_id)
     js_code = js_code.replace('REPLACE_THIS_WITH_KERNEL_ID', kernel_id)
     js_code = js_code.replace('REPLACE_THIS_WITH_COMM_TARGET_ID', comm_target_id)
-    # NOTE: we need to turn \ into \\ anywhere it exists, as otherwise we might get characters
-    # that are invalid JS, and we get a SyntaxError when trying to render the mitosheet
-    js_code = js_code.replace('REPLACE_THIS_WITH_SHEET_DATA_ARRAY', mito_backend.steps_manager.sheet_data_json.replace('\\', '\\\\'))
-    js_code = js_code.replace('REPLACE_THIS_WITH_ANALYSIS_DATA', mito_backend.steps_manager.analysis_data_json.replace('\\', '\\\\'))
-    js_code = js_code.replace('REPLACE_THIS_WITH_USER_PROFILE', mito_backend.get_user_profile_json().replace('\\', '\\\\'))
     js_code = js_code.replace('REPLACE_THIS_WITH_CSS', css_code_from_file)
+    # NOTE: we encode these as utf8 encoded byte arrays, so that we can avoid having to do complicated things with
```
naterush, line 326: Explain better

mitosheet/mitosheet/mito_backend.py
```diff
@@ -321,12 +322,19 @@ def get_mito_frontend_code(kernel_id: str, comm_target_id: str, div_id: str, mit
...
+    # NOTE: we encode these as utf8 encoded byte arrays, so that we can avoid having to do complicated things with 
+    # replacing \t, etc
+    def to_uint8_arr(string: str):
+        return np.frombuffer(string.encode("utf8"), dtype=np.uint8).tolist()
+    
+    js_code = js_code.replace('sheetDataBytes = new Uint8Array([]);', f'sheetDataBytes = new Uint8Array({to_uint8_arr(mito_backend.steps_manager.sheet_data_json)});')
+    js_code = js_code.replace('analysisDataBytes = new Uint8Array([]);', f'analysisDataBytes = new Uint8Array({to_uint8_arr(mito_backend.steps_manager.analysis_data_json)});')
+    js_code = js_code.replace('userProfileBytes = new Uint8Array([]);', f'userProfileBytes = new Uint8Array({to_uint8_arr(mito_backend.get_user_profile_json())});')
+
+    with open('out.txt', 'w+') as f:
```
naterush, line 335: Remove

naterush (commented):
Cleanup!

aarondr77:
Sweet, that Joel article is very helpful! Thanks for sharing. 

So I did a test with some other scary looking strings and a lot of them ended up rendering as invalid characters or not at all in Mito. They don't crash the sheet, but they are just invalid in our encoding, which is expected, I suppose. However, interestingly some of the strings are handled differently if I just print them out in Jupyter, like `'\this\is\text'`
```
import mitosheet
import pandas as pd
#df = pd.DataFrame({'A': ['`', "'", "'", '/this/is/text', '\this\is\text', '\this\ is\ text', '//this//is//text', '`this is text`', '&apos;', "\f{123}", '\r', '\f', '\fr', '<\a>']})
#mitosheet.sheet(df, analysis_to_replay="id-llihbcqvld")
```

Screenshot 2022-12-20 at 9 44 53 AM

Screenshot 2022-12-20 at 10 00 53 AM

naterush:
Screen Shot 2022-12-20 at 10 31 59 AM

This is on an older version of Mito from August. The text displays exactly the same. So not a regression, aka I good with this for now!

mitosheet/src/jupyterRender.tsx
```diff
@@ -3,16 +3,20 @@ import ReactDOM from 'react-dom';
...
-// We replace all of the following REPLACE_THIS_WITH... variables with 
-// the actual values they should be upon creation of the mitosheet. This is
-// how we get data needed at initialization from the mito backend to the 
-// frontend
+// We replace the following byte arrays with the real byte arrays of the utf8 encoded
+// JSON for the sheet data array, etc. We pass this encoded because the JSON parsing
+// when we don't gets really complicated trying to replace \t, etc
+const sheetDataBytes = new Uint8Array([]);
+const analysisDataBytes = new Uint8Array([]);
+const userProfileBytes = new Uint8Array([]);
```
aarondr77, line 11: Just to make this more clear, could we add a comment that says like "Do not make any edits to the following three lines since we are searching for this exact text in the mito_backend to replace it with the actual data if it exists.

aarondr77 (commented):
Other than that, once tests pass LGTM

PR #579, opened 2022-12-20
mitoinstaller: create option for turning off automatic jupyter launch

aarondr77:
# Description

Creates an optional argument to the installer that prevents the installer from automatically launching Jupyter at the end of the process, as this is breaking some users's startup scripts. https://github.com/mito-ds/monorepo/issues/578

# Testing

Run installer locally with `python3 -m mitoinstaller install --turn-off-jupyter-launch` and make sure Jupyter does not start

@naterush is there an easy way to actually test this functionality. My test leaves room to be desired. 

# Documentation

Yes. We could add documentation for this and --user options

### DIFF
--- a/mitoinstaller/mitoinstaller/installer_steps/final_installer_steps.py
+++ b/mitoinstaller/mitoinstaller/installer_steps/final_installer_steps.py
@@ -25,6 +25,10 @@ def replace_process_with_jupyter():
     """
     if is_running_test():
         return
+
+    if ('--turn-off-jupyter-launch' in sys.argv):
+        return 
+        
     # Get the prefered jupyter to launch, which we saved before
     prefered_jupyter = get_prefered_jupyter_env_variable()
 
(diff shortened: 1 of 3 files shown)


mitoinstaller/mitoinstaller/installer_steps/final_installer_steps.py
```diff
@@ -25,6 +24,11 @@ def replace_process_with_jupyter():
     """
     if is_running_test():
         return
+
+    if ('--turn-off-jupyter-launch' in sys.argv):
```
naterush, line 28: Can we name this `--no-jupyter-launch`? Turn off makes it seem like it _sets_ something, rather than just not launching this time.

Otherwise, LGTM.

naterush (commented):
minor name change requested, then LGTM!

naterush:
No way to test that is easy, I think. But this is ok for now.

PR #584, opened 2022-12-21
Snowflake import backend

naterush:
# Description

- Adds the `validate_snowflake_credentials` and `get_available_snowflake_options_and_defaults` api calls. Adds the `snowflake_import` step performer. 
- Adds the Snowflake-connector-python package to our test dependencies. It is only available on or after Python 3.7, so we take special care to not break out Python 3.6 tests. 
- Creates new GitHub secrets for connecting to our Snowflake testing instance, updates our Github workflows to use those secrets in our testing.  

# Testing

This is supposed to just be a backend PR, so see tests. 

Last week I invited you to join our testing snowflake environment. You should be able to add those credentials to a .env file and then use those credentials for testing. Otherwise, give me a call and I can give you the Pytest credentials. 

# Documentation

Not yet.


mitosheet/mitosheet/api/get_snowflake_connection.py
```diff
@@ -0,0 +1,160 @@
...
+
+        warehouse = _warehouse if _warehouse is not None else get_default_warehouse(ctx)
+        print('warehouse: ', warehouse)
+
+
+        # NOTE: These checks are not correct. This only works the first time. Consider this example.
+        # You validate your credentials and it gets all the default values and sets them in the query. 
+        # You then switch the database. The connection.schema is now still defined, but it does not belong to the 
+        # new database. So instead of just checking that its not none, we need to recognize that the database changed and
+        # then refresh everything beneath that.
```
aarondr77, line 60: This is out of date.

mitosheet/mitosheet/api/get_snowflake_columns.py
```diff
@@ -0,0 +1,20 @@
...
+
+def get_snowflake_columns(params: Dict[str, Any], steps_manager: StepsManagerType) -> str:
+        warehouse: str = params['warehouse']    
+        database: str = params['database']    
+        schema: str = params['schema']    
+        table: str = params['table']
+        
+        # TODO: implement this
+
+        return json.dumps(['abc', '123'])
```
aarondr77, line 20: Delete this

mitosheet/src/components/taskpanes/SplitTextToColumns/SplitTextToColumnsTaskpane.tsx
```diff
@@ -94,6 +94,7 @@ const SplitTextToColumnsTaskpane = (props: SplitTextToColumnsTaskpaneProps): JSX
     }, [props.startingColumnID])
 
     useEffect(() => {
+        console.log("USE EFFECT RAN")
```
aarondr77, line 97: remove

aarondr77:
Note: For the purposes of this PR I set up a testing Snowflake instance, it has only the default getting started data in it. I will delete that instance after this PR and we can set up a new instance for anything else we want to do because I accidentally committed my credentials.

mitosheet/mitosheet/api/get_available_snowflake_options_and_defaults.py
```diff
@@ -0,0 +1,156 @@
...
+    SNOWFLAKE_CONNECTOR_IMPORTED = True
+except ImportError:
+    SNOWFLAKE_CONNECTOR_IMPORTED = False
+
+# Load the .env file so we can access our pytest, read-only snowflake credentials
+load_dotenv()
+
+PYTEST_SNOWFLAKE_USERNAME = os.getenv('PYTEST_SNOWFLAKE_USERNAME')
+PYTEST_SNOWFLAKE_PASSWORD = os.getenv('PYTEST_SNOWFLAKE_PASSWORD')
+PYTEST_SNOWFLAKE_ACCOUNT = os.getenv('PYTEST_SNOWFLAKE_ACCOUNT')
```
aarondr77, line 24: This just makes development easy, but we should remove it

mitosheet/mitosheet/api/get_available_snowflake_options_and_defaults.py
```diff
@@ -0,0 +1,156 @@
...
+from typing import Any, Dict, List, Optional, Tuple
+from mitosheet.types import SnowflakeConnection, SnowflakeCredentials, SnowflakeImportParams, SnowflakeQueryParams, StepsManagerType
+import os
+from dotenv import load_dotenv
+
+try:
+    import snowflake.connector
+    SNOWFLAKE_CONNECTOR_IMPORTED = True
+except ImportError:
+    SNOWFLAKE_CONNECTOR_IMPORTED = False
```
aarondr77, line 17: explain why we do this

mitosheet/mitosheet/api/validate_snowflake_credentials.py
```diff
@@ -0,0 +1,65 @@
...
+from mitosheet.types import SnowflakeCredentials, StepsManagerType
+import os
+from dotenv import load_dotenv
+
+try:
+    import snowflake.connector
+    SNOWFLAKE_CONNECTOR_IMPORTED = True
+except ImportError:
+    SNOWFLAKE_CONNECTOR_IMPORTED = False
+
```
aarondr77, line 18: explain this

mitosheet/mitosheet/api/validate_snowflake_credentials.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+# Load the .env file so we can access our pytest, read-only snowflake credentials
+load_dotenv()
+
+PYTEST_SNOWFLAKE_USERNAME = os.getenv('PYTEST_SNOWFLAKE_USERNAME')
+PYTEST_SNOWFLAKE_PASSWORD = os.getenv('PYTEST_SNOWFLAKE_PASSWORD')
+PYTEST_SNOWFLAKE_ACCOUNT = os.getenv('PYTEST_SNOWFLAKE_ACCOUNT')
```
aarondr77, line 26: Remove this

mitosheet/mitosheet/api/validate_snowflake_credentials.py
```diff
@@ -0,0 +1,65 @@
...
+        return json.dumps({
+            'type': 'error',    
+            'error_message': 'snowflake-connector-python not accessible. Ensure it is installed.'
+        })
+
+    username = params['username']
+    password = params['password']
+    account = params['account']
+
+    username, password, account = PYTEST_SNOWFLAKE_USERNAME, PYTEST_SNOWFLAKE_PASSWORD, PYTEST_SNOWFLAKE_ACCOUNT # type: ignore
```
aarondr77, line 52: Remove

mitosheet/mitosheet/tests/api/test_get_available_snowflake_options_and_defaults.py
```diff
@@ -0,0 +1,130 @@
...
+                'schema': S_1,
+                'table': T_1,
+        },
+        'query_params': {
+                'columns': [C_1, C_2],
+                'limit': None 
+        }
+})
+
+    assert expected_return == response
```
aarondr77, line 129: Add an integration test here

mitosheet/mitosheet/errors.py
```diff
@@ -507,6 +507,19 @@ def make_invalid_simple_import_error(error_modal: bool=False) -> MitoError:
...
+def make_invalid_snowflake_import_error() -> MitoError:
+    """
+    Helper function for creating an invalid_snowflake_import_error.
+
+    Occures when the user tries to simple import and it fails.
+    """
+    return MitoError(
+        'invalid_snowflake_import_error',
+        'Cannot Query Database',
+        'We were unable to execute that query. Reach out to support for help.',
```
aarondr77, line 519: When we do more testing, let's figure out if the error message that Snowflake returns is helpful and maybe display that to the user here.

mitosheet/mitosheet/step_performers/snowflake_import.py
```diff
@@ -62,28 +66,29 @@ def execute(cls, prev_state: State, params: SnowflakeImportParams) -> Tuple[Stat
         username = credentials['username']
         password = credentials['password']
         account = credentials['account']
+        table = connection['table']
 
         # TODO: Remove before mering into dev
-        username, password, account = PYTEST_SNOWFLAKE_USERNAME, PYTEST_SNOWFLAKE_PASSWORD, PYTEST_SNOWFLAKE_ACCOUNT # type: ignore
+        # username, password, account = PYTEST_SNOWFLAKE_USERNAME, PYTEST_SNOWFLAKE_PASSWORD, PYTEST_SNOWFLAKE_ACCOUNT # type: ignore
```
aarondr77: Do this before merging

mitosheet/mitosheet/api/get_available_snowflake_options_and_defaults.py
```diff
@@ -0,0 +1,153 @@
...
+
+        credentials: SnowflakeCredentials = params['credentials']
+        connection: SnowflakeConnection = params['connection']
+
+        username = credentials['username']
+        password = credentials['password']
+        account = credentials['account']
+
+        # TODO: Remove before merging into dev
+        username, password, account = PYTEST_SNOWFLAKE_USERNAME, PYTEST_SNOWFLAKE_PASSWORD, PYTEST_SNOWFLAKE_ACCOUNT # type: ignore
```
aarondr77, line 52: Do this

mitosheet/mitosheet/api/validate_snowflake_credentials.py
```diff
@@ -0,0 +1,67 @@
...
+            'type': 'error',    
+            'error_message': 'snowflake-connector-python not accessible. Ensure it is installed.'
+        })
+
+    username = params['username']
+    password = params['password']
+    account = params['account']
+
+    # TODO: Remove before mering into dev
+    username, password, account = PYTEST_SNOWFLAKE_USERNAME, PYTEST_SNOWFLAKE_PASSWORD, PYTEST_SNOWFLAKE_ACCOUNT # type: ignore
```
aarondr77, line 54: Do this

mitosheet/mitosheet/code_chunks/snowflakeimport_code_chunk.py
```diff
@@ -0,0 +1,61 @@
...
+            'ctx = snowflake.connector.connect(',
+            f'{TAB}user=\'{username}\',',
+            f'{TAB}password=\'{password}\',',
+            f'{TAB}account=\'{account}\',',
+            f'{TAB}warehouse=\'{warehouse}\',',
+            f'{TAB}database=\'{database}\',',
+            f'{TAB}schema=\'{schema}\',',
+            ')',
+            '',
+
```
aarondr77, line 49: remove this line

mitosheet/mitosheet/step_performers/import_steps/snowflake_import.py
```diff
@@ -0,0 +1,132 @@
...
+        # TODO: We don't want to send these credentials because then they get saved to the analysis!!
+        # Instead, we can save them in the api call and read them in here. That is a weird flow because it 
+        # means the step has dependencies on the api call and the order of operations in a way we've never had before. 
+        # When the user reruns the analysis, if there is a snowflake_import, we need to have them reautnethicate first (?)
+        credentials: SnowflakeCredentials = get_param(params, 'credentials') 
+        connection: SnowflakeConnection = get_param(params, 'connection')
+        query_params: SnowflakeQueryParams = get_param(params, 'query_params')
+
+        # We make a new state to modify it
+        post_state = prev_state.copy() # TODO: update the deep copies
```
aarondr77, line 63: Remove the todo

mitosheet/mitosheet/step_performers/import_steps/snowflake_import.py
```diff
@@ -0,0 +1,132 @@
...
+
+        pandas_start_time = perf_counter()
+
+        username = credentials['username']
+        password = credentials['password']
+        account = credentials['account']
+        table = connection['table']
+
+        # TODO: Remove before mering into dev
+        # username, password, account = PYTEST_SNOWFLAKE_USERNAME, PYTEST_SNOWFLAKE_PASSWORD, PYTEST_SNOWFLAKE_ACCOUNT # type: ignore
```
aarondr77, line 73: Do this

mitosheet/setup.py
```diff
@@ -119,8 +119,13 @@
         extras_require = {
             'test': [
                 'pytest',
+                'pytest-mock',
```
naterush, line 122: We pay much of the cost of dependencies in terms of long-term complexity and maintaince, even when we only take them on as test dependencies. Moreover, I do not think that we need a full additional library to mock functions - there is Python built-in support for it.

Going forward, let's put _all_ dependencies through our dependency review process, even if they are just test/dev dependencies. For this one, I think writing this ourselves is <25 lines of code -- so let's def just do that!

Can you look into mocking functions without taking on additional dependencies here? You might be able to just use the unittest.mock.patch function, or potentially the `@patch` decorator. In either case, feel free to build some testing utilities that make it easier to mock in the future (potentially with the same interface as the current library you're using).

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -0,0 +1,478 @@
...
+    schema: string | undefined,
+    table: string | undefined
+}
+
+export type SnowflakeConnectionAPIReturn = {
+    warehouse: string | null, 
+    database: string | null, 
+    schema: string | null,
+    table: string | null
+}
```
naterush, line 49: Gross! Why don't we just make `SnowflakeConnection` (which we should rename) have `warehouse: string | undefined | null`?

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -0,0 +1,478 @@
...
+
+export type SnowflakeQueryParams = {
+	columns: string[],
```
naterush, line 52: Does our system work with weird column names? E.g. numbers or booleans?

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -0,0 +1,478 @@
...
+            StepType.SnowflakeImport, 
+            props.mitoAPI,
+            props.analysisData,
+    )
+    
+    const [openCredentialsSection, setOpenCredentialsSection] = useState(true);
+    const [availableSnowflakeOptionsAndDefaults, setAvailableSnowflakeOptionsAndDefaults] = useState<AvailableSnowflakeOptionsAndDefaults | undefined>(undefined);
+    const [liveUpdateNumber, setLiveUpdateNumber] = useState(0) 
+
+    const getAvailableOptionsAndDefaults = (newParams: SnowflakeImportParams): void => {
```
naterush, line 105: This feels like totally the wrong name for this function - it's not getting anything

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -0,0 +1,478 @@
...
+
+    //loading, edit, editApplied
+    const {params, setParams, edit} = useSendEditOnClick<SnowflakeImportParams, undefined>(
+            () => getDefaultParams(),
+            StepType.SnowflakeImport, 
+            props.mitoAPI,
+            props.analysisData,
+    )
+    
+    const [openCredentialsSection, setOpenCredentialsSection] = useState(true);
```
naterush, line 101: `openCredentialsSection` sounds like it's saying _which_ credentialsSection is open (aka, that there's multiple). A more clear name would be `credentialsSectionIsOpen` I think!

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -0,0 +1,478 @@
...
+    const [availableSnowflakeOptionsAndDefaults, setAvailableSnowflakeOptionsAndDefaults] = useState<AvailableSnowflakeOptionsAndDefaults | undefined>(undefined);
+    const [liveUpdateNumber, setLiveUpdateNumber] = useState(0)
```
naterush, line 103: We should probably remove all mention of live updating, given that this is not a live updating taskpane!

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -0,0 +1,478 @@
...
+
+    useEffect(() => {
+        if (liveUpdateNumber === 0) {
+            // Don't run on first render
+            return
+        }
+        void _getAvailableOptionsAndDefaults()
+    }, [liveUpdateNumber])
+
+    const _getAvailableOptionsAndDefaults = async () => {
```
naterush, line 118: I think this `get` function should also not set params -- that's a weird thing for a function to do!

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -0,0 +1,478 @@
...
+        if (params === undefined) {
+            // We don't expect this to ever happen because of the UI restrictions
+            return 
+        }
+        const availableSnowflakeOptionsAndDefaults = await props.mitoAPI.getAvailableSnowflakeOptionsAndDefaults(params.credentials, params.connection);
+
+        setAvailableSnowflakeOptionsAndDefaults(availableSnowflakeOptionsAndDefaults);
```
naterush, line 125: Setting here is weird

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -0,0 +1,478 @@
...
+
+        if (availableSnowflakeOptionsAndDefaults?.type === 'success') {
+            setParams((prevParams) => {
```
naterush, line 128: Setting here is also weird

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
naterush: Perhaps this function should be named `updateAvailableOptionsAndDefaults`?

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -0,0 +1,478 @@
...
+                                onChange={(newDatabase) => {
+                                    if (newDatabase === params['connection']['database']) {
+                                        return 
+                                    }
+                                    
+                                    const paramsCopy: SnowflakeImportParams = {...params}
+                                    const newParams = {
+                                        ...paramsCopy, 
+                                        'connection': {
+                                            ...paramsCopy.connection,
```
naterush, line 305: There is a huge chunk of duplicated code between the database, schema, and table updaters. Let's write a little helper that at least takes a db, option takes a schema, table, etc -- and then sets in the obvious way!

There's generally a lot of code in this taskpane that I think can be refactored into well-named helper functions - and I think we should do this so it's not a 400 line monsta

naterush (commented):
The tests you suggest sound great to me! I'd add a database with weird column names (e.g. spaces, or numbers, or weird characters).

The main things here:
1. let's make the relevant tests get skipped if you don't have the dependencies or passwords (all tests should pass, even if you don't work for Mito).
2. let's remove this new pytest-mock dependency. We can write the functionality we need in <25 lines of code with the Python standard library - let's not pick up dependencies for no reason!
3. Other minor renaming and cleanup.

The backend looks mostly great, and I'm very happy with how the architecture ended up. It looks very clean and extensible going forward!

I left some comments on the frontend as well, although I know that's not the focus of this PR. 

Ping me when it's ready for a final review :)

naterush:
Oh also make sure all tests are passing (including lints)

mitosheet/src/components/import/XLSXImportConfigScreen.tsx
```diff
@@ -142,6 +142,7 @@ function XLSXImportConfigScreen(props: XLSXImportConfigScreenProps): JSX.Element
                             searchable
                             height='medium'
                             toggleAllIndexes={(indexesToToggle) => {
+                                // TODO: this is broken. The toggle all is not working properly!
```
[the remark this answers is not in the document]
aarondr77: Opened issue here: https://github.com/mito-ds/monorepo/issues/598

.github/workflows/test-mitosheet-python.yml
```diff
@@ -48,6 +48,14 @@ jobs:
         flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
         # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: 'Create env file'
```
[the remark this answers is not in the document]
aarondr77: Yeah, we need the username, password and account for our Pytest snowflake instance in order to run the integration tests. This is the only way I found to create the credentials in a private way. What do you think?

mitosheet/mitosheet/api/get_available_snowflake_options_and_defaults.py
```diff
@@ -0,0 +1,153 @@
...
+        databases = cur.fetchall()
+        return [db[1] for db in databases]
+
+def get_schemas(ctx: Any, database: Optional[str]) -> List[str]:
+        # List all of the schemas in a particular database available to the user
+        cur = ctx.cursor().execute(f'SHOW SCHEMAS in {database}')
+        schemas = cur.fetchall()
+        return [s[1] for s in schemas]
+
+def get_tables(ctx: Any, database: Optional[str], schema: Optional[str]) -> List[str]:
```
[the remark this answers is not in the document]
aarondr77: I can look. I tried finding a types package for the connection, but I couldn't find one.

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
aarondr77: If you can help me understand why this is weird, I'm happy to apply those learnings and make a change.

aarondr77:
Adds these tests:

- [x] A database that has now rows in it
- [x] Invalid credentials 
- [x] Invalid table name 
- [x] Different types of columns headers 
- [x] Different types of columns

aarondr77:
@naterush can you take a look at the failing pytests. Each of them are failing for the same reason -- not correctly capturing the MitoError, but I've tried for a few hours to figure out what is wrong and I can't find it.

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
naterush: As above, function is misnamed.

naterush:
Screen Shot 2023-01-18 at 1 39 08 PM

I don't have snowflake installed currently, and when I try and run Mito this is the error I get. Mito should work 100% when snowflake is not installed (with the exception of that feature, ofc).

naterush:
^ speaking of which, can we somehow make a test run that doesn't use snowflake? So we know things still work without it?

naterush:
Screen Shot 2023-01-18 at 1 45 22 PM

Same with dotenv. **I do not think we need this dependency either, similar to pytest-mock.** It's also about 5 lines of code..

naterush:
In fact, why did we move to having a .env file at all? I'd prefer to don't make this a dependency or rely on it for a few reasons:
1. **We don't know if our users have .env files they use for other reasons.** We really should avoid _writing_ environment variables we don't have control over.
2. We don't need it. Just `export VAR_NAME='whatever'` on the command line -- and be done. We can do this both locally and in our Github action.
3. It's another dependency that we didn't evaluate (e.g. leading to everything from silly bugs like me not being able to run it to worse things we know not yet). 

Really, (1) is the most immediate one -- we simply don't know how this will interact with our current and future users code (many of which we know use something _similar_ to .env files -- something like environmental variable setting). I'd like to avoid any additional integration complexity here entirely!

naterush:
Just to make this really explicit: in the future, if you're finding your self pip or npm installing something, that's something that 100% should be coming from the specs! If it isn't from the spec, probably send a message about it, so we can evaluate it properly.

naterush:
Also, what is the dotenv library? I tried `pip install dotenv` and it failed. It's also not a maintained package, by the looks of it.

naterush:
Getting this weird error right off the bat:
Screen Shot 2023-01-18 at 2 04 26 PM
/213305471-3d01c0c7-4373-4bd4-a331-933d6ebf56d4.png">

I def do have pandas, but got this error. I followed it's debugging instructions (which I swore I already ran) and go this issue. Not sure why!

naterush:
# The error not being caught

The issue here is that we catch all errors that bubble up through the `mito_backend.receive_message` function! One really instructive way to figure this out is to go looking through the codebase where we have the line `with pytest.raises(MitoError) as e_info` -- and I found only 1 that sends a message. Specifically: `test_steps_manager.py`.

You can see that it has to call the underlying function (with the message) directly on the steps manager. This allows it to detect the error. 

I did a simple test updating to this for one of the tests, and it looked like it worked.

naterush:
The snowflake tests are _super_ slow. Running them takes `28.369` seconds, which is about as long as the rest of the tests combined (33.130 seconds). 

I generated performance data with https://pypi.org/project/pytest-profiling/ this package, following instructions. The output:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000   11.425   11.425 runner.py:109(pytest_runtest_protocol)
        1    0.000    0.000   11.424   11.424 runner.py:117(runtestprotocol)
        3    0.000    0.000   11.424    3.808 runner.py:217(call_and_report)
    17/11    0.000    0.000   11.424    1.039 _hooks.py:244(__call__)
    17/11    0.000    0.000   11.424    1.039 _manager.py:77(_hookexec)
    17/11    0.000    0.000   11.424    1.039 _callers.py:9(_multicall)
        3    0.000    0.000   11.423    3.808 runner.py:245(call_runtest_hook)
        3    0.000    0.000   11.423    3.808 runner.py:316(from_call)
        3    0.000    0.000   11.423    3.808 runner.py:260(<lambda>)
        1    0.000    0.000   11.344   11.344 runner.py:158(pytest_runtest_call)
        1    0.000    0.000   11.344   11.344 python.py:1787(runtest)
        1    0.000    0.000   11.344   11.344 python.py:188(pytest_pyfunc_call)
        1    0.000    0.000   11.344   11.344 test_snowflake_import.py:50(test_snowflake_import_integration)
        1    0.000    0.000   11.342   11.342 test_utils.py:29(wrapper)
        8    0.000    0.000   10.496    1.312 network.py:685(_post_request)
        8    0.000    0.000   10.495    1.312 network.py:759(fetch)
        8    0.000    0.000   10.494    1.312 network.py:821(_request_exec_wrapper)
        8    0.000    0.000   10.493    1.312 network.py:995(_request_exec)
        8    0.000    0.000   10.489    1.311 sessions.py:500(request)
        8    0.000    0.000   10.349    1.294 sessions.py:671(send)
```
It looks like it's the connection to snowflake! Which is good to know... hmm... I don't love that every time we run tests, we have to wait for a minute, rather than the previous 30 seconds. That's real annoying. Let me think for a bit.

naterush (commented):
Good work keeping this pushing! See my comments above errors (and let me know if there's anything else you're wondering about). Sorry you hit your head on that -- it's not a pleasant wart of our codebase certainly.

There's a few things I think we should do here before we merge this in that are listed in the review above. I don't think any of them are particularly hard fixups (probably less than 10 minutes on most, less than 1 hour on all) -- but feel free to give me a call and we can chat about what you're thinking if you have any questions! 

Performance-wise, I think most of the testing performance issues will be rectified if we reduce the number of calls we're doing the to database! 99.999% of the time, calls across a network are the bottleneck!

## On Splitting Out Snowflake Test

Given that there were bugs that manifested because we didn't have snowflake installed, I think we probably should ensure this doesn't happen again (especially b/c this is most of our users). 

This is, of course, the cost of a configurable system. You get an explosion of the number of cases that you need to test.

We'd much rather the snowflake integration break than the rest of Mito - and so overall, I think probably the most sensible thing to do here is to make it so the rests of our tests run without the snowflake package installed. 

The easiest way I can think to do this is to:
1. Remove the snowflake connector from the test dependencies. 
2. Create a new option in our `test-mitosheet-python.yml` file (like where we run tests on a bunch of different versions of Python) - that is something like `extra_installs` or something. And it can 1) install extra packages, and 2) potentially set some environment variables.

(2) might be a bit tricky, so if it's a bit too far out of your wheel house to feel like good bang, I'm happy to take it over. I'd offer to pair programmer it but video calls hurt my head still :-)

mitosheet/mitosheet/api/get_available_snowflake_options_and_defaults.py
```diff
@@ -0,0 +1,162 @@
...
+        database = _database if _database is not None else get_default_database(con)
+        schema = _schema if _schema is not None else get_default_schema(con, database)
+        table = _table if _table is not None else get_default_table(con, database, schema)
+
+        warehouses = get_warehouses(con)
+        databases = get_databases(con)
+        schemas = get_schemas(con, database)
+        tables = get_tables(con, database, schema)
+        columns = get_columns(con, database, schema, table)
+
```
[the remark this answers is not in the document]
aarondr77: Yeah we can update to this for now. In the future, we might want to query the database to get the user's default database.

aarondr77:
One of the snowflake tables that we're using in testing returns a dataframe with int8 columns, which is not the default int64 columns that you get when creating a new dataframe. I'm not sure how this interacts with the rest of our type system. I think we should further test that in a second PR and fix it if required. But no sense in complicating this PR.

aarondr77:
### WE NEED TO REMOVE [ActionEnum.SNOWFLAKEIMPORT]: before merging. Will do as last step after you review

mitosheet/mitosheet/api/api.py
```diff
@@ -171,6 +173,10 @@ def handle_api_event(
             result = get_render_count(params, steps_manager)
         elif event["type"] == "get_code_snippets":
             result = get_code_snippets(params, steps_manager)
+        elif event["type"] == "get_available_snowflake_options_and_defaults":
+            result = get_available_snowflake_options_and_defaults(params, steps_manager)
+        elif event["type"] == "validate_snowflake_credentials":
+            resutl = validate_snowflake_credentials(params, steps_manager)
```
aarondr77, line 179: resutl -> result

mitosheet/mitosheet/api/get_available_snowflake_options_and_defaults.py
```diff
@@ -0,0 +1,165 @@
...
+# Copyright (c) Saga Inc.
+# Distributed under the terms of the GPL License.
+
+import json
+from typing import Any, Dict, List, Optional, Tuple
+from mitosheet.types import MitoSafeSnowflakeConnection, SnowflakeCredentials, SnowflakeTableLocationAndWarehouseOptional, StepsManagerType
+
+# The snowflake-connector-python package is only available in Python > 3.6 
+# and is not distributed with the mitosheet package, so we make sure to 
+# note assume that the import will succeed.
```
aarondr77, line 13: note -> not

mitosheet/mitosheet/api/validate_snowflake_credentials.py
```diff
@@ -0,0 +1,55 @@
...
+# Copyright (c) Saga Inc.
+# Distributed under the terms of the GPL License.
+
+import json
+from typing import Any
+from mitosheet.types import SnowflakeCredentials, StepsManagerType
+
+# The snowflake-connector-python package is only available in Python > 3.6 
+# and is not distributed with the mitosheet package, so we make sure to 
+# note assume that the import will succeed.
```
aarondr77, line 13: note -> not

mitosheet/mitosheet/step_performers/import_steps/snowflake_import.py
```diff
@@ -0,0 +1,149 @@
...
+
+from mitosheet.state import DATAFRAME_SOURCE_IMPORTED, State
+from mitosheet.step_performers.step_performer import StepPerformer
+from mitosheet.step_performers.utils import get_param
+from mitosheet.types import ColumnHeader, SnowflakeCredentials, SnowflakeQueryParams, SnowflakeTableLocationAndWarehouse
+from mitosheet.utils import get_valid_dataframe_name
+
+# The snowflake-connector-python package is only available in Python > 3.6 
+# and is not distributed with the mitosheet package, so we make sure to 
+# note assume that the import will succeed.
```
aarondr77, line 22: note -> not

mitosheet/mitosheet/types.py
```diff
@@ -69,6 +69,18 @@
 
 PivotColumnTransformation = str
 
+# If the user does not have the snowflake.connector python package installed,
+# we take extra care to make sure that our mypy typing will still pass even though
+# that code is not accessible to the user.
+try:
+    from snowflake.connector import SnowflakeConnection 
+except ImportError:
+    # This is prety hacky and errors if we get rid of the #type: ignore,
```
aarondr77, line 78: errors -> results in a mypy error

aarondr77:
Two callouts for you to pay special attention to:
1. One of the snowflake tables that we're using in testing returns a dataframe with int8 columns, which is not the default int64 columns that you get when creating a new dataframe. I'm not sure how this interacts with the rest of our type system. I think we should further test that in a second PR and fix it if required. But no sense in complicating this PR. See line 255 of `test_snowflake_import.py`

2. See the comment at the top of types.py about conditionally creating the SnowflakeConnection mypy. I'm not sure how else to create this type, but feel free to give it a shot if you want to.

naterush:
I think it should interact with our type system mostly fine!

.github/workflows/test-mitosheet-python.yml
```diff
@@ -40,6 +41,11 @@ jobs:
         python -m pip install --upgrade pip
         pip install "pandas==${{ matrix.pandas-version }}"
         pip install -e ".[test, deploy]"
+    - name: Install optional feature dependencies
+      if: ${{ matrix.optional_feature_installs }}
+      run: |
+        cd mitosheet
+        pip install -e ".[optional_feature_dependencies]"
```
naterush, line 48: Sweet, this is a great

mitosheet/mitosheet/code_chunks/snowflake_import_code_chunk.py
```diff
@@ -0,0 +1,48 @@
...
+
+# Copyright (c) Saga Inc.
+# Distributed under the terms of the GPL License.
+from typing import List, Tuple, Dict
+from mitosheet.code_chunks.code_chunk import CodeChunk
+from mitosheet.state import State
+from mitosheet.transpiler.transpile_utils import TAB, param_dict_to_code
+from mitosheet.types import ColumnHeader, SnowflakeCredentials, SnowflakeQueryParams, SnowflakeTableLocationAndWarehouse
+
+class SnowflakeImportCodeChunk(CodeChunk):
```
naterush, line 13: Wow, the cleanup on this file is epic. This is seriously great!

mitosheet/mitosheet/step_performers/import_steps/snowflake_import.py
```diff
@@ -0,0 +1,149 @@
...
+        params: Dict[str, Any],
+        execution_data: Optional[Dict[str, Any]],
+    ) -> List[CodeChunk]:
+
+        table_loc_and_warehouse: SnowflakeTableLocationAndWarehouse = get_param(params, 'table_loc_and_warehouse')
+        connection_params_dict = get_connection_param_dict(get_param(params, 'credentials'), table_loc_and_warehouse)
+
+        query_params: SnowflakeQueryParams = get_param(params, 'query_params')
+        table = table_loc_and_warehouse['table']
+        sql_query = create_query(table, query_params)
```
naterush, line 110: There's one final refactor here. You should:
1. Return these variables (namely, `connection_params_dict` and `sql_query`) in execution data above.
2. Get it form the execution data in this function.

This means you have to redo _zero_ work -- which again reduces the chances we'll make silly bugs and forget to change things.

mitosheet/mitosheet/step_performers/import_steps/snowflake_import.py
```diff
@@ -0,0 +1,149 @@
...
+    database = table_loc_and_warehouse['database']
+    schema = table_loc_and_warehouse['schema']
+
+    all_params: Dict[str, str] = {}
+    all_params['user'] = username
+    all_params['password'] = password
+    all_params['account'] = account
+    all_params['warehouse'] = warehouse
+    all_params['database'] = database
+    all_params['schema'] = schema
```
naterush, line 140: This can be cut in half.

mitosheet/mitosheet/tests/step_performers/import_steps/test_snowflake_import.py
```diff
@@ -0,0 +1,336 @@
...
+    assert mito.dfs[0].equals(input_df_1)
+    assert mito.dfs[1].equals(snowflake_df)
+
+    mito.delete_dataframe(1)
+    assert len(mito.dfs) == 1
+    assert mito.dfs[0].equals(input_df_1)
+
+    # Remove the test file
+    os.remove(TEST_FILE_PATHS[0])
+    os.remove(TEST_FILE_PATHS[1])
```
naterush, line 336: We should add an optimization test as well!

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -0,0 +1,441 @@
...
+            StepType.SnowflakeImport, 
+            props.mitoAPI,
+            props.analysisData,
+    )
+    
+    const [credentialsSectionIsOpen, setCredentialsSectionIsOpen] = useState(true);
+    const [availableSnowflakeOptionsAndDefaults, setAvailableSnowflakeOptionsAndDefaults] = useState<AvailableSnowflakeOptionsAndDefaults | undefined>(undefined);
+    const [refreshDefaultsNumber, setRefreshDefaultsNumber] = useState(0) 
+
+    const refreshAvailableOptionsAndDefaults = (newParams: SnowflakeImportParams): void => {
```
naterush, line 96: This, I think, should just a setParams call. And we should change the names of setParams above. Will cleanup this code a bit.

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -0,0 +1,441 @@
...
+    const validateSnowflakeCredentials = async () => {
+        if (params === undefined) {
+            // We don't expect this to ever happen because of the UI restrictions
+            return 
+        }
+        
+        return await props.mitoAPI.validateSnowflakeCredentials(params.credentials)
+    }
+
+    const getNewParams = (database?: string | null, schema?: string | null, table?: string | null) => {
```
naterush, line 146: Need not be defined inline

naterush (commented):
A few more things to cleanup, I'll hop on em

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -0,0 +1,441 @@
...
+                        <Col>
+                            <Input 
+                                value={params.credentials.password} 
+                                type='password'
+                                onChange={(e) => {
+                                    const newPassword = e.target.value;
+                                    setParams((prevParams) => {
+                                        return {
+                                            ...prevParams, 
+                                            credentials: {
```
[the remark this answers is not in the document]
naterush: Interestingly, I found a bug in the setting of the new warehouse while updating to this -- a wrong key was being used. The partial updater function is actually more type safe than standard react!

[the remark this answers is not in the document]
naterush: Also -- just a reminder that `const paramsCopy: SnowflakeImportParams = {...params}` is not an actual copy, if you have nested objects (which we certainly do here).

naterush:
I'm doing some refactoring of the frontend code. I'd love to talk through it at some point. There's a bunch of opportunity for learning, I think, namely around how to -factor- functions. What should be included and what shoudn't. 

E.g. setCredentialsSectionIsOpen function should be called within the validateSnowflakeCredentials section, etc.

naterush:
@aarondr77 ok, a few things to call out:
1. I moved to using `window.structuredClone` to create copies. Check that this works in a couple places -- but notably it is able to handle undefined values! It requires running the setup scripts again, as it upgrade packages.
2. Check out the frontend refactors. I know this isn't a frontend PR, but there were some very low hanging fruit that a) cut a bunch of code, b) caught some bugs, and c) generally leads to a much more followable frontend. 

In practice, I was able to remove the effect triggered by the update number entirely - and move to just making a function call. This makes things much simpler, and easier to follow - and better function names help a lot as well.

Let me know if you want to chat about it at all. If this looks good, let's a) remove the action, and b) merge this guy!

PR #592, opened 2023-01-10
Strong type code chunks

naterush:
# Description

This makes code types way more strongly typed. Rather than passing everything through a dictionary that then has no typing on it, we simply actually pass the parameters we need to the code chunks!

This is step (1) in making our code chunks more robust to optimization. Next, we're going to make it so there is a _strong_ distinction in which params can change and which can't. This is going to happen by moving to only using `post_state`, and then passing the `post_state` to the `get_code` function. 

Once we do this, it will be _dramatically_ easier to reorder code chunks while _also_ being sure that we're doing so correctly!

I did this b/c it was a nice, not super brain power heavy way of programming!

# Testing

See our tests.

# Documentation

No.


aarondr77:
Reviewed https://github.com/mito-ds/monorepo/pull/596

naterush:
@aarondr77 this proposal is very sensible -- in fact, typing the entire params at once was the first thing that I tried! However, there are two reasons that it doesn't really work out:

1. Python's Type System sucks. Namely, making a class that has subclasses parametric in a type is... extremely challenging. I know this sounds like it should be not that hard, but I tried and failed to get things to type check with mypy for about 4 hours. I was unable to find any documentation on this online. 

Then I realized: let's not overcomplicate the type system. Why do we need to wrap these things in a params dictionary? The (by far) easiest, simplest, and _reccomended_ way to get 

2. The params are _not_ the same for the code chunks and step performers and will increasingly diverge. This is a big one, but generally, we're going to continue to refine the code chunks so that the params are quite different than those in the step performers. E.g. go look at the Excel Range Import. It's totally different. The step performers params are for ease of operation of the step performer. The code chunk is for the code chunk (and has different concerns, like being optimized). 

So -- we don't want one type for both of them! This fact + (1) above means that this is far and away the most sensible way of typing code chunks -- and it will protect us from a lot of silly errors going forward!

In the future, we _can_ make it so: 1) the step performers have strongly typed params that are checked with the frontend for consistency, so that 2) we can pass these more safely automatically to the code chunk. I think this part of your proposal is extremely sensible!

naterush:
Fwiw, having to use a line of code for each param ain't so bad. It's simple code. The part that does suck is that it isn't type checked!

naterush:
All of the failing tests are on excel range import which... isn't in this PR. So I'm just gonna chalk it up to some weird things I accidentally did with git, and merge these. We'll fix it up when we merge in the excel range import stuff.

PR #594, opened 2023-01-13
mitosheet: add working frontend for excel range imports

naterush:
# Description

Implements the frontend for: https://www.notion.so/trymito/Excel-Range-Import-b95bd6054a3a4d62851f028632d72242

This works reasonable well. Two notes:
1. This works a little weird with updating, and overwriting. This is a function of how the useSendEditOnClick function works -- I think it's ok for now. It works well with undo and redo. But at some point, we probably want to make it less surprising. 
2. We don't allow users to update TO this import during an update imports (even if the original import was this type). This is ok for now -- a) because no one uses that feature, and b) because this feature is also going to be very unused. So let's just not worry about it.

# Testing

1. Get there with a single sheet selected in the Excel import. I think this is a _fine_ workaround for now, but we can def massively improve this flow in the future.
2. Then, import some ranges!

# Documentation

When we release this, we needs some docs for it!


mitosheet/src/components/elements/ExpandableContentCard.tsx
```diff
@@ -0,0 +1,98 @@
...
+    children: React.ReactNode,
+    
+    isExpanded: boolean;
+    setExpanded: (isExpanded: boolean) => void;
+    
+    onDelete: () => void;
+    
+}
+
+const ExpandableContentCard = (props: ExpandableContentCardProps): JSX.Element => {
```
naterush, line 27: Document this

mitosheet/src/components/import/XLSXImportConfigScreen.tsx
```diff
@@ -276,6 +277,24 @@ function XLSXImportConfigScreen(props: XLSXImportConfigScreenProps): JSX.Element
                             </Col>
                         </Row>
                     }
+                    {params.sheet_names.length === 1 &&
```
naterush, line 280: Comment this

naterush (commented):
Cleanup!

aarondr77:
I think it's quite likely that people are going to configure one range import, press the import button to see if it worked correctly, and then add more. That's what I did. To support that workflow, it would be nice if pressing the import button updated the previous import instead of importing the data again.

aarondr77:
Once you get to the range import screen from by clicking the link the the xlsx config screen, there is no way to get back to the xlsx config screen. There should be a back button, so users can, for example, change which sheet they are importing from.

aarondr77:
Instead of removing this entirely if the user has multiple sheets selected, I think it should be disabled with an explanation. 
Screenshot 2023-01-16 at 5 46 20 PM

aarondr77:
I think the Start:End part of this error message is not helpful because you can do either A1:Z1000 or Z10000:A1. So its really just the format of column:number that we're trying to communicate, right?

Screenshot 2023-01-16 at 5 48 51 PM

aarondr77:
If you don't add a dataframe name, then it creates the same dataframe over and over. `df_`

Screenshot 2023-01-16 at 5 51 07 PM

aarondr77:
> This works a little weird with updating, and overwriting. 

What do you mean by this? Which part do you think is weird?

aarondr77:
> We don't allow users to update TO this import during an update imports 

Sounds good for now!

aarondr77:
Code looks good + functionality is working well overall. Feel free to go through this review and take what you want. Then LGTM.

PR #595, opened 2023-01-15
Excel range import upper left

naterush:
# Description

# Testing

# Documentation


naterush (commented):
Looks good!

naterush:
Note the branch this is merged into.

mitosheet/mitosheet/step_performers/import_steps/excel_range_import.py
```diff
@@ -13,13 +13,17 @@
...
+from mitosheet.errors import make_upper_left_corner_value_not_found_error
+from mitosheet.excel_utils import get_column_from_column_index, get_col_and_row_indexes_from_range, get_table_range_from_upper_left_corner_value
 from mitosheet.state import DATAFRAME_SOURCE_IMPORTED, State
 from mitosheet.step_performers.step_performer import StepPerformer
 from mitosheet.step_performers.utils import get_param
 from mitosheet.types import ExcelRangeImport
 from mitosheet.utils import get_valid_dataframe_name
 
+EXCEL_RANGE_IMPORT_TYPE_RANGE = 'range'
+EXCEL_RANGE_IMPORT_TYPE_UPPER_LEFT = 'upper left corner value'
```
naterush, line 25: Remove

naterush (commented):
Looks good!

mitosheet/mitosheet/excel_utils.py
```diff
@@ -52,4 +54,69 @@ def get_index_from_column(col: str) -> int:
...
         if c in string.ascii_letters:
             num = num * 26 + (ord(c.upper()) - ord('A')) + 1
-    return num - 1
\ No newline at end of file
+    return num - 1
+
+
+def get_table_range_from_upper_left_corner_value(file_path: str, sheet_name: str, value: Union[str, int, float, bool]) -> Optional[str]:
+    """
+    Given a string, this fu ction will look through the excel tab sheet_name at the given
```
aarondr77, line 62: fu ction -> function

mitosheet/mitosheet/excel_utils.py
```diff
@@ -52,4 +54,69 @@ def get_index_from_column(col: str) -> int:
...
+    dimension = sheet.calculate_dimension()
+    (min_search_col, min_search_row), (max_search_col, max_search_row) = get_col_and_row_indexes_from_range(dimension)
+    # Unfortunately, openpyxl indexes from 1, so we add one to treat everything in that range
+    (min_search_col, min_search_row), (max_search_col, max_search_row) = (min_search_col + 1, min_search_row + 1), (max_search_col + 1, max_search_row + 1)
+
+    # Loop over the columns one by one to find where this value is set
+    min_found_col_index, min_found_row_index = None, None
+    for col in sheet.iter_cols(min_row=min_search_row, max_row=max_search_row, min_col=min_search_col, max_col=max_search_col):
+        for cell in col:
+            if cell.value == value:
```
aarondr77, line 84: When looking at frontend, make sure we tell the user that it has to be an exact match with the entire cell value.

mitosheet/mitosheet/excel_utils.py
```diff
@@ -52,4 +54,69 @@ def get_index_from_column(col: str) -> int:
...
+                min_found_col_index, min_found_row_index = cell.column, cell.row
+        
+        # As soon as we find something, stop looking
+        if min_found_col_index is not None or min_found_row_index is not None:
+            break
+
+    if min_found_col_index is None or min_found_row_index is None:
+        return None
+
+    column = sheet[get_column_from_column_index(min_found_col_index - 1)] # We need to subtract 1 as we 1 index
```
aarondr77, line 94: subtract 1 as we 1 index -> subtract 1 as we 0 index

mitosheet/mitosheet/excel_utils.py
```diff
@@ -52,4 +54,69 @@ def get_index_from_column(col: str) -> int:
...
+    # as the length of the entire column
+    if max_found_row_index is None:
+        max_found_row_index = len(column)
+
+    # Then we find find where the rows are defined to
+    max_found_col_index = None
+    for row in sheet.iter_rows(min_row=max_found_row_index - 1, max_row=max_found_row_index - 1, min_col=min_found_col_index):
+        for cell in row:
+            if cell.value is None:
+                max_found_col_index = cell.column - 1 # minus b/c this is one past the end
```
aarondr77, line 114: This seems like a sensible separate utility function that we could break out and then call here

mitosheet/mitosheet/tests/step_performers/import_steps/test_excel_range_import.py
```diff
@@ -102,6 +153,167 @@ def test_import_with_defined_name_works():
...
+    with pd.ExcelWriter(TEST_FILE_PATH) as writer:
+        for r, df in zip(ranges, dfs):
+            ((startcol, startrow), _) = get_col_and_row_indexes_from_range(r)
+            df.to_excel(writer, sheet_name=TEST_SHEET_NAME, startrow=startrow, startcol=startcol, index=False)  
+
+
+    for r, df in zip(ranges, dfs):
+        upper_left_value = df.columns[0]
+        assert r == get_table_range_from_upper_left_corner_value(TEST_FILE_PATH, TEST_SHEET_NAME, upper_left_value)
+
```
aarondr77, line 214: Nice testing approach!

aarondr77:
Maybe this could say "exact cell value"?

Screenshot 2023-01-16 at 6 15 52 PM

aarondr77:
Can you vertically align this text?

Screenshot 2023-01-16 at 6 16 31 PM

aarondr77:
It's a bit strange that the `cell value` placeholder is not really an example, but instead of description whereas the `company_ids` placeholder is def an example. Maybe we could switch `company_ids` to `new dataframe name` for consistency  

Screenshot 2023-01-16 at 6 17 42 PM

aarondr77:
- Use the sample data from our customer 
- Set the upper left cell value as 1 
- Notice that it can't find the value although it exists in A6. I'm assuming this is a type bug

Screenshot 2023-01-16 at 6 19 52 PM

aarondr77:
The dataframe defining algorithm does not handle missing values. 

### Test case 1
- Use the sample data from the customer
- Delete the value in A35
- Import using `Customer ID` as the top left corner value
- Notice that the dataframe only contains one row instead of all 4

### Test case 2
- Use the sample data from the customer
- Add a value to E38
- Import using `Customer ID` as the top left corner value
- Notice that it doesn't pick up the new row. 

The algorithm requires that the first column of the dataframe be completely defined. We know that data often has missing values, so this is a poor assumption to make. Instead, we should end the dataframe when the entire row is missing values (which still could give false positives, but at least it much safer!)

aarondr77:
When the user switches between `Locate Dataframe By` options, it provably makes sense to clear the `value` input since they are irrelevant to each other and doesn't draw the user's attention that something switched

aarondr77:
When the value field is left blank, this error message is not that helpful. Could we update it to say something like: You left the value field blank ..... 

It would be additionally nice if the error message could open the correct card so that it was in context. If you have a few imports then it's hard to tell which one is wrong. 

Screenshot 2023-01-16 at 6 38 56 PM

aarondr77:
We could validate that all of the required fields have values in them before the user can press the `import ranges` button. Until then, the button is disabled.

aarondr77 (commented):
A couple important changes to get this to work on more realistic datasets, then LGTM.

mitosheet/mitosheet/excel_utils.py
naterush: Given that it's only called once, I'm down to factor all of this better later when we have more realistic datasets!

PR #596, opened 2023-01-16
Break out imports

naterush:
# Description

This PR breaks out imports, and deduplicates them. This is a very low-hanging bar (it was quite easy) to great effect -- scripts def feel much nicer now when you are importing multiple things!

We can iterate on this to improve how we're polluting the global namespace (which would be super worthwhile). See here: https://github.com/mito-ds/monorepo/issues/582

# Testing

Generate some code!

# Documentation

Nope.

### DIFF
--- a/mitosheet/mitosheet/code_chunks/add_column_set_formula_code_chunk.py
+++ b/mitosheet/mitosheet/code_chunks/add_column_set_formula_code_chunk.py
@@ -5,7 +5,7 @@
 # Distributed under the terms of the GPL License.
 
 from copy import copy
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 from mitosheet.code_chunks.code_chunk import CodeChunk
 from mitosheet.code_chunks.no_op_code_chunk import NoOpCodeChunk
@@ -35,7 +35,7 @@ class AddColumnSetFormulaCodeChunk(CodeChunk):
     def get_description_comment(self) -> str:
         return f'Added column {column_header_to_transpiled_code(self.column_header)}'
 
-    def get_code(self) -> List[str]:
+    def get_code(self) -> Tuple[List[str], List[str]]:
         column_headers = self.post_state.dfs[self.sheet_index].keys().tolist()
 
         python_code, _, _ = parse_formula(
@@ -50,7 +50,7 @@ class AddColumnSetFormulaCodeChunk(CodeChunk):
         column_header_index = self.column_header_index
         return [
             f'{self.df_name}.insert({column_header_index}, {transpiled_column_header}, {python_code})'
-        ]
+        ], []
 
     def get_edited_sheet_indexes(self) -> List[int]:
         return [self.sheet_index]
--- a/mitosheet/mitosheet/code_chunks/postprocessing/set_dataframe_format_code_chunk.py
+++ b/mitosheet/mitosheet/code_chunks/postprocessing/set_dataframe_format_code_chunk.py
@@ -282,9 +282,6 @@ def get_dataframe_format_code(state: State, sheet_index: int) -> Optional[str]:
         if line is None:
             continue
         dataframe_format_string += f"\\\n{TAB}{line}"
-
-    if 'np.where' in dataframe_format_string:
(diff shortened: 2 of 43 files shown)


naterush:
Note: merge the in after merging in the code chunk params PR

mitosheet/mitosheet/code_chunks/add_column_set_formula_code_chunk.py
```diff
@@ -5,68 +5,69 @@
...
 class AddColumnSetFormulaCodeChunk(CodeChunk):
 
+    def __init__(self, prev_state: State, post_state: State, sheet_index: int, column_id: ColumnID, column_header: ColumnHeader, column_header_index: int):
+        super().__init__(prev_state, post_state)
+        self.sheet_index = sheet_index
+        self.column_id = column_id
+        self.column_header = column_header
+        self.column_header_index = column_header_index
+
+        self.df_name = self.post_state.df_names[self.sheet_index]
```
aarondr77, line 30: So my understanding is that: we used to pass an untyped `param` argument to the code chunks. Then in the code chunk when we wanted to use the params, we needed to get the param from the untyped param, so we used a `get_param` function which just returned the param if it existed and if for some reason it did not, it returned None. 

Essentially, we relied on the type system from the frontend to ensure that the params we expected to be defined for each step were. And if for some reason it wasn't using the `get_param` function was a bit safer. 

Now, instead of using the untyped params, we are explicitly passing the params to the code chunk and setting those params as data fields in the class. For now, this doesn't do a ton for us, but in the future, we can define properties for these data fields like "this param cannot change". 

Alternatively to the change that we just made, we could strongly type our step performer params and then pass those to code chunks just as strongly typed param objects. To my understanding that would give us a very similar benefit to what we have here, but would also save us a ton of lines of code where we have to pass the params one by one to the code chunks and then set those params as data fields. 

With strongly typed params, we could also accomplish things like "this param cannot change" through the type system or if we want to create the classes still, we could unpack the params and create the data fields without having to use a line of code for each param.

mitosheet/mitosheet/code_chunks/postprocessing/set_dataframe_format_code_chunk.py
```diff
@@ -300,11 +300,11 @@ def get_display_name(self) -> str:
...
+    def get_code(self) -> Tuple[List[str], List[str]]:
         code = []
         for sheet_index in range(len(self.post_state.df_formats)):
             dataframe_format_code = get_dataframe_format_code(self.post_state, sheet_index)
             if dataframe_format_code is not None:
                 code.append(dataframe_format_code)
         
-        return code
\ No newline at end of file
+        return code, []
```
aarondr77, line 310: In this code chunk, we `import numpy as np`. We should move that to the new infrastructure too.

aarondr77 (commented):
LMK your thoughts on the suggestion below. No need to move to it, but would love to at least better understand the current state of your thinking, expecting that it will change and evolve as we do more of this! :)

mitosheet/mitosheet/code_chunks/step_performers/pivot_code_chunk.py
```diff
@@ -78,70 +79,88 @@ def build_args_code(
...
-            {'column_header': self.prev_state.column_ids.get_column_header_by_id(sheet_index, pf['column_id']), 'filter': pf['filter']}
-            for pf in pivot_filters_ids
+            {'column_header': self.prev_state.column_ids.get_column_header_by_id(self.sheet_index, pf['column_id']), 'filter': pf['filter']}
+            for pf in self.pivot_filters_ids
         ]
 
         # If there are no keys or values to aggregate on we return an empty dataframe. 
         if len(pivot_rows_with_transforms) == 0 and len(pivot_columns_with_transforms) == 0 or len(values) == 0:
-            return [f'{new_df_name} = pd.DataFrame(data={{}})']
+            return [f'{self.new_df_name} = pd.DataFrame(data={{}})'], ['import pandas as pd']
```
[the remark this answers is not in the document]
naterush: These types will change in the near future (specifically, to allow you to specify: what the package is, what name to import it under, and what specifically to import from that). We can fix this up then, but I'm fine to leave it for now!

PR #599, opened 2023-01-17
Export step

naterush:
# Description

Adds an export step. Mostly follows the spec here: https://www.notion.so/trymito/Export-Step-9de3658a13df46d682304116424ba047

There are a few differences:
1. A single step rather than multiple. This is simple and makes sense I think!
2. This is _not_ a post processing step. 

Per my comment on the specification, 
1) the formatting code doesn't need to be before this (it's unrelated formatting code),
2) users def want to export the _current state_ of their sheet, not whatever it ends up as if they keep editing (if they do keep editing)
3) if they don't keep editing, then the export code will be at the end of the analysis aways, so forcing it there will have no benefit!

I feel like not making a post processing step is clearly the right call. Happy to chat about any of any of this, ofc, give me a call if you so desire.

Also, let me know what you think about how I broke out Export into those two titles / if you can think of better names. Happy to think of better ways to factor this!

# Testing

Export some stuff!

# Documentation

We will need to write docs.


mitosheet/src/components/toolbar/Toolbar.tsx
```diff
@@ -192,10 +192,40 @@ const Toolbar = (
...
+
+                                    return {
+                                        ...prevUIState,
+                                        toolbarDropdown: undefined
+                                    }
+                                })
+                            }
+                            width='medium'
+                        >
+                            <DropdownItem title='Download as File' onClick={() => {props.setUIState(prevUIState => {
```
naterush, line 215: Think of better names

mitosheet/src/components/taskpanes/ExportToFile/ExportToFileTaskpane.tsx
```diff
@@ -0,0 +1,160 @@
...
+            <DefaultTaskpaneFooter>
+                <TextButton
+                    variant='dark'
+                    width='block'
+                    onClick={() => {
+                        edit();
+                    }}
+                    disabled={params.sheet_indexes.length === 0}
+                >
+                    Generate Export Code
```
naterush, line 153: What do we want this to do if you click it twice?

mitosheet/mitosheet/tests/step_performers/test_export_to_file.py
```diff
@@ -0,0 +1,129 @@
...
+    mito = create_mito_wrapper_dfs(*input_dfs)
+
+    mito.export_to_file(type, sheet_indexes, file_name)
+
+    for sheet_index, final_file_name in zip(sheet_indexes, final_file_names):
+        final_file_name = str(tmp_path / final_file_name)
+        assert os.path.exists(final_file_name)
+        assert pd.read_csv(final_file_name).equals(input_dfs[sheet_index])
+
+    check_dataframes_equal(mito)
```
naterush, line 68: TODO: I should check the files after this!

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -0,0 +1,47 @@
...
+        return f"Exports {len(self.sheet_index_to_export_location)} to file {self.file_name}"
+
+    def get_code(self) -> Tuple[List[str], List[str]]:
+        if self._type == 'csv':
+            return [
+                f'{self.post_state.df_names[sheet_index]}.to_csv("{export_location}", index=False)'
+                for sheet_index, export_location in self.sheet_index_to_export_location.items()
+            ], []
+        elif self._type == 'excel':
+            # If there is only one sheet being exported, we generate code that is as nice as we can manage
```
naterush, line 37: Better comment

naterush (commented):
Cleanup! Look at step performers.

mitosheet/mitosheet/step_performers/export_to_file.py
```diff
@@ -0,0 +1,110 @@
...
+
+def get_export_to_excel_sheet_index_to_sheet_name(state: State, file_name: str, sheet_indexes: List[int]) -> Dict[int, str]:
+    return {sheet_index: get_df_name_as_valid_sheet_name(state.df_names[sheet_index]) for sheet_index in sheet_indexes}
+
+
+def get_final_file_name(file_name: str, _type: str) -> str:
+    _, extension = os.path.splitext(file_name)
+    if extension == '':
+        if _type == 'csv':
+            return file_name + '.txt'
```
aarondr77, line 35: Why `.txt`  and not `.csv`?

aarondr77:
Can we make this bigger?

Screenshot 2023-01-17 at 5 19 09 PM

aarondr77:
If I set the `File Name` and for a csv file and then export multiple csv files, the file names are whacky.

In the example below, I set the file name to `ArgentinaCPI_export` and then it added the actual file names to it. 

```
# Exports 3 to file ArgentinaCPI_export_ArgentinaCPI.txt
loans.to_csv("ArgentinaCPI_export_loans.txt", index=False)
forestfires.to_csv("ArgentinaCPI_export_forestfires.txt", index=False)
ArgentinaCPI.to_csv("ArgentinaCPI_export_ArgentinaCPI.txt", index=False)
```

mitosheet/mitosheet/step_performers/export_to_file.py
aarondr77: We could give the user another .txt option if you think that's important. But I think the .csv option should generate a csv

mitosheet/src/components/taskpanes/ExportToFile/ExportToFileTaskpane.tsx
```diff
@@ -0,0 +1,161 @@
...
+                            return {
+                                ...prevParams,
+                                sheet_indexes: newSelectedSheetIndexes
+                            }
+                        })
+                    }}
+                />
+            </DefaultTaskpaneBody>
+            <DefaultTaskpaneFooter>
+                {editApplied && <p className='text-body-3'>Files and code written.</p>}
```
aarondr77, line 145: Can we update the `Files and code written.` result subtext to just say `Generated export code` 

And can we make it subtext instead of text-body-3

aarondr77:
- Click Generate Files
- Select Excel file
- Delete the automatically generated file name
- Press Import
- Get this error

Screenshot 2023-01-17 at 5 29 41 PM

aarondr77:
Let's add placeholder text to the filename input. It could be MitoExport, for example

aarondr77:
Could we add a little error/warning message here if the user adds the wrong file ending to their file name then it warns them. This is what other `Save As` wizards do

Screenshot 2023-01-17 at 5 33 58 PM

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -0,0 +1,47 @@
...
+from mitosheet.excel_utils import get_df_name_as_valid_sheet_name
+from mitosheet.state import State
+from mitosheet.types import ColumnID
+
+
+class ExportToFileCodeChunk(CodeChunk):
+
+    def __init__(self, prev_state: State, post_state: State, _type: str, file_name: str, sheet_index_to_export_location: Dict[int, str]):
+        super().__init__(prev_state, post_state)
+        self._type = _type
```
aarondr77, line 19: maybe we could use `export_type` so we don't need to use the underscore which is different from our other params

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -0,0 +1,47 @@
...
+                for sheet_index, export_location in self.sheet_index_to_export_location.items()
+            ], []
+        elif self._type == 'excel':
+            # If there is only one sheet being exported, we can avoid creating the pd.ExcelWriter
+            if len(self.sheet_index_to_export_location) == 1:
+                for sheet_index, export_location in self.sheet_index_to_export_location.items():
+                    return [f'{self.post_state.df_names[sheet_index]}.to_excel("{self.file_name}", sheet_name="{export_location}", index={False})'], []
+
+            return [f'with pd.ExcelWriter("{self.file_name}") as writer:'] + [
+                f'    {self.post_state.df_names[sheet_index]}.to_excel(writer, sheet_name="{export_location}", index={False})'
```
aarondr77, line 43: Want to use the TAB constant instead of just adding a tab here. This seems pretty fragile.

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -0,0 +1,47 @@
...
+                f'{self.post_state.df_names[sheet_index]}.to_csv("{export_location}", index=False)'
+                for sheet_index, export_location in self.sheet_index_to_export_location.items()
+            ], []
+        elif self._type == 'excel':
+            # If there is only one sheet being exported, we can avoid creating the pd.ExcelWriter
+            if len(self.sheet_index_to_export_location) == 1:
+                for sheet_index, export_location in self.sheet_index_to_export_location.items():
+                    return [f'{self.post_state.df_names[sheet_index]}.to_excel("{self.file_name}", sheet_name="{export_location}", index={False})'], []
+
+            return [f'with pd.ExcelWriter("{self.file_name}") as writer:'] + [
```
aarondr77, line 42: How come we're breaking this into two separate lists instead of just one list with all of the entries

mitosheet/src/components/LoadingIndicator.tsx
```diff
@@ -121,6 +121,9 @@ const getDisplayMessageForMessageType = (messageType: StepType | UpdateType): st
         case StepType.ExcelRangeImport: {
             return 'Excel Range Import'
         }
+        case StepType.ExportToFile: {
+            return 'Export to File'
```
aarondr77, line 125: This should say `Generating export code` or something like that.

mitosheet/src/components/elements/DataframeMultiSelect.tsx
```diff
@@ -0,0 +1,64 @@
...
+                indexesToToggle.forEach(sheetIndex => {
+                    if (newToggle) {
+                        addIfAbsent(newSelectedSheetIndexes, sheetIndex)
+                    } else {
+                        removeIfPresent(newSelectedSheetIndexes, sheetIndex)
+                    }
+                })
+                newSelectedSheetIndexes.sort() // Make sure these are in the right order;
+                props.onChange(newSelectedSheetIndexes);
+            }}
```
aarondr77, line 40: While we're at it, maybe we should make this a utility that we can use elsewhere. We have this code all over the codebase.

mitosheet/src/utils/actions.tsx
```diff
@@ -1281,11 +1302,32 @@ export const createActions = (
...
+                    return {
+                        ...prevUIState,
+                        currOpenTaskpane: {type: TaskpaneType.EXPORT_TO_FILE},
+                        selectedTabType: 'data'
+                    }
+                })
+            },
+            isDisabled: () => {return undefined}, // TODO
+            searchTerms: ['Export To File'],
+            tooltip: "Export To File"
```
aarondr77, line 1323: "Generate code to export data to file"

mitosheet/src/utils/actions.tsx
```diff
@@ -1281,11 +1302,32 @@ export const createActions = (
             searchTerms: ['CodeSnippets'],
             tooltip: "CodeSnippets"
         },
+        [ActionEnum.EXPORT_TO_FILE]: {
+            type: ActionEnum.EXPORT_TO_FILE,
```
aarondr77, line 1306: This action name and titles are pretty ambiguous. I think this should be about generating the code. Right now if you search for export, it's very confusing! 

Screenshot 2023-01-17 at 5 53 00 PM

mitosheet/src/utils/actions.tsx
```diff
@@ -441,6 +441,27 @@ export const createActions = (
             searchTerms: ['export', 'download', 'excel', 'csv'],
             tooltip: "Download dataframes as a .csv or .xlsx file."
         },
+        [ActionEnum.Export_Dropdown]: {
```
aarondr77, line 444: Having this makes the search for actions pretty confusing. Maybe we don't want this, or at least don't want to include it in the search results. (see other review comment)

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
aarondr77, line 43: Let's make sure to test this on windows. I think on Windows we need to prepend the path with `r' ....`

aarondr77 (commented):
A few small changes and then LGTM

mitosheet/mitosheet/step_performers/export_to_file.py
naterush: A .csv is a .txt file. But we can switch, nbd.

mitosheet/src/components/taskpanes/ExportToFile/ExportToFileTaskpane.tsx
naterush: I think it's good to say the files as well, but I updated the language to be more clear!

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
naterush: One is a regular list, and the other is a list comprehension! Sadly, you can't combine them into one (I tried initially).

mitosheet/src/components/elements/DataframeMultiSelect.tsx
naterush: Shore! Good callout for a refactor target :)

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
naterush: You only need to append paths with the raw variable if they have `\` in them -- which hopefully this won't (but who knows). Def should test this there!

mitosheet/src/components/elements/DataframeMultiSelect.tsx
naterush: We actually don't anymore! There are other similar functions, but none for dataframe selection -- so I'll leave for future work.

naterush:
Ok, this is ready to merge. Waiting to do so until after this weeks release (and I'm back with my lovely windows laptop), so I can test properly!

aarondr77:
> You only need to append paths with the raw variable if they have \ in them -- which hopefully this won't (but who knows). Def should test this there!

@naterush I bet that until we add the Export file location options too, users are going to pass a path to this. ... I did at least when testing this feature. So it might be worth adding the `r'` stuff so we generate valid paths in those cases.

naterush:
@aarondr77 how were you able to pass a path? I made it so the frontend disallows this

aarondr77:
Oooo nvm. I think originally you could, but I just tested and see that I'm no longer able to!

PR #601, opened 2023-01-19
Rolling window support

naterush:
# Description

This PR implements: https://www.notion.so/trymito/Rolling-Window-Formulas-ca15429824b44c6ab2c71ebc507316ca

_All work in the spec is implemented._ Notably, I did follow the plan of attack -- so I could open up multiple PRs, but I figured it'd be easier to just review one!

Generally, very happy with how things turned out. There's one medium sized thing we might want to address (let me know thoughts) -- when you open the formula editor, if you don't open it in the same row, it doesn't show you the proper offset (it only shows you the formula you submitted). This is related to if you write a formula, delete a row (or sort), and then edit the formula -- it's also out of date.

Let's:
1. Write a spec for how to handle this.
2. Ignore that issue in this PR (so feel free to not call it out, to avoid noise).

I think we can simply change the formula that the frontend gets from the backend to have some awareness of the offset, and where indexes are used. Then, it can show a different formula depending on the cell. **This is something we have to do for "every cell a different formula" anyways, so I think it's a really natural next specification.**

We can wait to merge this till we have that figured out, but let's get it ready otherwise!

# Testing

Write some formulas. Doh.

# Documentation

Maybe there are references to this in the docs?


mitosheet/mitosheet/types.py
```diff
@@ -106,11 +109,20 @@ class CodeSnippetEnvVars(TypedDict):
...
     ColumnHeaderWithFilter = Any # type:ignore
     ColumnIDWithPivotTransform = Any # type:ignore
     ColumnHeaderWithPivotTransform = Any # type:ignore
     ExcelRangeImport = Any # type:ignore
     CodeSnippet = Any # type:ignore
-    CodeSnippetEnvVars = Any # type:ignore
\ No newline at end of file
+    CodeSnippetEnvVars = Any # type:ignore
+    ParserMatch = Any # type:ignore
+    ParserMatchIndexLabel = Any # type:ignore
```
naterush, line 128: Remove

mitosheet/mitosheet/types.py
```diff
@@ -69,6 +69,9 @@
 
 PivotColumnTransformation = str
 
+RowOffset = int
+RelativeReference = Tuple[ColumnHeader, RowOffset]
```
naterush, line 73: Not needed

mitosheet/mitosheet/tests/test_parse.py
```diff
@@ -327,13 +364,376 @@
...
+        '=FUNC(aaaa)',
+        'A',
+        'a',
+        pd.DataFrame(columns=['aaa', 'aaaa'], index=pd.Index(['a', 'aa'])),
+        'df[\'A\'] = FUNC(df[\'aaaa\'])',
+        set(['FUNC']),
+        set(['aaaa'])
+    ),
+
+    # TODO: test where the types of the column header are same as index for datetime, and more etc
```
naterush, line 734: TODO: update these tests!

mitosheet/mitosheet/telemetry/anonymization_utils.py
```diff
@@ -58,10 +58,12 @@ def anonymize_formula(formula: str, sheet_index: int, steps_manager: Optional[St
         return anonymize_as_string(formula)
 
     # We just input a random address, as we don't use it
+    # TODO: Check this still works!
```
naterush, line 61: TODO

mitosheet/mitosheet/step_performers/column_steps/set_column_formula.py
```diff
@@ -234,8 +239,10 @@ def exec_column_formula(post_state: State, df: pd.DataFrame, sheet_index: int, c
     python_code, _, _ = parse_formula(
         spreadsheet_code, 
         column_header,
-        post_state.dfs[sheet_index].keys()
+        formula_label,
+        post_state.dfs[sheet_index]
     )
+    print("CDE", python_code)
```
naterush, line 245: Remove

mitosheet/mitosheet/parser.py
```diff
@@ -160,25 +172,28 @@ def safe_replace(
...
     )
 
     # Then, go through from the end to the start, and actually replace all the column headers
-    for (column_header, match) in column_header_match_tuples:
-        if column_header == old_column_header:
-            start = match.start()
-            end = match.end()
-            formula = formula[:start] + str(new_column_header) + formula[end:]
+    for parser_match in parser_matches:
+        if parser_match['type'] == 'column header':
```
naterush, line 184: Make these constants!

mitosheet/mitosheet/parser.py
```diff
@@ -219,18 +236,159 @@ def check_common_errors(
...
         formula: str,
-        column_headers: List[ColumnHeader],
-        string_matches: List
-    ) -> List[Tuple[ColumnHeader, Any]]:
+        formula_label: Union[str, bool, int, float], # Where the formula is written,
+        string_matches: List[Any],
+        df: pd.DataFrame
+    ) -> List[ParserMatch]:
     """
     Returns a list of the column header that is matched, as well as the
```
naterush, line 382: Update doc string

mitosheet/mitosheet/parser.py
```diff
@@ -256,36 +415,58 @@ def find_column_headers(match: Any) -> Any:
...
+                parser_matches.append({
+                    'type': 'column header',
+                    'match_range': match_range,
+                    'parsed': column_header,
+                    'unparsed': found_column_header,
+                    'row_offset': 0
+                })
+                return
+
+            # Second, check if the index matches under any condtions
```
naterush, line 433: Update comment

naterush (commented):
Cleanup!

aarondr77:
I think there is a bug and bad error message writing here:
1. `errors.py` is receiving an incorrect dtype for the params. They should both be dates, but this case is getting triggered
```
if (arg_one_type == 'str' and (arg_two_type == 'int' or arg_two_type == 'float')) or \
        (arg_two_type == 'str' and (arg_one_type == 'int' or arg_one_type == 'float')):
```
2. The error message should not tell you `wrap the number argument in a call to the VALUE function`. Instead, it should be the string argument! 

Screenshot 2023-01-22 at 2 43 45 PM

aarondr77:
We might want to figure out better fill_values depending on the dtype of the column. Filling the date time to 0 is probably more misleading than helpful... Maybe we should fill it with `np.NaT`. What do you think? Want to take a shot at specifying these in the next spec?

aarondr77:
When editing a formula, clicking on a column header to reference it no longer adds it to the formula. I expect that if I'm writing a formula in C5, and I click on the column B, then it will add the reference =B5 to the formula, for now. When we add full column formulas like `SUM(A:A), this might change. 

Not being able to click on the column header to reference it means that it's difficult to write an inplace formula. I would have to type out the column header and column index.

aarondr77:
Related to the above comment, if I use the suggested column header to write a formula, then it doesn't add the row reference either. I expect that it would add the row reference for whatever row I currently have. 

https://user-images.githubusercontent.com/18709905/213937767-90f96679-b765-4020-b4d4-74e7d4c5e082.mov

mitosheet/mitosheet/parser.py
```diff
@@ -256,69 +420,110 @@ def find_column_headers(match: Any) -> Any:
...
 
-            # Do not replace if it is part of a function, which means it has
-            # another ascii character before or after it. Or if it is part of
-            # a number. Or if it has a ( after it, then it's a function call
-            
-            if (start - 1 >= 0 and formula[start - 1].isalnum()) or \
-                (end < len(formula) and (formula[end].isalnum() or formula[end] == '(')):
-                return found_column_header
+            # First, we check if it's an unqualified column header with no index. This is just 
+            # if the characters after the header are
```
aarondr77, line 427: finish this comment too please!

mitosheet/mitosheet/parser.py
```diff
@@ -256,69 +420,110 @@ def find_column_headers(match: Any) -> Any:
...
+            if isinstance(column_header, str):
+                replace_string = f'{df_name}[\'{column_header}\']'
+            elif isinstance(column_header, datetime.datetime):
+                replace_string = f'{df_name}[pd.to_datetime(\'{column_header}\')]'
+            elif isinstance(column_header, datetime.timedelta):
+                replace_string = f'{df_name}[pd.to_timedelta(\'{column_header}\')]'
+            else:
+                replace_string = f'{df_name}[{column_header}]'
+
+            # If there is a row offset, then we use this to put the column header in an offset function
```
aarondr77, line 516: "put the column header in an offset function" -> "shift the series the correct number of rows".

mitosheet/mitosheet/saved_analyses/step_upgraders/set_column_formula.py
```diff
@@ -42,3 +42,40 @@ def upgrade_set_column_formula_1_to_2(step: Dict[str, Any], later_steps: List[Di
...
+            sheet_index: 0,
+            column_id: _column id_,
+            formula_label: 0,
+            old_formula: '=A',
+            new_formula: '=B',
+        }
+    }
+    """
+    params = step['params']
+    params['formula_label'] = 0
```
aarondr77, line 75: This is fine because none of the formulas have an offset in it. They are all column headers without indexes, so we won't calculate any offsets. 

Is the fact that the `0` might not be an index in the dataframe an issue?

mitosheet/mitosheet/parser.py
```diff
@@ -256,69 +420,110 @@ def find_column_headers(match: Any) -> Any:
...
-            # a number. Or if it has a ( after it, then it's a function call
-            
-            if (start - 1 >= 0 and formula[start - 1].isalnum()) or \
-                (end < len(formula) and (formula[end].isalnum() or formula[end] == '(')):
-                return found_column_header
+            # First, we check if it's an unqualified column header with no index. This is just 
+            # if the characters after the header are 
+            if is_column_header_without_index(formula, index, start, end):
+                parser_matches.append({
+                    'type': COLUMN_HEADER_MATCH_TYPE,
```
aarondr77, line 430: Maybe we should call this type `COLUMN_HEADER_WITHOUT_INDEX_MATCH_TYPE` . The name `COLUMN_HEADER_MATCH_TYPE` doesn't really tell the full story.

mitosheet/mitosheet/tests/step_performers/column_steps/test_set_column_formula.py
```diff
@@ -289,4 +290,96 @@ def test_set_column_formula_then_delete_diff_dataframe_not_optimizes():
...
+        0,
+        pd.DataFrame({'A': [1, 2], 'B': [4, 5], 'C': [1, 2]}, index=pd.RangeIndex(0, 2)),
+    ),
+    # Range index, negative offset
+    (
+        pd.DataFrame({'A': [1, 2], 'B': [4, 5]}, index=pd.RangeIndex(0, 2)),
+        '=A1',
+        'C',
+        0,
+        pd.DataFrame({'A': [1, 2], 'B': [4, 5], 'C': [2, 0]}, index=pd.RangeIndex(0, 2)),
```
aarondr77, line 319: This is actually a positive offset, right? We're referencing A1 in row 0.

mitosheet/mitosheet/tests/step_performers/column_steps/test_set_column_formula.py
```diff
@@ -289,4 +290,96 @@ def test_set_column_formula_then_delete_diff_dataframe_not_optimizes():
...
+    ),
+    # Range index, positive offset
```
aarondr77, line 321: This one is the negative offset.

mitosheet/src/components/endo/celleditor/cellEditorUtils.tsx
```diff
@@ -1,36 +1,56 @@
...
 import { FunctionDocumentationObject, functionDocumentationObjects } from "../../../data/function_documentation";
-import { ColumnHeader, ColumnID, EditorState, SheetData } from "../../../types";
+import { ColumnID, EditorState, MitoSelection, SheetData } from "../../../types";
 import { getDisplayColumnHeader, isPrimitiveColumnHeader, rowIndexToColumnHeaderLevel } from "../../../utils/columnHeaders";
+import { getColumnHeadersInSelection, getIndexLabelsInSelection } from "../selectionUtils";
 import { getCellDataFromCellIndexes } from "../utils";
 
 
+export const getSelectionFormulaString = (selections: MitoSelection[], sheetData: SheetData): string => {
+    // For each of the selections, we bu
```
aarondr77, line 11: comment missing ending

aarondr77 (commented):
Looks mostly good! A few small refactor suggestions. 

One thing to note here is that a huge portion of the parser code is to handle the different types of possible indexes and we don't even handle all of the different options. As we think about whether we want to keep the pandas index or use a Mito Row Number, we should take this into consideration.

aarondr77:
Your proposal of creating a spec to address the row index changing if you open the formula in a different location sounds good to me!

mitosheet/mitosheet/step_performers/column_steps/set_column_formula.py
```diff
@@ -271,7 +281,7 @@ def exec_column_formula(post_state: State, df: pd.DataFrame, sheet_index: int, c
     except Exception as e:
         # If this is the same formula as before, then it used to be valid and is not,
         # and so we let the user know they must have made some other change that made 
-        # in invalid
-        if spreadsheet_code == post_state.column_spreadsheet_code[sheet_index][column_id]:
-            raise make_invalid_formula_after_update_error()
+        # in invalid (TODO: broken)
+        #if spreadsheet_code == post_state.column_spreadsheet_code[sheet_index][column_id]:
+        #    raise make_invalid_formula_after_update_error()
         raise
```
naterush, line 287: Clean this up.

mitosheet/mitosheet/types.py
```diff
@@ -148,17 +151,40 @@ class CodeSnippetEnvVars(TypedDict):
...
+        parsed: Any
+        row_offset: RowOffset
+
+    class FrontendFormulaString(TypedDict):
+        type: str
+        string: str
+    class FrontendFormulaReference(TypedDict):
+        type: str
+        display_column_header: str
+        row_offset: int
```
naterush, line 167: Mark what these are. Make them constants, and test them.

mitosheet/mitosheet/utils.py
```diff
@@ -187,6 +187,8 @@ def df_to_json_dumpsable(
 
     # Import just before we use it to avoid circular imports
     from mitosheet.pro.conditional_formatting_utils import get_conditonal_formatting_result
+
+    print("FORMULA", column_formulas)
```
naterush, line 191: Remove.

mitosheet/src/components/endo/celleditor/CellEditor.tsx
```diff
@@ -176,19 +177,27 @@ const CellEditor = (props: {
...
+            props.editorState.pendingSelections, 
+            props.sheetData,
+            props.editorState.rowIndex
         );
 
-        // Strip the prefix, and append the suggestion
+        // Strip the prefix, and append the suggestion, and the current index label as well
         fullFormula = fullFormula.substr(0, fullFormula.length - suggestionReplacementLength);
         fullFormula += suggestion;
+        if (indexLabel !== undefined && indexLabel != null) {
```
naterush, line 188: Just check undefined.

mitosheet/src/components/endo/celleditor/cellEditorUtils.tsx
```diff
@@ -319,4 +355,56 @@ export const getDocumentationFunction = (formula: string): FunctionDocumentation
     } else {
         return matchingFunctions[0];
     }
+}
+
+export const getIndexLabelAtRowOffsetFromOtherIndexLabel = (sheetData: SheetData, indexLabel: any, rowOffset: number): any | undefined => {
```
naterush, line 360: getNewIndexLabel

mitosheet/src/components/endo/celleditor/cellEditorUtils.tsx
```diff
@@ -319,4 +355,56 @@ export const getDocumentationFunction = (formula: string): FunctionDocumentation
...
+            /**
+             * After adding the reference to the column header, we need to add the correct index label.
+             * Notably, this is the index label that is the formulaPart.rowOffset from the current indexLabel
+             */
+            const newIndexLabel = getIndexLabelAtRowOffsetFromOtherIndexLabel(sheetData, indexLabel, formulaPart.row_offset);
+            if (newIndexLabel !== undefined) {
+                formulaString += getDisplayColumnHeader(newIndexLabel);
+            } else {
+                /**
+                 * TODO: what to do if the index is out of bounds here? In the formula bar, we want to handle this
```
naterush, line 392: Leave a comment about this.

mitosheet/src/types.tsx
```diff
@@ -245,6 +245,10 @@ export type ConditionalFormattingResult = {
     'results': Record<ColumnID, Record<number | string, {color: string | undefined, backgroundColor: string | undefined} | undefined> | undefined>
 }
 
+type FormulaPart = {'type': 'string part', 'string': string}
```
naterush, line 248: TODO: test these types

mitosheet/mitosheet/tests/test_parse.py
```diff
@@ -399,4 +848,11 @@ def test_safe_replace(formula, old_column_header, new_column_header, column_head
 
 @pytest.mark.parametrize('formula,substring,contains', SAFE_CONTAINS_TESTS)
 def test_safe_contains(formula, substring, contains):
-    assert safe_contains(formula, substring, ['A', 'B']) == contains
\ No newline at end of file
+    assert safe_contains(formula, substring, ['A', 'B']) == contains
+
+
+@pytest.mark.parametrize("formula,column_header,formula_label,df,python_code,functions,columns", INDEX_TEST_CASES)
+def test_get_frontend_formula(formula,column_header,formula_label,df,python_code,functions,columns):
```
naterush, line 855: rename test

naterush: Mark the other function as just for testing

mitosheet/mitosheet/parser.py
```diff
@@ -411,4 +624,73 @@ def parse_formula(
         final_code = f'{df_name}[{transpiled_column_header}] = {code_with_functions}'
     else:
         final_code = f'{code_with_functions}'
-    return final_code, functions, column_header_dependencies
\ No newline at end of file
+    return final_code, functions, column_header_dependencies
+
+
+def get_frontend_formula(
```
naterush, line 630: Document

mitosheet/mitosheet/parser.py
```diff
@@ -411,4 +624,73 @@ def parse_formula(
...
+
+    return frontend_formula
+
+
+def get_backend_formula_from_frontend_formula(
+    frontend_formula: FrontendFormula,
+    formula_label: Union[str, bool, int, float],
+    df: pd.DataFrame,
+) -> str:
+    # TODO: write this as a testing utility
```
naterush, line 688: Cleanup

naterush (commented):
Cleanup and test!

mitosheet/mitosheet/types.py
```diff
@@ -148,17 +151,41 @@ class CodeSnippetEnvVars(TypedDict):
         MITO_CONFIG_CODE_SNIPPETS_URL: str
         MITO_CONFIG_CODE_SNIPPETS_SUPPORT_EMAIL: Optional[str]
 
+    class ParserMatch(TypedDict):
+        type: str # 'column header' | 'index label'
```
aarondr77, line 155: In the parser, we set the `type` to different values (below). Want to update this comment?

```
COLUMN_HEADER_MATCH_TYPE = 'column header match type'
INDEX_LABEL_MATCH_TYPE = 'index label match type'
```

mitosheet/mitosheet/parser.py
```diff
@@ -10,12 +10,23 @@
...
-from mitosheet.types import ColumnHeader
+from mitosheet.sheet_functions.types.utils import is_number_dtype, is_datetime_dtype, is_string_dtype
+from mitosheet.transpiler.transpile_utils import \
+    column_header_to_transpiled_code
+from mitosheet.types import (ColumnHeader, FrontendFormula, ParserMatch, ParserMatchRange,
+                             RowOffset)
+
+
+COLUMN_HEADER_MATCH_TYPE = 'column header match type'
+INDEX_LABEL_MATCH_TYPE = 'index label match type'
```
aarondr77, line 28: Can you leave a comment for each of these explaining what the match represents? My understanding is that if given the formula =A0 in B1, the parser matches would be:

```
{
                    'type': COLUMN_HEADER_MATCH_TYPE,
                    'match_range': (index_label_start, index_label_start + len(number_chars)),
                    'unparsed':'A',
                    'parsed': df['A'],
                    'row_offset': -1
}
{
                    'type': INDEX_LABEL_MATCH_TYPE,
                    'match_range': (index_label_start, index_label_start + len(number_chars)),
                    'unparsed':'0',
                    'parsed': 0,
                    'row_offset': -1
}

```

mitosheet/mitosheet/parser.py
```diff
@@ -256,69 +422,113 @@ def find_column_headers(match: Any) -> Any:
...
+            index_label_match = number_index_label_match or datetime_index_label_match or string_index_label_match or None
+            if index_label_match is not None:
+                parser_matches.append({
+                    'type': COLUMN_HEADER_MATCH_TYPE,
+                    'match_range': match_range,
+                    'parsed': column_header,
+                    'unparsed': found_column_header,
+                    'row_offset': index_label_match['row_offset']
+                })
+                parser_matches.append(index_label_match)
```
aarondr77, line 453: Since `COLUMN_HEADER_MATCH_TYPE` either has a row_offset of 0 or is followed by a `INDEX_LABEL_MATCH_TYPE`, we could remove the `row_offset` from the `COLUMN_HEADER_MATCH_TYPE` match type all together.

mitosheet/src/components/endo/celleditor/cellEditorUtils.tsx
```diff
@@ -1,36 +1,72 @@
...
+        const formulaIndexLabel = sheetData.index[rowIndex];
+        columnHeadersAndIndexLabels.push(getDisplayColumnHeader(columnHeader) + getDisplayColumnHeader(formulaIndexLabel));
+    })
+
+
+    selections.forEach(selection => {
+        // Skip these as we handle them above
+        if (isSelectionEntireSelectedColumn(selection)) {
+            return;
+        }
```
aarondr77, line 27: Why not to just move it here if we're doing the check anyways?

mitosheet/src/components/endo/celleditor/cellEditorUtils.tsx
```diff
@@ -319,4 +349,59 @@ export const getDocumentationFunction = (formula: string): FunctionDocumentation
...
+                formulaString += getDisplayColumnHeader(newIndexLabel);
+            } else {
+                /**
+                 * TODO: how do we want to handle the case where the newIndexLabel is undefined? This happens when 
+                 * we might have written a formula that references the row before, but now we're editing in the first row. 
+                 * 
+                 * E.g. B1 = A1 + A0, then go to B0. What do you see? B0 = A0 + A(?)
+                 * 
+                 * Currently, we handle this just by clamping to the first index, but this can be somewhat misleading. Let
+                 * me know what you think we should do here... we could try putting in the fill_value for the shift function...
```
aarondr77, line 396: Excel makes the formula `=#REF!`

I think you're right that its a bit misleading as is because if a user goes to figure our why its `NaN` which would be super reasonable thing to do, they'll end up changing the formula. 

In this example: eg. B1 = A1 + A0, then go to B0. What do you see? B0 = A0 + A(?)

Some ideas: 
= A0 + A-1
= A0 + #INVALID_REFERENCE
= #INVALID_REFERENCE
= #REF!

I would be fine with any of these.

aarondr77 (commented):
Nice! I proposed a few minor refactors and additional documentation that would be helpful to understand the new, more complex parser. 

+ There are a couple of review comments from my previous review that I think got accidentally passed over. 

Functionality is working great! Once you address this LGTM.

mitosheet/src/components/endo/celleditor/cellEditorUtils.tsx
naterush: Oh... nice.

mitosheet/mitosheet/parser.py
```diff
@@ -662,7 +664,7 @@ def get_frontend_formula(
                 })
 
             frontend_formula.append({
-                'type': 'reference',
+                'type': 'reference part',
```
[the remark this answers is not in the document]
naterush: I did this, but then I changed them back to Literal types. I added these types in the frontend. This is the best of both worlds, I think!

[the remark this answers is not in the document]
naterush: This is addressed in the next PR, so don't worry about it for now.

mitosheet/mitosheet/parser.py
```diff
@@ -219,18 +239,161 @@ def check_common_errors(
...
+    if parsed_label_index != -1:
+        row_offset: RowOffset = parsed_formula_label_index - parsed_label_index
+        return row_offset
+
+    return None
+
+def is_number_index(index: pd.Index) -> bool:
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        if isinstance(index, pd.RangeIndex) or isinstance(index, pd.Int64Index) or isinstance(index, pd.UInt64Index) or isinstance(index, pd.Float64Index):
```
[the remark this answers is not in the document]
naterush: I'm not sure. They are fine to keep though, I think.

mitosheet/mitosheet/parser.py
naterush: Close. I updated the comment.

naterush: You're off actually:
1. the match_range (has to be the range in the formula of the match), 
2. in the parsed column header, it just stays as A -- it would change if it were a number or something for example, and get parsed into an actual int)

Otherwise, is good! I added as a comment. They would also be returned in reverse order.

mitosheet/mitosheet/parser.py
naterush: I originally had something similar this, but you need to figure out _where_ the column headers are. So you need to reconstruct these matches in either case. 

Duplicating the row_offset in both cases is a little wacky, but in practice working with this API has led to me _always_ getting exactly what I need. The API is very pleasant to consume, so I think it makes sense to keep!

mitosheet/src/components/endo/celleditor/cellEditorUtils.tsx
naterush: I'm going to just make it so it puts a 0 in there. For most use cases, I think this is a reasonable thing to display and do - as it actually gives the users the input that is in there!

naterush:
Ok, this is ready to merge!

mitosheet/src/components/endo/celleditor/cellEditorUtils.tsx
naterush: This notably has the tradeoff that if the user goes and tries to submit this formula, it will change the formula in the cell. But I think this is a suitable tradeoff - as it has the benefit of actually being more clear when the user is looking at it!

PR #603, opened 2023-01-20
mitosheet: add env for turning off tours

aarondr77:
# Description

Adds an environment variable for turning off tours. Helpful for enterprise customers who run trainings and find the tour distracting during those sessions. 

Note: I did not update the Mito Config Version since we handle the default case where the Display Tours key is not set and nothing else changed. There's no benefit.  

# Testing

Set these environment variables and make sure that the intro and spreadsheet formula tours do not appear.
```
import os
os.environ['MITO_CONFIG_DISABLE_TOURS'] = 'True'
os.environ['MITO_CONFIG_VERSION'] = '2'
```

# Documentation

### DIFF
--- a/mitosheet/mitosheet/enterprise/mito_config.py
+++ b/mitosheet/mitosheet/enterprise/mito_config.py
@@ -16,6 +16,7 @@ MITO_CONFIG_SUPPORT_EMAIL = 'MITO_CONFIG_SUPPORT_EMAIL'
 MITO_CONFIG_CODE_SNIPPETS_SUPPORT_EMAIL = 'MITO_CONFIG_CODE_SNIPPETS_SUPPORT_EMAIL' 
 MITO_CONFIG_CODE_SNIPPETS_VERSION  = 'MITO_CONFIG_CODE_SNIPPETS_VERSION' 
 MITO_CONFIG_CODE_SNIPPETS_URL = 'MITO_CONFIG_CODE_SNIPPETS_URL'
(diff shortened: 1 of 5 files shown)


mitosheet/mitosheet/enterprise/mito_config.py
```diff
@@ -131,6 +135,11 @@ def get_support_email(self) -> str:
             return DEFAULT_MITO_CONFIG_SUPPORT_EMAIL
         return self.mec[MITO_CONFIG_SUPPORT_EMAIL]
 
+    def get_display_tours(self) -> bool:
+        if self.mec is None or self.mec[MITO_CONFIG_DISPLAY_TOURS] is None:
+            return True
+        return False if self.mec[MITO_CONFIG_DISPLAY_TOURS] == 'False' else True
```
naterush, line 141: I think we should have a helper function called `is_env_variable_set_to_true`. It should test the variable against:
1. 1 and 0
2. True and false (upper and lower)
3. t and f

Generally, people will put weird shit in this. We should accept the most common inputs

naterush (commented):
Can we flip the condition here? E.g. you set `MITO_CONFIG_DISABLE_TOURS=True` rather than the opposite? 

This makes it much more clear that having tours is the default. And I think is a better interface overall!

Let me know when you do, and then I'll test functionality!

naterush:
I think 1 should be in quotes

naterush:
This doesn't appear to be working for me.

naterush:
There's also an extra print to remove.

PR #604, opened 2023-01-20
Add a hint that a file is password protected to stop confusion

aarondr77:
# Description

When an Excel file is password protected, no sheets are displayed in the multi toggle box. This has causes confusion to our users. Now if there are no sheets returned, we just let them know that it might be because the file is password protected. 80/20

Supporting password protected files is not straight forward in pandas, so if we want to actually support these files, we'll need to look for new import methods. 

# Testing

Create a password protected Excel file and watch for the message. 

# Documentation

### DIFF
--- a/mitosheet/src/components/elements/MultiToggleBox.tsx
+++ b/mitosheet/src/components/elements/MultiToggleBox.tsx
@@ -18,8 +18,8 @@ import { Height, Width } from './sizes.d';
 const MAX_DISPLAYED = 10000;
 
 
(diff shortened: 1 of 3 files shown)


mitosheet/src/components/elements/MultiToggleBox.tsx
```diff
@@ -242,6 +248,7 @@ const MultiToggleBox = (props: {
                     loading={props.loading}
                     isSubset={props.isSubset}
                     message={props.message}
+                    _emptyMessage={props.emptyMessage}
```
naterush, line 251: Can we change the name of this param to emptyMessage? And then change the variable name used in the function?

naterush (commented):
Minor fix up, then LGTM

aarondr77:
Ready to merge after release

PR #609, opened 2023-01-24
Snowflake frontend

aarondr77:
# Description

Implements https://www.notion.so/trymito/Snowflake-Query-Import-INTERNAL-4a067d1d07ba436596ae8a64d510409b

# Testing

- Execute a snowflake import
- Create a new mitosheet and used the cached credentials to execute another snowflake import
- Restart your kernel and rerun the mitosheet. Use the update imports taskpane to revalidate your snowflake credentials 

# Documentation

Yes, we need to write docs during the release!


mitosheet/mitosheet/api/get_validate_snowflake_credentials.py
```diff
@@ -17,6 +17,16 @@
 except ImportError:
     SNOWFLAKE_CONNECTOR_IMPORTED = False
 
+# Global variable used to cache the snowflake credentials so that when users
+# open the snowflake taskpane multiple times in the same mito instantiation,
```
aarondr77, line 21: Update this to mention lifecycle of credentials. They exist for all mitosheets in this kernel until the kernel is restarted.

mitosheet/mitosheet/api/get_validate_snowflake_credentials.py
```diff
@@ -48,6 +58,11 @@ def get_validate_snowflake_credentials(params: SnowflakeCredentials, steps_manag
             'type': 'error',    
             'error_message': 'Invalid authentication information. Please try again.'
         })
+
+    # cache the snowflake credentials 
+    global cached_snowflake_credentials
+    cached_snowflake_credentials = params
+    print(cached_snowflake_credentials)
```
aarondr77, line 65: removed

mitosheet/mitosheet/api/get_test_imports.py
```diff
@@ -17,12 +18,16 @@
 CSV_IMPORT_ERROR = 'There was an error importing this CSV file. Make sure that the file exists, or select a different file or dataframe.'
 EXCEL_IMPORT_ERROR = 'There was an error importing this sheet. Make sure that the Excel file exists and contains this sheet, or select a different file or dataframe.'
 DATAFRAME_IMPORT_ERROR = 'There was an error importing this dataframe. Make sure that the dataframe is defined, or select a different file or dataframe.'
+SNOWFLAKE_IMPORT_ERROR = 'There was an error executing this query. Make sure that your credentials are valid and you have access to this table, or select a different file or dataframe.'
```
aarondr77, line 21: Figure out how to use this error message instead of the raised error message from the backend

mitosheet/mitosheet/tests/api/test_get_cached_snowflake_credentials.py
```diff
@@ -0,0 +1,43 @@
...
+def test_integration_get_cached_credentials():
+    mito = create_mito_wrapper_dfs()
+
+    get_validate_snowflake_credentials(TEST_SNOWFLAKE_CREDENTIALS, mito.mito_backend.steps_manager)
+
+    mito_2 = create_mito_wrapper_dfs()
+
+    credentials = get_cached_snowflake_credentials({}, mito_2.mito_backend.steps_manager)
+
+    assert credentials == 'null'
```
aarondr77, line 43: Update this test to expect it to pass!

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -353,9 +282,9 @@ const SnowflakeImportTaskpane = (props: SnowflakeImportTaskpaneProps): JSX.Eleme
                         </MultiToggleBox>
                         <TextButton
                             disabled={
-                                params.credentials.username.length === 0 || 
-                                params.credentials.password.length === 0 || 
-                                params.credentials.account.length === 0 || 
+                                // credentials.username.length === 0 || 
+                                // credentials.password.length === 0 || 
+                                // credentials.account.length === 0 ||
```
aarondr77, line 287: Figure out how we want to handle this

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportCard.tsx
```diff
@@ -128,6 +152,25 @@ const UpdateImportCard = (props: {
         })
     }
 
+    const getSnowflakeDropddownItemOrEmptyElement = (): JSX.Element =>  {
+        if (props.dataframeCreationData.step_type === 'snowflake_import') {
+            return (
+                <DropdownItem
+                    title='Enter Snowflake Credentials'
```
aarondr77, line 159: Make the dropdown wider or come up with a shorter name so this doesn't get cut off

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportsTaskpane.tsx
```diff
@@ -95,7 +100,9 @@ const UpdateImportsTaskpane = (props: UpdateImportsTaskpaneProps): JSX.Element =
 
             if (failedReplayData !== undefined) {
                 importData = await props.mitoAPI.getImportedFilesAndDataframesFromAnalysisName(failedReplayData.analysisName);
+                console.log('importData: ', importData)
```
aarondr77, line 103: remove these

mitosheet/src/components/taskpanes/taskpanes.tsx
```diff
@@ -48,6 +48,7 @@ export enum TaskpaneType {
     CODESNIPPETS = 'CodeSnippets',
     SNOWFLAKEIMPORT = 'SnowflakeImport',
     EXCEL_RANGE_IMPORT = 'Excel Range Import',
+    UPDATESNOWFLAKECREDENTIALSTASKPANE = 'UpdateSnowflakeCredentialsTaskpane',
```
aarondr77, line 51: remove

mitosheet/src/jupyter/api.tsx
```diff
@@ -725,6 +725,22 @@ export default class MitoAPI {
...
 
 
+    
+    async getCachedSnowflakeCredentials(): Promise<SnowflakeCredentials | undefined> {
+
+        const resultString = await this.send<string>({
+            'event': 'api_call',
+            'type': 'get_cached_snowflake_credentials',
+            'params': {}
+            // TODO: Make this a priority API call
```
aarondr77, line 735: Do this!

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportCard.tsx
```diff
@@ -86,13 +97,26 @@ export const getUpdateImportCardSubtitle = (updatedDataframeCreationData: Datafr
...
+    } else if (updatedDataframeCreationData.step_type === 'excel_range_import') {
         return (
             <div className='mt-3px'>
                 <span className='text-color-medium-gray-important'>Update to </span> {getSimpleNameSpan(updatedDataframeCreationData.params.range_imports[0].df_name)} <span className='text-color-medium-gray-important'>from </span> {getFileNameSpanFromFilePath(updatedDataframeCreationData.params.file_path)}
             </div>
         )
+    } else if (updatedDataframeCreationData.step_type === 'snowflake_import') {
+        return (
+            <div className='mt-3px'>
+                <span className='text-color-medium-gray-important'>Update to </span> {getSimpleNameSpan(updatedDataframeCreationData.params.table_loc_and_warehouse.table)} <span className='text-color-medium-gray-important'>from Snowflake </span>
```
aarondr77, line 109: Update this instead to say something like: Validated snowflake credentials

aarondr77 (commented):
Cleanup required

mitosheet/css/layout/CollapsibleSection.css
```diff
@@ -1,7 +1,14 @@
...
-
 .mito-blue-container {
     padding: 5px;
     background-color: var(--mito-light-blue);
     border-radius: 5px;
-}
\ No newline at end of file
+}
+
+.mito-collapsible-content-card-container {
```
aarondr77, line 7: Import this css file in AuthenticateToSnowflake

mitosheet/css/layout/CollapsibleSection.css
```diff
@@ -1,7 +1,14 @@
...
+    background-color: var(--mito-light-blue);
+    border-radius: 5px;
+}
+
+.mito-collapsible-content-card-subtext {
```
aarondr77, line 12: Change padding so this text is not so far away

mitosheet/mitosheet/api/get_imported_files_and_dataframes_from_current_steps.py
```diff
@@ -76,6 +77,16 @@ def get_import_data_with_single_import_list(step_type: str, params: Dict[str, An
             }
         } for range_import in params['range_imports']]
 
+    if step_type == SnowflakeImportStepPerformer.step_type():
+        print("HERE")
```
aarondr77, line 81: remove

mitosheet/mitosheet/errors.py
```diff
@@ -520,6 +520,19 @@ def make_invalid_snowflake_import_error() -> MitoError:
         False
     )
 
+def make_invalid_snowflake_credentials_error() -> MitoError:
```
aarondr77, line 523: Make sure we handle errors correctly in the frontend

mitosheet/src/components/elements/AuthenticateToSnowflakeCard.tsx
```diff
@@ -0,0 +1,160 @@
...
+    onCredentialsValidated: () => void;
+    isOpen: boolean
+}): JSX.Element => {
+
+    const [credentials, setCredentials] = useState<SnowflakeCredentials>(() => getDefaultCredentials())
+    const [snowflakeCredentialsValidityCheckResult, setSnowflakeCredentialsValidityCheckResult] = useState<SnowflakeCredentialsValidityCheckResult | undefined>(undefined)
+    const [loading, setLoading] = useState<boolean>(false)
+
+    useEffect(() => {
+        // On first render, load the cached credentials
```
aarondr77, line 36: ... if they exist

mitosheet/src/components/elements/AuthenticateToSnowflakeCard.tsx
```diff
@@ -0,0 +1,160 @@
...
+        setLoading(true)
+        void loadAndSetAndValidateCachedCredentials()
+    }, []);
+
+    const loadAndSetAndValidateCachedCredentials = async () => {
+        const cachedCredentials = await props.mitoAPI.getCachedSnowflakeCredentials()
+        if (cachedCredentials !== undefined) {
+            setCredentials(cachedCredentials)
+            void _validateSnowflakeCredentials(cachedCredentials)
```
aarondr77, line 45: Add an `await` here

aarondr77: Test by passing an incorrect account

mitosheet/src/components/elements/AuthenticateToSnowflakeCard.tsx
```diff
@@ -0,0 +1,160 @@
...
+        setLoading(false)
+    }
+
+    const validateSnowflakeCredentialsParams = async () => {
+        setLoading(true)
+        void _validateSnowflakeCredentials(credentials)
+    }
+
+    const _validateSnowflakeCredentials = async (credentials: SnowflakeCredentials) => {
+        const validityCheckResult = await props.mitoAPI.validateSnowflakeCredentials(credentials);
```
aarondr77, line 56: validityCheckResult -> credentialsValidityCheckResult

mitosheet/src/components/elements/LoadingCounter.tsx
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Mito
+
+import React, { useEffect, useState } from 'react';
+
+/*
+    Loading counter
+*/
+const LoadingCounter = (): JSX.Element => {
+    // We use a count to track the number of the time loading.
```
aarondr77, line 9: amount of time...

mitosheet/src/components/elements/LoadingCounter.tsx
```diff
@@ -0,0 +1,25 @@
...
+    const [timeLoading, setTimeLoading] = useState(1);
+
+    // Schedule a change to update the loading indicator, every .5 seconds
```
aarondr77, line 12: every 1 second

mitosheet/src/components/taskpanes/FileImport/FileImportTaskpane.tsx
```diff
@@ -110,6 +112,10 @@ function FileImportTaskpane(props: ImportTaskpaneProps): JSX.Element {
                 setImportState={setImportState}
             />
         )
+    } else if (importState.screen == 'authenticate_to_snowflake') {
```
aarondr77, line 115: Add comment explaining why this is empty

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -109,18 +110,20 @@ const getNewParams = (prevParams: SnowflakeImportParams, database?: string | nul
...
+            StepType.SnowflakeImport, 
+            props.mitoAPI,
+            props.analysisData,
     )
-    
+
+    const [credentialsValidated, setCredentialsValidated] = useState(false)
     const [credentialsSectionIsOpen, setCredentialsSectionIsOpen] = useState(true);
     const [availableSnowflakeOptionsAndDefaults, setAvailableSnowflakeOptionsAndDefaults] = useState<AvailableSnowflakeOptionsAndDefaults | undefined>(undefined);
+    const [loadingOptionsAndDefaults, setLoadingOptionsAndDefaults] = useState(false)
```
aarondr77, line 123: loadingAvailableOptionsAndDefaults

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -294,12 +249,15 @@ const SnowflakeImportTaskpane = (props: SnowflakeImportTaskpaneProps): JSX.Eleme
...
-                            Table
+                            <p className={classNames({'text-color-gray-disabled': loadingOptionsAndDefaults})}>
+                                Table
+                            </p>
                         </Col>
                         <Col>
                             <Select
                                 width="medium"
-                                value={params.table_loc_and_warehouse.table || 'None available'}
+                                value={loadingOptionsAndDefaults ? "Loading.." : params.table_loc_and_warehouse.table || 'None available'}
```
aarondr77, line 259: Remove this Loading ... since we use the disable now

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -351,22 +318,50 @@ const SnowflakeImportTaskpane = (props: SnowflakeImportTaskpaneProps): JSX.Eleme
...
-                                params.credentials.password.length === 0 || 
-                                params.credentials.account.length === 0 || 
+                                !credentialsValidated || // Activated when user first validates credentials. Its possible they've since invalidated their credentials.
                                 params.table_loc_and_warehouse.warehouse === undefined || 
                                 params.table_loc_and_warehouse.database === undefined || 
                                 params.table_loc_and_warehouse.schema === undefined ||
                                 params.table_loc_and_warehouse.table === undefined 
                             }
-                            disabledTooltip='Fields missing from the query. TODO: Cleanup'
+                            disabledTooltip='Fill out all required fieldsd'
```
aarondr77, line 350: fix typo

mitosheet/src/utils/actions.tsx
```diff
@@ -1344,26 +1344,26 @@ export const createActions = (
...
-        //                 currOpenTaskpane: {type: TaskpaneType.SNOWFLAKEIMPORT},
-        //                 selectedTabType: 'data'
-        //             }
-        //         })
-        //     },
-        //     isDisabled: () => {return undefined}, 
-        //     searchTerms: ['SQL', 'database', 'snowflake', 'import'],
-        //     tooltip: "Import dataframe from a Snowflake data warehouse"
-        // },
+        [ActionEnum.SNOWFLAKEIMPORT]: {
```
aarondr77, line 1347: Make sure this is in the dataframes dropdown

mitosheet/src/utils/actions.tsx
```diff
@@ -1344,26 +1344,26 @@ export const createActions = (
...
+            longTitle: 'Snowflake Import',
+            actionFunction: () => {
+                // We turn off editing mode, if it is on
+                setEditorState(undefined);
+
+                setUIState(prevUIState => {
+                    return {
+                        ...prevUIState,
+                        currOpenTaskpane: {type: TaskpaneType.SNOWFLAKEIMPORT},
+                        selectedTabType: 'data'
```
aarondr77, line 1359: The Snowflake Frontent should only be available if the user is on Enterprise ..... 

Should we update that?

aarondr77:
# Remaining polish 
- [ ] Figure out what we want to do with this for testing. Nobody is on the Enterprise version right now to test it with 
- [ ] When the user has cached credentials, don't flash the screen so many times on the way to opening the configuration options. Instead of opening the auth screen to do the cache, let's do it in the snowflake taskpane and then only open the screen if the user does not have cached validated credentials! 
- [ ] After the user has validated their credentials, reduce the delay between closing the auth card and opening the configuration card. This is currently a bit disorientating. It will just require changing the order of operations!

naterush:
An easy bit of code optimization I think we should do: two snowflake code chunks next to each other, with the same `connection_params`, then we can make only a single connection object, and have two different queries. This requires making sql_query a list of queries -- which makes sense to me! Each of them just makes a new dataframe. 

I have a feeling this will be a reasonably common thing to do with snowflake, it's a very simple optimization, and there are LARGE line savings!

Screen Shot 2023-01-27 at 9 13 04 AM

naterush:
I don't know how I got here (I'll try and replicate below), but this is a really bad error message. We should display the error that we got from the backend (and allow you to expand it, if that looks better). Not being able to dig in here makes it impossible to figure out what the heck I might have done wrong.

Screen Shot 2023-01-27 at 9 16 32 AM

naterush:
I can't replicate rn since I don't have credentials, but I think the bug above was a result of running 2 queries, then pressing undo twice, then trying to click and do the queries again.

mitosheet/src/components/elements/AuthenticateToSnowflakeCard.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+const getDefaultCredentials = (): SnowflakeCredentials => {
+    return {type: 'username/password', username: '', password: '', account: ''}
+}
+
+/**
+    * A card that handles all work required to authenticate a user to snowflake
+*/ 
+
+const AuthenticateToSnowflakeCard = (props: {
```
naterush, line 26: This card needs some cleanup that is very similar to what I requested on the SnowflakeImportTaskpane last time. With the comments below, the goal is to give you time to think through how to factor it a bit better -- of course, happy to talk about this on a call if it's helpful!

1. The functions are `loadAndSetAndValidateCachedCredentials`, `validateSnowflakeCredentialsParams`, and `_validateSnowflakeCredentials`. Similar feedback to the Snowflake Import taskpane last PR -- I cannot distinguish between these functions, not only what they do, but when I might call them. 
2. The effect sets loading to true, the function `loadAndSetAndValidateCachedCredentials` sets loading to false. But `loadAndSetAndValidateCachedCredentials` is only ever called by the effect!
3. The `validateSnowflakeCredentialsParams` param function does almost nothing - do we need it? If we get rid of it, how does this change how easy it is to name the functions?
4. Related to (3), what happens if we move the setLoading into the `_validateSnowflakeCredentals` function.

naterush (changes requested):
Some cleanup on the code here, and then I'll have to get new login info and test fully.

mitosheet/src/components/taskpanes/FileImport/FileImportTaskpane.tsx
```diff
@@ -39,6 +39,8 @@ export type ImportState = {
     filePath: string
 } | {
     screen: 'dataframe_import'
+} | {
+    screen: 'authenticate_to_snowflake'
```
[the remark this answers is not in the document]
naterush: Can we change the name of the taskpane too?

naterush:
Screen Shot 2023-01-27 at 10 14 08 AM

Similar error message improvements here. My life would have been a whole lot easier as a user trying to authenticate if:
1. The error message was displayed on the frontend.
2. If not, the error message was at least printed

I added the print you see in the screen shot. Can we make it so if the error message is not the standard just invalid password or username, we display it on the frontend?

naterush:
1. Auth with snowflake
2. Select a valid table, some columns, and click import
3. Click undo. 
4. All of the Warehouse, Database, and Schema utils switch to undefined. 

I'm not sure how in the weeds we have to get to fix that up, but ideally we'd not have them as undefined. Can you figure out if that's possible?

naterush:
Screen Shot 2023-01-27 at 10 20 42 AM

Why don't we just name Limit -> Max Rows, and then remove the ?

naterush:
Ok, I replicated the `We were unable to execute that query. Reach out to support for help.` error:
1. Open snowflake, authenticate.
2. Import from one table.
3. Import from a different table. 
4. Click undo. 
5. The toggle updates the table back to the previous, but not the columns. The net result is that trying to configure the column and import just results in errors. 

The error message from the backend would have been really helpful in resolving this problem, but we def should fix up how this interacts with Undo!

naterush:
Here's another way we could handle the cached snowflake credentials (that I think would be super sweet!):
1. Store them in the userProfile
2. Set them in the API call on the steps_manager

The benefit of this: if they were defined in the userProfile, when you open the SnowflakeImportTaskpane, we could skip over showing that screen (or indeed, having to make any API calls other than loading the defaults). This would also dramatically simplify the `AuthenticateToSnowflakeCard`, removing an entire API call from this (and removing the useEffect entirely!). 

Moreover, we'd remove the api call to get the chached credentials entirely. 

Getting the data to the step performer gets a little more challenging, but I think it's be reasonable to add the credentials to some state somewhere so you could access them there. 

Thoughts? A bit of a refactor, but I think the simplification across the frontend feels like it might be worth it to me.

mitosheet/src/components/taskpanes/FileImport/FileImportTaskpane.tsx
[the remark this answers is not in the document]
aarondr77: I think connecting to snowflake is what the users are thinking about doing, so `authenticate_to_snowflake` makes more sense to me

mitosheet/src/components/elements/AuthenticateToSnowflakeCard.tsx
aarondr77: This is cleaned up due to moving the credentials into the userProfiel

mitosheet/src/components/taskpanes/UpdateImports/UpdateImportCard.tsx
```diff
@@ -156,6 +201,7 @@ const UpdateImportCard = (props: {
                         closeDropdown={() => closeDropdown()}
                         width='medium'
                     >
+                        {getSnowflakeDropddownItemOrEmptyElement()}
```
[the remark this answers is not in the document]
aarondr77: Updated, but I'm not sure what you mean by the "the type system tries to encourage you to do just that". The type system just says `children: JSX.Element | JSX.Element[];`. This is extremely vague.

aarondr77:
TODO
- Put it behind enterprise wall
- Figure out best way to update the user.json through the api call so that the credentials are saved if the user opens taskpane, validates, closes taskpane, reopens it. 
- Optimize for dataframe delete, rename

aarondr77:
Opened a GitHub issue for minor credentials improvement

mitosheet/mitosheet/api/get_validate_snowflake_credentials.py
```diff
@@ -17,19 +17,36 @@
 except ImportError:
     SNOWFLAKE_CONNECTOR_IMPORTED = False
 
-def _validate_snowflake_credentials(username: str, password: str, account: str) -> bool:
+# Global variable used to cache the snowflake credentials so that users
+# only need to eneter credentials once were kernel's lifespan. This global variable is accessed
```
aarondr77, line 21: eneter -> enter

aarondr77: were -> per

mitosheet/package.json
```diff
@@ -68,10 +68,7 @@
     "lint": "eslint src/ --ext .ts,.tsx --fix",
 
     "prepare": "jlpm run clean && jlpm run build:all",
-    "prepack": "npm run build:all",
-
-    "build:docs:update_frontend": "python docs/make_function_docs.py update_frontend",
```
aarondr77, line 73: This was somehow duplicated in this PR, so just removing one of the duplicated sets...

naterush:
I can access this feature on non-pro accounts, just by searching for Snowflake in the help. I personally think we should do what we did with formatting, where:
1. We turn the taskpane to `requiresPro` (so I can't use it if I'm not Pro)
2. We make this visible in the Import dropdown

It has the benefit of communicating the value of Pro to users!

mitosheet/src/components/taskpanes/UpdateImports/updateImportsUtils.tsx
```diff
@@ -98,6 +98,31 @@ export const updateDataframeCreation = (
     setReplacingDataframeState(undefined);
 }
 
+export const updateAllSnowflakeImports = (
```
naterush, line 101: TODO: test this.

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -109,18 +110,42 @@ const getNewParams = (prevParams: SnowflakeImportParams, database?: string | nul
...
+            // to load and set the available options and defaults
+            let newParams: SnowflakeImportParams | undefined = undefined
+            if (typeof newParamsOrFunction === 'function') {
+                newParams = newParamsOrFunction()
+            } else {
+                newParams = newParamsOrFunction
+            }
+            if (newParams !== undefined) {
+                void loadAndSetOptionsAndDefaults(newParams)
+            }
```
naterush, line 130: This is an example of really bad API design (that you inherited!) as a function of types that I picked when I designed this. The consumer has to do a bunch of work to check if the new params are a type of function -- but in practice (if you actually look at what is returned), they never will be. 

If you just:
1. Change the type of `onUndoAndRedo?: (params: typeof defaultParams) => void` to have `params: ParamType`
2. Change `await mitoAPI.getParams<typeof defaultParams>(stepType, stepID, {});` to take `ParamType` as well.

Then, you don't need to do this checking. Which is... much more plesant!

naterush:
A bug:
1. Import twice from same table
2. Delete second dataframe
3. All code gets optimized out.

I think this is because `get_created_sheet_indexes` is now wrong! If there are two queries, it's should return two indexes!

naterush:
^ can you also add a test for this, so we catch it in the future!

naterush:
1. Authenicate
2. Start importing
3. Realize you signed in with the wrong account with wrong permissions
4. Go back to authenticating and change the params

If you authenticate incorrectly (or correctly), it doesn't invalidate the warehouse, etc params. Easy fix I think.

naterush:
I'd be down to lower case the dataframe names. Given that tables (as far as I can tell) always are, but that's really awful python code!

naterush:
Just because there was that semi-critical dataframe delete bug, I'll do one final check once you clean the above. Then, we'll get this merged! Almost there! Thanks for carrying this over the finish line.

aarondr77:
Discovered an existing bug in code optimization with deletes
https://github.com/mito-ds/monorepo/issues/619

naterush:
Do we want to add Snowflake Import to the Import button for pro users at least? I think yes.

PR #621, opened 2023-02-02
Fix spreadsheet formula bugs

naterush:
# Description

Fixes the bugs we found in usability testing:
1. Sheet crashing during column header renames
2. Taking function suggestion would add index header
3. Not being able to see the full summary statistic.

# Testing

Test each of these!

# Documentation

No.


aarondr77:
Functionality is working well, once tests pass LGTM

PR #626, opened 2023-02-03
mitosheet: allow users to set constants in formulas

naterush:
# Description

Set a specific cell index to a constant. 

# Testing

Try doing it. Note that we don't do type casting, which is a bit of a regression, but I think it's okey...
# Documentation

N/a


aarondr77:
I think we meant to fix up these comments in the previous PR, can you try to make them look a bit nicer?

Screenshot 2023-02-03 at 3 17 08 PM

PR #638, opened 2023-02-16
mitosheet: add prompt that never is too long

naterush:
# Description

Adds a `get_prompt` function that always is within the length bounds, intelligently. We will, of course, continue to improve this over time.

# Testing

Add some other tests.

# Documentation

No.

### DIFF
--- a/mitosheet/mitosheet/ai/prompt.py
+++ b/mitosheet/mitosheet/ai/prompt.py
@@ -123,7 +123,7 @@ df.replace(1, 2, inplace=True)
 ```"""
 
 
-def get_prompt(df_names: List[str], dfs: List[pd.DataFrame], current_selection: Optional[Selection], user_input: str):
+def get_prompt(df_names: List[str], dfs: List[pd.DataFrame], current_selection: Optional[Selection], user_input: str) -> str:
 
     if len(dfs) > 0:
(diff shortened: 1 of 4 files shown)


mitosheet/mitosheet/ai/prompt.py
```diff
@@ -0,0 +1,150 @@
...
+```"""
+    else:
+        return f"""Transformation: replace 1 with 2 in df
+Code:
+```
+df.replace(1, 2, inplace=True)
+```"""
+
+
+def get_prompt(df_names: List[str], dfs: List[pd.DataFrame], current_selection: Optional[Selection], user_input: str) -> str:
```
aarondr77, line 126: This only supports a single selection, but we let users have many selections. Do we want to support that? ie: "Apply this formula to all selected columns" could be a common command?

aarondr77 (commented):
Nice! This was pleasant to review since its such a small chunk!

mitosheet/mitosheet/ai/prompt.py
naterush: The selection object on the backend is very different than frontend. We're just going to send all the selected columns and rows to the backend, and that will be considered one selection.

naterush:
This is ready to merge when tests pass.

PR #639, opened 2023-02-16
Ai transformation pr 2

naterush:
# Description

Adds the `get_completion` PR with helpful errors. Nice.

# Testing

Add a new test, if you so wish.

# Documentation

No.


mitosheet/mitosheet/api/get_completion.py
```diff
@@ -0,0 +1,65 @@
...
+import os
+
+from mitosheet.ai.prompt import get_prompt
+
+
+import requests
+
+URL = 'https://api.openai.com/v1/completions'
+
+def get_completion(params: Dict[str, Any], steps_manager: StepsManagerType) -> str:
```
aarondr77, line 19: Maybe we can call this `get_ai_completion`.

mitosheet/mitosheet/api/get_completion.py
```diff
@@ -0,0 +1,65 @@
...
+                steps_manager.curr_step.final_defined_state.dfs,
+                selection,
+                user_input
+        )
+
+        OPENAI_API_KEY = os.environ.get('OPENAI_API_KEY')
+
+        if OPENAI_API_KEY is None:
+                return json.dumps({
+                        'error': 'Please aquire an OPENAI_API_KEY and set it as an enviornmental variable.'
```
aarondr77, line 34: aquire -> acquire
enviornmental -> environmental

It would be nice if we could point them to some documentation on how to do this. Since a lot of our users are new to Python, this is probably not enough guidance.

mitosheet/mitosheet/api/get_completion.py
```diff
@@ -0,0 +1,65 @@
...
+                })
+
+        data = {
+                "model": "text-davinci-003",
+                "prompt": prompt,
+                "max_tokens": 7,
+                "temperature": .2,
```
aarondr77, line 41: We should try testing some different temperatures at some point. OpenAI's playground uses a temperature of 0 for their Natural Language to Python playground.

mitosheet/mitosheet/api/get_completion.py
```diff
@@ -0,0 +1,65 @@
...
+
+        OPENAI_API_KEY = os.environ.get('OPENAI_API_KEY')
+
+        if OPENAI_API_KEY is None:
+                return json.dumps({
+                        'error': 'Please aquire an OPENAI_API_KEY and set it as an enviornmental variable.'
+                })
+
+        data = {
+                "model": "text-davinci-003",
```
aarondr77, line 38: We want to be using `code-davinci-002`

aarondr77 (commented):
LGTM after these fixes!

.github/workflows/test-mitosheet-python.yml
```diff
@@ -54,3 +54,4 @@ jobs:
         PYTEST_SNOWFLAKE_USERNAME: ${{ secrets.PYTEST_SNOWFLAKE_USERNAME }}
         PYTEST_SNOWFLAKE_PASSWORD: ${{ secrets.PYTEST_SNOWFLAKE_PASSWORD }}
         PYTEST_SNOWFLAKE_ACCOUNT: ${{ secrets.PYTEST_SNOWFLAKE_ACCOUNT }}
+        OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
```
aarondr77, line 57: Note that this only tests the case when the `OPENAI_API_KEY` is set. So we don't have any tests running in our CI when its not set

naterush: Since we don't conditionally install a dependency like snowflake, I'm fine with this!

mitosheet/mitosheet/api/get_completion.py
naterush: Yep, we can do this on the frontend!

PR #651, opened 2023-03-05
Ai transform usability tests fix

naterush:
# Description

Fixes: https://www.notion.so/trymito/AI-Usability-Test-dfac820795364684b37d7f3896a2a3d5?pvs=4

# Testing

Test all of these! Lots of changes, so I'm sure there's some new stuff.

# Documentation

No.


mitosheet/mitosheet/ai/ai_utils.py
```diff
@@ -40,17 +47,25 @@ def get_code_string_from_last_expression(code: str, last_expression: ast.stmt) -
...
-
-    last_expression_string = get_code_string_from_last_expression(code, last_expression)
+    
+    last_expression_string = get_code_string_from_last_expression(code)
 
+    if last_expression_string is None:
+        return code
+    
+    # We only want to replace the last instance, so we have to 
+    # reverse the string, replace the dataframe name with the reverse, and then reverse it back again
```
aarondr77, line 57: This comment should probably be moved down to the `replace_last_instance_in_string` function

mitosheet/mitosheet/ai/recon.py
```diff
@@ -141,14 +156,36 @@ def get_column_recon_data(old_df: pd.DataFrame, new_df: pd.DataFrame) -> ColumnR
     removed_columns = [ch for ch in old_columns_without_shared if ch not in renamed_columns]
 
     shared_columns = list(filter(lambda ch: ch in new_columns, old_columns))
-    modified_columns = [ch for ch in shared_columns if not old_df[ch].equals(new_df[ch])]
+
+    if not rows_added_or_removed:
+        modified_columns = [ch for ch in shared_columns if not old_df[ch].equals(new_df[ch])]
+    else:
+        # If rows were added or removed, then we don't want to detect every column as having changed
```
aarondr77, line 163: Ooo smart!

mitosheet/mitosheet/tests/ai/test_recon.py
```diff
@@ -320,138 +397,200 @@ def test_exec_for_recon(code, dfs, recon_data):
...
+            'column_recon': {
+                'created_columns': [],
+                'deleted_columns': [],
+                'modified_columns': [],
+                'renamed_columns': {}
+            }, 
+            'num_added_or_removed_rows': 0
+        }
+    ),
+    # Add rows, only sees rows as added and nothing modified
```
aarondr77, line 559: If a user adds/deletes rows and adds/deletes a dataframe we should show them that. Can we add a test?

mitosheet/src/components/taskpanes/AITransformation/AITransformationResultSection.tsx
```diff
@@ -97,7 +106,11 @@ const AITransformationResultSection = (props: AITransformationResultSectionProps
                     </div>
                 )
             })}
-            {!(result.last_line_value !== undefined && result.last_line_value !== null) && result.created_dataframe_names.length === 0 && Object.entries(result.modified_dataframes_column_recons).length === 0 && result.deleted_dataframe_names.length === 0 && 
+            {!(result.last_line_value !== undefined && result.last_line_value !== null)
```
aarondr77, line 109: We can change `{!(result.last_line_value !== undefined && result.last_line_value !== null) `
to `{(result.last_line_value === undefined || result.last_line_value === null)`

- result.last_line_value = undefined => true
- result.last_line_value = null => true 
- result.last_line_value  = something else => false

aarondr77:
"Instead of saving the prompt history in the taskpane, it would be nice to save it in the Mito state so when users reopen the taskpane, they can rerun a previous prompt." is checked off in the doc, but I don't see it implemented.

aarondr77:
Can we use the AI logo here and in the loading indicator 
Screenshot 2023-03-05 at 3 12 14 PM

aarondr77:
The `df.info()` works great! Can we make these work similarly `df.columns`, `df.shape`. Right now they add generated code, but they don't show up in the results section or in the mitosheet. 

Screenshot 2023-03-05 at 3 19 37 PM

aarondr77:
If there is an error here and the user updates the prompt and generates new code, we should clear the error

Screenshot 2023-03-05 at 3 27 11 PM

aarondr77:
When I hover over the examples, it would be nice to have the color change a bit like we do elsewhere with clickable things.

aarondr77:
It would be nice if the AI taskpane did not change when I renamed a column header, set a column formula, etc.

aarondr77:
After successfully executing code, it would be nice if there was a `successfully ran code` message for 2 seconds. It's not super obvious that anything happened in a lot of cases since the results section is hard to see.

PR #652, opened 2023-03-05
New privacy policy

naterush:
# Description

# Testing

# Documentation


mitosheet/mitosheet/user/upgrade.py
```diff
@@ -199,6 +199,18 @@ def upgrade_user_json_version_7_to_8(user_json_version_7: Dict[str, Any]) -> Dic
...
     return user_json_version_7
 
+def upgrade_user_json_version_8_to_9(user_json_version_8: Dict[str, Any]) -> Dict[str, Any]:
+    """
+    Adds the UJ_AI_PRIVACY_POLICY key, and sets it to False
+    """
+    # First, bump the version number
+    user_json_version_8[UJ_USER_JSON_VERSION] = 9
+
+    # Then, add the new enterprise key
```
aarondr77, line 209: comment out of date

mitosheet/src/components/taskpanes/AITransformation/AITransformationTaskpane.tsx
```diff
@@ -221,9 +251,31 @@ const AITransformationTaskpane = (props: AITransformationTaskpaneProps): JSX.Ele
...
+                        <p className="text-body-1">
+                            You can see Mito's <a className='text-underline' href='https://privacy.trymito.io/privacy-policy' target='_blank' rel="noreferrer">privacy policy here.</a>
+                        </p>
+                        <TextButton
+                            onClick={() => {
+                                void props.mitoAPI.updateAcceptAITransformationPrivacyPolicy();
+                            }}
+                            variant='dark'
+                        >
+                            Accept Privacy Policy for Mito AI
```
aarondr77, line 274: When they accept, can we log it? Seems like something that we'd want a record of.

aarondr77 (commented):
LGTM once tests pass. Before we do a wide-spread public release, I think we can improve the aesthetics of this privacy policy page to be more polished, transparent, and professional. Which might be particularly important for specific audiences. 

Relatedly, we'll probably want to do the work to give users more customization of what data they use. In my convo with a user on Friday, he thought that he wouldn't be able to use this on specific datasets because of the data collection. 

Something closer to the mockups you generated.

mitosheet/src/components/taskpanes/AITransformation/AITransformationTaskpane.tsx
naterush: We do log it -- all update events are logged by default.

naterush:
Sounds great. Looking forward to iterating on all of these things with the AI feature going forward!

PR #656, opened 2023-03-09
Add variables for turning off display of SQL and AI Transform in toolbar.

naterush:
# Description

It leaves the in the dropdown, which is a nice compromise.

# Testing

Set the variables.

# Documentation

We can define them.


aarondr77 (commented):
I think given our recent conversations, I think we should make AI transform and Snowflake not appear if on mitosheet-private at least. 

And maybe in addition, AI transform should not appear if on Mito Enterprise

PR #660, opened 2023-03-13
mitosheet: add get previous value to interface

naterush:
# Description

# Testing

# Documentation

### DIFF
--- a/mitosheet/mitosheet/public/v1/sheet_functions/misc_functions.py
+++ b/mitosheet/mitosheet/public/v1/sheet_functions/misc_functions.py
@@ -15,6 +15,7 @@ NOTE: This file is alphabetical order!
 import pandas as pd
 import numpy as np
 import datetime
+from mitosheet.is_type_utils import is_bool_dtype, is_float_dtype, is_int_dtype, is_string_dtype
 
 from mitosheet.public.v1.sheet_functions.sheet_function_utils import try_extend_series_to_index
 from mitosheet.public.v1.sheet_functions.types.decorators import filter_nans, convert_arg_to_series_type, handle_sheet_function_errors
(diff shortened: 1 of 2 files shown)


mitosheet/mitosheet/public/v1/sheet_functions/misc_functions.py
```diff
@@ -113,8 +115,63 @@ def get_element_type(element):
...
+    'error',
+    'error'
+)
+@convert_arg_to_series_type(
+    1,
+    'bool', 
+    'error', 
+    ('default', True)
+)
+def GETPREVIOUSVALUE(series: pd.Series, condition: pd.Series) -> pd.Series:
```
aarondr77, line 132: This makes sense for the v1/2 interface. When we move to v3, maybe we should make the following updates: `series: pd.Series` in our new world should probably be `series: pd.DataFrame` since the Excel syntax should ideally look like `A:A` instead of `A0`. 

The question is how would we handle the `condition`. In Excel, it would be something like this: We'll probably want to take whatever we do for COUNTIF and SUMIF

`GETPREVIOUSVALUE(Max_Balances:Max_Balances, "> 0")`

mitosheet/mitosheet/public/v1/sheet_functions/misc_functions.py
```diff
@@ -113,8 +115,63 @@ def get_element_type(element):
...
+    """
+
+    # Default to a different last occurence depending on the type
+    column_dtype = str(series.dtype)
+    last_occurrence: Any = -1
+    if is_int_dtype(column_dtype) or is_float_dtype(column_dtype):
+        last_occurrence = -1
+    elif is_string_dtype(column_dtype):
+        last_occurrence = ''
+    elif is_bool_dtype(column_dtype):
```
aarondr77, line 160: For datetimes, maybe we can make the default `NaT`? -1 isn't great

mitosheet/mitosheet/public/v1/sheet_functions/misc_functions.py
```diff
@@ -113,8 +115,63 @@ def get_element_type(element):
...
+        "function": "GETPREVIOUSVALUE",
+        "description": "Returns the value from series that meets the condition.",
+        "search_terms": ["ffill"],
+        "examples": [
+            "GETPREVIOUSVALUE(Max_Balances, Max_Balances > 0)",
+        ],
+        "syntax": "GETPREVIOUSVALUE(series, condition)",
+        "syntax_elements": [{
+                "element": "series",
+                "description": "The series to get the previous value from.."
```
aarondr77: Remove the double ..

mitosheet/mitosheet/tests/public/v1/sheet_functions/misc_functions/test_getpreviousvalue.py
```diff
@@ -0,0 +1,44 @@
...
+        pd.Series([1, 2, 3]), 
+        pd.Series([False, True, False]),  
+        pd.Series([-1, 2, 2])
+    ),
+    (
+        pd.Series(['a', 'b', 'c']), 
+        pd.Series([False, True, False]),  
+        pd.Series(['', 'b', 'b'])
+    ),
+    (
```
aarondr77, line 36: Try with other dtypes

aarondr77 (commented):
LGTM

PR #669, opened 2023-03-18
mitosheet: add LEFT function, error handling

naterush:
# Description

This PR adds a LEFT function, with some really nice error handling infrastructure that does dynamic type checking using the specified types. 

This is kinda epic.

# Testing

See tests.

# Documentation

No.

### DIFF
--- a/mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
+++ b/mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
@@ -8,18 +8,79 @@ Contains all functions that can be used in a sheet that operate on strings.
 
 NOTE: This file is alphabetical order!
 """
-from typing import Union
+from typing import Optional, Union
 import sys
 import numpy as np
 
 import pandas as pd
 
+from mitosheet.public.v3.errors import handle_sheet_function_errors
 from mitosheet.public.v3.sheet_functions.utils import get_final_result_series_or_primitive
-from mitosheet.public.v3.types.decorators import cast_values_in_arg_to_type
-from mitosheet.public.v3.types.sheet_function_types import StringFunctionReturnType
+from mitosheet.public.v3.types.decorators import cast_values_in_args_to_type, cast_values_in_arg_to_type
+from mitosheet.public.v3.types.sheet_function_types import StringInputType, StringRestrictedInputType, StringFunctionReturnType, IntRestrictedInputType
 
-@cast_values_in_arg_to_type('str')
-def CONCAT(*argv: pd.Series) -> StringFunctionReturnType:
+
+@cast_values_in_arg_to_type('string', 'str')
+@cast_values_in_arg_to_type('num_chars', 'int')
+@handle_sheet_function_errors
+def LEFT(string: StringRestrictedInputType, num_chars: IntRestrictedInputType=None) -> StringFunctionReturnType:
+    """
+    {
+        "function": "LEFT",
+        "description": "Returns a substring from the beginning of a specified string.",
+        "search_terms": ["left"],
+        "examples": [
+            "LEFT(A, 2)",
+            "LEFT('The first character!')"
+        ],
+        "syntax": "LEFT(string, [number_of_characters])",
+        "syntax_elements": [{
+                "element": "string",
+                "description": "The string or series from which the left portion will be returned."
+            },
+            {
+                "element": "number_of_characters [OPTIONAL, 1 by default]",
+                "description": "The number of characters to return from the start of string."
+            }
+        ]
+    }
+    """
+
+    if string is None:
+        return ''
+    else:
+
+        if num_chars is None:
+            num_chars = 1
+        elif isinstance(num_chars, pd.Series):
+            num_chars = num_chars.fillna(1)
+
+        if isinstance(string, str):
+            if isinstance(num_chars, int):
+                return string[:num_chars]
+            else:
+                print(num_chars)
+                return pd.Series(
+                    [string[:nc] for nc in num_chars],
+                    index=num_chars.index
+                )
+        else:
+            if isinstance(num_chars, int):
+                print("HERE!", num_chars, [s[:num_chars] for s in string])
+                return pd.Series(
+                    [s[:min(num_chars, len(s))] for s in string],
+                    index=string.index
+                )
+            else:
+                return pd.Series(
+                    [s[:nc] for s, nc in zip(string, num_chars)],
+                    index=string.index
+                )
+
+
+@cast_values_in_args_to_type('str')
+@handle_sheet_function_errors
+def CONCAT(*argv: StringInputType) -> StringFunctionReturnType:
     """
     {
         "function": "CONCAT",
--- a/mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
+++ b/mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
@@ -16,7 +16,7 @@ import pandas as pd
 
 from mitosheet.public.v3.errors import handle_sheet_function_errors
 from mitosheet.public.v3.sheet_functions.utils import get_final_result_series_or_primitive
-from mitosheet.public.v3.types.decorators import cast_values_in_args_to_type, cast_values_in_arg_to_type
+from mitosheet.public.v3.types.decorators import cast_values_in_all_args_to_type, cast_values_in_arg_to_type
 from mitosheet.public.v3.types.sheet_function_types import StringInputType, StringRestrictedInputType, StringFunctionReturnType, IntRestrictedInputType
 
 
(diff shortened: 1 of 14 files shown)


mitosheet/mitosheet/public/v3/types/sheet_function_types.py
```diff
@@ -6,7 +6,26 @@
 
 import pandas as pd
 
+from mitosheet.public.v3.rolling_range import RollingRange
 
-NumberFunctionReturnType = Union[pd.Series, int, float]
-BoolFunctionReturnType = Union[pd.Series, bool]
+# Input types
```
aarondr77, line 11: These seems super useful!

mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
```diff
@@ -8,18 +8,79 @@
...
+        if num_chars is None:
+            num_chars = 1
+        elif isinstance(num_chars, pd.Series):
+            num_chars = num_chars.fillna(1)
+
+        if isinstance(string, str):
+            if isinstance(num_chars, int):
+                return string[:num_chars]
+            else:
+                print(num_chars)
```
aarondr77, line 62: remove

mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
```diff
@@ -8,18 +8,79 @@
...
+                return pd.Series(
+                    [string[:nc] for nc in num_chars],
+                    index=num_chars.index
+                )
+        else:
+            if isinstance(num_chars, int):
+                print("HERE!", num_chars, [s[:num_chars] for s in string])
```
aarondr77, line 69: remove

mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
```diff
@@ -8,18 +8,79 @@
...
+                "element": "number_of_characters [OPTIONAL, 1 by default]",
+                "description": "The number of characters to return from the start of string."
+            }
+        ]
+    }
+    """
+
+    if string is None:
+        return ''
+    else:
```
aarondr77, line 51: Let's get rid of this else since we're returning in the previous case, so its less nested

mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
```diff
@@ -8,18 +8,79 @@
...
+from mitosheet.public.v3.types.decorators import cast_values_in_args_to_type, cast_values_in_arg_to_type
+from mitosheet.public.v3.types.sheet_function_types import StringInputType, StringRestrictedInputType, StringFunctionReturnType, IntRestrictedInputType
 
-@cast_values_in_arg_to_type('str')
-def CONCAT(*argv: pd.Series) -> StringFunctionReturnType:
+
+@cast_values_in_arg_to_type('string', 'str')
+@cast_values_in_arg_to_type('num_chars', 'int')
+@handle_sheet_function_errors
+def LEFT(string: StringRestrictedInputType, num_chars: IntRestrictedInputType=None) -> StringFunctionReturnType:
```
aarondr77, line 26: On our path to handling these functions more like Excel, let's change the parameters to match Excel. This touches the docs, so we can leave it till later again. 

Screenshot 2023-03-18 at 11 25 06 AM

mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
```diff
@@ -8,18 +8,79 @@
...
+    """
+
+    if string is None:
+        return ''
+    else:
+
+        if num_chars is None:
+            num_chars = 1
+        elif isinstance(num_chars, pd.Series):
+            num_chars = num_chars.fillna(1)
```
aarondr77, line 56: Excel handles empty values num_chars as returning empty. 

Screenshot 2023-03-18 at 11 28 06 AM

mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
```diff
@@ -8,18 +8,79 @@
...
+        else:
+            if isinstance(num_chars, int):
+                print("HERE!", num_chars, [s[:num_chars] for s in string])
+                return pd.Series(
+                    [s[:min(num_chars, len(s))] for s in string],
+                    index=string.index
+                )
+            else:
+                return pd.Series(
+                    [s[:nc] for s, nc in zip(string, num_chars)],
```
aarondr77, line 76: TODO: Make sure there is a text here where the num_chars in the series is longer than the string, since we are not doing the min check

mitosheet/mitosheet/public/v3/types/decorators.py
```diff
@@ -6,52 +6,52 @@
...
 from typing import Any, Callable
 
 import pandas as pd
 
 from mitosheet.public.v3.rolling_range import RollingRange
-from mitosheet.public.v3.types.utils import CONVERSION_FUNCTIONS
+from mitosheet.public.v3.types.utils import CONVERSION_FUNCTIONS, get_arg_cast_to_type
 from mitosheet.types import PrimitiveTypeName
 
+def cast_values_in_args_to_type(
```
aarondr77, line 18: I'd vote we call this `cast_values_in_all_args_to_type`

mitosheet/mitosheet/public/v3/types/decorators.py
```diff
@@ -6,52 +6,52 @@
...
-        isinstance(value, timedelta)
-        # TODO: nan here?
 
 def cast_values_in_arg_to_type(
+    arg_name: str,
     target_primitive_type_name: PrimitiveTypeName,
 ) -> Callable:
-
     def wrap(sheet_function):
+        print(sheet_function)
```
aarondr77, line 42: remove

mitosheet/mitosheet/public/v3/types/int.py
```diff
@@ -18,6 +21,9 @@ def cast_to_int(unknown: Union[str, int, float, bool, datetime, timedelta]) -> O
     elif isinstance(unknown, int):
         return unknown
     elif isinstance(unknown, float):
+        if np.isnan(unknown):
```
aarondr77, line 24: Instead of handling casting errors like this, what do you think about wrapping `cast_to_int` in a try, catch?

mitosheet/mitosheet/tests/public/v3/sheet_functions/string_functions/test_left.py
```diff
@@ -0,0 +1,57 @@
...
+    ([pd.Series([10000, 10.0, True, datetime(1997, 12, 22), timedelta(days=1)]), pd.Series([0, 1, 2, 3, 4, 5])], pd.Series(['', '1', 'Tr', '199', '1 da'])),
+    ([pd.Series([10000, 10.0, True, datetime(1997, 12, 22), timedelta(days=1)]), pd.Series([1000, 1000, 1000, 1000, 1000, 1000])], pd.Series(['10000', '10.0', 'True', '1997-12-22 00:00:00', '1 day, 0:00:00'])),
+]
+
+# There aren't really any invalid types for strings
+
+@pytest.mark.parametrize("_argv,expected", LEFT_VALID_TESTS)
+def test_concat(_argv, expected):
+    result = LEFT(*_argv)
+    print(result)
```
aarondr77, line 36: remove

mitosheet/mitosheet/tests/public/v3/sheet_functions/string_functions/test_left.py
```diff
@@ -0,0 +1,57 @@
...
+    if isinstance(result, pd.Series):
+        assert result.equals(expected)
+    else: 
+        assert result == expected
+
+
+
+
+# TODO: test that NaN values will lead to errors in the num_chars, but not elsewhere
```
aarondr77, line 45: I think we handle NaN values in num_chars by setting them = 1 right now, so this shouldn't error

mitosheet/mitosheet/tests/public/v3/sheet_functions/string_functions/test_left.py
```diff
@@ -0,0 +1,57 @@
...
+LEFT_INVALID_TESTS: Any = [
+    # Dataframes
+    ([pd.DataFrame({'a': ['a', 'b', 'c'], 'b': [1, 2, 3]}), pd.Series(['e','f','g'])]),
+
+    # Rolling ranges
+    ([RollingRange(pd.DataFrame({'B': ['a', 'b', 'c'], 'C': [1, 2, 3]}), 2, 0), True]),
```
aarondr77, line 51: I think each of these tests should actually be 2 tests. By looking at these tests, its not clear why they should be failing:

The first test should fail, it seems, both because LEFT does not support dataframe arguments, but also because the num_chars cannot be letters. 

Similar to the next one, we both don't support Rolling Ranges as the text or True as the num chars

aarondr77 (commented):
Looks mostly great! Mostly just some small cleanup and then LGTM!

mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
naterush: We can rename all of these when we do a docs pass.

mitosheet/mitosheet/public/v3/types/int.py
naterush: That's so much better!

PR #670, opened 2023-03-19
mitosheet: handle more formatting of mitosheet.sheet() call

aarondr77:
# Description

Handles more versions of the mitosheet.sheet call. 

We used to require that it be `analysis_to_replay="${analysis_name}")`. We now support the following:
- `analysis_to_replay="${analysis_name}")`
- `analysis_to_replay = "${analysis_name}")`
- `analysis_to_replay= "${analysis_name}")`
-`analysis_to_replay = "${analysis_name}" )`

# Testing

Test creating a new mitosheet and test starting a new analysis from the failed to replay analysis due to missing import screen.  

# Documentation

No.

### DIFF
--- a/mitosheet/src/utils/code.tsx
+++ b/mitosheet/src/utils/code.tsx
@@ -131,7 +131,9 @@ export function isMitoAnalysisCode(codeText: string): boolean {
     Returns true if the cell contains a mitosheet.sheet(analysis_to_replay={analysisName})
 */
 export function containsMitosheetCallWithSpecificAnalysisToReplay(codeText: string, analysisName: string): boolean {
-    return codeText.includes('sheet(') && codeText.includes(`analysis_to_replay="${analysisName}"`)
+    // Remove any whitespace from codeText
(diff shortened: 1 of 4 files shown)


mitosheet/src/utils/code.tsx
```diff
@@ -131,15 +131,19 @@ export function isMitoAnalysisCode(codeText: string): boolean {
...
 
 
 /* 
     Returns true if the cell contains a mitosheet.sheet(analysis_to_replay={analysisName})
 */
 export function containsMitosheetCallWithAnyAnalysisToReplay(codeText: string): boolean {
-    return isMitosheetCallCode(codeText) && codeText.includes(`analysis_to_replay=`)
+    // Remove any whitespace from codeText
+    const codeTextCleaned = codeText.replace(/\s/g, '');
+    return isMitosheetCallCode(codeTextCleaned) && codeTextCleaned.includes(`analysis_to_replay=`)
```
naterush, line 146: I think we should call the isMitosheetCallCode on the original text here - so we aren't making assumptions on how it handles things.

PR #671, opened 2023-03-22
V3 string functions

naterush:
# Description

Adds v3 of string functions. 

# Testing

I tried to test every combo, but I am sure I forgot some. Maybe try and add 2 tests per?

# Documentation

N/a.

### DIFF
--- a/mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
+++ b/mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
@@ -294,45 +294,51 @@ def MID(string: StringRestrictedInputType, start_loc: IntRestrictedInputType, nu
     elif isinstance(num_chars, pd.Series):
         num_chars = num_chars.fillna(0)
 
-    if isinstance(string, str):
-        if isinstance(start_loc, int) and isinstance(num_chars, int):
-            return string[start_loc - 1:start_loc + num_chars - 1]
-        elif isinstance(start_loc, int) and isinstance(num_chars, pd.Series):
-            return pd.Series(
-                [string[start_loc-1:start_loc+nc-1] for nc in num_chars],
-                index=num_chars.index
-            )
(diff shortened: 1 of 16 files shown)


mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
naterush, line 1: Very interestingly, we only gained about 20 lines of code. But we add so much more functionality that works so much better - I'm pretty proud of this refactor.

mitosheet/mitosheet/public/v3/sheet_functions/string_functions.py
```diff
@@ -73,46 +189,398 @@ def LEFT(string: StringRestrictedInputType, num_chars: IntRestrictedInputType=No
...
-        lambda previous_value, new_value: previous_value + new_value,
-        lambda previous_series, new_series: previous_series + new_series
-    )
+    if string is None:
+        return ''
+    elif isinstance(string, pd.Series):
+        string = string.fillna('')
+
+    if start_loc is None:
+        start_loc = 0
```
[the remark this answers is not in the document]
aarondr77: Oh wait. I think I might be an idiot and it is possible for these functions to receive `None` as the parameters. `FIND(None, None)` does not error.

mitosheet/mitosheet/tests/public/v3/sheet_functions/string_functions/test_clean.py
```diff
@@ -0,0 +1,55 @@
...
+
+from mitosheet.public.v3.sheet_functions.string_functions import LEN
+
+
+LEN_VALID_TESTS: Any = [
+    # Just constant tests
+    (['A'], 1),
+    (['ABC'], 3),
+    ([1.0], 3),
+    ([True], 4),
```
aarondr77, line 23: We should add tests where there are None constants all of the parameters

PR #672, opened 2023-03-24
V3 date functions

naterush:
# Description

# Testing

# Documentation

### DIFF
--- a/mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
+++ b/mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
@@ -20,7 +20,7 @@ import pandas as pd
 from mitosheet.public.v3.errors import handle_sheet_function_errors
 from mitosheet.public.v3.sheet_functions.utils import get_final_result_series_or_primitive, get_series_from_primitive_or_series
 from mitosheet.public.v3.types.decorators import cast_values_in_all_args_to_type, cast_values_in_arg_to_type
-from mitosheet.public.v3.types.sheet_function_types import AnyPrimitiveOrSeriesInputType, AnySeriesFunctionReturnType, AnySeriesInputType, BoolFunctionReturnType, BoolInputType, BoolRestrictedInputType
+from mitosheet.public.v3.types.sheet_function_types import AnyPrimitiveOrSeriesInputType, BoolFunctionReturnType, BoolInputType, BoolRestrictedInputType
 
 
 @cast_values_in_all_args_to_type('bool')
@@ -86,7 +86,7 @@ def BOOL(series: BoolRestrictedInputType) -> BoolFunctionReturnType:
 
 @cast_values_in_arg_to_type('condition', 'bool')
 @handle_sheet_function_errors
-def IF(condition: AnySeriesInputType, true_series: AnyPrimitiveOrSeriesInputType, false_series: AnyPrimitiveOrSeriesInputType) -> AnySeriesFunctionReturnType:
+def IF(condition: pd.Series, true_series: AnyPrimitiveOrSeriesInputType, false_series: AnyPrimitiveOrSeriesInputType) -> pd.Series:
     """
     {
         "function": "IF",
--- a/mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
+++ b/mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
@@ -19,9 +19,9 @@ import pandas as pd
 
 from mitosheet.public.v3.rolling_range import RollingRange
 from mitosheet.public.v3.errors import handle_sheet_function_errors
-from mitosheet.public.v3.sheet_functions.utils import get_final_result_series_or_primitive
-from mitosheet.public.v3.types.decorators import cast_values_in_all_args_to_type
-from mitosheet.public.v3.types.sheet_function_types import BoolFunctionReturnType, BoolInputType
+from mitosheet.public.v3.sheet_functions.utils import get_final_result_series_or_primitive, get_series_from_primitive_or_series
+from mitosheet.public.v3.types.decorators import cast_values_in_all_args_to_type, cast_values_in_arg_to_type
+from mitosheet.public.v3.types.sheet_function_types import AnyPrimitiveOrSeriesInputType, AnySeriesFunctionReturnType, AnySeriesInputType, BoolFunctionReturnType, BoolInputType
 
 
 @cast_values_in_all_args_to_type('bool')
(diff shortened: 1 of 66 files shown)


mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
```diff
@@ -60,10 +59,73 @@ def AND(*argv: BoolInputType) -> BoolFunctionReturnType:
...
+        ],
+        "syntax": "BOOL(series)",
+        "syntax_elements": [{
+                "element": "series",
+                "description": "An series to convert to boolean values, either True or False."
+            }
+        ]
+    }
+    """
+    if isinstance(series, bool):
```
aarondr77, line 82: One change we talked about in pair programming was updating these variable names from `series` to something like `series_or_primitive` or `bool_restricted_input_type` or something that makes it way more clear that its not necessarily a series.

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
```diff
@@ -60,10 +59,73 @@ def AND(*argv: BoolInputType) -> BoolFunctionReturnType:
...
+        return series
+
+    return series.fillna(False)
```
aarondr77, line 85: Maybe we should handle NaN values in our cast_to_bool function so that we can do it all in one place instead of trying to make sure we handle NaN values when eating to bools correctly/the same in each function

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
```diff
@@ -60,10 +59,73 @@ def AND(*argv: BoolInputType) -> BoolFunctionReturnType:
...
+
+@cast_values_in_arg_to_type('condition', 'bool')
+@handle_sheet_function_errors
+def IF(condition: pd.Series, true_series: AnyPrimitiveOrSeriesInputType, false_series: AnyPrimitiveOrSeriesInputType) -> pd.Series:
```
aarondr77, line 89: Why does the `condition` have to be a series. If I set the `condition` as `3<4` then it would just be a primitive and cast_values_in_arg_to_type would return a primitive.


Similarly, the return type should be able to be a primitive value instead of a series.

mitosheet/mitosheet/public/v3/sheet_functions/date_functions.py
```diff
@@ -0,0 +1,579 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# Copyright (c) Saga Inc.
+# Distributed under the terms of the GPL License.
+"""
+Contains all functions that can be used in a sheet that operate on strings.
```
aarondr77, line 7: "operate on strings" -> "operate on dates"

mitosheet/mitosheet/public/v3/sheet_functions/number_functions.py
```diff
@@ -55,10 +82,182 @@ def AVG(*argv: NumberInputType) -> NumberFunctionReturnType:
...
+                "element": "series_one",
+                "description": "The number series to convert to calculate the correlation."
+            }, {
+                "element": "series_two",
+                "description": "The number series to convert to calculate the correlation."
+            }
+        ]
+    }
+    """
+    if isinstance(s1, int) or isinstance(s1, float) or isinstance(s2, int) or isinstance(s2, float):
```
aarondr77, line 109: We could use the `is_primitive_value` helper function for this check. Or create a `is_primitive_number` function. 

We could use this in other functions like, `FLOAT`, `LOG`, etc.

mitosheet/mitosheet/public/v3/sheet_functions/number_functions.py
```diff
@@ -55,10 +82,182 @@ def AVG(*argv: NumberInputType) -> NumberFunctionReturnType:
...
+                "element": "base [OPTIONAL]",
+                "description": "The base of the logarithm to use. Defaults to the natural logarithm if no base is passed."
+            }
+        ]
+    }
+    """
+    
+    if base is None:
+        base = math.e
+
```
aarondr77, line 247: Excel uses 10 as the base value

Screenshot 2023-03-28 at 12 24 02 PM

mitosheet/mitosheet/public/v3/sheet_functions/number_functions.py
```diff
@@ -103,9 +302,129 @@ def MULTIPLY(*argv: NumberInputType) -> NumberFunctionReturnType:
...
+    }
+    """
+    if isinstance(series, int) or isinstance(series, float):
+        return 0
+        
+    return series.skew() # type: ignore
+
+@cast_values_in_arg_to_type('series', 'number')
+@handle_sheet_function_errors
+def STDEV(series: NumberRestrictedInputType) -> NumberFunctionReturnType:
```
aarondr77, line 401: According to our spec, `STDEV` should also support dataframes and rolling ranges.

mitosheet/mitosheet/public/v3/sheet_functions/number_functions.py
```diff
@@ -116,12 +435,72 @@ def SUM(*argv: NumberInputType) -> NumberFunctionReturnType:
...
+            }
+        ]
+    }
+    """
+    return series
+
+
+@cast_values_in_arg_to_type('series', 'number')
+@handle_sheet_function_errors
+def VAR(series: NumberRestrictedInputType) -> NumberFunctionReturnType:
```
aarondr77, line 463: This should also support dataframes and rolling ranges.

mitosheet/mitosheet/tests/public/v3/sheet_functions/bool_functions/test_bool.py
```diff
@@ -0,0 +1,37 @@
...
+
+BOOL_VALID_TESTS = [
+    ([1], True),
+    ([0], False),
+    ([pd.Series([1])], pd.Series([True])),
+    ([pd.Series([0])], pd.Series([False])),
+    ([pd.Series([0, 1, 0, 2])], pd.Series([False, True, False, True])),
+    ([pd.Series([0, 1, 0, np.NaN])], pd.Series([False, True, False, False])),
+    ([pd.Series(['Hi', 'Hello', 'Nah', 'Doh'])], pd.Series([False, False, False, False])),
+    ([pd.Series([np.NaN, 'Hi', 'Hello', 'Nah', 'Doh'])], pd.Series([False, False, False, False, False])),
```
aarondr77, line 27: Just calling out that usually, we've been testing the none case with the value `None` instead of np.NaN. It seems like it would not matter which we use for testing though

mitosheet/mitosheet/tests/public/v3/sheet_functions/date_functions/test_day.py
```diff
@@ -0,0 +1,33 @@
...
+DAY_TESTS = [
+    # Just constant tests
+    (['2000-1-2'], 2),
+
+    # Constants and series
+    ([pd.Series(data=['2000-1-2'])], pd.Series(2)),
+    ([pd.Series(data=['2000-1-2'])], pd.Series(2)),
+    ([pd.Series(data=['1/2/2000'])], pd.Series(2)),
+    ([pd.Series(data=['1/2/2000', '1/3/2000', '1/4/2000'])], pd.Series([2,3,4])),
+    ([pd.Series(data=['1/2/2000', None, '1/4/2000'])], pd.Series([2,None,4])),
```
aarondr77, line 24: I should add tests here for when the input it a datetime also for all of these tests.

aarondr77 (commented):
wooo woo. Thanks for guiding this out. To wrap this up, if you want to address the sheet function comments, I can finish up all of the testing comments.

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
naterush: There's only a few of these, so I think it's ok as-is.

mitosheet/mitosheet/public/v3/sheet_functions/number_functions.py
naterush: We actually can't, because then we loose type information. This is a limitation of the MyPy type system I do not normally work around. 

Type guards in TypeScript solve this exact problem.

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
naterush: Static conditions are kinda nonsense for this function (you know the answer), so I think we can leave this as-is. In which case, it also returns a series.

mitosheet/mitosheet/tests/public/v3/sheet_functions/bool_functions/test_bool.py
naterush: It does matter in the case of string columns, namely because we need to make sure that None's actually get converted to NaNs.

mitosheet/mitosheet/tests/public/v3/sheet_functions/number_functions/test_int.py
```diff
@@ -0,0 +1,56 @@
...
+    (['$(123123,12)'], [-123123.12]),
+    ([123], [123]),
+    ([123.123], [123.123]),
+    ([123.123000], [123.123]),
+    (['-$123,123.12 M'], [-123123120000]),
+    (['-$123,123.12 m'], [-123123120000]),
+    (['-$123,123.12 Mil'], [-123123120000]),
+    (['-$123,123.12 mil'], [-123123120000]),
+    (['-$123,123.12 Million'], [-123123120000]),
+    (['-$123,123.12 million'], [-123123120000]),
```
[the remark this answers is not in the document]
naterush: This should error. Int casting requires no NaN values.

mitosheet/mitosheet/tests/public/v3/sheet_functions/number_functions/test_var.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+import pandas as pd
+import numpy as np
+
+from mitosheet.public.v3.sheet_functions.number_functions import VAR
+
+# Raw function tests
+VAR_VALID_TESTS = [
+    ([pd.Series([1,2,3,4])], 1.6666666666666667),
```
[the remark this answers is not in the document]
naterush: It does not. It just errors in this case. It requires multiple inputs.

PR #676, opened 2023-03-28
mitosheet: fix bug with 8A as first value

naterush:
# Description

1. Create a mitosheet where upper left corner value is 8A.
2. Import it using range import. 

In previous versions, this would get cast to 8, which would not be found. This should work currently. 

# Testing

See above.

# Documentation

N/A.


aarondr77:
LGTM

Screenshot 2023-03-28 at 4 46 10 PM

mitosheet/src/components/taskpanes/ExcelRangeImport/ExcelRangeImportTaskpane.tsx
```diff
@@ -512,7 +514,8 @@ const ExcelRangeImportTaskpane = (props: ExcelRangeImportTaskpaneProps): JSX.Ele
 
                                 if (finalRangeImport.type === 'upper left corner value' && finalRangeImport.column_end_condition.type === 'num columns' && typeof finalRangeImport.column_end_condition.value === 'string') {
                                     const parsedValue = parseInt(finalRangeImport.column_end_condition.value);
-                                    if (!isNaN(parsedValue)) {
+                                    const isOnlyNumber = /^[+-]?\d+(\.\d+)?$/.test(finalRangeImport.column_end_condition.value);
```
aarondr77, line 517: Making this a util would be nice

PR #678, opened 2023-03-28
mitosheet: support views in snowflake

aarondr77:
# Description

Supports querying views from the snowflake import taskpane. 

# Testing

See tests + try importing from the new view "SIMPLE_PYTEST_TABLE_VIEW"

# Documentation

Update docs to say we support views.


naterush (changes requested):
An easy win. Since you're changing a parameter name, we need to upgrade the step, which requires:
1. updating the step version number in the step performer
2. adding a step upgrader
3. Adding a test

Since we don't have major usage on this feature... we can skip if you want. But it might break someone's analysis.

naterush:
Otherwise, the code is working well. I'll note that we're def at a point where this feels sluggish as a result of the network calls. At some point (once we have users), it'd be nice to polish this so it feels a bit snappier.

PR #680, opened 2023-03-30
V3 frontend

naterush:
# Description

Adds frontend support for new formulas. Deletes a bunch of old outdated tests as a result.

# Testing

Write some new formulas. Working pretty well!

# Documentation

We'll have to update formulas.


aarondr77:
Its printing `undefined` instead of the correct import statement

Screenshot 2023-03-30 at 6 35 43 PM

aarondr77:
We are messing up the logic when the user sets a range that has from a higher index to lower index. A5:A1. Notice that all of the following cells had the value 0.

https://user-images.githubusercontent.com/18709905/228979432-1a2670b4-72bd-4b26-b7c0-4754050bf32b.mov

It generated this code. 

Screenshot 2023-03-30 at 6 41 57 PM

aarondr77:
It's pretty difficult to select multiple columns. 

In Excel, if I select column header A, hold shift, and select column header B, it will make the range A:B

https://user-images.githubusercontent.com/18709905/228981444-be822176-6022-49b9-8c88-a04bc32673c4.mov

aarondr77:
I had some bug where I wrote the formula `SUM(annual_income:loan_amount)` and it just said processing, never having the loading indicator appear, or finishing execution. But then I could not replicate it.

aarondr77:
There seems to be some issues when replaying formulas that have ranges in them. 

If I just write the formula = SUM(A1), replaying it works fine. 

If I do a formula of the type =SUM(A:B), then I get this error. 


https://user-images.githubusercontent.com/18709905/228983300-d822e1c4-e05e-4a80-bd7b-29c7b18fb8dc.mov

If I do a formula of the type =SUM(A1:A5), then I get this error. 

Screenshot 2023-03-30 at 7 01 04 PM

aarondr77:
This is a pretty good error message! 

Screenshot 2023-03-30 at 7 10 00 PM

aarondr77:
I got a sheet crashing error:

- Open the cell editor
- Use the left arrow key to move the selection box over to the index column
- Get sheet crashing error 


https://user-images.githubusercontent.com/18709905/228985237-90f5383b-69e0-471c-bd9d-1462cb8ea60e.mov


Screenshot 2023-03-30 at 7 12 16 PM

naterush:
Screen Shot 2023-03-31 at 11 35 44 AM

There is a bug where formulas are not applied correctly if the index is out of order! This is pretty large.

aarondr77:
This error message could be dramatically improved if we said which column the values E is in. If E was not the first cell, it would be very difficult to correct this. 

Screenshot 2023-04-02 at 3 25 52 PM

aarondr77:
- Open the cell editor
- Write SUM(A4
- Then type : 
- Then use your mouse or arrow keys to select A1
- We don't fix this range to be A1:A4 in the cell editor or prior to submitting the formula, and therefore this formula does not work correctly. 

Screenshot 2023-04-02 at 3 54 01 PM

PR #684, opened 2023-04-04
Ai completion

naterush:
# Description

Adds Mito AI API.

# Testing

Don't set an Open AI API key. Use the feature.

# Documentation

Need to update instructions to include upgrade to Pro instructions.


mitosheet/mitosheet/api/get_ai_completion.py
```diff
@@ -12,9 +12,69 @@
...
+
+        headers = {
+                'Content-Type': 'application/json',
+        }
+
+        try:
+                res = requests.post(MITO_AI_URL, headers=headers, json=data)
+        except:
+                return json.dumps({
+                        'error': f'There was an error accessing the OpenAI API. This is likely due to internet connectivity problems or a firewall.'
```
naterush, line 66: Update this.

naterush (commented):
Cleanup!

aarondr77:
I got this error after only sending a few message... way less than 20 

Screenshot 2023-04-05 at 2 06 35 PM

aarondr77:
In the error message above, we should make the `upgrade to Mito Pro` text link to Stripe or the plans page

mitosheet/mitosheet/api/get_ai_completion.py
```diff
@@ -12,9 +12,69 @@
...
+def _get_ai_completion_from_mito_server(user_input: str, prompt: str) -> str:
+        user_email = get_user_field(UJ_USER_EMAIL)
+        user_id = get_user_field(UJ_STATIC_USER_ID)
+        num_completions = get_user_field(UJ_AI_MITO_API_NUM_USAGES)
+        if num_completions is None:
+                num_completions = 0
+
+        pro = is_pro()
+
+        if not pro and num_completions >= 3:
```
aarondr77, line 44: Update this to 20

mitosheet/mitosheet/api/get_ai_completion.py
```diff
@@ -12,9 +12,69 @@
...
+                "temperature": .2,
+                'max_tokens': 200,
+                'stop': ['```']
+        }
+
+
+def _get_ai_completion_from_mito_server(user_input: str, prompt: str) -> str:
+        user_email = get_user_field(UJ_USER_EMAIL)
+        user_id = get_user_field(UJ_STATIC_USER_ID)
+        num_completions = get_user_field(UJ_AI_MITO_API_NUM_USAGES)
```
aarondr77, line 38: I think accessing this the user profile is much slower on JupyterHub then we realize since we don't use JupyterHub that much. 

But only on JupyterHub do I ever see the loading indicator say things like `setting user profile`. So what if instead of accessing this each time, we cache reading it. And we could even write it only once every 10 times or something like that.

aarondr77: Or better yet, we could randomly decide if we want to write it or not so it's harder for users to abuse it. 

Write if: `math.random() > .2 or num_completions % 10 == 0`

aarondr77: Lastly, the UJ field is called ...num_usages. This variable is num_completions. Let's pick one

mitosheet/mitosheet/api/get_ai_completion.py
aarondr77: Maybe we can create a variable for the max number of Mito AI usages and then make sure it's in sync between the frontend and backend. Since we use it in a bunch of error messages. I'm sure we'll forget to update them all if we do

aarondr77 (commented):
Small updates then LGTM.

PR #685, opened 2023-04-04
mitosheet: fix bug with column headers in wrong order

aarondr77:
# Description

If we had the dataframe `df = pd.DataFrame({'A': [1,2,3], 'B': [4,5,6]})`, and we wrote the formula `=CONCAT(C:A)`, it would error because the dataframe sent to the rolling range would be empty. This occurred because we selected columns between C and A in the dataframe (which is none)

# Testing

See tests

# Documentation


naterush (changes requested):
Minor rephrase, then LGTM! Nice catch on this bug (and the other related one).

PR #689, opened 2023-04-09
Financial formulas

aarondr77:
# Description

It's a bit tricky to add these formulas with our existing infrastructure. See this explanation, in particular the comments in the NPV function. 

This is working except for range support.

One idea:

Use our existing infrastructure, where the `_previous_result` in the `get_new_result` function is a list of discounted cashflows or a series of lists of discounted cashflows (ie: cash flow / ((1+rate)^period). Then when we get the result back, calculate the total npv by summing them together. 

To figure out the period, we could take the length of the list in the  `_previous_result`

# Testing

# Documentation


naterush:
Closing.

PR #690, opened 2023-04-10
Keyboard Formula Navigation

aarondr77:
# Description

Implements the keyboard formula navigation improvements from https://www.notion.so/trymito/Cell-Editing-UI-1955396ade8244f58cca4d2c55755de7

# Testing

- Start writing a formula and write `:`, then use arrow keys to scroll in the grid instead of in the formula
- Use `=` to enter cell editing mode and notice that it creates a new formula that starts with `=` instead of appending `=` to the end of the cell editor. Just like Excel
- Press `F2` (function + F2) to enter cell editing mode. Notice that it opens the formula to the current formula without any changes. Just like Excel. However, we also support the same interaction with `Enter`

Tested on Mac + windows

# Documentation


naterush (approved):
Looks good. Merging.

PR #692, opened 2023-04-10
mitosheet: improve getDocumentationFunction to take selectionStart

naterush:
# Description

As title says.

# Testing

Use it. It's def better!

# Documentation

No.


aarondr77:
The documentation updates if I use my mouse to select a new place in the formula, but not if I use my arrow keys to scroll to a new place. 

https://user-images.githubusercontent.com/18709905/231227465-927689ec-ae8f-46b4-be05-cc52baa2dc92.mov

aarondr77:
After fixing that LGTM. This is sweet!

PR #693, opened 2023-04-10
Resize cell editor

naterush:
# Description

When we have more than 30 characters, get bigger. This is a very simple heuristic. Feel free to make it slightly smarter if you want, but honestly I think it's probably pretty good!

# Testing

Write some long formulas. The only limitation is that you can't use string constants with new lines in them (which is fine, you couldn't do this before and it's never come up!).

# Documentation

No.


naterush:
Also what should tab do? Anything weird? Did I miss anything else?

aarondr77:
The only issue with this is that it covers so much space, which is especially important now that you need that space to click on / navigate to other cells to reference. 

Screenshot 2023-04-11 at 12 00 40 PM

mitosheet/src/components/endo/celleditor/CellEditor.tsx
```diff
@@ -493,7 +498,7 @@ const CellEditor = (props: {
                 In the dropdown box, we either show an error, a loading message, suggestions
                 or the documentation for the last function, depending on the cases below
             */}
-            <div className='cell-editor-dropdown-box' style={{width: props.editorState.editorLocation === 'cell' ? `${CELL_EDITOR_WIDTH}px` : '300px'}}>
+            <div className='cell-editor-dropdown-box' style={{width: `${cellEditorWidth}px`}}>
```
aarondr77, line 501: I don't think we want this in the case of the formula bar because it covers up so much of the data and column headers. Can we just leave this as the regular width in the formula bar?


Screenshot 2023-04-11 at 12 18 39 PM

aarondr77 (commented):
LGTM after fixing these two points. The cell editor growing is super nice. Great call on this one

PR #698, opened 2023-04-11
mitosheet: log pro and enterprise feature upgrade actions

aarondr77:
# Description

Logs when the user is prompted to upgrade to pro or enterprise. 
Logs when the user clicks on the upgrade button to explore the plans. 

# Testing

- Use Mito Open Source and click on the snowflake import button 

# Documentation

No.


mitosheet/src/pro/taskpanes/ConditionalFormatting/ConditionalFormattingTaskpane.tsx
```diff
@@ -130,6 +130,8 @@ const ConditionalFormattingTaskpane = (props: ConditionalFormattingTaskpaneProps
                 userProfile={props.userProfile} 
                 requiresPro
                 requiresProMessage="Conditional formatting is a Mito Pro feature. Please upgrade to use this feature."
+                mitoAPI={props.mitoAPI}
+                featureName="conditional_formatting"
```
naterush, line 134: Can we collapse these into one type. So the props can be:

```
requiresPro?: {
    message: string,
    featureName: string,
    mitoAPI: MitoAPI
}
```

And same for enterprise. That way you don't have to remember to provide all of these optional arguments together.

mitosheet/src/pro/taskpanes/SetDataframeFormat/SetDataframeFormatTaskpane.tsx
```diff
@@ -93,6 +93,8 @@ const SetDataframeFormatTaskpane = (props: SetDataframeFormatTaskpaneProps): JSX
                 userProfile={props.userProfile}
                 requiresPro
                 requiresProMessage="Setting the dataframe format is a Mito Pro feature. Please upgrade to use this feature."
+                mitoAPI={props.mitoAPI}
+                featureName="color_dataframe"
```
naterush, line 97: Let's call this Set Dataframe Format or whatever the taskpane is called, so it's easier to find.

naterush (changes requested):
Some type-level cleanup, then LGTM.

PR #714, opened 2023-04-20
Excel range import sheet index

naterush:
# Description

Adds excel range import by sheet. This results in a cleaner UI, and is entirely backwards compatible. The polish I added is also totally cool - so I'm proud of that!

# Testing

Use it.

# Documentation

No.


mitosheet/mitosheet/code_chunks/step_performers/import_steps/excel_range_import_code_chunk.py
```diff
@@ -102,13 +120,15 @@ def get_display_name(self) -> str:
     
     def get_description_comment(self) -> str:
         
-        return f"Imported {len(self.range_imports)} dataframes from {self.sheet_name} in {self.file_path}"
+        return f"Imported {len(self.range_imports)} dataframes from {self.sheet['value']} in {self.file_path}"
```
naterush, line 123: Clean this up

naterush (commented):
Cleanup

aarondr77:
I was a little bit confused be

Screenshot 2023-04-21 at 11 04 59 AM

aarondr77:
This is sweet! I like these messages

Screenshot 2023-04-21 at 1 18 22 PM

aarondr77:
In the input, I can't type `-1` because it's a number. So I need to use the toggles to get to a negative index. 

If we fix this, we should make it a component so we can use it elsewhere.

aarondr77:
Let's at least make these the same width so they look nicer. 

Screenshot 2023-04-21 at 1 40 48 PM

But also, it would be nice to group them together visually so its more obvious that they effect each other. Making them in one row could be reasonable. Or using some other visual indicator like:

<h3> Select Sheet
<bold> Select sheet by -> Input
<bold> Sheet Index -> Input

mitosheet/src/components/taskpanes/ExcelRangeImport/ExcelRangeSheetSelection.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const params = props.params;
+    const setParams = props.setParams;
+    const sheet_name = props.sheet_name;
+
+    return (
+        <>
+            <Row justify="space-between" align="center">
+                <Col>
+                    <p className="text-header-3">
+                            Find Sheet By
```
aarondr77, line 27: Maybe instead of "Find" we use "Select". IMO symmetry here is helpful in grouping these together.

mitosheet/src/components/taskpanes/ExcelRangeImport/ExcelRangeSheetSelection.tsx
```diff
@@ -0,0 +1,119 @@
...
+                    >
+                        <DropdownItem
+                            title='Sheet Name'
+                            id="sheet name"
+                            subtext="Select a sheet by name to import multiple ranges from."
+                        />
+                        <DropdownItem
+                            title='Sheet Index'
+                            id="sheet index"
+                            subtext="Select a sheet by index to import multiple ranges from. Negative sheet indexes are allowed, and are counted from the end of the sheet list."
```
aarondr77, line 53: "Select a sheet by index to import multiple ranges from. Negative sheet indexes are allowed, and are counted from the end of the sheet list." -> "Select a sheet by index in the Excel file. Use a negative index to count backwards from the end of the sheet list."

mitosheet/mitosheet/public/v2/excel_utils.py
```diff
@@ -132,7 +146,7 @@ def get_read_excel_params_from_range(range: str) -> Tuple[int, int, str]:
     return start_row_index, nrows, usecols
 
 
-def convert_csv_file_to_xlsx_file(csv_path: str, sheet_name: str) -> str:
+def convert_csv_file_to_xlsx_file(csv_path: str, sheet_name: Union[str, int]) -> str:
```
aarondr77, line 149: We should add a comment here too that these parameters can not change order for backwards compatibility. It would be nice to test that in fact.

aarondr77 (commented):
LGTM after small fixes.

mitosheet/mitosheet/public/v2/excel_utils.py
naterush: Added test.

PR #716, opened 2023-04-23
mitosheet: stop making infinite divs for dropdowns

naterush:
# Description

We had a bug where we were previously creating a _new_ div on every rerender
of every dropdown. This bug has been in Mito for a very long time. It's not
clear that the observered bugs we found were a function of this issue - but 
it is def not good news! This PR fixes it up.

# Testing

First, try the dev branch of Mito. Import some data. Open the console
and run the code `document.body.querySelectorAll('div').length`. Then scroll
around in Mito, and then run `document.body.querySelectorAll('div').length` 
again. You will see the number of divs has increased!

Now, try the same steps on this branch. You will see that here, we do not
increase the number of divs over time.

# Documentation

### DIFF
--- a/mitosheet/mitosheet/mito_frontend.js
+++ b/mitosheet/mitosheet/mito_frontend.js
@@ -21158,7 +21158,9 @@ For more info, visit https://reactjs.org/link/mock-scheduler`);
       left: void 0
     });
     const dropdownAnchor = (0, import_react4.useRef)(null);
-    const dropdownContainerElement = (0, import_react4.useRef)(document.createElement("div"));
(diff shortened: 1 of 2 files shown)


aarondr77:
These are successive `document.body.querySelectorAll('div').length` -> `scroll` -> `document.body.querySelectorAll('div').length` on Safari on dev

Screenshot 2023-04-23 at 9 23 08 PM

These are successive `document.body.querySelectorAll('div').length` -> `scroll` -> `document.body.querySelectorAll('div').length` on Safari on this branch 

Screenshot 2023-04-23 at 9 26 56 PM

PR #717, opened 2023-04-23
Fix insane css

naterush:
# Description

This PR does a few nice things with our CSS:
1. Standardizes a lot of CSS names, to avoid overlap, and to make it easier
	to figure out what is what. Nice!
2. Gets rid of a _bunch_ of height hacks that we previously used all across 
	the top divs. There was so much cruft here I cut out.

Both of these will go a long way in making this all easier to maintain (and
in the case of the insane bug we found last week, easier to debug).
# Testing

Make sure Mito works the same on all browsers.

# Documentation

No.


mitosheet/css/mito.css
```diff
@@ -7,13 +7,13 @@
     The pure css selectors are used for the JLab button, and the id selection is used for the notebook 
     button, since we're not able to set the class name of our button in notebooks. 
 */
-:root .toolbar-mito-button-class, #mito-toolbar-button-id {
+:root .toolbar-mito-button-class, #mito-mito-toolbar-button-id {
```
aarondr77, line 10: We want mito-mito-....? Not just mito-...?

mitosheet/src/components/Mito.tsx
```diff
@@ -860,23 +860,23 @@ export const Mito = (props: MitoProps): JSX.Element => {
 
     // Check which taskpanes are open
     const taskpaneOpen = uiState.currOpenTaskpane.type !== TaskpaneType.NONE;
-    const graphTaskpaneOpen = uiState.currOpenTaskpane.type === TaskpaneType.GRAPH && uiState.selectedTabType === 'graph';
-    const narrowTaskpaneOpen = taskpaneOpen && !graphTaskpaneOpen;
+    const wideTaskpaneOpen = uiState.currOpenTaskpane.type === TaskpaneType.GRAPH && uiState.selectedTabType === 'graph';
```
aarondr77, line 863: I think this name might be worse, but wideTaskpane just makes it sound like it's a taskpane but a bit wider, where in actuality, the graphTaskpane / wideTaskpane is actually a full sheet! Which is horrible, but it is.

aarondr77 (commented):
LGTM after minor updates

PR #718, opened 2023-04-23
Frontend profiling

naterush:
# Description

This PR used the chrome dev-tools to find some easy wins on frontend 
performance (not that Mito is slow... but just for fun). 

Namely, it reduces the amount of work dropdowns do if they aren't displayed.
It also removes a bunch of spreadsheet functions from the help dropdown,
as they were almost never used and generally just extra work for no benefit.

# Testing

# Documentation


aarondr77:
The formula bar is MIA

Screenshot 2023-04-23 at 9 02 22 PM

aarondr77 (commented):
Nice cleanup! Since that formula bar issue exists, I'll do another round of testing before merge. We should make sure to test these heavily on windows too!

PR #719, opened 2023-04-23
mitosheet: make code minify

naterush:
# Description

This PR fixes up our esbuild (and hacks to get data into js_code variable) 
so that we now work with minified code. 

In practice, this means that an empty Mito sheet is about 50% of the size on
disk when saved in a notebook. This should reduce issues where users cannot
save their notebooks because they are too large!

(BTW, in the future, clear outputs will resolve this problem!)

# Testing

See before and after screenshot of size with 10 mitosheets rendered and
saved within a single notebook.

# Documentation

 No.


aarondr77:
LGTM

PR #730, opened 2023-04-27
Dog food 4 27 23

aarondr77:
# Description

- update column header transform taskpane language from `old/new` to `find/replace`
- improve download location language again
- code optimization for column headers 
- prevent duplicated column headers from bricking the sheet as a result of bulk column header renaming 

# Testing

# Documentation


mitosheet/mitosheet/step_performers/column_headers_transform.py
```diff
@@ -63,7 +64,16 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
                     new_columns.append(col)
         else:
             raise ValueError(f'Unknown transformation type: {transformation["type"]}')
-        
+
+        if len(set(new_columns)) < len(new_columns):
```
naterush, line 68: @aarondr77 why don't you just use the new deduplicate functionality here? We could just have it option in this step, since it's likely so rare.

mitosheet/src/components/toolbar/Toolbar.tsx
```diff
@@ -236,18 +236,26 @@ const Toolbar = (
...
+                                onClick={() => {props.setUIState(prevUIState => {
+                                    return {
+                                        ...prevUIState,
+                                        currOpenTaskpane: {type: TaskpaneType.DOWNLOAD}
+                                    }
+                                })
+                            }}/>
+                            <DropdownItem 
+                                title='Download File when Executing Code' 
+                                subtext='Download the file to Jupyter each time you run the generated code'
```
naterush: This could be more clear.

naterush (approved):
Minor request, feel free to ignore.

PR #731, opened 2023-04-27
Improve mitosheet call robustness

aarondr77:
# Description

Support all of the following mitosheet.sheet calls

```
mitosheet.sheet    (   )    
```
```
mitosheet.sheet    (
    analysis_to_replay    =    "id-uzqdyuepse"
)  
```
```
mitosheet.sheet    (  [tab]  )    
```
```
mitosheet.sheet(   "FILE NAME.csv"       ) 
```
```
mitosheet.sheet("FILE 'NAME.csv")
```
# Testing

# Documentation


naterush:
This is broken on file names that have spaces in them. Can we do the following:
1. Rename removeWhiteSpace to removeWhitespaceInPythonCode 
2. Never pass cleaned code to isMitosheetCallCode (like we do in tryWriteAnalysisToReplayParameter) - so it's clear that the caller need not be responsible to clean before they do so. E.g. just move the code cleaning down a line into the if statement there 
3. Clean up the removeWhitespaceInPythonCode to not remove whitespace in file names

Thanks!

mitosheet/src/utils/code.tsx
```diff
@@ -136,4 +136,28 @@ export function containsMitosheetCallWithAnyAnalysisToReplay(codeText: string):
 */
 export function containsGeneratedCodeOfAnalysis(codeText: string, analysisName: string): boolean {
     return isMitoAnalysisCode(codeText) && codeText.includes(analysisName);
-}
\ No newline at end of file
+}
+
+export function removeWhitespaceInPythonCode(codeText: string): string {
```
naterush, line 141: Just add a little doc string

naterush (commented):
Minor cleanup

PR #741, opened 2023-05-04
Excel range import improvements

naterush:
# Description

1. Allow you to specify multiple start or end conditions without erroring. takes whatever it hits first.
2. Add # of consecutive empty cells in the first column
3. Fixed bugs where if you had an end condition that was contained in the first cell, it would just read in one cell
4. Fixed bug where if you had an empty cell in the middle of your range, it would stop if you had a contains

# Testing

See new tests. Add some more if you can think of good ones. This boy is complex :) 

# Documentation

No.


aarondr77:
> Allow you to specify multiple start or end conditions without erroring. takes whatever it hits first.

I can't figure out how to use this functionality. Is it pushed?

https://user-images.githubusercontent.com/18709905/236319986-3ef6c554-e7fe-4019-8ecd-1623d216306e.mov

PR #748, opened 2023-05-09
Turn on code config for mitosheet private

aarondr77:
# Description

Turn on code config for mitosheet private, without using the mito_config because this isn't a publicly configurable thing. We'd rather handle this ourselves. 

# Testing

- Make sure the code config is not turned on when on pro
- Make sure the code config is turned on when using enterprise
- Make sure the code config is turned on when you set the package_name to 'mitosheet-private'

# Documentation


naterush:
Don't we have mitosheet-private handle config themselves? I am concerned about increasing complexity on our end dealing with different packages, tbh. There's so many options...

PR #749, opened 2023-05-13
Add user defined functions

naterush:
# Description

Pass in user defined functions with `sheet_functions=[LIST_OF_NAMES]` to the mitosheet.sheet call. Nice.

# Testing

We should add some tests. Use it and let me know what you think though. Also, we should make this Pro.

# Documentation

Yes.


aarondr77:
This is confusing.. why do the functions need to be all uppercase?

Screenshot 2023-05-16 at 3 52 25 PM

PR #751, opened 2023-05-16
Ai fixes

naterush:
# Description

Fixes the results of our usability tests:
- Store chat history between opens of taskpane (and runs!). This is... so much better. 
- Fix bug where wrong result would show up
- Makes chat height work much better if there are multiple lines but not enter spots
- Make it clear you can enter your own examples
 
# Testing

Use the new stuff.

# Documentation

N/A.


aarondr77:
The spacing on this is a bit funky looking. Ideas:
- Make the list of modified element a bulleted list
- Add more top-margin to the `How did Mito AI Assistant do?`
- Make the emojis a bit smaller to reflect the smaller text size. 

Screenshot 2023-05-17 at 5 31 43 PM

aarondr77:
This is working very nicely!

PR #755, opened 2023-05-19
Custom imports

naterush:
# Description

This should be reviewed after custom functions. It adds custom importers, and a way to a access them.

# Testing

Use the feature!

# Documentation

Yes enterprise docs.


mitosheet/mitosheet/step_performers/user_defined_import.py
```diff
@@ -0,0 +1,91 @@
...
+    def step_version(cls) -> int:
+        return 1
+
+    @classmethod
+    def step_type(cls) -> str:
+        return 'user_defined_import'
+
+    @classmethod
+    def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Optional[Dict[str, Any]]]:
+        print("PARAMS", params)
```
aarondr77: remove

mitosheet/mitosheet/step_performers/user_defined_import.py
```diff
@@ -0,0 +1,91 @@
...
+        importer: str = get_param(params, 'importer')
+
+        # We make a new state to modify it
+        post_state = prev_state.copy()
+
+        pandas_start_time = perf_counter()
+        
+        print("NAMES", importer, post_state.user_defined_importers, [f.__name__ for f in post_state.user_defined_importers])
```
aarondr77: remobve

mitosheet/mitosheet/step_performers/user_defined_import.py
```diff
@@ -0,0 +1,91 @@
...
+        return 1
+
+    @classmethod
+    def step_type(cls) -> str:
+        return 'user_defined_import'
+
+    @classmethod
+    def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Optional[Dict[str, Any]]]:
+        print("PARAMS", params)
+        importer: str = get_param(params, 'importer')
```
aarondr77, line 51: its unclear from this variable name what `importer` is. Is it the name of the function? Is it the actual import code?

mitosheet/mitosheet/tests/step_performers/test_userdefinedimport.py
```diff
@@ -0,0 +1,35 @@
...
+
+USERDEFINEDIMPORT_TESTS = [
+    (
+        [
+            pd.DataFrame({'A': [1, 2, 3], 'B': [1.0, 2.0, 3.0], 'C': [True, False, True], 'D': ["string", "with spaces", "and/!other@characters"], 'E': pd.to_datetime(['12-22-1997', '12-23-1997', '12-24-1997']), 'F': pd.to_timedelta(['1 days', '2 days', '3 days'])})
+        ],
+        "put", 
+        "params", 
+        "here",
+        [
```
aarondr77: These tests aren't actually testing the importer, right?

mitosheet/src/components/taskpanes/UserDefinedImport/UserDefinedImportTaskpane.tsx
```diff
@@ -0,0 +1,127 @@
...
+                    width='block'
+                    onClick={() => {
+                        if (params !== undefined) {
+                            edit(params);
+                        }
+                    }}
+                    disabled={params === undefined}
+                    disabledTooltip={NO_IMPORTERS_MESSAGE}
+                >
+                    Import Ranges
```
aarondr77, line 123: This should probably say something like "Import Data"

aarondr77:
We should add it here too

Screenshot 2023-05-19 at 3 33 57 PM

aarondr77:
If users create a mitosheet like this with multiple lines for the import, then it doesn't create an analysis id. This might be fixed in the other PR, I'm not sure. 

Screenshot 2023-05-19 at 3 46 48 PM

aarondr77:
I think a very common error users are going to get is not having a package installed. ie: requests. It would be helpful to case on this error and display it within the taskpane.

mitosheet/mitosheet/tests/step_performers/test_userdefinedimport.py
aarondr77: We should add tests for 1) A custom import working 2) A custom import adding multiple files 3) A custom import that has a timeout in it and the sheet still working??

aarondr77:
LGTM after fixed up

aarondr77 (commented):
LGTM after cleaned up and more tests

aarondr77 (commented):
LGTM after cleaned up and more tests

mitosheet/mitosheet/tests/step_performers/test_userdefinedimport.py
naterush: Custom imports currently can only add one - otherwise, I added these tests. No timeout for now (slows down tests).

naterush:
LGTM when tests pass.

PR #760, opened 2023-05-22
Parameter variables

aarondr77:
# Description

# Testing

# Documentation


mitosheet/mitosheet/tests/test_utils.py
```diff
@@ -1586,6 +1586,10 @@ def create_mito_wrapper(*args: Union[pd.DataFrame, str], arg_names: Optional[Lis
     mito_backend = get_mito_backend(*args)
     test_wrapper =  MitoWidgetTestWrapper(mito_backend)
 
+    print("ARGS")
+    print(*args)
+    print(arg_names)
```
naterush, line 1591: Remove

mitosheet/mitosheet/transpiler/transpile_utils.py
```diff
@@ -246,11 +246,35 @@ def convert_script_to_function(steps_manager: StepsManagerType, imports: List[st
...
-    # Then, add the function call
+    # Build the params and variables taking special care to ensure that dataframes and file paths 
+    # that are passed as parameters to the function. 
+    params = []
+    num_variables_created = 0
+    saturdated_params_dict = _get_params_dict_for_function_call(steps_manager, function_params)
+    saturated_params_names_list = list(saturdated_params_dict.keys())
+    saturdated_params_values_list = list(saturdated_params_dict.values())
+
+    for idx, param_name in enumerate(saturated_params_names_list):
```
naterush, line 257: I think you can change this to the following:
```
for param_name, param_value in _get_params_dict_for_function_call(steps_manager, function_params).items():
     if param_name in function_params:
          params.append(param_name)
          final_code.append(f"{param_name} = {param_value}")
     else:
          params.append(param_value)
```

This allows us to get rid of 3 variables that store intermediate state, and cut down the number of lines by about 2x. 

Also, we can remove `num_variables_created` and just replace that with the length of function_params, right?

Finally, let's rename `params` to something more descriptive. I think it should be called `final_params_to_call_function_with`?

naterush (changes requested):
Easy refactors that will cut the complexity of this code in half, I think?

PR #765, opened 2023-05-23
Ai recon 1

aarondr77:
# Description

Adds recon as per: https://www.notion.so/trymito/AI-Recon-cead1b6d5c9f45dbae76273b4404f926

# Testing

Use it!

# Documentation

Yeah, when we do the docs for AI!


mitosheet/src/components/taskpanes/Merge/MergeInvalidSelectedColumnsError.tsx
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) Mito
```
aarondr77, line 1: Delete this file

aarondr77 (commented):
cleanup

mitosheet/src/components/taskpanes/AITransformation/AITransformationResultSection.tsx
```diff
@@ -64,7 +67,7 @@ const AITransformationResultSection = (props: AITransformationResultSectionProps
                                 }
                             })
                         }}>
-                        <span>Created:</span> <span className="text-underline">{dfName}</span>  ({numRows} rows, {numColumns} columns)
+                        <span className={classNames({'text-color-recon-created': props.isMostRecentResult})}>Created:</span> <span className="text-underline">{dfName}</span>  ({numRows} rows, {numColumns} columns)
```
naterush, line 70: I think we can always highlight this text, no?

mitosheet/src/components/taskpanes/AITransformation/AITransformationResultSection.tsx
```diff
@@ -117,6 +121,43 @@ const AITransformationResultSection = (props: AITransformationResultSectionProps
...
+                    justify="space-between" 
+                    onClick={() => {
+                        props.setUIState(prevUIState => {
+
+                            if (prevUIState.dataRecon !== undefined) {
+                                return {
+                                    ...prevUIState, 
+                                    dataRecon: undefined
+                                }
+                            }
```
naterush, line 136: Why does this make sense? Maybe a comment here to clarify

mitosheet/src/components/taskpanes/AITransformation/AITransformationTaskpane.tsx
```diff
@@ -140,8 +140,41 @@ const AITransformationTaskpane = (props: AITransformationTaskpaneProps): JSX.Ele
             taskpaneBodyRef.current.scrollTop = taskpaneBodyRef.current.scrollHeight;
         }
 
+        // Update the UIState for Recon
+        props.setUIState(prevUIState => {
+
+            if (previousParamsAndResults.length === 0) {
+                return{
```
naterush, line 148: Space

mitosheet/src/components/taskpanes/AITransformation/aiUtils.tsx
```diff
@@ -39,4 +40,25 @@ export const getChatHeight = (userInput: string, chatInputRef: React.MutableRefO
     } else {
         return chatOrScrollMax;
     }
-}
\ No newline at end of file
+}
+
+
+export const reconIsColumnCreated = (columnHeader: ColumnHeader, uiState: UIState, sheetData: SheetData): boolean => {
```
naterush, line 46: Rather than these functions taking UIState, can they just take an optional recon - I think this makes more sense.

mitosheet/src/types.tsx
```diff
@@ -850,6 +851,14 @@ export interface UIState {
         // TODO: Move the other popups (loading, tour, fast forward) to use this infrastructure
         [PopupLocation.TopRight]: PopupInfo 
     }
+    dataRecon: { 
+        created_dataframe_names: string[],
```
naterush, line 855: Intesting that we do all of these as strings. Is this a function what we return from the backend?

mitosheet/css/sitewide/colors.css
```diff
@@ -30,4 +30,10 @@
 
     /* Color for success message */
     --mito-green: #10710D;
+
+    /* Colors used in recon */
+    --mito-recon-created-background-color: #96D394;
+    --mito-recon-created-text-color: #10710D;
+    --mito-recon-modified-background-color: #EDDAA2;
+    --mito-recon-modified-text-color: #ECBB4A;
```
naterush, line 38: One thing on colors: I think columns should have colors alternating per row, even if they are highlighted green. 

E.g. we need to add `--mito-recon-created-text-color-light` or something, and odd rows become this color, while odd rows are `--mito-recon-created-text-color`. Otherwise, it's much harder to parse the data.

mitosheet/src/components/endo/ColumnHeader.tsx
```diff
@@ -150,14 +154,21 @@ const ColumnHeader = (props: {
         />
     )
 
+    //If there is a dataRecon set, highlight the column headers that have been created or renamed
+    const isColumnCreated = reconIsColumnCreated(columnHeader, props.uiState, props.sheetData)
```
naterush, line 158: This is an O(num_headers * num_created) operation (every header is checking against every created header). 

Note sure this a problem, but can you check on a huge dataset and see the performance impact?

mitosheet/src/components/endo/GridData.tsx
```diff
@@ -63,15 +65,20 @@ const GridData = (props: {
...
+                            const isColumnCreated = reconIsColumnCreated(columnHeader, props.uiState, sheetData)
+                            const isColumnModified = reconIsColumnModified(columnHeader, props.uiState, sheetData)
+
                             const className = classNames('mito-grid-cell', 'text-unselectable', {
                                 'mito-grid-cell-selected': cellIsSelected,
                                 'mito-grid-cell-conditional-format-background-color': conditionalFormat?.backgroundColor !== undefined,
                                 'mito-grid-cell-hidden': props.editorState !== undefined && props.editorState.rowIndex === rowIndex && props.editorState.columnIndex === columnIndex,
-                                'right-align-number-series': isNumberDtype(columnDtype)
+                                'right-align-number-series': isNumberDtype(columnDtype),
+                                'created-recon-background-color': isColumnCreated,
```
naterush, line 80: So for odd and event, you just need to add two more cases, and then check if it's an odd or even row index. Pretty simple I think.

mitosheet/src/components/taskpanes/AITransformation/AITransformationTaskpane.tsx
```diff
@@ -140,8 +140,41 @@ const AITransformationTaskpane = (props: AITransformationTaskpaneProps): JSX.Ele
...
+
+            return {
+                ...prevUIState,
+                dataRecon: newDataRecon
+            }
+        })
+
     }, [previousParamsAndResults.length, taskpaneState.type])
 
+    useEffect( () => () => {
```
naterush, line 170: Can you make the function body and return statement explicit? The double arrow is kool, but took me a long while to parse (I wrote a long comment being confused about it).

mitosheet/src/types.tsx
```diff
@@ -850,6 +851,14 @@ export interface UIState {
         // TODO: Move the other popups (loading, tour, fast forward) to use this infrastructure
         [PopupLocation.TopRight]: PopupInfo 
     }
+    dataRecon: {
```
naterush, line 854: Let's make a type for the data recon, and share the type with AITransformationResult. Peep the following fancy thing!

Screen Shot 2023-05-23 at 1 18 26 PM

naterush (changes requested):
A few minor refactors, and then a few more thoughts:

1. I think the recon color should take priority over the color in the section. It looks pretty strange otherwise. It looks good if we change the color (see test here):

Screen Shot 2023-05-23 at 1 24 20 PM

I think we can leave the sheet tabs as they are!

2. Some changes to the eyeball that I think make sense:
- Moving it to the upper corner, and moving the text to a title (see screenshots below - removes a lot of text)
- Always keeping it visible, just disabling it (and changing the tooltip) if it's not the first recon.
Screen Shot 2023-05-23 at 1 25 18 PM
Screen Shot 2023-05-23 at 1 25 26 PM

With this little bit of polish, I think it'll make this feature feel a lot better - but it's an awesome thing so far!

mitosheet/src/types.tsx
aarondr77: Hmm tbh I'm not sure. I just reused the types that we already had.

mitosheet/src/components/endo/ColumnHeader.tsx
aarondr77: Its a bit hard to test on huge dataframes because of https://github.com/mito-ds/monorepo/issues/768, but its not an issue when we have 1000 rows and 40 columns.

mitosheet/src/components/taskpanes/AITransformation/AITransformationResultSection.tsx
aarondr77: I wanted to only highlight in the last one so it was clear what is highlighted in the sheet.

mitosheet/src/components/taskpanes/AITransformation/AITransformationResultSection.tsx
aarondr77: Comment added -- if the user clicks the eye when the recon is already displayed, then we want to turn off the recon.

naterush:
LGTM once tests pass.

PR #778, opened 2023-05-30
mitosheet: move eyeball to the right

naterush:
# Description

Our taskpane is getting busy. So this moves the eyeball up and to the right. 

# Testing

# Documentation


aarondr77:
See it flashing when the state changes 
https://github.com/mito-ds/monorepo/assets/18709905/2fc17adc-9d3f-4018-af42-23865a7e18d4

aarondr77:
Can we improve the spacing a bit?

Screenshot 2023-05-30 at 12 48 00 PM

PR #779, opened 2023-05-30
Fix bug to rename column headers to duplicated

naterush:
# Description

You could get the mitosheet in a broken state if you wrote "rename all column headers to A".

# Testing

Check this works now.

# Documentation

No.


mitosheet/mitosheet/ai/recon.py
```diff
@@ -135,6 +135,16 @@ def get_modified_dataframe_recon_data(old_df: pd.DataFrame, new_df: pd.DataFrame
     old_columns = old_df.columns.to_list()
     new_columns = new_df.columns.to_list()
 
+    # If there are duplicated columns in the new dataframe, we throw an error
+    # as we don't support this elsewhere in the codebase
+    if len(new_columns) != len(set(new_columns)):
```
aarondr77, line 140: I got this error, which is not the error generated here, but also not an error I've gotten before when using Mito AI. I've never seen it generate a modal before. Do you think that is caused by this change?

Some things to fix:
1. Instead of the modal, we should handle the error in place like we usually do with Mito AI
2. ^ should also make it so that the prompt and result are displayed. See in this example the prompt disappeared and we are back to the example. 

Screenshot 2023-05-30 at 1 06 51 PM

aarondr77 (commented):
I'm not sure that this is working as I expected. See below. Then LGTM.

PR #780, opened 2023-05-30
mitosheet: fix up wrong highlight on error

naterush:
# Description

If you had some AI transformations, closed the taskpane, and then opened it, and then entered a prompt that caused an error when executing, the completion would be marked as successful - which would cause the previous prompt recon to be highlighted before the new prompt was finished (because it errored). This confused me big-time, as it showed an old-recon when I thought the current one had errored! (I guess that's good in showing I use the recon!).

Fixed this just by tracking successful execution, rather than successful completion.

# Testing

Test the above scenario.

# Documentation

No.


mitosheet/src/components/taskpanes/AITransformation/AITransformationTaskpane.tsx
```diff
@@ -241,6 +235,7 @@ const AITransformationTaskpane = (props: AITransformationTaskpaneProps): JSX.Ele
                     previousFailedCompletions.push([completionOrError.completion, possibleError])
                 } else {
                     setTaskpaneState({type: 'default'});
+                    setSuccessfulExecutionSinceOpen(true); // Mark a successful execution
```
aarondr77, line 238: Since this gets set after the `previousParamsAndResults` is set and in a race with `taskpaneState`, the first render does not have the recon turned on. 

I fixed this by adding `successfulExecutionSinceOpen` to the useEffect triggers.

PR #793, opened 2023-06-02
Move `mitosheet` to just `mitosheet`

naterush:
# Description

The mitosheet package now no longer has a switch.py. This breaks out `mitosheet-private` deploy for now, but we'll fix that at another time!

# Testing

All tests should pass!

# Documentation

N/A


naterush:
I think tests are just OpenAI flakiness!

mitosheet/mitosheet/user/utils.py
```diff
@@ -110,24 +112,11 @@ def should_upgrade_mitosheet() -> bool:
     NOTE: if the user clicks the upgrade button in the app, then we change the upgraded 
     date to this date, so that the user doesn't get a bunch of annoying popups. This just
     pushes back when they are annoyed to upgrade!
+
+    NOTE: THIS FUNCTION IS NOT IN USE ANY MORE. WE NO LONGER TELL USERS TO UPGRADE
+    AS THERE IS NO REAL REASON FOR THEM TO DO SO, AND IT IS ANNOYING.
```
aarondr77, line 117: If we're getting rid of this, then let's delete the whole function so its not hanging around our codebase. Want to do this in a separate PR?

mitosheet/setup.py
```diff
@@ -39,237 +39,123 @@
...
+    url                     = package_json["homepage"],
+    author                  = package_json["author"]["name"],
+    author_email            = package_json["author"]["email"],
+    description             = package_json["description"],
+    license                 = "GNU Affero General Public License v3",
+    long_description="""
+    To learn more about Mito, checkout out our documentation: https://docs.trymito.io/getting-started/installing-mito\n\n
+    Before installing Mito \n\n
+    1. Check that you have Python 3.6 or above. To check your version of Python, open a new terminal, and type python3 --version. If you need to install or update Python, restart your terminal after doing so.\n\n
+    2. Checkout our terms of service and privacy policy. By installing Mito, you're agreeing to both of them. Please contact us at aarondr77 (@) gmail.com with any questions.\n\n
```
aarondr77: aarondr77 (@) gmail.com -> aaron@sagacollab.com

mitosheet/setup.py
```diff
@@ -39,237 +39,123 @@
...
+    Installation Instructions \n\n
+    For more detailed installation instructions, see our documentation: https://docs.trymito.io/getting-started/installing-mito\n\n
+    1. pip install mitosheet\n\n
+    2. Launch JupyterLab 3.0 and open a new notebook\n\n
+    3. In the notebook, run the following code:\n\n
+    from mitosheet import sheet\n\n
```
aarondr77: from mitosheet import sheet -> import mitosheet

aarondr77 (commented):
Nice cleanup. The structure we're moving to seems so much better. Thanks for taking the time to explain it to me, also! 

One thing is just I don't think we want to go along time with the mitosheet-private deployment process broken. A few ideas:
1. Let's prioritize figuring out how to set the mitosheet-private environment variables and test it
2. Let's start the conversation with them about what the next upgrade is going to entail so they have time to prepare for that before we want to deliver them new features or fix bugs. Can you start this communication?

aarondr77:
I removed `python switch.py mitosheet` from our dev setup scripts, btw

PR #794, opened 2023-06-02
mitosheet: remove random code options from taskpane

naterush:
# Description

I don't know how I merged these. Woo.

# Testing

See they are gone.

# Documentation

N/A.


mitosheet/src/components/taskpanes/CodeOptions/CodeOptionsTaskpane.tsx
```diff
@@ -92,39 +91,6 @@ const CodeOptionsTaskpane = (props: CodeOptionsTaskpaneProps): JSX.Element => {
                         setCodeOptions={setCodeOptions}
                     />
                 </CollapsibleSection>
```
aarondr77, line 93: I think we can take this out of a CollapsibleSection again, then LGTM

PR #802, opened 2023-06-13
Parameterize export path

naterush:
# Description

Adds export to file path. Has to change how we handle multiple CSV exports so things don't break if you change imports -- but working well!

Also fixes some bug in displaying the paths on the frontend. This functionality was pretty easy to add, shoutout this infrastructure!

# Testing

Use it. Change some imports maybe too.

# Documentation

N/A.


aarondr77:
I think it would be super helpful to 1) Not have these both listed as file paths, but maybe have one listed as `import file path` and the other as `export file path`. and 2) Maybe have a divider line in the dropdown between the import and export file paths?

Maybe not 2. But def 1. 

Screenshot 2023-06-13 at 8 05 05 PM

aarondr77 (commented):
LGTM once fixed up

PR #804, opened 2023-06-17
fix: save scroll position in session storage

Udit-takkar:
# Description

Fixes: https://github.com/mito-ds/monorepo/issues/94

I have saved the scroll position of sheet in the session storage.

There is one more method that i tried which was changing the type of gridState.scrollPosition to Record<number, Scrollposition> where key would be sheetIndex. so we store all scrollPositions by key but the session storage method is much more simpler.

https://github.com/mito-ds/monorepo/assets/53316345/02c31fd0-d61f-4598-8254-123b27e4063f

# Testing

1) Import two data frames with enough data size to scroll in.
2) Now scroll on first sheet to the end and check second sheet if it is still at the previous position or not

# Documentation


mitosheet/mitosheet/mito_frontend.css
Udit-takkar, line 1: i am not sure about changes in mito_frontend.css and mito_frontend.css

naterush:
@Udit-takkar this looks awesome! I'm going to dig into this before the end of the week - currently mostly Afk so might be a bit slower turn around than normal.

Thanks for getting this up.

Udit-takkar:
Thanks @naterush

naterush:
@Udit-takkar so sorry for the later review here, and thanks for the bump on this by opening another PR. It's the best possible way you could have bumped me :) 

I think this PR suffers from the same issues the other does -- that we shouldn't be storing things in local storage if we can avoid it, as it leads to confusing state duplication. 

What were the complexities with storing this in gridState? I guess that's specific to one sheet currently... hmm. I wonder if GridState itself should just be an array, one for each sheet?

Udit-takkar:
@naterush i used session storage because so that even when i refresh the page my scroll position is persisted and it was much simpler to implement.

I don't remember exactly what difficulty i faced with the other approach of storing position in state but there was many bugs and each cases. I'll just try again and come up with a approach without session storage.

naterush:
You rock! Happy to chat about it more if its helpful, to ideate with you on solutions. Let me know.

PR #805, opened 2023-06-27
Snowflake role

aarondr77:
# Description

Select the `role` using Snowflake import

# Testing

Use the Pytest account and try switching between the two roles and make sure it works properly. 

# Documentation


mitosheet/mitosheet/api/get_available_snowflake_options_and_defaults.py
```diff
@@ -61,11 +61,20 @@ def get_available_snowflake_options_and_defaults(params: Dict[str, Any], steps_m
...
         con = con_or_exception
 
+        _role = table_loc_and_warehouse.get('role')
         _warehouse = table_loc_and_warehouse.get('warehouse')
         _database = table_loc_and_warehouse.get('database') 
         _schema = table_loc_and_warehouse.get('schema')
         _table_or_view = table_loc_and_warehouse.get('table_or_view')
 
+        roles = get_roles(con)
+        role = _role if _role is not None else get_default_role(roles)
```
naterush, line 71: Let's not make this go and do another DB lookup _every time_. The performance here is already pretty bad. 

you can just move the `get_roles` call down to the next line, as a `get_default_role` parameter.

mitosheet/src/components/taskpanes/SnowflakeImport/SnowflakeImportTaskpane.tsx
```diff
@@ -239,7 +274,12 @@ const SnowflakeImportTaskpane = (props: SnowflakeImportTaskpaneProps): JSX.Eleme
                                 value={params.table_loc_and_warehouse.database || 'None available'}
                                 disabled={loadingAvailableOptionsAndDefaults}
                                 onChange={(newDatabase) => {
-                                    const newParams = getNewParams(params, newDatabase)
+                                    const newParams = getNewParams(
+                                        params, 
+                                        params.table_loc_and_warehouse.role, 
+                                        params.table_loc_and_warehouse.role, 
+                                        newDatabase
```
naterush, line 281: Yeah. So this looks wrong -- but I also think it's probably wrong as a result of the types on this function -- all strings!

My thinking is that we should create a new function here that uses types similar to the stuff we use in graph, where there are partial updates, and we pass things by name (the reason we have this separate function is to set things to undefined, if I remember correctly). 

Thoughts?

naterush (changes requested):
Two minor refactor opportunities, then ping me and I'll test.

naterush (changes requested):
Minor request for changes.

naterush:
LGTM once you attempt above.

mitosheet/mitosheet/api/get_available_snowflake_options_and_defaults.py
aarondr77: Okay, unfortunetly this is taking longer than expected, so will merge as is.

PR #806, opened 2023-06-28
Move mito into folder

naterush:
# Description

- Moves into a separate folder in `mitosheet/src` for easier refactoring.
- Makes Jupyter utilities optional props to the `Mito` package

# Testing

Normal testing.

# Documentation

N/A.


aarondr77:
This PR seems to introduce a new comm setup bug for notebooks only. See below that I can create a comm on dev, but not here.  

On this branch:
https://github.com/mito-ds/monorepo/assets/18709905/039c828b-0e7c-457d-a6be-933545d396ff

On dev:

https://github.com/mito-ds/monorepo/assets/18709905/fd6292ce-7a78-4924-84d8-37f15a358c07

mitosheet/src/mito/components/Mito.tsx
```diff
@@ -7,25 +7,24 @@ import React, { Fragment, useCallback, useEffect, useRef, useState } from 'react
...
-import '../../css/sitewide/fonts.css';
-import '../../css/sitewide/height.css';
-import '../../css/sitewide/hover.css';
-import '../../css/sitewide/icons.css';
-import '../../css/sitewide/margins.css';
-import '../../css/sitewide/paddings.css';
-import '../../css/sitewide/scroll.css';
-import '../../css/sitewide/text.css';
-import '../../css/sitewide/widths.css';
+import '../../../css/sitewide/all-modals.css';
```
aarondr77, line 10: Maybe this file should come out of the `components` subfolder and as a top layer in the `mito` folder

mitosheet/src/mito/components/Mito.tsx
```diff
@@ -713,6 +721,8 @@ export const Mito = (props: MitoProps): JSX.Element => {
                     setCurrPathParts={setCurrPathParts}
     
                     failedReplayData={uiState.currOpenTaskpane.failedReplayData}
+
+                    overwriteAnalysisToReplayToMitosheetCall={props.jupyterUtils?.overwriteAnalysisToReplayToMitosheetCall}
```
aarondr77, line 725: Instead of passing these potentially undefined utilities throughout the app, maybe we should handle the undefined case in one place here. Would make types easier everywhere else.

aarondr77: Currently if these are undefiend, I can keep using Mito, which is risky. Someone uses Mito for while and never gets an analysis_id or generated code, so they lose all of their work. 

There used to be a similar problem if the analysis_id had spaces around it, for example, and it caused users to lose their work!

aarondr77 (commented):
Nice work! I can see the path to streamlit! Two things here to cleanup IMO. 

1. Making sure this works in notebook
2. Error handling when the jupyterUtils are not defined

mitosheet/src/mito/components/Mito.tsx
naterush: There are many cases where you _don't want_ the analysis ID to be written! Namely, if you are not in Juypter, you don't want the frontend to write this ID at all. 

In our streamlit case, the author of the streamlit app is going to be responsible for configuring how the analysis gets saved. But this is to say: it is both valid and legit to have these undefined and not what to do anything with them!

naterush:
@aarondr77 ok -- I was able to replicate the bug with Mito not showing up, but I moved some things around and I am now getting the same behavior as on dev. 

Wanna try pulling this branch and testing again? This is def something we should sort (and generally fix our NB extension), but I'm ok with this PR as long as it doesn't make things worse.

aarondr77:
This behavior is a bit different from the dev behavior... neither of them work the first time you run mitosheet.sheet() call but on dev the mitosheet.sheet() call didn't terminate until I got bored and reran it, then it worked. Here, it ran, terminated and didn't create a mitosheet until I reran it. 

https://github.com/mito-ds/monorepo/assets/18709905/183cdb6c-be8e-4b46-a1a0-707015d93849

aarondr77:
This is what I get on notebook now. 

https://github.com/mito-ds/monorepo/assets/18709905/31eb1791-0299-45b3-a20b-740c638ed097


This is the console message I get when opening the notebook, before running any code. The Mito extension failed to load. 
Screenshot 2023-07-07 at 2 55 16 PM

aarondr77:
Okay, I think it's working?? I'm having a hard time testing it because sometimes it works and sometimes it doesn't.... idk if something is wrong with my development environment -- I'm not used to switching back and forth between branches to test things on Jupyter notebooks, so possibly there is something whacky going on there?

PR #808, opened 2023-07-06
Add filename field to export

marthacryan:
## Description
This PR adds a "file name" input field into the export task pane:
image

It displays a placeholder and is optional for the user, defaulting to "MitoExport":
image

It validates for forbidden characters in filenames:
image

## Todo
- [x] Add validation
  - [x] Prevent form submission when field is invalid
- [ ] Add documentation
- [ ] Add comments to code
- [x] Update css to align all input fields

### DIFF
--- a/mitosheet/mitosheet/api/get_available_snowflake_options_and_defaults.py
+++ b/mitosheet/mitosheet/api/get_available_snowflake_options_and_defaults.py
@@ -61,10 +61,19 @@ def get_available_snowflake_options_and_defaults(params: Dict[str, Any], steps_m
 
         con = con_or_exception
 
+        _role = table_loc_and_warehouse.get('role')
         _warehouse = table_loc_and_warehouse.get('warehouse')
         _database = table_loc_and_warehouse.get('database') 
         _schema = table_loc_and_warehouse.get('schema')
         _table_or_view = table_loc_and_warehouse.get('table_or_view')
+
+        roles = get_roles(con)
+        role = _role if _role is not None else get_default_role(roles)
+
+        # After getting the role, use it so that we get the correct list of 
+        # databases, schemas, tables, etc.
+        if con is not None:
+                con.cursor().execute(f'USE ROLE {role}')
 
         warehouses = get_warehouses(con)
         warehouse = _warehouse if _warehouse is not None else get_default_warehouse(warehouses)
@@ -82,10 +91,10 @@ def get_available_snowflake_options_and_defaults(params: Dict[str, Any], steps_m
 
         columns = get_columns(con, database, schema, table_or_view)
 
-
         return {
                 'type': 'success',    
                 'config_options': {
+                        'roles': roles,
                         'warehouses': warehouses,    
                         'databases': databases,    
                         'schemas': schemas,    
@@ -93,13 +102,27 @@ def get_available_snowflake_options_and_defaults(params: Dict[str, Any], steps_m
                         'columns': columns
                 },
                 'default_values': {
+                        'role': role,
                         'warehouse': warehouse,
                         'database': database,
                         'schema': schema,
                         'table_or_view': table_or_view,
                 },
         }
-                
+
+def get_roles(con: MitoSafeSnowflakeConnection) -> List[str]:
+        if con is None:
+                return []
+
+        # We use SHOW GRANTS to show all of the roles granted to the user. 
+        # Instead of SHOW ROLES which shows all of the roles in the account.
+        cur = con.cursor().execute('SHOW GRANTS')   
+
+        if cur is None:
+                return []
+        
+        roles = cur.fetchall() # type: ignore
+        return [role[1] for role in roles] # type: ignore
 
 def get_warehouses(con: MitoSafeSnowflakeConnection) -> List[str]:
         if con is None: 
@@ -178,6 +201,9 @@ def get_columns(con: MitoSafeSnowflakeConnection, database: Optional[str], schem
 
         columns = cur.fetchall() # type: ignore
         return [column[2] for column in columns] # type: ignore
+
+def get_default_role(roles: List[str]) -> Optional[str]:
+        return roles[0] if len(roles) > 0 else None
 
 def get_default_warehouse(warehouses: List[str]) -> Optional[str]:
         return warehouses[0] if len(warehouses) > 0 else None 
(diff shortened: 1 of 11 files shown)


naterush:
@marthacryan the frontend tests tend to be flakier than is ideal - so if only one of them is failing (as is the case currently) -- feel free to ignore.

mitosheet/src/types.tsx
```diff
@@ -825,8 +825,9 @@ export interface MitoStateUpdaters {
     setUIState: React.Dispatch<React.SetStateAction<UIState>>,
 }
 
-export interface CSVExportState { exportType: 'csv' }
-export interface ExcelExportState { exportType: 'excel', sheetIndexes: number[] }
+export interface ExportState { fileName?: string, exportType: 'csv' | 'excel' }
+export interface CSVExportState extends ExportState { exportType: 'csv' }
+export interface ExcelExportState extends ExportState { exportType: 'excel', sheetIndexes: number[] }
```
naterush, line 830: These are pleasant types! We rarely use extends, but just b/c we forget about them. Nice!

mitosheet/src/components/taskpanes/Download/DownloadTaskpane.tsx
```diff
@@ -169,13 +199,35 @@ const DownloadTaskpane = (props: DownloadTaskpaneProps): JSX.Element => {
...
+                    </Row>
+                    { invalidFileNameWarning !== undefined && <Row justify='end' align='end'> <p className="text-color-error">{invalidFileNameWarning}</p> </Row> }
+                    { props.uiState.exportConfiguration.exportType === 'excel' && 
                         <ExcelDownloadConfigSection 
                             dfNames={props.dfNames}
                             mitoAPI={props.mitoAPI}
                             userProfile={props.userProfile}
                             sheetDataArray={props.sheetDataArray}
-                            exportState={props.uiState.exportConfiguration}
+                            exportState={props.uiState.exportConfiguration as ExcelExportState}
```
naterush, line 230: So sad TS can't type check this </3

mitosheet/src/components/elements/TextButton.tsx
```diff
@@ -70,6 +70,11 @@ interface TextButtonProps {
         * @param [disabledTooltip] -- Message to display as tooltip when button is disabled 
     */
     disabledTooltip?: string
+
+    /**
+     * Boolean to indicate whether the button is showing error. Impacts color.
```
naterush, line 75: This doesn't appear to impact color any more than disabled. Can we rip this prop out and just pass things through disabled for now?

mitosheet/src/components/taskpanes/Download/DownloadTaskpane.tsx
```diff
@@ -21,6 +22,18 @@ import DefaultEmptyTaskpane from '../DefaultTaskpane/DefaultEmptyTaskpane';
 import DefaultTaskpaneFooter from '../DefaultTaskpane/DefaultTaskpaneFooter';
 
 
+const INVALID_CHARACTERS_IN_FILENAME = [
```
naterush, line 25: Given that we define this constant, and the code that consumes it twice across both download taskpanes, how would you feel about moving this into a shared utility like `getInvalidFileNameError(fileName: string): string | undefined`?

mitosheet/src/components/taskpanes/Download/DownloadTaskpane.tsx
```diff
@@ -207,13 +259,14 @@ const DownloadTaskpane = (props: DownloadTaskpaneProps): JSX.Element => {
...
-                    disabled={exportString === '' }
+                    error={!!invalidFileNameWarning}
+                    disabled={!!invalidFileNameWarning || exportString === '' }
                     href={exportHRef} 
                     download={exportName}
                     onClick={onDownload}  
                 >
                     
-                    {exportString === '' ? (<>Preparing data for download <LoadingDots /></>) : `Download ${props.uiState.exportConfiguration.exportType === 'csv' ? 'CSV file': 'Excel workbook'}`}
+                    {invalidFileNameWarning ? <>Invalid input.</> : exportString === '' ? (<>Preparing data for download <LoadingDots /></>) : `Download ${props.uiState.exportConfiguration.exportType === 'csv' ? 'CSV file': 'Excel workbook'}`}
```
naterush, line 269: Maybe slightly more helpful: 

```suggestion
                    {invalidFileNameWarning ? <>{invalidFileNameWarning}</> : exportString === '' ? (<>Preparing data for download <LoadingDots /></>) : `Download ${props.uiState.exportConfiguration.exportType === 'csv' ? 'CSV file': 'Excel workbook'}`}
```

naterush (changes requested):
This looks awesome -- great work getting this up today. Your code is clean, works well, and touches the exact bit of code that makes sense. Couldn't ask for more --  whenever I touch new codebases I find myself mucking about with random/wrong internals :) 

Two minor changes that I missed when specifying, but I think would be nice polish before merging: 

## Reordering/renaming the fields

Namely, so they match between the two download taskpanes. I think keeping the ExportToFileTaskpane as is, and then:
1. changing the order of file name and file type, 
2. renaming from export type to file type in this taskpane 

would be a nice easy way to standardize. 

## Button flickering

The `URL.createObjectURL(new Blob(` code reruns whenever you edit the file name, which leads to a weird flickering (which is worse with larger files).

Probably the easiest thing to do would be to store the exportHRef with a `useState`, and then just update it when you load a new exportString (might even be able to stop storing exportString). There are probably other ways of handling this with a useEffect or something -- so feel free to take whatever approach makes the most sense for you. 

Let me know thoughts on these, if you think other approaches might be better! Sorry for missing em in the spec :}

mitosheet/src/components/taskpanes/Download/DownloadTaskpane.tsx
marthacryan: I guess what I was thinking is that if we add any more validation that we'd want it to be more general (in case there were multiple validation errors). But now that I'm thinking about it, I do think that generally when a form is invalid, the submission button becomes disabled but the actual text stays the same, and then you'd have more info in a tooltip or add some red text above the submission button. Maybe we can brainstorm a little about the design

mitosheet/src/types.tsx
marthacryan: Thanks!

marthacryan:
> 1. changing the order of file name and file type,
> 2. renaming from export type to file type in this taskpane
> 
> would be a nice easy way to standardize.

Updated!

> Probably the easiest thing to do would be to store the exportHRef with a `useState`, and then just update it when you load a new exportString (might even be able to stop storing exportString). There are probably other ways of handling this with a useEffect or something -- so feel free to take whatever approach makes the most sense for you.

Updated the code to just use `exportHref` as a state variable instead of `exportString` (because `exportString` was only being used by `exportHref`). Let me know how that looks!

naterush:
Flickering is reduced! It's still sometimes here, I think because the useDebouncedEffect has the entire export state as it's dependency (which includes the file name). Luckily, the debouncing reducing the impact of this.

LGTM! I'll squash those merge conflicts :-)

PR #809, opened 2023-07-06
Adds parameters

naterush:
# Description

This allows you to specify parameters to import. E.g.

```
def get_data(nnnnnn, i: int, f: float, s: str):
    import pandas as pd
    
    return pd.DataFrame({'A': [nnnnnn, i, f, s]})

mitosheet.sheet(importers=[get_data], analysis_to_replay="id-gwoviyfngk")
```

Notably, we don't support all param types yet. Also, errors only popup when you click import (e.g. if you try and pass "ABC" to an integer param) - this is because I tried my best to keep the casting in one location, which I think makes sense. 

There's some weirdness with params of types `any`, but I think it's ok for now and is a good step forward for the conversations we want to have. 

# Testing

Use it!

# Documentation

We can add this to a foundational page, potentially? Or at least the docs.


mitosheet/mitosheet/tests/step_performers/test_user_defined_import.py
```diff
@@ -24,26 +24,48 @@ def create_df_diff() -> pd.DataFrame:
...
+        {},
         [
             create_df(), create_df_diff()
         ]
     ),
+    (
+        [],
+        "get_df_with_params",
+        {
+            'n': '1',
```
naterush, line 55: Try with a weirder object, like a dictionary

mitosheet/src/components/taskpanes/UserDefinedImport/UserDefinedImportTaskpane.tsx
```diff
@@ -34,10 +36,33 @@ const getDefaultParams = (
...
+    paramType: UserDefinedImporterParamType
+): string => {
+    if (paramType === 'str') {
+        return 'string'
+    } else if (paramType == 'float') {
+        return 'float'
+    } else if (paramType == 'int') {
+        return 'int'
+    } else {
+        return 'unknown'
```
naterush, line 65: This isn't helpful. Just remove it.

mitosheet/mitosheet/code_chunks/user_defined_import_code_chunk.py
```diff
@@ -4,16 +4,36 @@
...
 # Distributed under the terms of the GPL License.
-from typing import List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple
 from mitosheet.code_chunks.code_chunk import CodeChunk
-from mitosheet.types import ColumnID
+from mitosheet.types import ColumnID, UserDefinedImporterParamType
 from mitosheet.state import State
+from mitosheet.transpiler.transpile_utils import column_header_to_transpiled_code
+
+def get_transpiled_importer_params(importer_params_and_type_and_value: Dict[str, Tuple[UserDefinedImporterParamType, str]]) -> str:
```
naterush, line 13: What if we just passed through the right values, and called the transpile on them? I think this should work -- as long as the eval is right. Add the dict test first.

naterush (commented):
Minor cleanup

aarondr77:
- After using the custom import, if you go to update data, you get this.
- After using the custom import, if you clear the analysis, the data disappears 

Let's make sure we're doing the right import accounting for these imports.  

Screenshot 2023-07-07 at 9 45 26 AM

mitosheet/src/components/taskpanes/UserDefinedImport/UserDefinedImportTaskpane.tsx
```diff
@@ -109,6 +136,60 @@ const UserDefinedImportTaskpane = (props: UserDefinedImportTaskpaneProps): JSX.E
                                 </Select>
                             </Col>
                         </Row>
+                        {userDefinedImporter?.parameters !== undefined && Object.keys(userDefinedImporter?.parameters).length > 0 && 
+                            <Row justify='start' align='center' title='TODO'>
+                                <Col>
+                                    <p className='text-header-3'>
+                                        Params
```
aarondr77, line 143: Maybe change the text to `Set Param Values` and then have ? that explains what this section is. The end users are probably going to not really understand this.

aarondr77:
Maybe update this text to: 

```
Custom Importers is how you can access your functions that create dataframes from within the Mito Spreadsheet. 

To access a custom importer in Mito, pass the function to the mitosheet.sheet() call like this:

mitosheet.sheet(importers=[get_data])
```

Screenshot 2023-07-07 at 10 25 52 AM

aarondr77:
We should make these required fields 

Screenshot 2023-07-07 at 10 39 20 AM

aarondr77 (commented):
Pretty cool! This is a good start to showing enterprises how they can plug and play functionality into Mito

aarondr77:
Let's make sure to add this to the website plans page underneath `User Defined Functions`. 

And then when we get the foundational page outline back from SEO team, we can make a full foundational page about this stuff

mitosheet/src/components/taskpanes/UserDefinedImport/UserDefinedImportTaskpane.tsx
```diff
@@ -109,6 +136,60 @@ const UserDefinedImportTaskpane = (props: UserDefinedImportTaskpaneProps): JSX.E
...
+                                </Col>
+                            </Row>
+                        }
+                        {userDefinedImporter?.parameters !== undefined && Object.entries(userDefinedImporter.parameters).map(([paramName, paramType]) => {
+                            const paramValue = params.importer_params[paramName];
+                            
+                            let inputElement = null;
+                            if (paramType === 'str' || paramType === 'any' || paramType == 'int' || paramType == 'float') {
+                                inputElement = (
+                                    <Input
```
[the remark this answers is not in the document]
naterush: I'd rather do not default value support than potentially be misleading. Let's leave default value support for another sprint on this, if we want to flesh it out more due to perceived interest.

naterush:
There are 4 users in the past month who have updated imports in the middle of an analysis. As far as I can tell, there is one user who is updating imports when an analysis fails. 

I think we should not invest more in this feature currently. It's a huge maintain burden, and given that we also have no one (and will have no one) using this feature for a bit, I think it's a solidly bad use of time!

If we get users on either updating imports or users on this, we can make them work well together. I'll open an issue.

mitosheet/mitosheet/step_performers/user_defined_import.py
```diff
@@ -80,6 +82,8 @@ def get_user_defined_importer_params_from_frontend_params(f: Callable, frontend_
                 user_defined_importer_params[param_name] = int(param_value)
             elif param_type == 'float':
                 user_defined_importer_params[param_name] = float(param_value)
+            elif param_type == 'bool':
+                user_defined_importer_params[param_name] = 'true' in param_value.lower()
```
aarondr77, line 86: We could use the utility `is_env_variable_set_to_true` to do this check and support things like `T` also, which might be helpful

mitosheet/src/components/taskpanes/UserDefinedImport/UserDefinedImportTaskpane.tsx
```diff
@@ -137,13 +150,7 @@ const UserDefinedImportTaskpane = (props: UserDefinedImportTaskpaneProps): JSX.E
...
                         </Row>
                         {userDefinedImporter?.parameters !== undefined && Object.keys(userDefinedImporter?.parameters).length > 0 && 
-                            <Row justify='start' align='center' title='TODO'>
-                                <Col>
-                                    <p className='text-header-3'>
-                                        Params
-                                    </p>
-                                </Col>
-                            </Row>
+                            <LabelAndTooltip tooltip="">Import Parameters</LabelAndTooltip>
```
aarondr77: Missing tooltip here

mitosheet/src/components/taskpanes/UserDefinedImport/UserDefinedImportTaskpane.tsx
```diff
@@ -168,16 +175,32 @@ const UserDefinedImportTaskpane = (props: UserDefinedImportTaskpaneProps): JSX.E
...
                                     />
                                 )
-                            } else {
-                                // TODO: in the future, handle other types (e.g. booleans or Unions here)
+                            } else if (paramType === 'bool') {
+                                inputElement = (
+                                    <Toggle
+                                        value={paramValue.toLowerCase().includes('true')}
+                                        onChange={() => {
+                                            const newValue = !paramValue.toLowerCase().includes('true');
```
aarondr77: Why aren't we just using actual True and False values here instead of checking if the param Value includes true?

aarondr77 (commented):
Small cleanup then LGTM

aarondr77:
After using the custom import, if you clear the analysis, the data disappears

mitosheet/src/components/taskpanes/UserDefinedImport/UserDefinedImportTaskpane.tsx
naterush: I'll lead some documentation. We want to do all the parsing in one location for all these params, and since we support all sorts of Python objects (e.g. dicts), which we can't parse on the frontend, we just do all the parsing on the backend and treat everything on the frontend as a string.

naterush:
LGTM when tests pass.

PR #816, opened 2023-07-07
Add filter for string contains case insensitive

marthacryan:
Adds a string filter condition for contains w/ case insensitive. 

For comparison, these are the results with contains:
image

And these are with contains case insensitive:
image


mitosheet/mitosheet/tests/step_performers/test_filter.py
```diff
@@ -259,6 +260,12 @@
         "2",
         pd.DataFrame(data={"A": ["12", "122222"]}),
     ),
+    (
+        pd.DataFrame(data={"A": ["abcDEF", "ABCdef", "def", "dEf"]}),
+        FC_STRING_CONTAINS_CASE_INSENSITIVE,
+        "aBc",
+        pd.DataFrame(data={"A": ["abcDEF", "ABCdef"]}),
+    ),
```
naterush, line 268: Two more tests that might be nice: 
1. Combining two contains case insensitive with an And
2. Combinion two contains case insensitive with an Or

Just so we cover all our cases here and make sure they are working properly!

naterush (changes requested):
Great work, it's working perfect! Hopefully you got a chance to poke around the backend a little bit as well. 

A coupla tests we could add to cover our bases, and then feel free to merge into dev once tests pass (no need for another review)!

PR #818, opened 2023-07-09
Dogfood - 7/9

naterush:
# Description

- mitosheet: make rename select correct tab when multiple mitosheets
- stops params from switching orders if you had multiple in your code config taskpane
- A bug I couldn't fix with renaming graph tabs through dropdowns (issue here: https://github.com/mito-ds/monorepo/issues/817)

# Testing

If you want to do a 30 minute workflow, that would be fantastic as well - there are lots of refactors going into this change. I tested literally every bit of functionality, but a double check would be sweet. 

# Documentation

A bunch of changes. TODO.


aarondr77:
Just noting, but don't need to fix now. 

https://github.com/mito-ds/monorepo/issues/819

aarondr77:
1. Create a graph 
2. Go into full screen mode, notice that the graph does not change sizes
3. Edit the graph, notice that the graph still does not change size

Screenshot 2023-07-09 at 8 29 46 PM

aarondr77:
In the right hand corner of the formula bar, there is the icon to resize the formula bar, but we don't allow it. We should hide the resizer 

Screenshot 2023-07-09 at 8 33 43 PM

aarondr77:
When using the formula bar + clicking on column headers to reference them, I'm only able to add the column I am editing the formula of. 

https://github.com/mito-ds/monorepo/assets/18709905/7248f6de-aaa1-47c1-90da-4981abe5d64e

aarondr77:
Click on an example AI prompt that goes to the second line in the text area

See that the text area doesn't expand until you type more

https://github.com/mito-ds/monorepo/assets/18709905/e8fbdf25-6ef1-4b71-aa4c-0fc84c5257ac

aarondr77:
Sometimes using `write code snippet to notebook` makes the notebook do something weird. It looks like it closes Mito and reopens it?? 

https://github.com/mito-ds/monorepo/assets/18709905/89408ac2-493f-4bc1-8e92-d0bf78d74837

aarondr77:
We should either add a hint here that says something like "The first sheet is at index 0" or change it so if the user does 1, it selects the first sheet

Screenshot 2023-07-09 at 9 01 50 PM

aarondr77:
> mitosheet: make rename select correct tab when multiple mitosheets

This is working

aarondr77:
> stops params from switching orders if you had multiple in your code config taskpane

This is working

PR #825, opened 2023-07-11
Add support for highlighting border between columns when dragging

marthacryan:
This PR implements the following behavior: when the user is dragging a column header across the page to reorder, the UI highlights the border between columns that it would be placed in to emphasize to the user where the column is being dropped. 

image

**Note**: Deleted the `mito_frontend.*` files because they're now in the gitignore.


mitosheet/src/mito/components/endo/ColumnHeaders.tsx
```diff
@@ -88,6 +89,11 @@ const ColumnHeaders = (props: {
                         e.preventDefault()
                         e.persist()
 
+                        // First we handle highlighting the border
```
aarondr77, line 92: The `onDragOver` event handler is also triggered when we are resizing columns, so in order to only show the highlighted border when we're reordering columns, we need to case on the data transfer operation. You can see an example of this a few lines down in the `onDrop` function. 

https://github.com/mito-ds/monorepo/assets/18709905/731aa45d-85bd-4ec6-9647-5c68e8159d18

mitosheet/src/mito/components/endo/ColumnHeaders.tsx
```diff
@@ -88,6 +89,11 @@ const ColumnHeaders = (props: {
...
+                        const {columnIndex} = getIndexesFromXAndY(e.clientX, e.clientY)
+                        props.setUIState({ ...props.uiState, highlightedColumnIndex: columnIndex });
```
aarondr77, line 94: Every time we update the UIState, the entire grid gets rerendered, so this is working well on small datasets, but when I tested it on a larger dataset, it becomes really laggy. 

Here's a few refactor ideas that might make this more performant: 

1. One simple idea is to only update the `highlighitedColumnIndex` if it changes. That should reduce a ton of the rerenders! 
2. Instead of using the UIState to keep track of the highlightedColumnIndex, I wonder if we could instead just keep track that we're in `reorder mode` and apply a class like `highlight-border` to the column that is hovered over. The challenge here would be figuring out based on where the cursor is what column to apply the class too because we don't have access to the drag event in `GridData.tsx` and it would be hard to link the column header that we are hovering over to the column, I think.. 

There's a few large datasets in the monorepo that you can use for testing purposes -- you can find them in the mitosheet/datasets/ folder

aarondr77 (commented):
Nice job with the spec + implementation. This makes reordering columns a much more pleasant experience. 

This is working really well when the grid has only a few columns, but when the data gets larger, it becomes a big laggy. I left a couple of refactor ideas that might be helpful, but happy to brainstorm more on this!

mitosheet/src/mito/components/endo/ColumnHeaders.tsx
naterush: Haven't looked at the code (so excuse me popping in here), but another idea here I thought I'd throw out -- would it be possible to just use state that we already have to get the reorder mode information?

If we already store the dragged column index somewhere, we could just highlight whatever it is over potentially (I am not sure we do -- so this might just be silly). 

Another possible optimization here might be storing the highlightedColumnIndex within the ColumnHeaders component -- so only this rerenders -- rather than in the UI state.

marthacryan: > Every time we update the UIState, the entire grid gets rerendered, so this is working well on small datasets, but when I tested it on a larger dataset, it becomes really laggy.

I'm having some trouble reproducing the laggy-ness (even in the `datasets/large-datasets` directory), is there a specific spreadsheet you're seeing that behavior for? Maybe it's the way I'm doing it? 

> Here's a few refactor ideas that might make this more performant:
> 
> 1. One simple idea is to only update the `highlighitedColumnIndex` if it changes. That should reduce a ton of the rerenders!

Just tried that! It worked well for me so I'll try pushing that in case it fixes the laggy-ness for you

> 2. Instead of using the UIState to keep track of the highlightedColumnIndex, I wonder if we could instead just keep track that we're in `reorder mode` and apply a class like `highlight-border` to the column that is hovered over. The challenge here would be figuring out based on where the cursor is what column to apply the class too because we don't have access to the drag event in `GridData.tsx` and it would be hard to link the column header that we are hovering over to the column, I think..

Maybe I'm misunderstanding, but I think the issue with this is that the cells are rendered as rows, so we don't have a "column" to add a drag handler / class onto. So I think we'd need to re-render either way? But I'm not totally sure! 

Also - we do keep track of a "reorder mode" but I'm not sure if it helps much because it mostly keeps track of whether we're readjusting the size of the column or reordering the columns. 

> would it be possible to just use state that we already have to get the reorder mode information?

Yep! See above answer 
 
> If we already store the dragged column index somewhere, we could just highlight whatever it is over potentially (I am not sure we do -- so this might just be silly).

We're not keeping track of it that I know of - we're using the `onDrop` function to handle the actual re-ordering, so what happens while you're dragging isn't really involved with that.  

> Another possible optimization here might be storing the highlightedColumnIndex within the ColumnHeaders component -- so only this rerenders -- rather than in the UI state.

We need that variable for all of the cells in that column to also have a highlighted border - unless we want to only highlight the border between column headers? I could see an argument for that

mitosheet/src/mito/components/endo/GridData.tsx
```diff
@@ -93,7 +93,7 @@ const GridData = (props: {
                                     className={className} key={columnIndex}
                                     style={{
                                         width: `${cellWidth}px`,
-                                        ...getBorderStyle(props.gridState.selections, props.gridState.copiedSelections, rowIndex, columnIndex, sheetData.numRows),
+                                        ...getBorderStyle(props.gridState.selections, props.gridState.copiedSelections, rowIndex, columnIndex, sheetData.numRows, props.uiState.highlightedColumnIndex),
```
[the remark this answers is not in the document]
marthacryan: Ah totally! Just missed adding that as an argument in that file.

mitosheet/src/mito/components/endo/ColumnHeaders.tsx
marthacryan: Actually just checked and it only works when you drop the header on the same level as a header, so I think it might be more intuitive if we only highlight the border between headers - should I go ahead and add that?

marthacryan: Ok just tried that out and it's a lot harder to see - especially when you're trying to drop the column on top of the header already. 
image

marthacryan: Update from conversation with Aaron: the lagginess is fixed but we noticed that when you drop the header on the *grid*, the highlighted column ends up staying highlighted. So I'm going to try to add an `onDragOver` or `onDrop` handler for the `EndoGrid` and see if I can fix it up. In another PR, we might try to update the logic to allow users to drop the column header over the `EndoGrid` to reorder columns.

mitosheet/src/mito/components/endo/ColumnHeaders.tsx
marthacryan: Shoot - just realized (as is described in this comment:) https://github.com/mito-ds/monorepo/blob/f43b8fcab2c157325fc6c0b108679c9272e8039b/mitosheet/src/mito/components/endo/ColumnHeaders.tsx#L80) we won't be able to access the operation data from that function. Not sure what the best solution is for that one. Any ideas?

marthacryan: At least since my recent change, the highlighted column won't stay there? Maybe I can check for operation in the border style function as a workaround? Or add the operation into the uistate object?

marthacryan: Oh lol nevermind, there actually is a state object to track that in this file already. Just updated!

PR #826, opened 2023-07-11
Add a link to documentation in the cell editor dropdown

marthacryan:
Adds a link to the [documentation](https://docs.trymito.io/how-to/interacting-with-your-data/mito-spreadsheet-formulas) on formulas in the cell editor dropdown:
image
image


mitosheet/dev/macsetup.sh
marthacryan, line 1: Not sure why this is happening - maybe because I ran `chmod +x dev/macsetup.sh`?

mitosheet/css/endo/CellEditor.css
```diff
@@ -41,6 +41,12 @@
     position: absolute;
 }
 
+.jp-RenderedHTMLCommon .mito-container a.cell-editor-dropdown-link {
```
aarondr77, line 44: Usually we just use a single class name to identify elements to apply css to. I think here you could get away with just using `.cell-editor-dropdown-link`. 

The nice part about just using `.cell-editor-dropdown-link` is that if we use this component in streamlit, for example, the class would still apply even if the `.jp-RenderedHTMLCommon` isn't there. 

Are there advantages to using this more specific class selection technique that aren't on our radar right now?

mitosheet/css/endo/CellEditor.css
```diff
@@ -41,6 +41,12 @@
...
+    margin: 0.5em;
```
aarondr77, line 45: Across the codebase right now, we only use `px` instead of `em`. It probably isn't best practice (in the trymito.io codebase we use `em` instead), but for consistency right now, I think it might be best to stick with `px` just because other things are hardcoded for the correct `px` settings currently. 

At some point we should do a css refactor, it needs a bit of care.

mitosheet/src/mito/components/endo/celleditor/CellEditorDropdown.tsx
```diff
@@ -210,6 +211,14 @@ const CellEditorDropdown = (props: {
                     </div>
                 </div>
             }
+            {/* Always display a link to the documentation */}
+            <a
+                className='text-body-2 text-body-2-link cell-editor-dropdown-link'
```
aarondr77, line 216: btw, we created a utility that helps us avoid missing space bugs in class names. The function is called `classNames` and it just takes each of these strings as arguments and returns this exact string. 

It also lets you pass dictionaries to it with the format {<class name> : <True or False>} so that you can conditionally apply css classes.

aarondr77 (commented):
This is working great + I think it's such a good idea for discoverability! yay using Mito and fixing the issues we run into. 

Just a few small comments to take a look at. 

I just reran the tests, so let's see if they pass this time!

mitosheet/src/mito/components/endo/celleditor/CellEditorDropdown.tsx
```diff
@@ -210,6 +211,14 @@ const CellEditorDropdown = (props: {
...
+                target='_blank'
+                href={FORMULA_DOC_LINK}
```
aarondr77, line 218: Addressing this `Using target="_blank" without rel="noreferrer" (which implies rel="noopener") is a security risk in older browsers` should make the linter pass

mitosheet/css/endo/CellEditor.css
marthacryan: Ah yeah that was because it was being overridden by the JupyterLab stuff (I usually try to make it as brief as possible too!). I saw this a little in other parts of the CSS so I think it might be something you've run into before: https://github.com/mito-ds/monorepo/blob/de5decfdd6c935c33705f53f21713de53e983e7b/mitosheet/css/mito.css#L20

PR #828, opened 2023-07-12
mitosheet: add basic streamlit support

naterush:
# Description

This adds basic support for streamlit in Mito. See the testing instructions below. 

All features should be working as expected, with a few minor exceptions (documented here: https://www.notion.so/trymito/Feature-Tests-2748271f28234549a0a2d98eced8a0e9?pvs=4). 

# Testing

To test, first `pip install streamlit`
```
npm run watch:streamlit
streamlit run mitosheet/streamlit/streamlit.py
```

# Documentation


mitosheet/mitosheet/streamlit/streamlit.py
```diff
@@ -0,0 +1,156 @@
...
+# function "_component_func", with an underscore prefix, because we don't want
+# to expose it directly to users. Instead, we will create a custom wrapper
+# function, below, that will serve as our component's public API.
+
+# It's worth noting that this call to `declare_component` is the
+# *only thing* you need to do to create the binding between Streamlit and
+# your component frontend. Everything else we do in this file is simply a
+# best practice.
+
+if not _RELEASE:
```
naterush, line 25: Remove this

mitosheet/mitosheet/streamlit/streamlit.py
```diff
@@ -0,0 +1,156 @@
...
+
+    mito_backend, responses = get_mito_backend(args, key)
+    sheet_data_json = mito_backend.steps_manager.sheet_data_json,
+    analysis_data_json = mito_backend.steps_manager.analysis_data_json,
+    user_profile_json = mito_backend.get_user_profile_json()
+
+    msg = message_passer_component(key=key + 'message_passer')
+    print(msg)
+    if msg is not None:
+        print("\n\n\n MESSAGE")
```
naterush, line 127: Remove

mitosheet/mitosheet/streamlit/streamlit.py
```diff
@@ -0,0 +1,156 @@
...
+    # There's no need to do this in our simple example - but it's an option.
+    return component_value
+
+
+# Create a wrapper function for the component. This is an optional
+# best practice - we could simply expose the component function returned by
+# `declare_component` and call it done. The wrapper allows us to customize
+# our component's API: we can pre-process its input args, post-process its
+# output value, and add a docstring for users.
+def mito_component(*args, key=None):
```
naterush, line 100: Move this into it's own file, called `mito_component.py`

mitosheet/mitosheet/user/location.py
```diff
@@ -118,4 +118,12 @@ def is_jupyterlite() -> bool:
     except ImportError:
         pass
 
-    return False
\ No newline at end of file
+    return False
+
+def is_streamlit() -> bool:
+    try:
+        # TODO: make this more robust
```
naterush, line 125: Do this

naterush (commented):
Some cleanup, to get this ready for review

mitosheet/src/streamlit/MitoMessagePasser.tsx
```diff
@@ -0,0 +1,52 @@
...
+ *   function.
+ * - When the setComponentValue function is called, the Python script re-runs from the top.
+ * - This means that when we render the component in the Python script, we have not yet 
+ *   received the data from the Mito iframe, as the component value is the return value
+ *   of the Python component function. 
+ * - To get around this, we use this component to listen for messages from the Mito iframe
+ *   and then pass them to the Python script using the setComponentValue function.
+ * - As this component is always rendered before the Mito iframe, we can be sure that
+ *   the Python script will have received the message from the Mito iframe before it 
+ *   renders the Mito iframe.
```
aarondr77, line 24: So my understanding is this: 

1. Every time the user makes an edit to their data (or even something like loading the summary statistics) we send that message to the backend through the setComponentValue which is the only way that the Mito streamlit component can communicate with the backend. But, every time the setComponentValue function is called, it triggers the entire streamlit app to re-render from top to bottom. Which means that the Mito streamlit component will re-render before it gets back the response from the backend. 
2. To solve this issue, we created another component, the Mito Message Passer. When the Mito component needs to communicate with the backend, instead of directly calling the setComponentValue function, we tell the Mito Message Passer to call the setComponentValue instead. 

After that I'm a little bit confused. Primarily because: 
1. If the problem is that anytime the setComponentValue function is called, the entire streamlit app rerenders, then why does this approach fix that? We are still calling the setComponentValue function.

aarondr77: Okay, I did some experimentation in a template streamlit component and I think this is my understanding to the above. 

When we call the setComponentValue function, the streamlit app starts pre-rendering right away, but doesn't finish rerendering until the component is finished updating. 

So I assume that what we're doing is that the MitoMessagePasser's value is actually the response of each message we pass. That way, because the MitoMessagePasser is rendered before the Mito Streamlit Component, we are sure to get the message back. 

The reason that this works is because we're actually setting the value of the component to the result, instead of just trying to get a response. 

I feel like that is on the right track, but not the full answer.

aarondr77: With that understanding, it seems like another solution (not that we should do it, but just for my understanding) would be that instead of having API.tsx calls return responses, we send all of the responses through the components props, just like we do with SheetData, for example. 

That way, the component won't finish rerendering until it gets the data back. 

I know that there are other issues, like debouncing, and the fact that we don't actually want to rerender the entire app each time is possible (or at least make that configurable).

naterush: When you call setComponentValue, the value appears in the Python script as the return value of the component function. The component function is also what takes the props. So we need the response before we call the component function, but we only get the message _after_ we call the component function (it's the return value, as I just said). 

The MitoMessagePasser lives before the Mito Component, and since we pass messages through this, it allows us to get messages before rending the Mito Component, and then pass the responses back with props.

mitosheet/mitosheet/api/get_defined_df_names.py
```diff
@@ -40,6 +41,35 @@ def __exit__(self, *args):
...
+            if isinstance(value, pd.DataFrame):
+                dfs.append(name)
+
+    return dfs
+
+
 
 def get_defined_df_names(params: Dict[str, Any], steps_manager: StepsManagerType) -> List[str]:
-    return get_df_names()
+    print("HERE")
```
naterush, line 72: Remove

mitosheet/mitosheet/streamlit/mito_component.py
```diff
@@ -27,6 +35,18 @@ def send(response):
...
         mito_backend.mito_send = send
 
+        if df_names is not None:
+            mito_backend.receive_message(
+                {
+                    'event': 'update_event',
+                    'id': get_new_id(),
+                    'type': 'args_update',
+                    'params': {
+                        'args': args
```
naterush, line 45: df_names

mitosheet/src/mito/components/import/DataframeImportScreen.tsx
```diff
@@ -50,6 +50,7 @@ const DataframeImportScreen = (props: DataframeImportTaskpaneProps): JSX.Element
         [],
         async () => {
             const response = await props.mitoAPI.getDefinedDfNames()
+            console.log("RESPONSE", response)
```
naterush, line 53: Remove

mitosheet/src/streamlit/MitoMessagePasser.tsx
```diff
@@ -25,7 +27,38 @@ import React, { ReactNode } from "react"
...
  * It's a bit of a hack, but it uses no undocumented APIs, and it works!
  */
-class MitoMessagePasser extends StreamlitComponentBase {
+class MitoMessagePasser extends StreamlitComponentBase<{messageQueue: any[], isSending: boolean}> {
+    timer: null | NodeJS.Timeout;
+    
+    constructor(props: any) {
+        super(props);
+        this.state = {
+          messageQueue: [],
```
naterush, line 36: TODO: explain all this

mitosheet/src/streamlit/MitoStreamlitWrapper.tsx
```diff
@@ -14,7 +14,8 @@ interface State {
 // Max delay is the longest we'll wait for the API to return a value
 // There is no real reason for these to expire, so we set it very high
 // at 5 minutes
-const MAX_DELAY = 5 * 60_000;
+//const MAX_DELAY = 5 * 60_000;
+const MAX_DELAY = 10_000;
```
naterush, line 18: Move back

aarondr77:
The `&apos;` in the text on this taskpane is displayed, instead of rendering as `'`

Screenshot 2023-07-18 at 9 29 21 AM

aarondr77:
Not such a bad thing, but the delimiter section here used to be much wider 

Screenshot 2023-07-18 at 9 35 37 AM

aarondr77:
Summary Stats are black and is wider than expected

Screenshot 2023-07-18 at 9 38 25 AM

aarondr77:
On Chrome and Safari when you exit full screen mode by pressing escape, the Mito sheet disappears from the app until you reload the page

https://github.com/mito-ds/monorepo/assets/18709905/cde7dd2e-9a92-4a6a-91da-53517cd3bd1c

aarondr77:
This dropdown is cut off. I think we assume that our dropdowns can expand beyond the boundary of the sheet. But here it cannot. 

Screenshot 2023-07-18 at 10 47 06 AM

aarondr77:
Toggling off the range slider also removes the y axis labels

https://github.com/mito-ds/monorepo/assets/18709905/bf03614d-75f2-4e7b-88dd-c98ffe243039

aarondr77:
`Write code to notebook` doesn't do anything in the code snippets, but it says it does. We should probably hide/disable that option?

https://github.com/mito-ds/monorepo/assets/18709905/7645a601-6c7a-4b83-852c-ed13c0c73fe4

mitosheet/mitosheet/streamlit/mito_component.py
```diff
@@ -0,0 +1,123 @@
...
+        This component simply passes messages from the frontend to the backend,
+        so that the backend can process them before it is rendered.
+        """
+        component_value = _message_passer_component_func(key=key)
+        return component_value
+
+
+    def mito_component( # type: ignore
+            *args: Union[pd.DataFrame, str, None], 
+            importers: Optional[List[Callable]]=None,
```
aarondr77, line 65: If we have `importers` we should have `sheet_functions` too

mitosheet/mitosheet/streamlit/mito_component.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        component_value = _message_passer_component_func(key=key)
+        return component_value
+
+
+    def mito_component( # type: ignore
+            *args: Union[pd.DataFrame, str, None], 
+            importers: Optional[List[Callable]]=None, 
+            df_names: Optional[List[str]]=None,
+            key=None
```
aarondr77, line 67: I briefly tested that if I have two different keys than I can have two mitosheets in the same dashboard and it was working

mitosheet/mitosheet/streamlit/streamlit.py
```diff
@@ -0,0 +1,20 @@
+
+import streamlit as st
+from mitosheet.streamlit.mito_component import mito_component
+
+st.set_page_config(layout="wide")
```
aarondr77, line 5: I should change this and see if it works

aarondr77:
Narrow mode: Send arrow is cut off

Screenshot 2023-07-18 at 11 05 01 AM

aarondr77:
Narrow mode: Excel import, a lot of stuff is cutoff here

Screenshot 2023-07-18 at 11 06 19 AM

aarondr77:
Narrow mode: Mito Enterprise button is cutoff

Screenshot 2023-07-18 at 11 08 02 AM

aarondr77:
Narrow mode: x is cutoff

Screenshot 2023-07-18 at 11 09 25 AM

aarondr77:
When I go into full screen mode, it doesn't change from narrow taskapen to wide taskpane

aarondr77:
`Current Value` is cut off

Screenshot 2023-07-18 at 11 17 19 AM

aarondr77:
Going to stop looking for narrow taskpane issues now and will come back to it after we fix these up!

mitosheet/mitosheet/user/location.py
```diff
@@ -118,4 +118,13 @@ def is_jupyterlite() -> bool:
...
 
-    return False
\ No newline at end of file
+    return False
+
+def is_streamlit() -> bool:
+    try:
+        # TODO: have to handle versions of streamlit that don't have this
+        # or it's in different places: https://discuss.streamlit.io/t/how-to-check-if-code-is-run-inside-streamlit-and-not-e-g-ipython/23439/8
+        from streamlit.runtime.scriptrunner import get_script_run_ctx
```
aarondr77, line 127: We should create out a TODO list testing release procedure for different cases like this. And we should also make sure that we have good logging so we can identify that there are errors and can fix them

aarondr77:
The font is quite different in streamlit vs jupyter

Streamlit:
Screenshot 2023-07-18 at 11 22 12 AM

Jupyter: 
Screenshot 2023-07-18 at 11 22 43 AM

mitosheet/src/mito/components/icons/SuggestedStyle3Icon.tsx
```diff
@@ -5,7 +5,7 @@ import React from 'react';
 
 const SuggestedStyle3Icon = (): JSX.Element => {
     return (
-        <svg width="58" height="46" viewBox="0 0 58 46" fill="none" xmlns="http://www.w3.org/2000/svg">
+        <svg width="50" height="46" viewBox="0 0 58 46" fill="none" xmlns="http://www.w3.org/2000/svg">
```
aarondr77, line 8: I'm fine leaving this, but just calling out that this looks worse in Jupyter as a result. 

Now that we're supporting different size task panes, we should make sure to think about that in our Figma designs! 

Screenshot 2023-07-18 at 11 25 50 AM

aarondr77:
In Jupyter, these dropdowns have way more vertical space between the items then they used to

Screenshot 2023-07-18 at 11 29 36 AM

mitosheet/src/mito/components/taskpanes/UserDefinedImport/UserDefinedImportTaskpane.tsx
```diff
@@ -71,7 +71,13 @@ const getParamTypeDisplay = (
     }
 }
 
-const NO_IMPORTERS_MESSAGE = 'You have not defined any importers. To define importers, pass them to the mitosheet.sheet call with the `importers` parameter. An importer is just a function that returns a pandas dataframe.';
+let NO_IMPORTERS_MESSAGE = 'You have not defined any importers. An importer is just a function that returns a pandas dataframe.';
+if (isInStreamlit()) {
+    NO_IMPORTERS_MESSAGE += ' You can define importers in the mito_component call with the `importers` parameter.';
```
aarondr77, line 76: We could add a similar check / message in the update files taskpane where we say

Screenshot 2023-07-18 at 11 33 06 AM

mitosheet/src/streamlit/MitoMessagePasser.tsx
```diff
@@ -0,0 +1,121 @@
...
+            clearTimeout(this.processMessageQueueTimer);
+        }
+    }
+    
+    processQueue = () => {
+        if (this.state.messageQueue.length > 0) {
+            // Send one message
+            const message = this.state.messageQueue[0];
+            Streamlit.setComponentValue(message);
+            console.log("JUST SENT", message, Date.now())
```
aarondr77, line 70: remove this?

aarondr77 (commented):
Nice the functionality is working really well in Streamlit and it looks awesome! A few things here to fix up, mostly around the narrow taskpane. 

Ping me, and I'll do another polish review!

mitosheet/mitosheet/user/location.py
naterush: If we have tests for streamlit (which we will), then we will catch this. I wonder what version of streamlit we should go back to.

naterush:
Ok, on the graphing cutoff stuff, it seems like the entire content div is just shifted over? I am very unsure of why this would be, and am having trouble replicating or resolving it. 

Maybe tracking down colors for graphs will illuminate things.

aarondr77:
The selected sheet should title should be white

Screenshot 2023-07-22 at 10 07 32 AM

aarondr77:
Mito Enterprise text should be white

Screenshot 2023-07-22 at 10 08 58 AM

aarondr77:
The colors in the sheet used to be: 
- Column Header: Blue
- Data: striped gray and white

Now it is:
- Column Header: Blue
- Data: Striped **blue** and white

I think we should def move back to the gray and white stripes!


Screenshot 2023-07-22 at 10 11 20 AM

Screenshot 2023-07-22 at 10 10 31 AM

Screenshot 2023-07-22 at 10 10 57 AM

aarondr77:
Probably same as the other issue, so will stop posting after this, but the text on top of the dark blue should be white: 

Screenshot 2023-07-22 at 10 16 37 AM

aarondr77:
The disabled text button should look like this:

Screenshot 2023-07-22 at 10 18 21 AM


But in streamlit, it looks like this: 

Screenshot 2023-07-22 at 10 17 40 AM

aarondr77:
The white arrows in the select are missing:

Screenshot 2023-07-22 at 10 19 51 AM

aarondr77:
I don't think "Group date by" text is supposed to be bold

Screenshot 2023-07-22 at 10 22 44 AM

aarondr77:
The border on this modal did not used to be so prominent

Screenshot 2023-07-22 at 10 24 04 AM

aarondr77:
The `dtype` text used to be + should be darker

Screenshot 2023-07-22 at 10 24 38 AM

aarondr77:
Filter icons should be purple when filtered

Screenshot 2023-07-22 at 10 25 23 AM

aarondr77:
Taskapnes like `merge` and `concat` which only make sense when having multiple sheets in the sheet used to open to a taskpane that says "You need to have multiple sheets to use this". Now, the buttons just don't do anything. 

We should go back to our previous way of handling it, or if we cannot, then we should disable the buttons with a disabled tooltip

https://github.com/mito-ds/monorepo/assets/18709905/634817ee-7fa2-426d-982f-40bd2dac40ed

aarondr77:
Maybe not new, but the text button should never shrink vertically like this!

https://github.com/mito-ds/monorepo/assets/18709905/69637040-25a0-4617-b124-0b858f780e15

aarondr77:
I think these dropdowns used to / should be taller so we can see more! 

Screenshot 2023-07-22 at 10 30 04 AM

aarondr77:
When you open the format colors taskpane, it shows the colors as all black. 

https://github.com/mito-ds/monorepo/assets/18709905/01b1a0fa-548e-46eb-b71b-f0369aa44ef9

aarondr77:
If you set the text color, it does not update! 

(same video as above)
https://github.com/mito-ds/monorepo/assets/18709905/2bba2a5b-cc2f-458f-ada2-fa3c5f85d7ef

aarondr77:
This should be fixed in a previous comment, but noting that the color presets do not match the sheet anymore because we changed the stripe color

Screenshot 2023-07-22 at 10 35 03 AM

aarondr77:
I know graph color is already on your radar, but just marking here again. Having multiple series that are all black without being able to change it is pretty bad. 


Screenshot 2023-07-22 at 10 37 15 AM

aarondr77:
The OFF toggle text is too white

Screenshot 2023-07-22 at 10 38 23 AM

aarondr77:
In Streamlit, the `See All Functionality` looks like this: 

Screenshot 2023-07-22 at 10 41 38 AM

In Jupyter, the `See All Functionality` looks like this: 

Screenshot 2023-07-22 at 10 42 12 AM

The Jupyter version is much better.

aarondr77:
In Jupyter + streamlit, the formal bar input has a gray background. 

Screenshot 2023-07-22 at 10 44 39 AM

aarondr77:
The disabled select should not have a purple arrow


Screenshot 2023-07-22 at 10 45 42 AM

aarondr77:
The disabled input's text should be lighter 

Screenshot 2023-07-22 at 10 46 16 AM

aarondr77:
This text should be white

Screenshot 2023-07-22 at 10 48 59 AM

aarondr77:
The cursor on Jupyter changes correctly to show that it's clickable. The cursor on Streamlit does not -- its just text selection


https://github.com/mito-ds/monorepo/assets/18709905/ab0fae4c-8eb4-4a3c-888a-b23265e21e80

aarondr77:
This Add Filter button looks disabled, but it shouldn't! 

Screenshot 2023-07-22 at 10 55 13 AM

aarondr77:
Narrow Taskpane: 

These buttons are slightly overlapping

Screenshot 2023-07-22 at 10 57 34 AM

aarondr77:
Narrow Taskpane: 

Screenshot 2023-07-22 at 10 58 07 AM

aarondr77:
The `Where` is cutoff

Screenshot 2023-07-22 at 10 58 58 AM

aarondr77:
The text here should look disabled when I open it. 

https://github.com/mito-ds/monorepo/assets/18709905/f0257d32-9b6e-4fe4-a393-f5f4010147ea

aarondr77:
On Streamlit only  

1. Click the More decimals button once. 
2. See it finished
3. Click it again
4. See it finished

Now, 

1. Click the More decimal button twice in a row to add two decimals
2. See that the first one finishes
3. See that the second one never finishes


https://github.com/mito-ds/monorepo/assets/18709905/0d6498e7-342d-4917-848f-f3be5f35f2fb

aarondr77:
If I delete a bunch of rows in a row, then it will delete all of them except hang on deleting the last one 

https://github.com/mito-ds/monorepo/assets/18709905/92abaaf3-ce1c-441b-9028-830f924236e2

aarondr77:
Can we vertically align the text to the input?

Screenshot 2023-07-22 at 11 17 30 AM

aarondr77:
Similar to above, sometimes when I refresh the browser, I get two updating user profile events and then the second one just sits there forever. I can still use the app, but the loading indicator never goes away

Screenshot 2023-07-22 at 11 18 02 AM

aarondr77:
You can turn on wide mode as the user of a streamlit app. A really nice polish could be to detect apps in narrow mode and then have a hint that says "This is much better in wide mode. Turn it on here."

https://github.com/mito-ds/monorepo/assets/18709905/fdc3d356-ec28-489d-862d-b09f1f1eda55

aarondr77:
The Pencil icon that we default to needs a white version

Screenshot 2023-07-22 at 11 31 22 AM

aarondr77:
The color themes don't match with what we end up displaying because the stripe color is different

Screenshot 2023-07-22 at 11 32 18 AM

aarondr77:
Modal text should be white

Screenshot 2023-07-22 at 11 33 26 AM

aarondr77:
Should we make dark mode dropdowns?

Screenshot 2023-07-22 at 11 35 10 AM

aarondr77:
A bunch of formatting stuff here too

Screenshot 2023-07-22 at 11 54 00 AM

aarondr77:
Its strange that the selects change size depending on the content 

https://github.com/mito-ds/monorepo/assets/18709905/5db4e08d-4635-4db7-b082-be9e3cf31e7c

aarondr77:
The signup modal overflows on narrow

Screenshot 2023-07-22 at 12 00 05 PM

aarondr77:
The `Back` button coloring kinda looks disabled.. 


Screenshot 2023-07-22 at 12 01 08 PM

aarondr77:
Maybe the blue should be red?

Screenshot 2023-07-22 at 12 01 56 PM

aarondr77:
Narrow taskapne `Access Code` text is cut off

Screenshot 2023-07-22 at 12 02 19 PM

aarondr77:
This text is cutoff on narrow

Screenshot 2023-07-22 at 12 03 24 PM

aarondr77:
The font being larger than the FAQ titles is not great

Screenshot 2023-07-22 at 12 03 43 PM

aarondr77:
The loading circle is cut off on narrow

Screenshot 2023-07-22 at 12 04 16 PM

aarondr77:
Yellow modified color really obscures the data

Screenshot 2023-07-22 at 12 04 45 PM

naterush:
# Messages Being Lost

I have confirmed that all of the loading indicator stuff (where messages appear to never resolve) is in-fact because messages are being lost. 

This is a result of streamlit deboucing setComponentValue updates. I can't figure out _how_ they do this. We're going to leave this in for the beta, and move forward without fixing it for now. 

This week, it will be one of the big things we resolve. I reacted to all of these above with a celebration emoji.

naterush:
# Set Dataframe Colors

This feature is currently broken on streamlit. Leaving for now, struggling to fix it up.

PR #832, opened 2023-07-13
mitosheet: helpful error with single equals

aarondr77:
# Description

1. Write the formula `=A=B`
2. The mitosheet sets the new column =B, but it also sets the column A equal to B
3. In addition, the code that it generates is invalid. 

https://github.com/mito-ds/monorepo/assets/18709905/cafd0e1a-9b42-4122-824c-5c9412c00320

We now throw a helpful error unless the formula has a `=` that is:
1. At the front of the formula
2. Part of multiple = sign
3. In a string 
4. Part of a column header
5. !=, >=, <=

An enterprise customer had multiple users run into this bug during a training

# Testing

Run the tests and try it out

# Documentation

No


mitosheet/mitosheet/step_performers/column_steps/set_column_formula.py
```diff
@@ -290,6 +290,8 @@ def exec_column_formula(
         
         # Exec the code, where the df is the original dataframe
         # See explination here: https://www.tutorialspoint.com/exec-in-python
+
+        print('python_code: ', python_code)
```
marthacryan, line 294: Is this a debugging thing?

mitosheet/mitosheet/parser.py
```diff
@@ -109,6 +109,38 @@ def safe_contains(
...
+        '!=',
+        '>=',
+        '<=',
+    ]
+
+    # Remove all occurences of string_to_ignore from the formula
+    for string_to_ignore in strings_to_ignore:
+        formula = formula.replace(string_to_ignore, '')
+
+    pattern = r"(^|[^'\"=])\=(?![\=])([^'\"=]|$)"
```
marthacryan, line 141: Maybe you could add a little explanation of this regex? I'm not great with regexes but is it possible to add in some of the logic above about ignoring strings into this regex if we're already making a regex?

marthacryan:
Looks great! I just had one question about the regex stuff that might be my own lack of understanding of regexes. It's probably worth leaving a comment to explain it either way though

mitosheet/mitosheet/parser.py
aarondr77: Good call! I'll add some documentation here

marthacryan (approved):
LGTM!

PR #840, opened 2023-07-19
Add support for exporting header color to xlsx

marthacryan:
Adds the ability to export the background color and the font color of the header row when exporting to excel. 

Todo:
- [x] Prevent setting the header for columns that are empty

Before and after the PR:
image

### DIFF
--- a/mitosheet/mitosheet/api/get_dataframe_as_excel.py
+++ b/mitosheet/mitosheet/api/get_dataframe_as_excel.py
@@ -6,6 +6,7 @@
 import base64
 import io
 from typing import Any, Dict
+from openpyxl import Workbook
 
 import pandas as pd
 from mitosheet.excel_utils import get_df_name_as_valid_sheet_name
@@ -24,23 +25,21 @@ def get_dataframe_as_excel(params: Dict[str, Any], steps_manager: StepsManagerTy
     # We write to a buffer so that we don't have to save the file
     # to the file system for no reason
     buffer = io.BytesIO()
-    with pd.ExcelWriter(buffer) as writer:
+    with pd.ExcelWriter(buffer, engine="openpyxl") as writer:
+        workbook = writer.book
         for sheet_index in sheet_indexes:
-            # First, write the tab in the Excel file
+            # Get the dataframe and sheet name
             df = steps_manager.dfs[sheet_index]
             df_name = steps_manager.curr_step.df_names[sheet_index]
+            sheet_name = get_df_name_as_valid_sheet_name(df_name)
 
-            # We reset the default style so that we can apply our own custom style
-            pd.io.formats.excel.ExcelFormatter.header_style = None
-            df.to_excel(
-                writer, 
-                sheet_name=get_df_name_as_valid_sheet_name(df_name),
-                index=False
-            )
+            # Write the dataframe to the sheet
+            df.to_excel(writer, sheet_name, index=False)
 
-            # Adding formatting is a pro feature
+            # Add formatting to the sheet for pro users
+            format = steps_manager.curr_step.df_formats[sheet_index]
             if is_pro_user:
-                add_formatting_to_excel_sheet(writer, steps_manager, sheet_index)
+                add_formatting_to_excel_sheet(workbook, sheet_name, format)
     
     # Go back to the start of the buffer
     buffer.seek(0)
--- a/mitosheet/mitosheet/api/get_dataframe_as_excel.py
+++ b/mitosheet/mitosheet/api/get_dataframe_as_excel.py
@@ -25,7 +25,6 @@ def get_dataframe_as_excel(params: Dict[str, Any], steps_manager: StepsManagerTy
     # to the file system for no reason
     buffer = io.BytesIO()
     with pd.ExcelWriter(buffer, engine="openpyxl") as writer:
-        workbook = writer.book
         for sheet_index in sheet_indexes:
             # Get the dataframe and sheet name
             df = steps_manager.dfs[sheet_index]
@@ -37,8 +36,8 @@ def get_dataframe_as_excel(params: Dict[str, Any], steps_manager: StepsManagerTy
 
             # Add formatting to the sheet for pro users
             format = steps_manager.curr_step.df_formats[sheet_index]
-            if is_pro_user:
-                add_formatting_to_excel_sheet({ workbook, sheet_name, format })
+            if is_pro_user and format['headers']: 
+                add_formatting_to_excel_sheet(writer, sheet_name, format['headers']['backgroundColor'], format['headers']['color'])
     
     # Go back to the start of the buffer
     buffer.seek(0)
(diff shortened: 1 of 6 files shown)


mitosheet/mitosheet/pro/download/formatting.py
```diff
@@ -18,24 +18,21 @@ def add_formatting_to_excel_sheet(
...
-        format_data = format_data_map[column_id]
-        excel_column = get_excel_range_from_column_index(column_index)
-
-        if (is_number_dtype(column_dtype)):
-            format = workbook.add_format({'num_format': 'General'})
-            worksheet.set_column(excel_column, None, format)
\ No newline at end of file
+    # Create the formatting object for styling headers
+    headerFormat = workbook.add_format({ "border": 1, "bold": True })
+    if format.get('headers').get('color') is not None:
```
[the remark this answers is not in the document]
naterush: Another option is to add a default value of an empty dictionary to the .get call. This is probably less clear though.

naterush (changes requested):
Good work getting this up! The code is well architected and clear. I caught a few lurking bugs I expand on below. 

Once we fix them up, can we add some tests in the `mitosheet/tests/api` folder? I think we can just make sure that:
1. If you try and download an Excel file without formatting, you will get a non-empty response 
2. If you try and download an Excel file with formatting, you will get a non-empty response

These two tests would have likely caught the bugs I found below, and will protect us from any similar bugs in the future.

aarondr77:
Hey @marthacryan can you check out our CLA here: https://github.com/mito-ds/monorepo/blob/dev/CLA.md

If you agree, just hit me back with "I have read the CLA Document and I hereby sign the CLA"

marthacryan:
@aarondr77 I have read the CLA Document and I hereby sign the CLA

mitosheet/mitosheet/public/v3/formatting.py
```diff
@@ -0,0 +1,36 @@
...
+    format = params['format']
+    sheet = workbook.get_sheet_by_name(sheet_name)
+    
+    # Add formatting to the header row    
+    header_name = f"{sheet_name}_Header"
+    if format.get('headers'):
+        header_format = NamedStyle(name=header_name)
+        # Add font and background colors to the header format
+        if format['headers']['color']:
+            font_color = format['headers']['color'][1:]
```
naterush, line 25: Can you leave a comment explaining the 1: -- assuming this is chopping a # of the hex, but it would be great to confirm.

mitosheet/mitosheet/public/v3/formatting.py
```diff
@@ -0,0 +1,36 @@
+from openpyxl import Workbook
+from openpyxl.worksheet.worksheet import Worksheet
+
+from openpyxl.styles import Font, PatternFill
+from openpyxl.styles import NamedStyle
+
+def add_formatting_to_excel_sheet( params: dict[Workbook, str, dict]) -> None:
```
naterush, line 7: Let's make the parameters explicitly named variables to the function. A good format sounds like:

```
def add_formatting_to_excel_sheet(
     writer: pd.ExcelWriter, 
     sheet_name: str,
     header_background_color: Optional[str]=None,
     header_color: Optional[str]=None,
):
```
A few notes on the params: 
1. It's taking a writer, rather than the book (I think this is a clearer interface - and what one expects).
2. It takes all the specific formatting variables as explicitly arguments. 

The approach you took is very react (e.g. everything is wrapper in props), and results in something that is easy to call with our current params object. If we move to this new version I propose, it's a bit harder to call, but we get the benefits of a strongly typed function and a _very clear_ public interface. Having a very clear public interface is important as this is exposed as a public function to users!

We'll also have to do a bit more work to transpile the function params - but it's not too bad. You can see utilities for doing that in `excel_range_import_code_chunk.py` file -- peep the `param_dict_to_code` helper.

naterush (changes requested):
I propose a refactor to the type of the formatting function that I think will be clearer and safer going forward. Let me know what you think!

Could you also some tests for this code path? I think the obvious one is just a file can be gotten from the API if it has formatting applied (this test might already exist!). In future PRs -- try and add an automated test if you can -- we generally try and write tests for all backend features!

naterush:
The code isn't running for me -- there's some error occurring that's getting swallowed by our error handling in the API. Wanna add a test and see if it catches the error?

mitosheet/mitosheet/public/v3/formatting.py
marthacryan: Done!

mitosheet/mitosheet/tests/api/get_dataframe_as_excel.py
```diff
@@ -0,0 +1,72 @@
...
+    # Check that the excel string is not empty
+    assert excel_string != ''
+
+# This tests exporting as excel without formatting
+def test_export_to_excel_sheet_no_formatting():
+    # Create a mito wrapper with data
+    test_wrapper = create_mito_wrapper_with_data(['abc'])
+    test_wrapper.add_column(0, 'B')
+    test_wrapper.add_column(0, 'C')
+    test_wrapper.set_dataframe_format(0, {
```
naterush, line 42: I think this test should just entirely avoid setting dataframe format. So you can just delete this line of code. That way, it tests that defaults work and don't result int errors.

mitosheet/mitosheet/api/get_dataframe_as_excel.py
```diff
@@ -24,20 +24,20 @@ def get_dataframe_as_excel(params: Dict[str, Any], steps_manager: StepsManagerTy
...
 
-            # Adding formatting is a pro feature
-            if is_pro_user:
-                add_formatting_to_excel_sheet(writer, steps_manager, sheet_index)
+            # Write the dataframe to the sheet
+            df.to_excel(writer, sheet_name, index=False)
+
+            # Add formatting to the sheet for pro users
+            format = steps_manager.curr_step.df_formats[sheet_index]
+            if is_pro_user and format['headers']:
```
naterush, line 39: This if returns false if you're not Pro, so the tests aren't actually testing this code path. 

We can fix this by adding a line that says `if (is_pro_user or is_running_test()) and ...`

naterush (changes requested):
The functionality is working perfectly -- great work! 

Tests are almost perfect - I think there's two lines of fix up. Feel free to fix them (and verify that the right code branch is running in the test) -- and then once you're confident they are good, feel free to merge it!

mitosheet/mitosheet/tests/api/get_dataframe_as_excel.py
marthacryan: Ah totally, good idea

mitosheet/mitosheet/api/get_dataframe_as_excel.py
marthacryan: Ah true I added that but it must have gotten lost between git stashes!

PR #841, opened 2023-07-20
Add header formatting to xlsx export code chunk

marthacryan:
Adds header font color and background color into the code chunk. Uses the new public api from #840.


naterush (changes requested):
Nice work getting this up! I'll pause on reviewing this for now, as it will likely need to factor in the changes from the other PR in accessing the API and the issues with no. 

Tests for this code as well seems legit too :)

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -10,6 +10,33 @@
 from mitosheet.state import State
 from mitosheet.transpiler.transpile_utils import TAB, column_header_to_transpiled_code
 
+# This is a helper function that generates the code for formatting the excel sheet
+def get_format_code(state: State) -> list:
+    code = []
+    formats = state.df_formats
+    for sheetIndex in range(len(formats)):
+        sheet_name = state.df_names[sheetIndex]
+        format = formats[sheetIndex]
```
naterush, line 19: for sheet_name, format in zip(state.df_names, formats):

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -10,6 +10,33 @@
...
+        # If there is no formatting, we skip trying to access the colors
+        if format.get('headers') is None:
+            continue
+        
+        # Prepares the header font color and background color to be used in the API call
+        # So either they should be surrounded by quotes, or they should be None
+        header_font_color = None
+        header_background_color = None
+        if (format['headers'].get('color') is not None):
+            header_font_color = f'"{format["headers"]["color"]}"'
```
naterush, line 30: We have a really badly named utility called `column_header_to_transpiled_code` that you can call on these values and it will put them in a string. So it would be like:

```
header_font_color = column_header_to_transpiled_code(format["headers"]["color"])
```

mitosheet/mitosheet/tests/test_transpile.py
```diff
@@ -702,16 +703,39 @@ def test_transpiled_with_export_to_xlsx_single():
...
+    with pd.ExcelWriter(path, engine="openpyxl") as writer:
+        df.to_excel(writer, sheet_name="df", index=False)
     
     return df
 
 path = r"te'st.xlsx"
 
 df = function(df, path)"""
 
+def test_transpiled_with_export_to_xlsx_format():
```
naterush, line 718: I think you can move this to the `test_export_to_file.py` file. We usually just test code chunks with their corresponding step.

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -10,6 +10,33 @@
...
+        if (format['headers'].get('backgroundColor') is not None):
+            header_background_color = f'"{format["headers"]["backgroundColor"]}"'
+        
+        # If both are None, we skip this sheet
+        if header_font_color is None and header_background_color is None:
+            continue
+        code.append(f'{TAB}add_formatting_to_excel_sheet(writer, "{sheet_name}", {header_background_color}, {header_font_color})')
```
naterush, line 37: I'd love if this code transpiled as:
```
add_formatting_to_excel_sheet(writer, "sheet name", header_background_color="00000", header_font_color="XXXXXX")
```

That way, users are able to tell which color is which when they are editing the code. I'm also just thinking that it will:
1. Make it much easier to read as we add more formatting options. 
2. Allow us to reorder the parameters if we want to do so in the function, without breaking users code. 

Let me know what what you think! This is where the `param_dict_to_code` utility could be useful -- it's in `transpile_utils.py` -- check it out!

naterush (changes requested):
Overall, really clean approach. I love how simple the code ends up being when we expose this in a public interface. I have a few minor refactoring ideas that I think could be cool to invest in. 

A few more tests I think it would be awesome to add:
1. Adding formatting to 2 or more sheets
2. Adding formatting to one sheet, but not to the other
3. No formatting at all, still get valid code

Ping me whenever for a re-review :)

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
marthacryan: Oh cool! I'll add that

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
marthacryan: Oh yeah that makes sense! I'm a little confused about how using `param_dict_to_code` would work here - does the code I just pushed look ok?

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -10,6 +10,30 @@
...
+        header_background_color = None
+        if (format['headers'].get('color') is not None):
+            header_font_color = column_header_to_transpiled_code(format["headers"]["color"])
+        if (format['headers'].get('backgroundColor') is not None):
+            header_background_color = column_header_to_transpiled_code(format["headers"]["backgroundColor"])
+        
+        # If both are None, we skip this sheet
+        if header_font_color is None and header_background_color is None:
+            continue
+        code.append(f'{TAB}add_formatting_to_excel_sheet(writer, "{sheet_name}", header_background_color={header_background_color}, header_font_color={header_font_color})')
```
naterush, line 34: Ok so one thing the `param_dict_to_code` could be useful for is not having to manually write all these out. 

```
>>> param_dict_to_code({'header_background_color': 'ffffff', 'header_font_color': 'xxxxx'}, as_single_line=True)
"header_background_color='ffffff', header_font_color='xxxxx'"
```

Let's move to this, I think. That way, we can just build a params dict and then pass it to this function without having to update variable names in the future.

naterush (changes requested):
Things are working great! One minor code refactor, and then LGTM once tests are passing (no need to request new review). 

Good work getting this finished!

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
marthacryan: Just pushed to use that! How does that look?

PR #846, opened 2023-07-26
Add row formatting to excel exports

marthacryan:
Follow-up PR to #840 and #841 to add row formatting to both excel exports that are "download now" and excel exports as code chunks. Adds 4 new arguments to the public API for `add_formatting_to_excel_sheet`: `even_background_color`, `even_font_color`, `odd_font_color`, `odd_background_color`.


mitosheet/mitosheet/api/get_dataframe_as_excel.py
```diff
@@ -40,7 +40,16 @@ def get_dataframe_as_excel(params: Dict[str, Any], steps_manager: StepsManagerTy
             # Add formatting to the sheet for pro users
             format = steps_manager.curr_step.df_formats[sheet_index]
             if allow_formatting and format['headers']:
```
naterush, line 42: It feels like this None check might need to be expanded to check the other things, no? Actually -- see the `.get('key', {})` trick below -- we could use it here and avoid all of the None checking here!

naterush: Might be good to add a test for header color but not for row color

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -18,16 +18,26 @@ def get_format_code(state: State) -> list:
...
-        # So either they should be surrounded by quotes, or they should be None
-        params = {}
-        if (format['headers'].get('backgroundColor') is not None):
-            params['header_background_color'] = format["headers"]["backgroundColor"]
-        if (format['headers'].get('color') is not None):
-            params['header_font_color'] = format["headers"]["color"]
+        if format.get('headers') is not None:
+            # Prepares the header font color and background color to be used in the API call
+            # So either they should be surrounded by quotes, or they should be None
+            params = {}
```
naterush, line 24: This probably all could be rewritten as the following:
```
params = {
     'header_background_color': format.get('headers', {}).get('backgroundColor')
     'header_font_color': format.get('headers', {}).get('color'),
     ....

}
```

The two tricks here:
1. Creating the param in one place, which is clean. 
2. Using default arguments to .get to define a new dict, so you can then continue and look up on it if it doesn't exist. 

I think this could cut like 10 or so lines out of this function.

naterush (commented):
Quick incremental review with some tips. Ping me whenever when tests are ready :)

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
marthacryan: Ahh that is the syntax I was looking for! That's awesome I'll definitely add it

mitosheet/mitosheet/api/get_dataframe_as_excel.py
marthacryan: Fixed! In the tests, I did include format objects w/ headers and not rows, as well as format objects with both

mitosheet/mitosheet/transpiler/transpile_utils.py
```diff
@@ -92,7 +92,7 @@ def column_header_map_to_string(column_header_map: Dict[ColumnHeader, ColumnHead
         return result
 
 
-def param_dict_to_code(param_dict: Dict[str, Any], level: int=0, as_single_line: bool=False) -> str:
+def param_dict_to_code(param_dict: Dict[str, Any], level: int=0, as_single_line: bool=False, tab_level: int=1) -> str:
```
naterush, line 95: Nice! Very useful.

mitosheet/mitosheet/tests/step_performers/test_export_to_file.py
```diff
@@ -14,6 +14,37 @@
 from mitosheet.tests.test_utils import check_dataframes_equal, create_mito_wrapper
 from mitosheet.tests.decorators import pandas_post_1_2_only, python_post_3_6_only
 
+DF_FORMAT_HEADER = {
```
naterush, line 17: Tests look good!

naterush (changes requested):
Code looks real clean. Functionality works great. Maybe one more test for row formatting but no header formatting? 

Then LGTM - no need for more review!

PR #847, opened 2023-07-27
Add greater than conditional format to excel export

marthacryan:
This adds the ability to export (through code chunks) conditional formats for the "greater than" condition. The generated code looks like this:
```
    add_formatting_to_excel_sheet(writer, "df", df, 
        conditional_formats=[
            {'columns': ['A'], 'filters': [{'condition': 'greater', 'value': 5}], 'font_color': '#e72323', 'background_color': '#ffffff'},
            {'columns': [], 'filters': [{'condition': 'not_empty', 'value': ''}], 'font_color': None, 'background_color': None}
        ]
    )
```

### DIFF
--- a/mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
+++ b/mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
@@ -12,11 +12,24 @@ from mitosheet.transpiler.transpile_utils import TAB, column_header_to_transpile
 
 from mitosheet.transpiler.transpile_utils import param_dict_to_code
 
+from mitosheet.excel_utils import get_column_from_column_index
+
 # This is a helper function that generates the code for formatting the excel sheet
 def get_format_code(state: State) -> list:
     code = []
     formats = state.df_formats
-    for sheet_name, format in zip(state.df_names, formats):
+    for sheet_index, (sheet_name, format) in enumerate(zip(state.df_names, formats)):
+        # We need to convert the column IDs to column letters for conditional formats to export to excel
+        conditional_formats = format.get('conditional_formats')
+        column_ids = state.column_ids.get_column_ids(sheet_index=sheet_index)
+        if conditional_formats is not None:
+            for conditional_format in conditional_formats:
+                # Create new object to store the columns in the excel format
+                conditional_format['columns'] = []
+                for column_id in conditional_format.get('columnIDs'):
+                    column = get_column_from_column_index(column_ids.index(column_id))
+                    conditional_format['columns'].append(column)
+
         # If there is no formatting, we skip trying to access the colors
         params = {
             'header_background_color': format.get('headers', {}).get('backgroundColor'),
@@ -25,6 +38,7 @@ def get_format_code(state: State) -> list:
             'even_font_color': format.get('rows', {}).get('even', {}).get('color'),
             'odd_background_color': format.get('rows', {}).get('odd', {}).get('backgroundColor'),
             'odd_font_color': format.get('rows', {}).get('odd', {}).get('color'),
+            'conditional_formats': conditional_formats
         }
         param_dict = {
             key: value for key, value in params.items()
--- a/mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
+++ b/mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
@@ -12,25 +12,21 @@ from mitosheet.transpiler.transpile_utils import TAB, column_header_to_transpile
 
 from mitosheet.transpiler.transpile_utils import param_dict_to_code
 
-from mitosheet.excel_utils import get_column_from_column_index
+from mitosheet.utils import add_columns_to_condition_formats
 
 # This is a helper function that generates the code for formatting the excel sheet
 def get_format_code(state: State) -> list:
     code = []
     formats = state.df_formats
     for sheet_index, (sheet_name, format) in enumerate(zip(state.df_names, formats)):
-        # We need to convert the column IDs to column letters for conditional formats to export to excel
-        conditional_formats = format.get('conditional_formats')
-        column_ids = state.column_ids.get_column_ids(sheet_index=sheet_index)
-        if conditional_formats is not None:
-            for conditional_format in conditional_formats:
-                # Create new object to store the columns in the excel format
-                conditional_format['columns'] = []
-                for column_id in conditional_format.get('columnIDs'):
-                    column = get_column_from_column_index(column_ids.index(column_id))
-                    conditional_format['columns'].append(column)
-
-        # If there is no formatting, we skip trying to access the colors
+        # We need to convert the column IDs to column letters
+        # for conditional formats to export to excel
+        conditional_formats = add_columns_to_condition_formats(
+            format.get('conditional_formats'),
+            df=state.dfs[sheet_index],
+            column_ids=state.column_ids,
+            sheet_index=sheet_index
+        )
         params = {
             'header_background_color': format.get('headers', {}).get('backgroundColor'),
             'header_font_color': format.get('headers', {}).get('color'),
--- a/mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
+++ b/mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
@@ -23,7 +23,7 @@ def get_format_code(state: State) -> list:
         # for conditional formats to export to excel
         conditional_formats = get_conditional_formats_objects_to_export_to_excel(
             format.get('conditional_formats'),
-            column_ids=state.column_ids,
+            column_id_map=state.column_ids,
             sheet_index=sheet_index
         )
         params = {
--- a/mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
+++ b/mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
@@ -12,7 +12,7 @@ from mitosheet.transpiler.transpile_utils import TAB, column_header_to_transpile
 
 from mitosheet.transpiler.transpile_utils import param_dict_to_code
 
-from mitosheet.utils import add_columns_to_condition_formats
+from mitosheet.utils import get_conditional_formats_objects_to_export_to_excel
 
 # This is a helper function that generates the code for formatting the excel sheet
 def get_format_code(state: State) -> list:
@@ -21,7 +21,7 @@ def get_format_code(state: State) -> list:
     for sheet_index, (sheet_name, format) in enumerate(zip(state.df_names, formats)):
         # We need to convert the column IDs to column letters
         # for conditional formats to export to excel
-        conditional_formats = add_columns_to_condition_formats(
+        conditional_formats = get_conditional_formats_objects_to_export_to_excel(
             format.get('conditional_formats'),
             df=state.dfs[sheet_index],
             column_ids=state.column_ids,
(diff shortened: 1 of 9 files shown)


mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -12,11 +12,24 @@
...
+    for sheet_index, (sheet_name, format) in enumerate(zip(state.df_names, formats)):
+        # We need to convert the column IDs to column letters for conditional formats to export to excel
+        conditional_formats = format.get('conditional_formats')
+        column_ids = state.column_ids.get_column_ids(sheet_index=sheet_index)
+        if conditional_formats is not None:
+            for conditional_format in conditional_formats:
+                # Create new object to store the columns in the excel format
+                conditional_format['columns'] = []
+                for column_id in conditional_format.get('columnIDs'):
+                    column = get_column_from_column_index(column_ids.index(column_id))
```
naterush, line 30: You're missing a step here! Remember that we need to go from ColumnID => ColumnHeader => int => excel column. 

Notably, ColumnID is a Mito internal concept that allows us to have a static id for a specific column, even if you rename or sort it. It's useful in a couple of places, but generally it's just really nice to be able to refer to a column by a specific ID over the full lifecycle of an analysis. 

This means that the column_ids in the the column_ids list that you have -- there are no explicit ordering constraints on it. This is just whatever order the column ids happened to be in when they were first created. 

Instead, here's what you want to do:
```
for column_id in conditional_format.get('columnIDs'):
     column_header = state.column_ids.get_column_header_by_id(sheet_index, column_id)
     column_index = state.dfs[sheet_index].columns.tolist().index(column_header)
     .... 
```
Let me know if this doesn't make sense. The primary thing to callout is that the _only place_ you can reliably get the index of a column in a dataframe is by checking the dataframe, which requires the actual column header, which you can get from the column id. 

Let me know if any of this is unclear! Maybe you could write up your understanding of this as a longer comment in the `ColumnIDMap` class - that way future Mito programmers can figure it out easier - and we can debug your understanding more if necessary :)

mitosheet/mitosheet/public/v3/formatting.py
```diff
@@ -70,4 +72,18 @@ def add_formatting_to_excel_sheet(
                     if row % 2 == 0:
                         sheet.cell(row=row, column=col).style = even_name
                     else:
-                        sheet.cell(row=row, column=col).style = odd_name
\ No newline at end of file
+                        sheet.cell(row=row, column=col).style = odd_name
+
+        # Add conditional formatting
+        if conditional_formats is None:
```
naterush, line 78: Maybe let's break this out into a little helper?

naterush (changes requested):
Hopefully helpful!

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -12,11 +12,24 @@
...
     formats = state.df_formats
-    for sheet_name, format in zip(state.df_names, formats):
+    for sheet_index, (sheet_name, format) in enumerate(zip(state.df_names, formats)):
+        # We need to convert the column IDs to column letters for conditional formats to export to excel
+        conditional_formats = format.get('conditional_formats')
+        column_ids = state.column_ids.get_column_ids(sheet_index=sheet_index)
+        if conditional_formats is not None:
+            for conditional_format in conditional_formats:
+                # Create new object to store the columns in the excel format
+                conditional_format['columns'] = []
```
naterush, line 28: Rather than editing the conditional formatting object, let's make a new object. In it, we can include:
1. The columns (you do this)
2. The filter condition and values (so just copying this over)

This should be pretty simple, but will protect us from accidentally editing objects we're using elsewhere in our codebase with fields we don't expect. 

I think you're on the right track with what to pass to the `add_formatting_to_excel_sheet` function, though! We def want to give it columns rather than column ids or indexes or anything -- users will find that so clear.

marthacryan: Sounds good!

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
marthacryan: Oh ok good to know! I'll try to add some comments to clarify that in the `ColumnIDMap` - From looking at the comments in that file, I suspected that I wasn't getting the index correctly. I do think it was kind of hard to wrap my head around where exactly to get the actual index of the column, so it'd be good to document that a bit

mitosheet/mitosheet/public/v3/formatting.py
marthacryan: Good idea

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
marthacryan: Added a short comment in that class, but lmk if you think I should elaborate / add more in other places

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
marthacryan: Added that in, how does it look?

mitosheet/mitosheet/utils.py
```diff
@@ -141,6 +143,32 @@ def dfs_to_array_for_json(
     return new_array
 
 
+def add_columns_to_condition_formats(
```
naterush, line 146: I think a more informative name would be something like `get_conditional_formats_objects_to-add_to_excel` -- add is a bit overloaded in this namespace / PR.

mitosheet/mitosheet/utils.py
```diff
@@ -141,6 +143,32 @@ def dfs_to_array_for_json(
...
+            'filters': conditional_format.get('filters'),
+            'font_color': conditional_format.get('color'),
+            'background_color': conditional_format.get('backgroundColor')
+        }
+        export_cond_formats.append(new_format)
+        for column_id in conditional_format.get('columnIDs'):
+            column_header = column_ids.get_column_header_by_id(sheet_index, column_id)
+            column_index = df.columns.tolist().index(column_header)
+            column = get_column_from_column_index(column_index)
+            new_format['columns'].append(column)
```
naterush, line 169: Another option for the public interface: 
1. Make the `columns` passed to the `add_formatting_to_excel_sheet` function in the conditional format the actual column_headers rather than the excel column names
2. Pass the dataframe into the`add_formatting_to_excel_sheet` - and do this logic for finding the excel column inside that function. 

This would be nice because it:
1. Doesn't expose users to ColumnIDs.
2. Makes it really easy for users to edit the generated code if they want. - they wouldn't need to go backwards from Excel column -> their data column. 

What do you think?

marthacryan: Ah yeah that definitely makes sense. Added!

mitosheet/mitosheet/tests/step_performers/test_export_to_file.py
```diff
@@ -264,6 +309,64 @@ def test_transpiled_with_export_to_xlsx_format_rows_no_header():
...
 
+
+# This tests when the user exports a dataframe with row formatting without header formatting.
+def test_transpiled_with_export_to_xlsx_conditional_format():
+    df = pd.DataFrame({'A': [1, 2, 3]})
+    mito = create_mito_wrapper(df, arg_names=['df'])
+    mito.set_dataframe_format(0, DF_FORMAT_CONDITIONAL)
+    filename = 'test_format_conditional.xlsx'
+    mito.export_to_file('excel', [0], filename)
+    assert "\n".join(mito.transpiled_code) == """from mitosheet.public.v3 import *
```
naterush, line 320: Currently, we're testing that the generated code is correct. This is an OK first step -- but it doesn't test the real thing we want to, which is that the cell color in the Excel is correct. 

I asked ChatGPT to get the cell color from an Excel file - it gave me this:
```
import pandas as pd
from openpyxl import load_workbook

# Step 1: Read the Excel file using Pandas
file_path = 'path/to/your/excel_file.xlsx'
df = pd.read_excel(file_path, sheet_name='Sheet1')

# Step 2: Load the workbook using openpyxl
wb = load_workbook(file_path)

# Step 3: Get the specific cell whose background color you want to read
sheet_name = 'Sheet1'
cell_address = 'A1'  # Replace this with the desired cell address

sheet = wb[sheet_name]
cell = sheet[cell_address]

# Step 4: Access the background color of the cell
background_color = cell.fill.start_color.index

# Step 5: Convert the color index to the RGB format
# The background_color variable will be something like 'FFFF0000'
# where the first two characters represent transparency, and the last six characters represent the RGB color.
rgb_color = background_color[2:]
# If you need a "#" before the RGB value (e.g., "#FF0000"), you can add it like this:
# rgb_color = "#" + rgb_color

# Step 6: Now you can use the RGB color value as needed
print("Background color of cell {} is: {}".format(cell_address, rgb_color))
```

Want to try writing a helper function of the type `get_cell_background_color(file_name, sheet, row, column) -> str` -- and then we can use it in tests to make sure that we're getting correct values? 

I think it's worth investing in some good testing infrastructure before we scale to the rest of the conditional format conditions.

mitosheet/mitosheet/tests/step_performers/test_export_to_file.py
```diff
@@ -42,6 +42,51 @@
     "conditional_formats": []
 }
 
+CONDITIONAL_FORMATS = [
```
naterush, line 45: Check out pytest.mark.parameterize -- and how we use it through our codebase in mitosheet/tests. I think it's going to be super helpful for scaling tests to multiple filter conditions - especially if we have the above mentioned utility.

naterush (changes requested):
A thought on a testing utility that's going to make scaling these tests hopefully much easier. Overall, this code is super clean and sweet - I love it!

The functionality is working perfect for this condition. Once you get that utility written and used in your test - feel free to merge :)

mitosheet/mitosheet/tests/step_performers/test_export_to_file.py
marthacryan: Ok sounds good!

mitosheet/mitosheet/step_performers/export_to_file.py
```diff
@@ -17,6 +17,10 @@
 from mitosheet.step_performers.step_performer import StepPerformer
 from mitosheet.step_performers.utils import get_param
 
+from mitosheet.utils import write_to_excel
```
naterush, line 20: Potentially, because we're using this as part of the step performer, I'd want to version it with the rest of the public interface. 

But in this case, we aren't actually exposing this function to users. Also, there are less strong backwards compatibility constraints b/c it won't cause errors in the rest of the analysis to e.g. change the formatting. So I'm cool with this as is!

mitosheet/mitosheet/tests/step_performers/test_export_to_file.py
```diff
@@ -264,6 +283,115 @@ def test_transpiled_with_export_to_xlsx_format_rows_no_header():
...
+    df.to_excel(writer, sheet_name="df", index=False)
+    add_formatting_to_excel_sheet(writer, "df", df, 
+        conditional_formats=[
+            {{'columns': {column_ids}, 'filters': {filters}, 'font_color': '{font_color}', 'background_color': '{background_color}'}}
+        ]
+    )
+
+df_styler = df.style\\
+    .apply(lambda series: np.where(series > {filters[0]['value']}, 'color: {font_color}; background-color: {background_color}', None), subset={column_ids})
+"""
```
naterush, line 337: Can we use the `get_cell_formatting` function to test that the correct cells have color in this test (and the other conditional formatting test as well)? I think adding the cell index to the parameterization, and then expecting it to be the colors you set for the background colors would be cool.

naterush (changes requested):
Minor cleanup on tests, then LGTM!

mitosheet/mitosheet/step_performers/export_to_file.py
marthacryan: Sounds good!

mitosheet/mitosheet/tests/step_performers/test_export_to_file.py
marthacryan: Oh oops! Just added that, good catch

PR #848, opened 2023-07-27
Streamlit refactors

naterush:
# Description

1. Adds our own hashing function. See issue on streamlit here that this is a workaround for: https://github.com/streamlit/streamlit/issues/7086
2. We were double delivering messages in multiple cases - because setComponentValue persists between runs, even if the backend changes. This fixes that. 

# Testing

1. Renames aren't detected by the caching function in streamlit. We need to write our own caching function.
2. Two mitosheets that feed into eachother. Add column in first, then delete column in second, then add new column in first. The second sheet has code that is out of date with the sheet.
3. 
# Documentation

### DIFF
--- a/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
+++ b/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
@@ -16,6 +16,10 @@ def _get_dataframe_hash(df: pd.DataFrame) -> bytes:
     2. The values of the dataframe
     3. The index of the dataframe
     4. The order of all of these
+
+    This is necessary due to the issues described here: https://github.com/streamlit/streamlit/issues/7086
+    where streamlit default hashing is not ideal for pandas dataframes, as it misses some column header and
+    reordering changes. 
     """
     try:
         return hashlib.md5(
--- a/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
+++ b/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
@@ -133,9 +133,13 @@ try:
             final dataframes. The second element is a list of lines of code
             that were executed in the Mito spreadsheet.
         """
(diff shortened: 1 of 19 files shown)


mitosheet/mitosheet/streamlit/v1/spreadsheet.py
```diff
@@ -1,12 +1,40 @@
...
 import os
+import pickle
 from typing import Any, Dict, List, Callable, Optional, Tuple, Union
 
 import pandas as pd
 
 from mitosheet.mito_backend import MitoBackend
 from mitosheet.utils import get_new_id
 
+def _get_dataframe_hash(df: pd.DataFrame) -> bytes:
```
aarondr77, line 12: I think adding to the comment why we need this would be dope

aarondr77 (commented):
Those two issues are fixed up! Nice work! LGTM

mitosheet/mitosheet/streamlit/v1/spreadsheet.py
```diff
@@ -70,6 +104,7 @@ def spreadsheet( # type: ignore
             sheet_functions: Optional[List[Callable]]=None, 
             importers: Optional[List[Callable]]=None, 
             df_names: Optional[List[str]]=None,
+            import_folder: Optional[str]=None,
```
aarondr77, line 107: Since this is a big vulnerability, maybe we should not make it an optional parameter, but instead should make users specifically turn it off. Otherwise, we should ensure that the default cause should be safe.

aarondr77: Ooo well I see the feature is turned off in streamlit if you don't set the import_folder so that seems good!

aarondr77 (commented):
LGTM!

PR #851, opened 2023-07-31
Excel round

aarondr77:
# Description

# Testing

# Documentation

### DIFF
--- a/mitosheet/mitosheet/public/v3/sheet_functions/number_functions.py
+++ b/mitosheet/mitosheet/public/v3/sheet_functions/number_functions.py
@@ -537,8 +537,6 @@ def ROUND(arg: NumberRestrictedInputType, decimals: Optional[IntRestrictedInputT
     arg = get_series_from_primitive_or_series(arg, index).fillna(np.nan)
     decimals = get_series_from_primitive_or_series(decimals, index).fillna(0)
 
-    # TODO: Should we cast to an int if we are rounding to 0?
-
     return pd.Series([excel_round(num, decimal) for num, decimal in zip(arg, decimals)])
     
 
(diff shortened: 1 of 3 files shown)


mitosheet/mitosheet/public/v3/sheet_functions/number_functions.py
```diff
@@ -488,21 +488,39 @@ def ROUND(arg: NumberRestrictedInputType, decimals: Optional[IntRestrictedInputT
     }
     """
 
+    def excel_round(x: Optional[Union[int, float]], decimals: int) -> Optional[Union[int, float]]:
+        if pd.isna(x): 
+            return None 
+
+        factor = 10 ** decimals
+        x *= factor
```
naterush, line 496: This needs some serious documentation. What is going on here? Is there a better name we can pick for factor?

naterush (changes requested):
Ok, I think there's a more standard Pythonic way of doing this - namely using the Python standard library of `decimal`. It's less code, more readable, and way less likely to be broken in some way we don't understand. 

I wonder about performance -- so perhaps test it on a larger dataset against current implementation and see how bad it is...


```
import decimal

def excel_round(number, decimals=0):
    context = decimal.getcontext()
    context.rounding = decimal.ROUND_HALF_UP
    number = decimal.Decimal(number)
    rounded_number = round(number, decimals)
    return float(rounded_number)
```

Check out docs here: https://docs.python.org/3/library/decimal.html

naterush:
^ I did it. Feel free to review, and then LGTM. Maybe test perf.

aarondr77:
@naterush Thanks for cleaning this up!

PR #853, opened 2023-07-31
Add other number comparison conditional formats to export with excel

marthacryan:
Adds these conditionals for number fields to the exported formats for excel: 
* less than
* less than or equal
* greater than or equal
* equal
* not equal

### DIFF
--- a/mitosheet/mitosheet/public/v3/formatting.py
+++ b/mitosheet/mitosheet/public/v3/formatting.py
@@ -38,10 +38,6 @@ def add_conditional_formats(
             
             if cond_fill is None and cond_font is None:
                 continue
-            elif cond_fill is None:
-                column_conditional_rule = CellIsRule(operator=operator, font=cond_font, formula=[f'{filter["value"]}'])
-            elif cond_font is None:
-                column_conditional_rule = CellIsRule(operator=operator, fill=cond_fill, formula=[f'{filter["value"]}'])
             else:
                 column_conditional_rule = CellIsRule(operator=operator, fill=cond_fill, font=cond_font, formula=[f'{filter["value"]}'])
             
--- a/mitosheet/mitosheet/public/v3/formatting.py
+++ b/mitosheet/mitosheet/public/v3/formatting.py
@@ -9,7 +9,7 @@ from pandas import DataFrame, ExcelWriter
 from mitosheet.excel_utils import get_column_from_column_index
 
 # Object to map the conditional formatting operators to the openpyxl operators
-operator_map = {
+CONDITIONAL_TO_OPENPYXL_OPERATOR_MAP = {
     'greater': 'greaterThan',
     'less': 'lessThan',
     'number_exactly': 'equal',
@@ -26,7 +26,7 @@ def add_conditional_formats(
     for conditional_format in conditional_formats:
         for filter in conditional_format.get('filters'):
             # Start with the greater than condition
-            operator = operator_map[filter['condition']]
+            operator = CONDITIONAL_TO_OPENPYXL_OPERATOR_MAP[filter['condition']]
             if operator is None:
                 continue
             cond_fill = None
--- a/mitosheet/mitosheet/public/v3/formatting.py
+++ b/mitosheet/mitosheet/public/v3/formatting.py
@@ -29,9 +29,22 @@ def add_conditional_formats(
             operator = operator_map[filter['condition']]
             if operator is None:
                 continue
-            cond_fill = PatternFill(start_color=conditional_format['background_color'][1:], end_color=conditional_format['background_color'][1:], fill_type='solid')
-            cond_font = Font(color=conditional_format['font_color'][1:])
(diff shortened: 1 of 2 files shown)


mitosheet/mitosheet/public/v3/formatting.py
```diff
@@ -16,11 +26,12 @@ def add_conditional_formats(
     for conditional_format in conditional_formats:
         for filter in conditional_format.get('filters'):
             # Start with the greater than condition
-            if filter['condition'] != 'greater':
+            operator = operator_map[filter['condition']]
```
naterush, line 29: This will error if the condition isn't in the map - e.g. non-number conditions. We should us a .get on the operator_map.

mitosheet/mitosheet/public/v3/formatting.py
```diff
@@ -8,6 +8,16 @@
 
 from mitosheet.excel_utils import get_column_from_column_index
 
+# Object to map the conditional formatting operators to the openpyxl operators
+operator_map = {
```
naterush, line 12: In python, we usually make constants all caps -- so `OPERATOR_MAP`. Also, this could be more informatively named. Can you come up with a name that tells us specifically what's in the map?

naterush (changes requested):
This code errors when you set a background color but not a font, or a font color but not a background color. This would def be a good thing to test - setting one, but not the other. 

I added this to the testing doc - I think it's easy to miss - but generally testing a "partial application" is a really common way of catching errors. We as app devs can make assumptions - but often users just use a piece of what we give them.

Ping me whenever for a re-review!

mitosheet/mitosheet/public/v3/formatting.py
marthacryan: Ok! I added it as `CONDITIONAL_TO_OPENPYXL_OPERATOR_MAP`, is that too long?

mitosheet/mitosheet/public/v3/formatting.py
marthacryan: Done!

mitosheet/mitosheet/public/v3/formatting.py
```diff
@@ -8,19 +8,43 @@
...
-            cond_fill = PatternFill(start_color=conditional_format['background_color'][1:], end_color=conditional_format['background_color'][1:], fill_type='solid')
-            cond_font = Font(color=conditional_format['font_color'][1:])
-            column_conditional_rule = CellIsRule(operator="greaterThan", font=cond_font, fill=cond_fill, formula=[f'{filter["value"]}'])
+            elif cond_fill is None:
+                column_conditional_rule = CellIsRule(operator=operator, font=cond_font, formula=[f'{filter["value"]}'])
+            elif cond_font is None:
+                column_conditional_rule = CellIsRule(operator=operator, fill=cond_fill, formula=[f'{filter["value"]}'])
+            else:
+                column_conditional_rule = CellIsRule(operator=operator, fill=cond_fill, font=cond_font, formula=[f'{filter["value"]}'])
+
```
naterush, line 47: I think these cases can be collapsed into just two cases. The way that optional arguments work in Python is that they are most often set to be equal to None by default. In this case, looking at the CellIsRule, we have the definition:
```
def CellIsRule(operator=None, formula=None, stopIfTrue=None, font=None, border=None, fill=None):
```

Thus, if the cond_fill is None, and you pass it to the file -- this will be like you didn't pass it at all (as the default is None). 

So I think all of this can be collapsed into two cases - one where both cond_fill and cond_font are None, and the other case where you just pass both arguments.

mitosheet/mitosheet/tests/step_performers/test_export_to_file.py
```diff
@@ -289,39 +294,86 @@ def test_transpiled_with_export_to_xlsx_format_rows_no_header():
...
+        "=="
+    ),
+    (
+        ['A'],
+        [{'condition': 'number_not_exactly', 'value': 4}],
+        None, 
+        '#001a58',
+        'A2',
+        'B2',
+        "!="
```
naterush, line 345: Great additional tests!

naterush (changes requested):
Functionality is working great. One proposed refactor that will cut a bit of duplicated code (and hopefully is educational about default arguments), and then LGTM!

mitosheet/mitosheet/public/v3/formatting.py
marthacryan: Ah true! Just added

mitosheet/mitosheet/public/v3/formatting.py
```diff
@@ -8,19 +8,43 @@
 
 from mitosheet.excel_utils import get_column_from_column_index
 
+# Object to map the conditional formatting operators to the openpyxl operators
+CONDITIONAL_TO_OPENPYXL_OPERATOR_MAP = {
```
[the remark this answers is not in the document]
marthacryan: Cool!

PR #871, opened 2023-08-03
Add number formatting into excel export

marthacryan:
Adds formatting to the excel export for:
* Precision of decimal points
* Currency
* Accounting
* Percentage
* Scientific Notation

Creates this API call: 
```
    add_formatting_to_excel_sheet(writer, "final_sheet_name", final_sheet_name, 
        number_formats={
            "FFMC": '($0.0)'
        }
    )
```

Todo:
- [x] Add tests
- [x] Clean up API call

### DIFF
--- a/mitosheet/mitosheet/public/v3/formatting.py
+++ b/mitosheet/mitosheet/public/v3/formatting.py
@@ -18,7 +18,6 @@ from mitosheet.types import (
     FC_STRING_ENDS_WITH, FC_STRING_EXACTLY, FC_STRING_NOT_EXACTLY,
     FC_STRING_STARTS_WITH, FC_STRING_CONTAINS_CASE_INSENSITIVE)
 
-PLACEHOLDER = 'FORMAT_STRING_PLACEHOLDER'
 
 # Object to map the conditional formatting operators to the excel formulas
 CONDITION_TO_COMPARISON_FORMULA: Dict[str, str] = {
@@ -115,7 +114,7 @@ def add_conditional_formats(
 
 
 def add_number_formatting(
-    number_formats: Optional[Dict[str, Any]],
+    number_formats: Optional[Dict[str, str]],
     sheet: Worksheet,
     df: DataFrame
 ) -> None:
@@ -126,11 +125,7 @@ def add_number_formatting(
         column = get_column_from_column_index(column_index)
         cell_range = f'{column}2:{column}{sheet.max_row}'
         for cell in sheet[cell_range]:
-            precision = number_format.get('precision', 0)
-            decimal_string = f'0.{precision*"0"}' if precision > 0 else '0'
-            format_string = number_format.get('type', PLACEHOLDER)
-            full_format_string = format_string.replace(PLACEHOLDER, decimal_string)
-            cell[0].number_format = full_format_string
+            cell[0].number_format = number_format
 
 
 def add_formatting_to_excel_sheet(
--- a/mitosheet/mitosheet/public/v3/formatting.py
+++ b/mitosheet/mitosheet/public/v3/formatting.py
@@ -1,4 +1,4 @@
-from typing import Optional, Dict, Tuple
+from typing import Any, Optional, Dict
 
 from openpyxl.styles import Font, PatternFill
 from openpyxl.styles import NamedStyle
@@ -17,6 +17,8 @@ from mitosheet.types import (
     FC_NUMBER_NOT_EXACTLY, FC_STRING_CONTAINS, FC_STRING_DOES_NOT_CONTAIN,
     FC_STRING_ENDS_WITH, FC_STRING_EXACTLY, FC_STRING_NOT_EXACTLY,
     FC_STRING_STARTS_WITH, FC_STRING_CONTAINS_CASE_INSENSITIVE)
+
+PLACEHOLDER = 'FORMAT_STRING_PLACEHOLDER'
 
 # Object to map the conditional formatting operators to the excel formulas
 CONDITION_TO_COMPARISON_FORMULA: Dict[str, str] = {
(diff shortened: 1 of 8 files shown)


mitosheet/src/mito/types.tsx
```diff
@@ -184,11 +184,11 @@ export interface ColumnFilterMap {
...
-    CURRENCY = 'currency',
-    ACCOUNTING = 'accounting',
-    PERCENTAGE = 'percentage',
-    SCIENTIFIC_NOTATION = 'scientific notation'
+    PLAIN_TEXT = 'FORMAT_STRING_PLACEHOLDER',
+    CURRENCY = '$FORMAT_STRING_PLACEHOLDER',
+    ACCOUNTING = '($FORMAT_STRING_PLACEHOLDER)',
+    PERCENTAGE = 'FORMAT_STRING_PLACEHOLDER%',
+    SCIENTIFIC_NOTATION = 'FORMAT_STRING_PLACEHOLDERE+0'
 }
```
naterush, line 192: Ok! These changes require you to touch a new part of the codebase -- the part that deals with upgrading analyses. 

# Why upgrading analyses

Imagine that a user uses Mito 0.1. and sets the format of some number column to `plain text`. This is saved in a file in the `.mito/saved_analyses/{analysis_id}.json` file on their computer. 

Then, they go and upgrade to Mito 0.2 -- the version that makes the above changes. That means that when the analysis is replayed, the number format that gets set is `plain text` -- but notably, the frontend no longer detects that as a valid number formatting, and so the number wouldn't be properly formatted.

**Whenever we change the structure or contents of a step performer, we have to do an upgrade, to make sure that previous user analyses don't break.** Notably, in this case, we need to upgrade the `set_dataframe_format.py` step!

There is a whole bunch of infrastructure for managing these upgrades (relevant files are `mitosheet/saved_analyses/step_upgraders/set_dataframe_format.py` and `mitosheet/saved_analyses/schema_utils.py` if you're interested). 

# Should we upgrade in this case

I'm wondering -- since we have to do transformations on the FORMAT_STRING_PLACEHOLDER anyways -- might we want to just keep the `NumberColumnFormatEnum` the same for now? 

You'll have to change the backend a bit (but not by a ton) just to have an If statement for each type of these - but otherwise the logic will be very similar to what you have. 

Then, if we want to move closer to Excel formatting strings over time, we can do it in the following way:
1. Add a new number format type: 'excel format string'. Notably, this will also take a format string as well an additional parameter on top of the precision. 
2. Furthermore, as this is a _new parameter_ and not an _old one_ -- we won't break anyone's old analyses. 
3. We can keep all the old code paths around for a while, until we think no one is using them anymore.

I am still excited about building a TSX function called `getFormattedNumber(num: number, excel_format_string: string) -> string` -- but I think that it might be a bigger lift than is worth it for now (especially with having to upgrade). So perhaps let's just make these changes as small as possible. 

To make it very clear, I think we should:
1. Change `NumberColumnFormatEnum` back to previous values.
2. Add if statements to the backend to build these strings.
3. Change the interface in the way you said (so you're passing the final Excel formatting strings to the add_formatting.. function)

naterush (commented):
Bit of a long comment here, but took the opportunity for some knowledge sharing. happy to clarify more, let me know!

mitosheet/src/mito/types.tsx
marthacryan: Got it! That all definitely makes sense. I'll make those changes!

mitosheet/mitosheet/tests/step_performers/test_export_to_file.py
```diff
@@ -86,6 +87,54 @@ def get_cell_conditional_formatting(
...
+        {
+            "A": {
+                "type": ACCOUNTING,
+                "precision": 1
+            }
+        },
+        "($0.0)"
+    )
+]
+
```
naterush, line 137: Here's two tests that aim to flex this new feature to find some bugs:
1. Two columns with number formats. 
2. Formatting with number formats and conditional formats.

mitosheet/mitosheet/public/v3/formatting.py
```diff
@@ -112,6 +113,21 @@ def add_conditional_formats(
...
+    df: DataFrame
+) -> None:
+    if number_formats is None:
+        return
+    for column_header, number_format in number_formats.items():
+        column_index = df.columns.tolist().index(column_header)
+        column = get_column_from_column_index(column_index)
+        cell_range = f'{column}2:{column}{sheet.max_row}'
+        for cell in sheet[cell_range]:
+            cell[0].number_format = number_format
```
naterush, line 128: Clean and nice.

mitosheet/mitosheet/transpiler/transpile_utils.py
```diff
@@ -113,21 +113,20 @@ def param_dict_to_code(param_dict: Dict[str, Any], level: int=0, as_single_line:
...
     for key, value in param_dict.items():
+        # If we're defining a dict entry, we need to add quotes around the key and use a colon
+        key_definition = f'"{key}": ' if is_dict_entry else f'{key}='
         if isinstance(value, dict):
             # Recurse on this nested param dictonary
-            code_chunk = f"{key} = {param_dict_to_code(value, level=level + 1)}"
+            code_chunk = f"{key_definition}{{{f'{param_dict_to_code(value, level=level + 1, is_dict_entry=True)}{NEWLINE_CONSTANT}{TAB_CONSTANT * (level+1)}'}}}"
         elif isinstance(value, list):
-            code_chunk = f"{key}={column_header_to_transpiled_code(value, tab_level=tab_level + 1)}"
+            code_chunk = f"{key_definition}{column_header_to_transpiled_code(value, tab_level=tab_level + 1)}"
```
naterush, line 126: Sweet, this util is getting more and more useful!

naterush (changes requested):
This is working awesome - and the code is super clean! Two more tests to stress test things a bit more, and once those pass LGTM!

PR #872, opened 2023-08-03
Fix exporting excel with multiple sheets bug

marthacryan:
This updates an existing test to catch the following bug and fixes the bug:
To replicate:

* Create two tabs in the mitosheet.
* Format both of them.
* Go to export when running generated code, and select Excel as the file type
* Export just the second sheet
* See that the generated code makes calls to add_formatting_to_excel_sheet twice, despite only one tab being exported.

### DIFF
--- a/mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
+++ b/mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
@@ -15,36 +15,35 @@ from mitosheet.transpiler.transpile_utils import param_dict_to_code
 from mitosheet.utils import get_conditional_formats_objects_to_export_to_excel
 
 # This is a helper function that generates the code for formatting the excel sheet
-def get_format_code(state: State) -> list:
-    code = []
+def get_format_code(state: State, sheet_index: int) -> list:
     formats = state.df_formats
-    for sheet_index, (sheet_name, format) in enumerate(zip(state.df_names, formats)):
-        # We need to convert the column IDs to column letters
-        # for conditional formats to export to excel
-        conditional_formats = get_conditional_formats_objects_to_export_to_excel(
-            format.get('conditional_formats'),
-            column_id_map=state.column_ids,
-            sheet_index=sheet_index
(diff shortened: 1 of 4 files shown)


mitosheet/mitosheet/tests/step_performers/test_export_to_file.py
```diff
@@ -599,17 +599,12 @@ def test_transpiled_with_export_to_xlsx_format_two_sheets():
     mito.set_dataframe_format(0, DF_FORMAT_HEADER)
     mito.set_dataframe_format(1, DF_FORMAT_HEADER_AND_ROWS)
     filename = 'test_format_two.xlsx'
-    mito.export_to_file('excel', [0, 1], filename)
+    mito.export_to_file('excel', [1], filename)
     assert "\n".join(mito.transpiled_code) == """from mitosheet.public.v3 import *
 import pandas as pd
 
 with pd.ExcelWriter(r\'test_format_two.xlsx\', engine="openpyxl") as writer:
```
naterush, line 606: Test looks good! This is one of the issues with copying generated code back into a test -- it can be easy to not read thoroughly and then miss an important bit.

mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -15,36 +15,35 @@
 from mitosheet.utils import get_conditional_formats_objects_to_export_to_excel
 
 # This is a helper function that generates the code for formatting the excel sheet
-def get_format_code(state: State) -> list:
-    code = []
+def get_format_code(state: State, sheet_index: int) -> list:
```
naterush, line 18: What if it took sheet_indexes: List[int]? Then the call-site could be get_format_code(post_state, list(self.sheet_index_to_export_location.keys()) -- and the code would be pretty much the same as before. Probably much simpler than the complex comprehension and filter statement below.

naterush (changes requested):
One minor refactor, and then LGTM once tests pass!

PR #873, opened 2023-08-03
mitosheet: add more code config options that make recon app easier

naterush:
# Description

Ian into this when creating the recon app. If you are Mito Enterprise, and you want the function to transpile as code - you can configure through the Code > Code Options taskpane. 

However, in Streamlit, you often want this script to _not_ call the code it generates. This is pretty reasonable, I think -- so I added a configuration option for it. 

In the future, we should make the code options configurable through the `spreadsheet` interface - you should be able to by default turn on the function generation.

# Testing

Use this feature, make sure it's not calling the function it generates.

# Documentation

We should add to docs

### DIFF
--- a/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
+++ b/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
@@ -7,6 +7,7 @@ from typing import Any, Dict, List, Callable, Optional, Tuple, Union
 import pandas as pd
 
 from mitosheet.mito_backend import MitoBackend
+from mitosheet.types import CodeOptions
 from mitosheet.utils import get_new_id
 
 def _get_dataframe_hash(df: pd.DataFrame) -> bytes:
@@ -83,6 +84,7 @@ try:
             *args: Union[pd.DataFrame, str, None], 
             _importers: Optional[List[Callable]]=None, 
             _sheet_functions: Optional[List[Callable]]=None, 
+            _code_options: Optional[CodeOptions]=None,
             import_folder: Optional[str]=None,
             df_names: Optional[List[str]]=None,
             session_id: Optional[str]=None,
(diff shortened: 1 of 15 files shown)


mitosheet/mitosheet/streamlit/v1/spreadsheet.py
```diff
@@ -83,6 +84,7 @@ def _get_mito_backend(
             *args: Union[pd.DataFrame, str, None], 
             _importers: Optional[List[Callable]]=None, 
             _sheet_functions: Optional[List[Callable]]=None, 
+            _code_options: Optional[CodeOptions]=None,
```
naterush, line 87: For reviewer -- noting here that I'm aiming to allow app developers to specify code options rather than requiring them from the user in the future. This isn't exposed yet, fully.

marthacryan:
This looks great!! The code is really clear. I have some feedback on the UI but I think some of it could/should be in another PR:
* (This probably should go into this PR) I think it would make sense to disable the toggle in the same way that you're disabling the other input fields for here: https://github.com/mito-ds/mito/blob/6f5e94ca0d19e77f855462589d7a0fa51ea229ec/mitosheet/src/mito/components/taskpanes/CodeOptions/CodeOptionsTaskpane.tsx#L76 I could even see an argument for just hiding those fields when the "Generate Function" toggle is set to off but that seems like it'd be better in a separate PR
* I noticed that there was kind of a long lag when you press the toggle before it actually switched on/off. Do you know why that might be happening? That might require a larger refactor (or actually be the desired behavior) but it did make the UX a little confusing for me because I ended up clicking the toggle again because I thought I might've not actually clicked it, and then it just toggled back off because of the two clicks. 
* We might want to use a different tooltip from what the built-in "title" field provides, because I had to hover so long over the question marks next to the fields that I actually gave up until I saw in the code that it was set through the title field.

PR #874, opened 2023-08-04
Add toggle to export formatting for excel

marthacryan:
Adds a toggle to the download taskpane when the file type is excel:
image
image

(Maybe should be changed) CTA for non-pro users:
image

Todo:
- [x] Add toggle to export as code taskpane
- [x] Add CTA for non-pro users
- [ ] Add tooltip to explain that performance is slower when formatting is on
  - I tried adding a tooltip but the component we have is a little question mark icon and it looked a little weird next to the pro icon so I left it out. I can add that in if you think it's a good idea though!

Questions:
- How do we handle when CSV is the file type? (Right now it just only appears as a field to the user when excel is the file type)
  - Keep as is
- What should the CTA be for non-pro users? Right now I have a really plain text that says "Upgrade to Pro" - not saying that should be the design but I'm not sure what should be 
  - Solution: disabling the toggle

### DIFF
--- a/mitosheet/mitosheet/api/get_dataframe_as_excel.py
+++ b/mitosheet/mitosheet/api/get_dataframe_as_excel.py
@@ -20,7 +20,8 @@ def get_dataframe_as_excel(params: Dict[str, Any], steps_manager: StepsManagerTy
     sheet_indexes = params['sheet_indexes']
 
     # Formatting is a Mito pro feature, but we also allow it for testing
-    allow_formatting = is_pro() or is_running_test()
+    export_formatting = params.get('export_formatting', False)
+    allow_formatting = (is_pro() and export_formatting) or is_running_test()
 
     # We write to a buffer so that we don't have to save the file
     # to the file system for no reason
--- a/mitosheet/mitosheet/api/get_dataframe_as_excel.py
+++ b/mitosheet/mitosheet/api/get_dataframe_as_excel.py
@@ -21,7 +21,7 @@ def get_dataframe_as_excel(params: Dict[str, Any], steps_manager: StepsManagerTy
 
     # Formatting is a Mito pro feature, but we also allow it for testing
     export_formatting = params.get('export_formatting', False)
-    allow_formatting = (is_pro() and export_formatting) or is_running_test()
+    allow_formatting = (is_pro() or is_running_test()) and export_formatting
 
     # We write to a buffer so that we don't have to save the file
     # to the file system for no reason
--- a/mitosheet/mitosheet/tests/test_utils.py
+++ b/mitosheet/mitosheet/tests/test_utils.py
@@ -595,7 +595,8 @@ class MitoWidgetTestWrapper:
                 'params': {
                     'type': type,
                     'sheet_indexes': sheet_indexes,
-                    'file_name': file_name
+                    'file_name': file_name,
+                    'export_formatting': True
                 }
             }
         )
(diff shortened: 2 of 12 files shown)


mitosheet/mitosheet/api/get_dataframe_as_excel.py
```diff
@@ -20,7 +20,8 @@ def get_dataframe_as_excel(params: Dict[str, Any], steps_manager: StepsManagerTy
     sheet_indexes = params['sheet_indexes']
 
     # Formatting is a Mito pro feature, but we also allow it for testing
-    allow_formatting = is_pro() or is_running_test()
+    export_formatting = params.get('export_formatting', False)
```
naterush, line 23: A new param! Cool. You'll need to do some step upgrading. I'll post some documentation on this in a moment.

mitosheet/mitosheet/tests/test_utils.py
```diff
@@ -595,7 +595,8 @@ def export_to_file(
                 'params': {
                     'type': type,
                     'sheet_indexes': sheet_indexes,
-                    'file_name': file_name
+                    'file_name': file_name,
+                    'export_formatting': True
```
naterush, line 599: Let's make this a parameter to the function, and make it default to True. 

Then, can we add a test to make sure things still work (and you don't get formatting) when this is set to False!

mitosheet/mitosheet/api/get_dataframe_as_excel.py
```diff
@@ -20,7 +20,8 @@ def get_dataframe_as_excel(params: Dict[str, Any], steps_manager: StepsManagerTy
...
+    allow_formatting = (is_pro() and export_formatting) or is_running_test()
```
naterush, line 24: When you write a test for when formatting is turned off, you'll have to update this statement possible - the or might be off

mitosheet/src/mito/components/taskpanes/ExportToFile/ExportToFileTaskpane.tsx
```diff
@@ -133,6 +138,27 @@ const ExportToFileTaskpane = (props: ExportToFileTaskpaneProps): JSX.Element =>
                         </Select>
                     </Col>
                 </Row>
+                {params.type === 'excel' &&
+                <Row justify='space-between' align='center'>
+                    <Col style={{ display: 'flex' }}>
+                        <p className="text-header-3">Export with formatting</p>&nbsp;
+                        <ProIcon/>
```
naterush, line 145: I think we only need to display the Pro icon when the user is not on Pro.

mitosheet/src/mito/components/taskpanes/Download/DownloadTaskpane.tsx
```diff
@@ -199,16 +204,33 @@ const DownloadTaskpane = (props: DownloadTaskpaneProps): JSX.Element => {
...
-                            exportState={props.uiState.exportConfiguration as ExcelExportState}
-                            setUIState={props.setUIState}
-                            newlyFormattedColumns={newlyFormattedColumns}
-                            setNewlyFormattedColumns={setNewlyFormattedColumns}
-                        />  
+                        <div>
+                            <Row justify='space-between' align='center'>
+                                <Col style={{ display: 'flex' }}>
+                                    <p className="text-header-3">Export with formatting</p>&nbsp;
+                                    <ProIcon/>
```
naterush, line 211: Same as above -- let's only display this in the case that the user is not on Pro.

naterush (changes requested):
A bit of code cleanup. I'm going to push some documentation on the who/what/when/where/why of step upgrading in a moment!

naterush:
All the functionality is working great! Check out the minor refactors -- as well as the step upgrading work that is necessary so we don't break users analyses. 

Feel free to ping me with any questions -- or improve the documentation yourself as you get into it, if you think it can be clarified more :)

PR #875, opened 2023-08-05
streamlit: return an ordered dict so users can rely on order

naterush:
# Description

In my building of real life apps, I don't just wanna access by name - I also want to access by index. An OrderedDict splits the difference!

# Testing

Use it.

# Documentation

No.

### DIFF
--- a/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
+++ b/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 import hashlib
 import json
 import os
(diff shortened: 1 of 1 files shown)


marthacryan (approved):
This definitely seems like a good idea to me - LGTM!

marthacryan:
Ok I'm actually a little confused about the use of the returned objects in the `spreadsheet` function. I see we're using the `code`, but where are the `new_dfs` being used?

naterush:
Sorry for the confusion! The `streamlit.py` file is just a little testing application we commit (maybe we shouldn't even commit it). 

Feel free to edit `streamlit.py` to test the new functionality - specifically, that the order is maintained!

marthacryan:
Got it - just tested and it's working as expected! LGTM

PR #880, opened 2023-08-10
Fix format code chunk bug for long dataframe names

marthacryan:
Fixes a bug when dataframe names are long where the generated code for calling `to_excel` uses a different name for the dataframe from the call to `add_formatting_to_excel` and throws an error when run. This switches to using the same value between those two calls

### DIFF
--- a/mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
+++ b/mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
@@ -21,8 +21,7 @@ from mitosheet.utils import (
 def get_format_code(state: State, sheet_indexes: Dict[int, str]) -> list:
     code = []
     formats = state.df_formats
-    for sheet_index in sheet_indexes.keys():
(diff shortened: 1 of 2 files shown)


mitosheet/mitosheet/code_chunks/export_to_file_code_chunk.py
```diff
@@ -21,8 +21,7 @@
 def get_format_code(state: State, sheet_indexes: Dict[int, str]) -> list:
     code = []
     formats = state.df_formats
-    for sheet_index in sheet_indexes.keys():
-        sheet_name = state.df_names[sheet_index]
+    for sheet_index, export_location in sheet_indexes.items():
```
naterush, line 24: `sheet_indexes` is probably a badly named variable - everywhere else in the codebase, it is a list of sheet_indexes. 

Want to rename this to the name of the variable you are passing into this function?

naterush (changes requested):
Super minor cleanup.

PR #889, opened 2023-08-16
Prevent export when 'export as code' is selected

marthacryan:
Currently when the user selects "Download File when Executing Code", the code downloads when the user generates the code. This is not the expected behavior for the user, so I just removed it.


mitosheet/mitosheet/step_performers/export_to_file.py
```diff
@@ -97,10 +97,7 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
             for sheet_index, file_name in sheet_index_to_export_location.items():
                 post_state.dfs[sheet_index].to_csv(file_name, index=False)
```
naterush, line 98: Looks like we write CSV files too - wanna remove that as well? 

Also, let's add a test that no files get creates _until_ the code gets run?

Then - LGTM.

naterush (changes requested):
Minor fix up, a test (for CSV and Excel) then LGTM.

PR #890, opened 2023-08-18
Add IFS formula

marthacryan:
Adds support for the IFS formula. 

Todo: 
- [x] Add tests
- [ ] Add ability to use TRUE as a condition (which would basically be the "else" case)
  - [ ] This can be tackled in another PR


mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
```diff
@@ -123,6 +123,52 @@ def IF(condition: pd.Series, true_series: AnyPrimitiveOrSeriesInputType, false_s
...
+                "description": "The value the function returns if condition is True."
+            }
+        ]
+    }
+    """
+    # Go through all of the conditions. The first condition that is true for
+    # *each cell* should be selected. 
+    results = pd.Series()
+    for index, condition in enumerate(argv):
+        # The args are alternating between the conditions and the values corresponding to them.
```
naterush, line 155: I think that we can actually do input sanity checking. Namely, all series with an even index should be boolean (or a boolean series).

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
```diff
@@ -123,6 +123,52 @@ def IF(condition: pd.Series, true_series: AnyPrimitiveOrSeriesInputType, false_s
...
+            # If it is, use the "true_series" to fill the value in the end series.
+            # If it isn't save this for the next series by using "None"
+            true_series = get_series_from_primitive_or_series(argv[index+1], condition.index)
+            # Add the new data into the series
+            new_series = pd.Series(
+                data=[true_series.loc[i] if c else None for i, c in condition.items()],
+                index=condition.index
+            )
+            # Combine the series to fill in the None values in the final series
+            results = results.combine_first(new_series)
```
naterush, line 169: Cool -- this is a Pandas function I didn't know about.

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
naterush: You can see an example of how a sheet function might raise an error in the `SUMPRODUCT` function in the v3 interface.

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
```diff
@@ -123,6 +123,52 @@ def IF(condition: pd.Series, true_series: AnyPrimitiveOrSeriesInputType, false_s
...
+        #   only used as a function of the conditions. 
+        if isinstance(condition, pd.Series):
+            # For each cell, check if the value is true.
+            # If it is, use the "true_series" to fill the value in the end series.
+            # If it isn't save this for the next series by using "None"
+            true_series = get_series_from_primitive_or_series(argv[index+1], condition.index)
+            # Add the new data into the series
+            new_series = pd.Series(
+                data=[true_series.loc[i] if c else None for i, c in condition.items()],
+                index=condition.index
```
naterush, line 166: Yeah, this does look really slow. A cool thing that might work -- using the boolean series as indexers, like this: 

Screen Shot 2023-08-20 at 9 43 30 AM

`s1` is the condition, and because we just use it to select only some of the `true_series`, we can then just put this series subset (with only some of the index labels defined) into the combine_first function.

I think this will be much faster!

```

File: mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py [hunk 4]
```diff
@@ -123,6 +123,52 @@ def IF(condition: pd.Series, true_series: AnyPrimitiveOrSeriesInputType, false_s
...
+        ]
+    }
+    """
+    # Go through all of the conditions. The first condition that is true for
+    # *each cell* should be selected. 
+    results = pd.Series()
+    for index, condition in enumerate(argv):
+        # The args are alternating between the conditions and the values corresponding to them.
+        # If this is a series, we assume it is the condition. We skip the values because they're
+        #   only used as a function of the conditions.
```
naterush, line 157: I don't think we should do this! I think that even numbered args should be treated as conditions (if possible), and odd number args should be treated as results. 

This means the code below needs to figure out if the true_series is really a series or if it might be just a constant.

naterush (changes requested):
I left a few thoughts on the approach, but generally you're barking up the right tree. 

I think it probably makes sense to write a good set of tests (including all varied data types you're like to see -- series and constants alike) before going any farther. 

In general, sheet functions are an extremely good target for test driven development, as they have incredibly static interfaces.

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
marthacryan: Ah ok didn't know that syntax! That'll be really helpful

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
marthacryan: Just added it (and a test for the error)! How does that look?

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
```diff
@@ -123,6 +124,59 @@ def IF(condition: pd.Series, true_series: AnyPrimitiveOrSeriesInputType, false_s
...
+        "syntax_elements": [{
+                "element": "boolean_condition",
+                "description": "An expression or series that returns True or False values. Valid conditions for comparison include ==, !=, >, <, >=, <=."
+            },
+            {
+                "element": "value_if_true",
+                "description": "The value the function returns if condition is True."
+            }
+        ]
+    }
```
naterush, line 149: This needs to be updated, I think.

naterush (changes requested):
A few more tests that I think would be awesome to add -- ping me when these are up and I'll do a complete re-review!

mitosheet/mitosheet/tests/public/v3/sheet_functions/bool_functions/test_ifs.py
```diff
@@ -0,0 +1,89 @@
...
+
+import pytest
+import pandas as pd
+from mitosheet.errors import MitoError
+
+from mitosheet.public.v3.sheet_functions.bool_functions import IFS
+
+# Raw function tests
+
+IFS_TESTS = [
```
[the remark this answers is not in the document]
marthacryan: Added a test for values that are different datatypes! And for when one value is a series and the other is a constant.

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
```diff
@@ -123,6 +124,76 @@ def IF(condition: pd.Series, true_series: AnyPrimitiveOrSeriesInputType, false_s
...
+            # This is the "else" case for the IFS formula.
+            elif condition == True:
+                base_index = next(iter(s.index for s in argv if isinstance(s, pd.Series)), None)
+                if base_index is None:
+                    error = MitoError(
+                        'invalid_args_error',
+                        'IFS',
+                        f"IFS requires at least one boolean series condition.",
+                        error_modal=False
+                    )
```
naterush, line 177: This feels like an unnecessary limitation. Will describe an implementation that might remove this in a second -- let me know what you think.

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
```diff
@@ -123,6 +124,76 @@ def IF(condition: pd.Series, true_series: AnyPrimitiveOrSeriesInputType, false_s
...
+                    results = true_series
+                else:
+                    # Combine the series to fill in the None values in the final series
+                    results = results.combine_first(true_series)
+            # If this is an even indexed argument, it should be a boolean series or True. If it isn't, we throw an error.
+            else:
+                error = MitoError(
+                    'invalid_args_error',
+                    'IFS',
+                    f"IFS requires a boolean series condition as the first argument, but got {condition} instead.",
```
naterush, line 191: Nice -- user-facing error messages! Since our users are new to Python, they don't know what series are -- also the {condition} might be a huge series, and so this error message would be insanely long. 

Can you try and think of another error message that would be more helpful to our users? Hint -- talking about the "5th" argument to the function is something users get. They also probably get the term "boolean".

mitosheet/mitosheet/tests/public/v3/sheet_functions/bool_functions/test_ifs.py
```diff
@@ -0,0 +1,104 @@
...
+@pytest.mark.parametrize("_argv, expected", IFS_TESTS)
+def test_ifs_direct(_argv, expected):
+    result = IFS(*_argv)
+    if isinstance(result, pd.Series):
+        assert result.equals(expected)
+    else: 
+        assert result == expected
+
+
+IFS_INVALID_TESTS = [
```
naterush, line 97: Another test -- what happens if you pass an 3 arguments? Can we make this report a good error message. 

Also, can you make this test_invalid_args error test actually test that the error messages coming back are informative? 

You're capturing the error info with `e_info`, so this shouldn't be too hard!

naterush (changes requested):
## A simpler implementation

There's quite a bit of duplicated code that results from handling constants and series separately -- and I have a bit of trouble following all the nested cases. 

Here's a little proposal on how to reduce some of this code -- let me know what you think. 

```
base_index = next(iter(s.index for s in argv if isinstance(s, pd.Series)), None)
    # If index is None, we're dealing with all constants
    if base_index is None:
        for index, arg in enumerate(argv):
            if index % 2 == 0:
                if arg:
                    return argv[index+1]
        return 0 # TODO: what should our default return be

    else:
        # Otherwise, we have at least one series -- so we can go through and turn all of the constants into series.
        # This dramatically reduces the amount of casing we have to do
        argv = tuple([get_series_from_primitive_or_series(arg, base_index) if not isinstance(arg, pd.Series) else arg for arg in argv])
        results = pd.Series(index=base_index)

        for index, condition in enumerate(argv):
            if index % 2 == 0:
                if condition.dtype != bool:
                    raise Error # TODO: make this legit
                
                # If it is, use the "true_series" to fill the value in the result series
                true_series = argv[index+1]
                new_series = true_series[condition]
                results = results.combine_first(new_series)
                

        return results
```

I didn't run the above code, so I'm sure there are some kinks in it to work out -- but I think the strategy employed is pretty legit: 
1. If there are all constants, then handle them in one case
2. If there is a series, turn _everything_ into a series. 

The net result is that 15 lines of code cut -- but I also think the resulting code is dramatically easier to understand and follow, as there are only two cases, with no additional nested IFS. 

FWIW, I didn't have this solution in-mind until reading this code -- I just got sorta confused between all the cases, and so took a shot at reducing to the minimal number of cases. I think that this sort of "taking a step back" and seeing if any dramatic simplifications are possible is something that can be hard to remember to do -- but it can be really useful to review your own PR and ask yourself that very question, to see if a refactor (before merge) will lead to easier to maintain code going forward. 

## Formula Writing Guide

Based on what you learned from this PR, it would be awesome if you could write a little "formula writing guide". I think it could include learnings about testing, writing the actual formula, how it gets used (e.g. why you can't find out where they are called anywhere), and also building the frontend docs. 

On top of being useful for future Mito contributors, this will also give us another chance to debug your understand of both how formulas work, as well as iterate on how to write them effectively!

marthacryan:
@naterush That worked perfectly! Thanks for the refactor thoughts, I agree that it got a little hard to read. Definitely makes sense to take a step back to think about readability in the future

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
```diff
@@ -123,6 +124,76 @@ def IF(condition: pd.Series, true_series: AnyPrimitiveOrSeriesInputType, false_s
...
+        ],
+        "syntax": "IFS(boolean_condition_1, value_if_true, boolean_condition_2, value_if_true)",
+        "syntax_elements": [{
+                "element": "boolean_condition",
+                "description": "An expression or series that returns True or False values. Valid conditions for comparison include ==, !=, >, <, >=, <=."
+            },
+            {
+                "element": "value_if_true",
+                "description": "The value the function returns if condition is True."
+            }, ...
```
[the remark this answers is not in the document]
marthacryan: How does the updated docstring look? It compiled correctly with the doc script! I realized that some paths needed to be updated in the doc script!

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
marthacryan: Updated! How does that look?

mitosheet/mitosheet/tests/public/v3/sheet_functions/bool_functions/test_ifs.py
marthacryan: Added!

mitosheet/mitosheet/tests/public/v3/sheet_functions/bool_functions/test_ifs.py
```diff
@@ -0,0 +1,104 @@
...
+        pd.Series([1, pd.Timestamp('2017-01-04')])
+    ),
+    (
+        [
+            pd.Series([True, False]), pd.Series(['option1', 'option2']),
+            pd.Series([False, True]), 'option3',
+        ],
+        pd.Series(['option1', 'option3'])
+    ),
+    (
```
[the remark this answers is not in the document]
marthacryan: I added a couple test cases that I think cover what you're asking for!

mitosheet/mitosheet/public/v3/sheet_functions/README.md
```diff
@@ -0,0 +1,28 @@
+# Guide to Writing Sheet Functions
+
+## Context on Formulas
+From frontend to backend (in order), a formula will have the following order of operations:
+1. The user edits a cell to include a formula. They'll see some information in a dropdown below the cell editor that's autogenerated from a docstring in the python function for that formula. If you edit that docstring or create a new function, you'll want to run `mitosheet/docs/make_function_docs.py update_frontend` to update the information displayed in that dropdown. 
+2. Validation is run on the frontend to check for things like the validity of the columns used as input and whether the formula exists.
```
naterush, line 6: This happens on the backend, in the parser!

mitosheet/mitosheet/public/v3/sheet_functions/README.md
```diff
@@ -0,0 +1,28 @@
...
+3. The parser takes the raw string that the user created, and converts it to python code. This is a good step for debugging! You can see what the parser is generating by looking at the code chunk, and you can see exactly what your python function is going to be getting as input.
+4. **Arguments to the formula function:** When you write a formula in the mitosheet, users can use three things as arguments: a column, a cell, or a constant value. As such, when you write a formula's arguments, you'll need to handle both the case that an argument is a series, and the case that the argument is a constant. There are helper functions to make this more manageable! Namely, the `get_series_from_primitive_or_series` function will make it easy to make sure you're exclusively dealing with series to simplify the code.
```
naterush, line 8: They can also be a range reference! E.g. if someone references `A1:A4` - in your case, you didn't really need to handle these. 

RollingRange is the name of the object the parser generates and passes in this case, in the `rolling_range.py` file.

mitosheet/mitosheet/public/v3/sheet_functions/README.md
```diff
@@ -0,0 +1,28 @@
...
+    error_modal=False
+)
+```
+6. **Return values for formula functions:** As is explained above ^^, a formula can be applied to an entire column or to a single cell, so you're going to want to return the same type as you're receiving as input. When you get a series, return a series with the same index. When you get a constant, you're returning a constant. 
+
+## Testing 
+* Test the valid cases *and* the invalid cases. If you're using `MitoError`, test for the content of the error!
+* Add tests for different variations of series and constants. If you use a constant as input, are you getting a constant as output? Is that true of series? 
+* Add tests for different data types - especially if you're not casting the types for your function (see #4 above). 
+* Add tests for unusual amounts of arguments. Do you have a limited number of arguments? Should it throw an error if so?
```
naterush, line 28: These look good!

mitosheet/mitosheet/public/v3/sheet_functions/bool_functions.py
```diff
@@ -123,6 +124,66 @@ def IF(condition: pd.Series, true_series: AnyPrimitiveOrSeriesInputType, false_s
...
+        for index, arg in enumerate(argv):
+            if index % 2 == 0:
+                if arg == True:
+                    return argv[index+1]
+        return None
+
+    else:
+        # Otherwise, we have at least one series -- so we can go through and turn all of the constants into series.
+        argv_series = tuple([get_series_from_primitive_or_series(arg, base_index) for arg in argv])
+        results = pd.Series(index=base_index, dtype='object')
```
naterush, line 168: What if you made the dtype the same dtype as the first results series? I think this will fix your issues with dtypes...

naterush (changes requested):
This looks pretty good! I think this PR is ready to merge once CI is passing -- 

In terms of the writing a formula guide, I think adding an actual "how to write" section would be helpful. Mostly thinking with a) start with tests, b) write the formula, c) write more tests, d) write and generate the doc string, and e) then test it in the sheet.

Once these are good, LGTM!

mitosheet/mitosheet/public/v3/sheet_functions/README.md
marthacryan: Ah oops

PR #893, opened 2023-08-21
Dogfood 8 21

aarondr77:
# Description

- [x] Adds a CTA to upgrade to pro/enterprise if you are exporting formatting 
- [x] Use the correct default precision depending on whether the number is a float or int
- [x] Only use `()` around a number in accounting mode if the number is negative
- [x] When we use accounting, currency, or percent mode, we now use commas to separate numbers in the excel file like we do in the mitosheet. ie: `$1,000` instead of `1,000`
- [x] Adds a helpful loading message when the exported excel file is getting generated

# Testing

See the updated tests + add export to an Excel file 

# Documentation

Nope


mitosheet/mitosheet/api/api.py
```diff
@@ -48,7 +48,7 @@
 # so the API calls are handled in the main thread, to make printing easy.
 # In newer versions of JupyterLab, to see these print statements:
 # View > Show Log Console > in the console set Log Level to Debug
-THREADED = True
```
marthacryan, line 51: Should you set this back to True?

mitosheet/mitosheet/streamlit/streamlit.py
```diff
@@ -7,5 +7,5 @@
 st.set_page_config(layout="wide")
 
 from mitosheet.streamlit.v1 import spreadsheet
-new_dfs, code = spreadsheet('https://raw.githubusercontent.com/plotly/datasets/master/tesla-stock-price.csv')
```
marthacryan, line 10: Is this meant for this PR?

mitosheet/src/mito/components/elements/Toggle.tsx
```diff
@@ -47,9 +52,14 @@ const Toggle = (props: ToggleProps): JSX.Element => {
 
     const disabled = props.disabled === true
     const className = props.className !== undefined ? props.className : ''
+    const title = props.title !== undefined ? props.title : undefined
```
marthacryan, line 55: I think this is functionally equivalent to just using `props.title` without the ternary operator

marthacryan (changes requested):
Other than the small debugging line to change and the line breaking at a weird place this all looks really good! Great catches!

mitosheet/mitosheet/tests/step_performers/test_export_to_file.py
```diff
@@ -490,7 +523,7 @@ def test_transpiled_with_export_to_xlsx_format_rows_no_header():
         '#b9abff',
         'B2',
         'C2',
-        True
+        False
```
marthacryan, line 526: I think this might be breaking the CI - for some reason when I run this test locally, the numpy import isn't added, but on the CI it is?

PR #894, opened 2023-08-22
mitosheet_private: create new mitosheet_private package

aarondr77:
# Description

Creates a new mitosheet-private python package that is just a wrapper around mitosheet. When mitohseet_private is installed, the mitosheet package is uses Pro. 
 
# Testing

- Create an old mitosheet-private deployment. Then move to this and make sure the analysis still works. 

# Documentation


mitosheet-helpers/mitosheet_private/setup.cfg
```diff
@@ -0,0 +1,7 @@
+[metadata]
```
naterush, line 1: I think we can just use a setup.cfg, a no setup.py. Setup.py is actually in the process of being deprecated (and it's more confusing than a setup.cfg anyways). 

Wanna move over to setup.cfg?

mitosheet/mitosheet/user/utils.py
```diff
@@ -32,6 +32,12 @@
 except ImportError:
     MITOSHEET_HELPER_ENTERPRISE = False
 
+try:
```
naterush, line 35: I think there might be other checks of mitosheet private. I'll investigate now...

naterush, line 35: I think there might be other checks of mitosheet private. I'll investigate now...

mitosheet-helpers/mitosheet_private/setup.cfg
naterush: You probably can post the setup.py in chat gpt and it'll likely nail it :)

naterush (commented):
Nice! Glad you got to setup the package structure. Try moving it to `setup.cfg` for future proofing and simplicity, and then ping me and I can merge.

mitosheet/mitosheet/user/utils.py
aarondr77: Don't see anything else

PR #896, opened 2023-08-22
Add excel export to website

marthacryan:
Adds excel exporting to https://www.trymito.io/spreadsheet-automation as a card.


trymito.io/pages/spreadsheet-automation.tsx
```diff
@@ -97,6 +97,28 @@ const SpreadsheetAutomation: NextPage = () => {
                 <Image src={'/explore_data_visually.png'} alt='Explore data with Mito' width={500} height={250} layout='responsive'/>
               </div>
             </div>
+            <div className={pageStyles.subsection}>
+              <div className={classNames(textImageSplitStyles.functionality_media, 'display-desktop-only-inline-block')}>
+                <Image src={'/export_to_excel.png'} alt='Mito generated code' width={500} height={250} layout='responsive'/>
+              </div>
+              <div className={textImageSplitStyles.functionality_text}>
+                <h1>
+                  Compatible with <span className='text-highlight'>Excel</span>
```
aarondr77, line 106: I think using "Compatible" sells ourself a bit short. Everything is compatible with Excel in some form! 

Our unique value is that we can generate Excel files. So maybe we could go with something like:

1. Create presentation-ready Excel files 
2. Generate formatted Excel files

trymito.io/pages/spreadsheet-automation.tsx
```diff
@@ -97,6 +97,28 @@ const SpreadsheetAutomation: NextPage = () => {
...
+                </h1>
+                <p>
+                  Easily create Excel files with Mito.
```
aarondr77, line 109: I think this is a good point. But maybe the one missing thing is just "why do I want to generate Excel files with Mito". So maybe we could update this to:

Ideation of these two paragraphs:

p1: Apply conditional formatting, set table colors, and format numbers using Mito's. Don't waste time trying to understand the xlsxwriter API -- it's confusing even to engineers! 
p2: Learn more about generating presentation-ready Excel files.

aarondr77 (commented):
Yay! Nice first contribution to the website! Code looks perfect. It's nice that all of the layout is taken care of and you can just copy and paste the other section. 

Just a few wording ideations -- feel free to use it as inspiration if you have other ideas! 

Screenshot 2023-08-22 at 2 47 54 PM

This image is a little bit crunched -- I think since we're trying to show off the formatting in particular, we could show less data. ie: a view like this: 

Screenshot 2023-08-22 at 2 51 26 PM

PR #898, opened 2023-08-25
Add support for TRUE and true as constants

marthacryan:
Adds ability to use TRUE and true in formulas in the same way that you can currently use True.

### DIFF
--- a/mitosheet/mitosheet/parser.py
+++ b/mitosheet/mitosheet/parser.py
@@ -832,7 +832,6 @@ def replace_true_values(
         else:
             return match.group()
     ret = re.sub(r'TRUE|true', replace_true_values_internal, formula)
-    print(ret)
     return ret
     
 
--- a/mitosheet/mitosheet/parser.py
+++ b/mitosheet/mitosheet/parser.py
@@ -817,6 +817,25 @@ def replace_newlines_and_tabs(
     return re.sub(r'\n|\t', replace_newlines_and_tabs_internal, formula)
     
 
+    
+def replace_true_values(
+        formula: str,
+    ) -> str:
+    """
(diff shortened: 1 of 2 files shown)


mitosheet/mitosheet/tests/test_parse.py
```diff
@@ -76,6 +76,24 @@ def get_string_data_for_df(columns: List[Any], length: int) -> Dict[Any, Any]:
...
+    ),
+    (
+        '=true',
+        'B',
+        0,
+        pd.DataFrame(get_string_data_for_df(['B'], 2)),
+        'df[\'B\'] = True',
+        set([]),
+        set([])
+    ),
```
naterush, line 96: One additional test to see how it performs if TRUE is a column header, then LGTM!

mitosheet/mitosheet/parser.py
```diff
@@ -817,6 +817,25 @@ def replace_newlines_and_tabs_internal(match):
...
+    """
+    string_matches = get_string_matches(formula)
+
+    def replace_true_values_internal(match):
+        if not match_covered_by_matches(string_matches, (match.start(), match.end())):
+            return 'True'
+        else:
+            return match.group()
+    ret = re.sub(r'TRUE|true', replace_true_values_internal, formula)
+    print(ret)
```
naterush, line 835: Extra bring

mitosheet/mitosheet/tests/test_parse.py
naterush: Also the column header true. I'm fine if we detect this as the column header or as the constant -- let's just be consistent.

naterush (commented):
Two additional tests then LGTM!

PR #899, opened 2023-08-27
Streamlit improvements

naterush:
# Description

I spent a large portion of the last two weeks making Streamlit dashboards. There were a few things that I really wanted, based on my experience building Streamlit dashboards for automation use cases. 

To make these changes, I added the `return_type` parameter to the spreadsheet component. I'm not sure this is the best way to support multiple return types -- thoughts on this approach?

### Code Configuration

I am constantly looking to get a function back from the mitosheet, so I can run it on new data. But I don't know exactly what file names / paths the user is going to be importing. So I made it possible to generate params for any functions of a specific type -- nice!

### Returning the current selection 

Another thing I really wanted to do was have some part of the streamlit application react to what was selected in the mitosheet. This is ... now possible. Just use the `return_type='selection'` -- very nice!

# Testing

1. Test code configuration.
2. Test all the new return types. Maybe add a 

# Documentation

Nothing yet.


mitosheet/src/streamlit/MitoStreamlitWrapper.tsx
```diff
@@ -160,10 +169,64 @@ class MitoStreamlitWrapper extends StreamlitComponentBase<State> {
...
+
+            // Set a timer to process the next message after a delay
+            this.processMessageQueueTimer = setTimeout(this.processSetComponentValueQueue, DELAY_BETWEEN_SET_COMPONENT_VALUES);
+        } else {
+            // Otherwise, we have processed the full queue
+            this.setState({ isSettingComponentValue: false });
+        }
+    };
+
+    setComponentValue = (value: any) => {
```
naterush, line 206: One thing I'm wondering -- can we drop some of the selections? I think yes. I'm going to get rid of all this debouncing infrastructure.

naterush (commented):
Cleanup

mitosheet/mitosheet/transpiler/transpile_utils.py
```diff
@@ -225,6 +256,9 @@ def convert_script_to_function(
     final_code += imports
     final_code.append("")
 
+    # Get the final function param
+    function_params = get_final_function_params_from_code_options(steps_manager, function_params)
```
aarondr77, line 260: It's hard to tell from looking at this what the function does, especially given the return is called `function_params` but it's also one of the params.

aarondr77 (commented):
A few small cleanups + some thoughts on a different sensible API. Your call then LGTM

mitosheet/mitosheet/streamlit/v1/spreadsheet.py
```diff
@@ -218,21 +289,43 @@ def spreadsheet( # type: ignore
...
+            return ordered_dict, code
+        elif return_type == 'selection':
+            return get_selected_element(final_state.dfs, selection)
+        elif return_type == 'default_list':
+            return final_state.dfs, code
+        elif return_type == 'dfs':
+            return ordered_dict
+        elif return_type == 'code':
+            return code
+        elif return_type == 'dfs_list':
```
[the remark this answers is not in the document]
naterush: We can just make it so you have to manually turn on the selection (or it adds it in later) if you call the `get_selection). Leaving the interface as is for now.

PR #901, opened 2023-09-05
Add support for Azure OpenAI, Palm, Replicate, Sagemaker (100+LLMs) - using litellm 

ishaan-jaff:
# Description

This PR adds support for the above mentioned LLMs using LiteLLM https://github.com/BerriAI/litellm/ 
Example
```python
from litellm import completion

## set ENV variables
os.environ["OPENAI_API_KEY"] = "openai key"
os.environ["COHERE_API_KEY"] = "cohere key"

messages = [{ "content": "Hello, how are you?","role": "user"}]

# openai call
response = completion(model="gpt-3.5-turbo", messages=messages)

# cohere call
response = completion("command-nightly", messages)

# anthropic call
response = completion(model="claude-instant-1", messages=messages)
```

# Testing

# Documentation

### DIFF
--- a/mitosheet/mitosheet/api/get_ai_completion.py
+++ b/mitosheet/mitosheet/api/get_ai_completion.py
@@ -8,6 +8,7 @@ import os
 from typing import Any, Dict, List, Optional, Tuple
 
 import requests  # type: ignore
+from litellm import completion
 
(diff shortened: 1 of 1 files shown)


naterush:
@ishaan-jaff thanks for the PR -- we're not looking to take on any additional dependencies at this time, but best of luck with your project!

PR #903, opened 2023-09-05
Add support for cross-sheet references in VLOOKUP

marthacryan:
Adds the ability to reference **a range of headers** in a different sheet for *specifically* **the "where" argument in VLOOKUP**. This looks like `{SHEET}!{HEADER}:{HEADER}`. 

A side effect of adding this is **updating the arguments that `parse_formula` takes** to include all dataframes and dataframe names. 

Todo:
- [x] Prevent usage of cross-sheet references outside of this specific case
  * Add validation that helps users with the syntax. We're not allowing rollingrange so this could be confusing
- [x] Add more tests
- [x] Update all calls to `parse_formula` to reflect new arguments


mitosheet/mitosheet/parser.py
```diff
@@ -412,50 +413,73 @@ def get_raw_parser_matches(
...
+            raw_parser_matches.append({
+                'type': '{SHEET}',
+                'substring_range': (formula.index(f'{sheet_name}!'), formula.index(f'{sheet_name}!') + len(sheet_name)),
+                'unparsed': f'{sheet_name}!',
+                'parsed': sheet_name,
+                'row_offset': 0
+            })
+
+            # Update to look at the column headers in the other sheet
+            column_headers = dfs[sheet_index].columns.to_list()
```
naterush, line 481: I think this won't work if you have a formula with column headers from the original dataframe, and a cross sheet reference. It'll miss the column header references to the original dataframe (as it's no longer searching for those column headers). 

Maybe we can just add the `column_headers` to the list of column headers? And make sure to deduplicate? A bit of a hack, but seems like it would work.

mitosheet/mitosheet/tests/test_parse.py
```diff
@@ -1134,6 +1134,41 @@ def test_parse(formula, column_header, formula_label, df, python_code, functions
...
 
+# Right now, VLOOKUP is the only formula that allows cross-sheet referencing. In the future, 
+# other cross-sheet references in parser can be added here. 
+VLOOKUP_TESTS = [
+    (
+        '=VLOOKUP(A0, df_2!A:B, 2)',
+        'B',
+        0,
+        [
+            pd.DataFrame(
```
naterush, line 1145: Let's change the basic test to make sure the two dataframes don't share any columns - else we'll run into the bug we hit above.

naterush (commented):
A few minor bugs in this code, but the overall structure / shape of this code looks pretty awesome!

mitosheet/mitosheet/tests/test_parse.py
```diff
@@ -1134,6 +1134,156 @@ def test_parse(formula, column_header, formula_label, df, python_code, functions
         )
 
 
+# Right now, VLOOKUP is the only formula that allows cross-sheet referencing. In the future, 
+# other cross-sheet references in parser can be added here. 
+VLOOKUP_TESTS = [
+    (
+        '=VLOOKUP(A0, df_2!C:D, 2)',
```
naterush, line 1141: Some more parser tests to add:
1. VLOOKUP inside some other function (like a SUM or something else, with other references to the sheet).
2. Some headers in dataframes other than just strings (e.g. some numbers or booleans in the cross sheet reference.


The other major set of tests we should add: set column formula tests!

We have the ability to test the entire setting formula loop from start to finish. Notably, this sort of backend testing would mean you get to spend less time manually writing formulas in the frontend -- this is the "lead with tests" idea we've talked about before.

I'll let you write these tests yourself - many of the parser tests likely will transport over, but also there are some new, more realistic scenarios that might be nice to test!

mitosheet/mitosheet/parser.py
```diff
@@ -234,6 +236,13 @@ def check_common_errors(
             error_modal=False
         )
 
+    for sheet_name in df_names:
+        if safe_contains(formula, f'{sheet_name}!', column_headers) and not safe_contains_function(formula, 'VLOOKUP', column_headers):
+            raise make_invalid_formula_error(
+                formula,
+                f'Cross-sheet references are only allowed in calls to VLOOKUP',
```
naterush, line 243: This is a great user-facing error!

mitosheet/mitosheet/parser.py
```diff
@@ -656,48 +717,47 @@ def get_header_index_header_index_match(formula: str, raw_parser_matches: List[R
...
+        df_names,
+        sheet_index
     )
 
     match_index = 0
     while match_index < len(raw_parser_matches):
-
-        header_index_header_index_match = get_header_index_header_index_match(formula, raw_parser_matches, match_index)
-        if header_index_header_index_match is not None:
+        if (header_index_header_index_match := get_header_index_header_index_match(formula, raw_parser_matches, match_index)) is not None:
```
naterush, line 727: The := syntax doesn't work on Python 3.6 -- you'll sadly have to define these on a separate line </3. This might be why CI is failing.

mitosheet/mitosheet/parser.py
```diff
@@ -656,48 +717,47 @@ def get_header_index_header_index_match(formula: str, raw_parser_matches: List[R
...
-        header_match = get_header_match(formula, raw_parser_matches, match_index)
-        if header_match is not None:
+        elif (header_match := get_header_match(formula, raw_parser_matches, match_index)) is not None:
             parser_matches.append(header_match)
             match_index += 1
-            continue
+        else:
+            raise make_invalid_formula_error(
+                formula,
+                'Something went wrong while parsing the formula. Please contact support.',
```
naterush, line 745: I don't think we need to tell them to contact support. Let's just tell them to edit the formula and try again -- sometimes users write wrong formulas, and then fix them!

mitosheet/mitosheet/parser.py
```diff
@@ -767,6 +832,17 @@ def get_header_header_selection_code(column_header_one: ColumnHeader, column_hea
             transpiled_column_header = column_header_to_transpiled_code(column_header)
             replace_string = f'{df_name}[{transpiled_column_header}]{get_shift_string(column_dtype, row_offset)}' # type: ignore
         
+        elif match_type == '{SHEET}!{HEADER}:{HEADER}':
+            (sheet_name, column_header_one, column_header_two) = match['parsed']
+            cross_sheet_df = dfs[df_names.index(sheet_name)]
+            # Ensure that the column headers are in the order they appear in the dataframe
+            first_column_header, second_column_header = (column_header_one, column_header_two) if cross_sheet_df.columns.get_loc(column_header_one) <= cross_sheet_df.columns.get_loc(column_header_two) else (column_header_two, column_header_one)
```
naterush, line 839: If we don't have a test for this line of code (in `test_parser.py`) -- let's add one as well.

naterush (changes requested):
Overall, looking great - I'm so hype for this feature. I think the only major thing is making sure to test the full setting column formula loop in `test_set_column_formula.py` -- now that we support cross sheet references.

Let's try and get this merged before the sprint is over!

naterush (changes requested):
Good work getting this up. Requesting a few more tests, to catch the bugs caught down below.

## Possible Error Case

Screen Shot 2023-09-11 at 8 40 10 AM

Am I missing something, or is the above screenshot returning the wrong results? Shouldn't the key `3` be filled in with `b`, based on the starting dataframes?

Screen Shot 2023-09-11 at 8 41 37 AM

this looks like a VLOOKUP bug, given that this works when I add in the value 1. Let's add this as a test case to VLOOKUP directly. 

## Indexes are Broken

Screen Shot 2023-09-11 at 8 43 22 AM

naterush (changes requested):
Working great. Good cleanup, and nice tests. One minor thing -- can you remove the extra parameter (`False`) from the examples for VLOOKUP in the documentation box? 

Then, LGTM!

PR #907, opened 2023-09-07
User defined function doc string

naterush:
# Description

There are some bugs in our documentation - namely, if you take a suggestion, it won't display the documentation for that formula until you type another character. It's a race condition that's hard to fix for now.

# Testing

```
def ADD_ONE(col):
    """Adds one to a column"""
    return col + 1

mitosheet.sheet(sheet_functions=[ADD_ONE])
```
# Documentation

Update our function docs.


aarondr77:
If we look at the docstrings that we use in our spreadsheet formulas or the one that I created for my function, they are pretty long. And our cell editor does not handle them well if they are not the right structure. This looks bad. 

Maybe we could have an expand button that handles it better? 

Screenshot 2023-09-07 at 1 04 08 PM

aarondr77:
If you use the docstring format that we request and the name of the function is not the actual name of the function, we don't match on the actual name of the function!

Screenshot 2023-09-07 at 1 07 08 PM

aarondr77:
Similarly to ^ . Here when I typed `CHECK` it showed me the function name that did not actually exist. Given that we require users to have uppercase function names, I could imagine that the function param might be a proper case version of it... 
Screenshot 2023-09-07 at 1 08 26 PM

aarondr77:
We aren't using the description in the comments. We want to do that, right?

Screenshot 2023-09-07 at 1 14 22 PM

naterush:
@aarondr77 I'm explicitly making the choice to not handling badly formatting doc strings in our format, as we are the primary users of this feature and I don't think polish here will really move the needle on this feature being useful.

PR #909, opened 2023-09-08
Fix escape from cell editor bug

marthacryan:
# Description
This fixes a small bug where when the focus moves out of the cell editor, using "escape" to close the cell editor stops working. 

# Testing
1. Open the cell editor
2. Click anywhere outside of the editor
3. Press "escape" 
4. The cell editor should close

### DIFF
--- a/mitosheet/src/mito/Mito.tsx
+++ b/mitosheet/src/mito/Mito.tsx
@@ -1001,6 +1001,7 @@ export const Mito = (props: MitoProps): JSX.Element => {
             ref={mitoContainerRef} 
             tabIndex={0}
             onKeyDown={(e) => {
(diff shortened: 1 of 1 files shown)


mitosheet/src/mito/Mito.tsx
```diff
@@ -1000,6 +1000,11 @@ export const Mito = (props: MitoProps): JSX.Element => {
             data-jp-suppress-context-menu 
             ref={mitoContainerRef} 
             tabIndex={0}
+            onKeyDown={(e) => {
+                if (e.key === 'Escape') {
```
naterush, line 1004: Can you add the PR description comment here, just so we know why this is here?

PR #910, opened 2023-09-11
Add support for cross-sheet formulas on the frontend

marthacryan:
Adds the ability to select columns in different sheets **when the user is editing a VLOOKUP formula**. 

This changes the `editorState` to include `sheetIndex`.

Also updates the behavior of selecting a range of columns to allow for the user to shift and click another header to select the range. 

Todo:
- [x] When the user presses "enter" in the cell editor, it should refer to the `editorConfig.sheetIndex`. Right now it assumes you're adding this to the sheet you currently have open

### DIFF
--- a/mitosheet/src/mito/Mito.tsx
+++ b/mitosheet/src/mito/Mito.tsx
@@ -409,12 +409,6 @@ export const Mito = (props: MitoProps): JSX.Element => {
         if (source !== undefined && source === DFSource.Pivoted && uiState.currOpenTaskpane.type === TaskpaneType.NONE) {
             void openEditedPivot()
         }
-
-        // Close the cell editor if it is open
-        if (editorState !== undefined && !editorState.formula.includes('VLOOKUP')) {
-            setEditorState(undefined)
-        }
-
     }, [uiState.selectedSheetIndex])
 
     // Store the prev open taskpane in a ref, to avoid triggering rerenders
--- a/mitosheet/src/mito/Mito.tsx
+++ b/mitosheet/src/mito/Mito.tsx
@@ -411,7 +411,7 @@ export const Mito = (props: MitoProps): JSX.Element => {
         }
 
         // Close the cell editor if it is open
-        if (editorState !== undefined) {
+        if (editorState !== undefined && !editorState.formula.includes('VLOOKUP')) {
             setEditorState(undefined)
         }
 
--- a/mitosheet/src/mito/components/endo/EndoGrid.tsx
+++ b/mitosheet/src/mito/components/endo/EndoGrid.tsx
@@ -238,11 +238,17 @@ function EndoGrid(props: {
                     editorState.pendingSelections?.inputSelectionEnd,
                     selectionEnd
                 )
+
+                // If the user is holding down the shift key, we want to extend the selection
+                // rather than starting from scratch with a new selection
                 let startingColumnIndex = props.editorState?.pendingSelections?.selections[0].startingColumnIndex ?? columnIndex;
                 let endingColumnIndex = props.editorState?.pendingSelections?.selections[0].endingColumnIndex ?? columnIndex;
-                if (startingColumnIndex > columnIndex) {
+                if (e.shiftKey && startingColumnIndex > columnIndex) {
                     startingColumnIndex = columnIndex;
-                } else if (endingColumnIndex < columnIndex) {
+                } else if (e.shiftKey && endingColumnIndex < columnIndex) {
+                    endingColumnIndex = columnIndex;
+                } else if (!e.shiftKey) {
+                    startingColumnIndex = columnIndex;
                     endingColumnIndex = columnIndex;
                 }
                 const newSelection: MitoSelection[] = [{
--- a/mitosheet/src/mito/components/endo/EndoGrid.tsx
+++ b/mitosheet/src/mito/components/endo/EndoGrid.tsx
@@ -244,6 +244,7 @@ function EndoGrid(props: {
                     endingRowIndex: rowIndex !== undefined ? rowIndex : -1,
                     startingColumnIndex: columnIndex,
                     endingColumnIndex: columnIndex,
+                    sheetIndex: sheetIndex,
                 }]
 
                 // Select the column that was clicked on, as they do in Excel
@@ -306,6 +307,7 @@ function EndoGrid(props: {
                             endingRowIndex: rowIndex,
                             startingColumnIndex: columnIndex,
                             endingColumnIndex: columnIndex,
+                            sheetIndex: sheetIndex,
                         })
                         return {
                             ...gridState,
@@ -333,6 +335,7 @@ function EndoGrid(props: {
                                     endingRowIndex: rowIndex,
                                     startingColumnIndex: columnIndex,
                                     endingColumnIndex: columnIndex,
+                                    sheetIndex: sheetIndex,
                                 })
                                 return {
                                     ...gridState,
@@ -349,6 +352,7 @@ function EndoGrid(props: {
                                 endingRowIndex: rowIndex,
                                 startingColumnIndex: columnIndex,
                                 endingColumnIndex: columnIndex,
+                                sheetIndex: sheetIndex,
                             })
                             setGridState((gridState) => {
                                 return {
@@ -367,6 +371,7 @@ function EndoGrid(props: {
                                         endingRowIndex: rowIndex,
                                         startingColumnIndex: columnIndex,
                                         endingColumnIndex: columnIndex,
+                                        sheetIndex: sheetIndex,
                                     }]
                                 }
                             })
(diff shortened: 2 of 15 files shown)


mitosheet/src/mito/components/endo/celleditor/cellEditorUtils.tsx
```diff
@@ -9,26 +9,29 @@ import { getCellDataFromCellIndexes } from "../utils";
...
-export const getSelectionFormulaString = (selections: MitoSelection[], sheetData: SheetData): string => {
+export const getSelectionFormulaString = (selections: MitoSelection[], sheetData: SheetData, sheetIndex: number): string => {
     // For each of the selections, we turn them into a string that goes into the formula
     const selectionStrings: string[] = []
 
     selections.forEach(selection => {
-
+        let sheetIndexStr = '';
+        if (sheetIndex !== selection.sheetIndex) {
+            sheetIndexStr = `${sheetData.dfName}!`;
```
naterush, line 19: Can we rename this variable to `dfName`?

mitosheet/src/mito/Mito.tsx
```diff
@@ -411,7 +411,7 @@ export const Mito = (props: MitoProps): JSX.Element => {
         }
 
         // Close the cell editor if it is open
-        if (editorState !== undefined) {
+        if (editorState !== undefined && !editorState.formula.includes('VLOOKUP')) {
             setEditorState(undefined)
```
naterush, line 415: I actually think we should add the cross-sheet reference even if the user doesn't have VLOOKUP written. Namely, if they can't add cross cell references they will be very confused why. Instead, if we let them add it - and then throw that helpful error you have on the backend, they will learn why that doesn't work!

mitosheet/src/mito/components/endo/EndoGrid.tsx
```diff
@@ -238,12 +238,19 @@ function EndoGrid(props: {
...
                     selectionEnd
                 )
-
+                let startingColumnIndex = props.editorState?.pendingSelections?.selections[0].startingColumnIndex ?? columnIndex;
+                let endingColumnIndex = props.editorState?.pendingSelections?.selections[0].endingColumnIndex ?? columnIndex;
+                if (startingColumnIndex > columnIndex) {
+                    startingColumnIndex = columnIndex;
+                } else if (endingColumnIndex < columnIndex) {
+                    endingColumnIndex = columnIndex;
+                }
```
naterush, line 247: I'm confused by what this work is here. Can you document the why?

mitosheet/src/mito/components/endo/FormulaBar.tsx
```diff
@@ -35,10 +35,10 @@ const FormulaBar = (props: {
 
     const rowIndex = props.selection.startingRowIndex
     const colIndex = props.selection.startingColumnIndex
-
-    const {columnHeader, columnFormula, cellValue, columnFormulaLocation} = getCellDataFromCellIndexes(props.sheetData, rowIndex, colIndex);
+    const sheetData = props.sheetDataArray[props.sheetIndex];
```
naterush, line 38: Shouldn't we be using the editorState.sheetIndex here? I think we actually probably are -- but in that case, let's remove the sheetIndex prop from the FormulaBar and just use the editorState.sheetIndex everywhere in this file for clarify. Otherwise, we have two variables confusingly referring to the same thing.

mitosheet/src/mito/components/endo/celleditor/cellEditorUtils.tsx
```diff
@@ -41,20 +44,15 @@ export const getSelectionFormulaString = (selections: MitoSelection[], sheetData
...
 */
 export const getFullFormula = (
-    formula: string, 
-    pendingSelections: {
-        selections: MitoSelection[],
-        inputSelectionStart: number,
-        inputSelectionEnd: number,
-    } | undefined,
+    editorState: EditorState,
     sheetData: SheetData,
```
naterush, line 48: Ok -- this function call gets really confusing. previously, sheetData was the sheet being edited. Now it's the sheet that happens to be selected at the time of calling this function - which is a weird and hard to express invariant that isn't super clear. 

My feeling: we should make `getFullFormula` take the sheetDataArray, and then pass that down to the `getSelectionFormulaString`. Then, we can get the dfName in this case more clearly -- and it's really obvious what this function is doing / requires. 

It removes the need to remember _which_ sheetData you need to pass to `getFullFormula` -- is it the sheetData of the formula, or the sheetData of the currently selected sheet? I'd rather not think about it.

mitosheet/src/mito/components/endo/celleditor/CellEditor.tsx
```diff
@@ -485,10 +494,11 @@ const CellEditor = (props: {
...
         }
 
-        const columnID = props.sheetData.data[props.editorState.columnIndex].columnID;
-        const columnHeader = props.sheetData.data[props.editorState.columnIndex].columnHeader;
-        const formula = getFullFormula(props.editorState.formula, props.editorState.pendingSelections, props.sheetData)
-        const formulaLabel = props.sheetData.index[props.editorState.rowIndex];
+        const editorSheetData = props.sheetDataArray[props.editorState.sheetIndex];
+        const columnID = editorSheetData.data[props.editorState.columnIndex].columnID;
+        const columnHeader = editorSheetData.data[props.editorState.columnIndex].columnHeader;
+        const formula = getFullFormula(props.editorState, sheetData)
```
naterush, line 500: Here's a great location where I'd rather not have to think about -- is this sheetData of selected sheet or sheet data of editor sheet?

mitosheet/src/mito/components/endo/celleditor/CellEditor.tsx
```diff
@@ -477,6 +477,15 @@ const CellEditor = (props: {
         // Don't refresh the page
         e.preventDefault();
 
+        if (props.sheetIndex !== props.editorState.sheetIndex) {
```
naterush, line 480: Add a little doc comment explaining why we switch back

mitosheet/src/mito/components/endo/celleditor/CellEditor.tsx
```diff
@@ -105,7 +106,7 @@ const CellEditor = (props: {
             // at the _end_ of them!
             if (props.editorState.pendingSelections !== undefined) {
                 // TODO: use the correct funciton, rather than JSON.stringify
```
naterush, line 108: I think this comment is out of date and can be removed.

naterush (changes requested):
Functionality is working great! Reqesting a bit of interface cleanup + clarification, but this is such an awesome feature. 

+ 10 that Shift + Click is going to make this way better -- as of now, it's just a bit annoying :) 

Good work!

mitosheet/src/mito/components/endo/celleditor/cellEditorUtils.tsx
```diff
@@ -9,26 +9,33 @@ import { getCellDataFromCellIndexes } from "../utils";
 import { CELL_EDITOR_DEFAULT_WIDTH, CELL_EDITOR_MAX_WIDTH } from "./CellEditor";
 
 
-export const getSelectionFormulaString = (selections: MitoSelection[], sheetData: SheetData): string => {
+export const getSelectionFormulaString = (selections: MitoSelection[], sheetData: SheetData, sheetIndex: number): string => {
```
naterush, line 12: This interface still confuses me. Is this the sheetIndex of the selectedSheet? Or the sheetIndex of the edited sheet? Can we rename to be more clear?

mitosheet/src/mito/components/endo/EndoGrid.tsx
```diff
@@ -239,11 +239,24 @@ function EndoGrid(props: {
                     selectionEnd
                 )
 
+                // If the user is holding down the shift key, we want to extend the selection
+                // rather than starting from scratch with a new selection
+                let startingColumnIndex = props.editorState?.pendingSelections?.selections[0].startingColumnIndex ?? columnIndex;
+                let endingColumnIndex = props.editorState?.pendingSelections?.selections[0].endingColumnIndex ?? columnIndex;
+                if (e.shiftKey && startingColumnIndex > columnIndex) {
```
naterush, line 246: Nice catch!

naterush (changes requested):
This works great! Nice, sweet, totally cool changes that add a very cool feature. We're so close to the finish line here!

There's one minor bit of polish that I think would be nice to add. I think it's going to be super common to write a formula of the following type:
```
VLOOKUP(column, other_df_name!other_column0:other_other_column0, 2)
```

Where you get `other_column0` by clicking on a cell in the sheet (e.g. the column header is `other_column` and the index is `0`). 

In this case, the parser doesn't throw a particularly helpful error. I don't think this would be to hard to add - as it's just two cases (you hit an index on the first column after a sheet, or a second column after a sheet). 

Can we add a better error message for these two cases, as well as a test for each of em? Then, LGTM! Woo woo! Great work getting this out.

PR #914, opened 2023-09-14
Add basic search/highlight functionality

marthacryan:
Adds a search bar that highlights all matches in cells and column headers to the search value. You can open this either through View > Search (maybe I should move this but not sure where), through the toolbar button, or through cmd+F.

Updated view menu:
image

Added toolbar button:
image

Blank search bar:
image

Search bar w/ highlighting:
image

Todo:
- [x] Fix cmd+F (for some reason it's not working!)
- [x] highlight all occurrences of the string rather than just the first occurrence (maybe optional?)
- [x] Add search to column headers
- [x] Grab focus when the search bar is opened


aarondr77:
Checkout this video. Searching in Mito doesn't handle number formats the same way that Excel does. 

In future PR's the most important related thing is going to be making sure that Mito highlights the same values that are actually getting replaced! Since the values in the backend are not formatted, ie: `100,000` that is displayed in the frontend is actually just 100000, it's probably easiest to require the user to search for unformatted values, like Excel. 

What do you think? 

https://github.com/mito-ds/mito/assets/18709905/af937683-9722-4766-a4e6-bc9ca2bae28e

aarondr77:
Style nitpick, but maybe we could get rid of this dark left border? We don't use that anywhere else in the app so it kinda sticks out

Screenshot 2023-09-15 at 1 50 41 PM

aarondr77:
I like this color on hover!

Screenshot 2023-09-15 at 1 51 54 PM

aarondr77:
Ooo also re adding the button to the toolbar, let's make sure that it works on Streamlit in narrow mode too cause I think the toolbar is getting really full

mitosheet/css/elements/SearchBar.css
```diff
@@ -0,0 +1,21 @@
...
+    display: flex;
+    align-items: center;
+    margin-left: 5px;
+}
+
+.mito-close-search-button:hover {
+    background-color: var(--mito-highlight-light);
+}
+
+.mito-search-bar {
```
naterush, line 13: Move class up to top as it's the top thing...

mitosheet/css/elements/SearchBar.css
```diff
@@ -0,0 +1,21 @@
...
+    position: absolute;
+    top: 72px;
+    right: 0;
+    background-color: var(--mito-background-highlight);
+    box-shadow: 1px 4px 6px var(--mito-light-gray);
```
naterush, line 18: Is there a parametric color variable you can use for the shadow? E.g. other than -light-gray, which might change per theme?

mitosheet/src/mito/Mito.tsx
```diff
@@ -1058,6 +1073,12 @@ export const Mito = (props: MitoProps): JSX.Element => {
                             {getCurrOpenTaskpane()}
                         </div>
                     }
+                    {uiState.currOpenSearch.isOpen &&
```
naterush, line 1076: Very nice and pleasant, good organization to add to the `uiState` type.

mitosheet/src/mito/components/endo/ColumnHeader.tsx
```diff
@@ -171,7 +172,8 @@ const ColumnHeader = (props: {
                 'endo-column-header-text',
                 {
                     'endo-column-header-container-selected': selected,
-                    'recon': isColumnCreated || isColumnRenamed
+                    'recon': isColumnCreated || isColumnRenamed,
+                    'mito-search-match': matchesSearch,
```
naterush, line 176: I don't see this class anywhere. Forget to commit it?

mitosheet/src/mito/components/endo/ColumnHeader.tsx
```diff
@@ -82,7 +82,8 @@ const ColumnHeader = (props: {
     // Get the pieces of the column header. If the column header is not a MultiIndex header, then
     // lowerLevelColumnHeaders will be an empty array
     const { lowerLevelColumnHeaders, finalColumnHeader } = getColumnHeaderParts(columnHeader);
-    const borderStyle = getBorderStyle(props.gridState.selections, props.gridState.copiedSelections, -1, props.columnIndex, props.sheetData.numRows, props.uiState.highlightedColumnIndex);
+    const matchesSearch = !!props.uiState.currOpenSearch.searchValue && (finalColumnHeader + '').toLowerCase().includes(props.uiState.currOpenSearch.searchValue.toLowerCase());
```
naterush, line 85: Wanna factor all of this logic into a function, and share it between column header and grid data?

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1094,6 +1094,32 @@ export const createActions = (
...
         },
+        [ActionEnum.OpenSearch]: {
+            type: ActionEnum.OpenSearch,
+            shortTitle: 'Search',
+            longTitle: 'Search',
+            actionFunction: () => {
+                // We turn off editing mode, if it is on
+                setEditorState(undefined);
+                if (uiState.currOpenSearch.isOpen) {
+                    document.getElementById('mito-search-bar-input')?.focus();
```
naterush, line 1105: Almost. The problem is that if there are two mito sheets, this won't get the right one in every case. 

You can use the `mitoContainerRef` that is passed to this function, and then search inside that though - and this will work!

mitosheet/src/plugin.tsx
```diff
@@ -319,22 +319,22 @@ function activateMitosheetExtension(
...
     */
     app.commands.addKeyBinding({
-        command: 'mitosheet:focus-on-search',
+        command: 'mitosheet:open-search',
         args: {},
         keys: ['Accel F'],
         selector: '.mito-container'
     });
-    app.commands.addCommand('mitosheet:focus-on-search', {
+    app.commands.addCommand('mitosheet:open-search', {
```
naterush, line 327: Where does this command get called?

mitosheet/src/mito/components/toolbar/Toolbar.tsx
```diff
@@ -408,6 +408,11 @@ const Toolbar = (
 
                     <div className="toolbar-vertical-line"></div>
 
+                    <ToolbarButton
+                        id='mito-open-search' // NOTE: this is used to click the open search button in plugin.tsx
```
naterush, line 412: Wanna define it as a constant, and then import it where you need it?

mitosheet/src/mito/components/endo/ColumnHeader.tsx
marthacryan: Oh yeah I think that's from exploring the "jump to next match" functionality - that can be removed from this PR

marthacryan: Although maybe it'd be good to have a special background color or something? Not sure how prominent we want the highlighting to be

mitosheet/src/plugin.tsx
marthacryan: Directly above this code ^^ `app.commands.addKeyBinding({ command: 'mitosheet.open-search' ... })`

marthacryan:
> ## The Search Icon Doesn't Respect Streamlit Colors

Fixed

> ## Float search is a bit buggy

Might need to discuss this a little more

> ## Sizing issues
> If you make the mitosheet small, icons get overlapped - and you can't click the graph button. Your call on what to do here, but valid options include: shrinking icons more when it gets small, hiding the search icon... any ideas are valid. Feel free to make a call

The plan is to add an extra dropdown with overflow buttons when the viewport is narrow, in a separate PR.

mitosheet/src/mito/components/endo/utils.tsx
```diff
@@ -77,6 +77,21 @@ export const getTypeIdentifier = (columnDtype: string): string => {
...
+/**
+ * A helper function to check if a cell or column header matches a search value
+ * @param value - The value that you want to check against. This can be a cell value or a column header
+ * @param searchValue - The search value that you want to match
+ * @returns Boolean indicating if the value matches the search value
+ */
+export const checkMatchesSearch = (value: any, searchValue?: string) => {
+    if (searchValue === undefined || searchValue === '' || value === undefined) return false;
+    // Trailing spaces are ignored
+    searchValue = searchValue?.toLocaleLowerCase().trim();
```
naterush, line 89: what is toLocaleLowerCase?

mitosheet/src/mito/components/toolbar/Toolbar.tsx
```diff
@@ -26,7 +26,11 @@ import ToolbarViewDropdown from './ToolbarViewDropdown';
 import { ToolbarButtonType } from './utils';
 import ToolbarCodeDropdown from './ToolbarCodeDropdown';
 
-const Toolbar = (
+export const MITO_TOOLBAR_OPEN_SEARCH_ID = 'mito-open-search';
+export const MITO_TOOLBAR_UNDO_ID = 'mito-undo-button';
+export const MITO_TOOLBAR_REDO_ID = 'mito-redo-button';
```
naterush, line 31: Nice cleanup!

mitosheet/src/mito/components/endo/GridData.tsx
```diff
@@ -71,6 +71,7 @@ const GridData = (props: {
 
                             const isColumnCreated = reconIsColumnCreated(columnHeader, props.uiState.dataRecon, sheetData)
                             const isColumnModified = reconIsColumnModified(columnHeader, props.uiState.dataRecon, sheetData)
+                            const matchesSearch = checkMatchesSearch(cellData.toString(), props.uiState.currOpenSearch.searchValue);
```
naterush, line 74: If we always cast the cellData to a string, wanna type the function as such?

naterush (changes requested):
Working so great! Super minor cleanup then LGTM :)

mitosheet/src/mito/components/endo/utils.tsx
marthacryan: It's supposed to be more friendly with other languages: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase. It'll output the same thing as `toLowerCase` most of the time

PR #918, opened 2023-09-15
Streamlit import improvements

naterush:
# Description

Makes it so:
1. the `function` return type for streamlit supports mito defined functions
2. The `function` return type of streamlit can optional include the imports for custom defined functions. we don't want this on by default for JLab, but we might for Streamlit.

# Testing

Use it. 

# Documentation

No.


mitosheet/mitosheet/transpiler/transpile.py
```diff
@@ -87,10 +88,18 @@ def transpile(
     # We then deduplicate the imports, keeping the same order as originally
     final_imports_code = deduplicate_array(imports_code)
 
+    # If the user wants the custom imports to be included in the script, we include them here
+    if steps_manager.code_options['import_custom_python_code']:
```
aarondr77, line 92: Currently, we don't create the spreadsheet widget, and our tests are failing, if `code_options` is not none, but you don't set `import_custom_python_code`. 

I think since there are so many code options at this point, we should set default values for them and render the sheet. 

The most helpful default here seems like "If you are returning a function or the function code, then `import_custom_python_code` should be `True`.

aarondr77 (commented):
An idea for improving this interface, then LGTM.

mitosheet/mitosheet/transpiler/transpile.py
naterush: Passing this information down is a bit of a hassle, and gets a bit confusing with the default. Keeping the minor refactor of just defaulting to False everywhere for now.

PR #922, opened 2023-09-20
Add ability to jump to the next match

marthacryan:
Adds the ability to jump to the next match to your search in the spreadsheet. 

Todo:
- [x] Add support for column headers
- [x] When you reset the field to empty string, the matches don't reset
- [x] Add arrows for next/previous match
- [x] Shift+enter doesn't work
- [x] Code cleanup and comments
- [x] backend tests

https://github.com/mito-ds/mito/assets/6673460/28b4a4d6-8979-4d28-8c4f-0b941cf16f1f


mitosheet/css/elements/SearchBar.css
```diff
@@ -11,12 +11,12 @@
 
 .mito-search-bar span {
     margin-left: 5px;
-    min-width: 9ch;
+    min-width: 5ch;
```
naterush, line 14: we usually use pixels - I'm fine with this

mitosheet/mitosheet/api/get_search_matches.py
```diff
@@ -0,0 +1,44 @@
...
+
+# Copyright (c) Saga Inc.
+# Distributed under the terms of the GPL License.
+import re
+import pandas as pd
+from typing import Any, Dict
+from mitosheet.types import StepsManagerType
+
+
+def get_search_matches(params: Dict[str, Any], steps_manager: StepsManagerType) -> str:
```
naterush: This type is off, methinks. Doesn't return a string

mitosheet/mitosheet/api/get_search_matches.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # Then, sum the matches:
+    total_number_matches = matches.sum()
+
+    # Then, add the column names to the matches:
+    total_number_matches += len([col for col in df.columns if re.search(search_regex,str(col)) is not None])
+
+    # Find the indices of cells containing the search value
+    # Only search the first 1500 rows because the editor only shows the first 1500 rows. 
+    cell_matches = [{'row': i, 'col': j} for i in range(min(1500,len(df.index))) for j in range(len(df.columns)) if (re.search(search_regex,str(df.iloc[i, j])) is not None)]
```
naterush, line 37: I think we should call these `rowIndex` and `colIndex` to be consistent with the rest of our codebase. Otherwise, it's not clear if it's the header or the index or what.

mitosheet/mitosheet/api/get_search_matches.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # Find the indices of columns containing the search value
+    column_matches = [{'row': -1, 'col': j} for j in range(len(df.columns)) if (re.search(re.compile(search_value, re.IGNORECASE),str(df.columns[j])) is not None)]
```
naterush, line 40:
```suggestion
    column_matches = [{'row': -1, 'col': j} for j, column in enumerate(df.columns) if (re.search(re.compile(search_value, re.IGNORECASE),str(column)) is not None)]
```

mitosheet/mitosheet/api/get_search_matches.py
naterush: I think `.iat` might be faster -- see here: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.iat.html#pandas.DataFrame.iat

also, if you save the num_rows and num_columns you can avoid the len calculations running every loop. For large datasets, this is a significant amount of computation - so let's break them out of the list comprehension and save them above

mitosheet/src/mito/Mito.tsx
```diff
@@ -1073,13 +1076,6 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
                         </div>
                     }
-                    {uiState.currOpenSearch.isOpen &&
-                        <SearchBar
-                            uiState={uiState}
-                            setUIState={setUIState}
-                            mitoAPI={mitoAPI}
-                        />
-                    }
                 </div>
```
naterush, line 1079: Hmm, where do we display it?

mitosheet/src/mito/api/api.tsx
```diff
@@ -254,10 +259,10 @@ export class MitoAPI {
     /*
         Returns a string encoding of the CSV file to download
     */
-    async getTotalNumberMatches(sheetIndex: number, searchValue: string): Promise<MitoAPIResult<string>> {
-        return await this.send<string>({
+    async getTotalNumberMatches(sheetIndex: number, searchValue: string): Promise<MitoAPIResult<SearchResults>> {
```
naterush, line 262: `getSearchMatches`

mitosheet/src/mito/components/endo/GridData.tsx
```diff
@@ -71,8 +70,15 @@ const GridData = (props: {
 
                             const isColumnCreated = reconIsColumnCreated(columnHeader, props.uiState.dataRecon, sheetData)
                             const isColumnModified = reconIsColumnModified(columnHeader, props.uiState.dataRecon, sheetData)
-                            const matchesSearch = checkMatchesSearch(cellData.toString(), props.uiState.currOpenSearch.searchValue);
 
+                            // Check if the cell is a search match. 
+                            const matchesSearch = !!props.uiState.currOpenSearch.matches.find((value) => {
```
naterush, line 75: Oh nice! This allows us to reuse the logic, so we can be sure backend is the single source of truth

mitosheet/src/mito/components/endo/ColumnHeader.tsx
```diff
@@ -82,7 +82,15 @@ const ColumnHeader = (props: {
     // Get the pieces of the column header. If the column header is not a MultiIndex header, then
     // lowerLevelColumnHeaders will be an empty array
     const { lowerLevelColumnHeaders, finalColumnHeader } = getColumnHeaderParts(columnHeader);
-    const matchesSearch = checkMatchesSearch(finalColumnHeader + '', props.uiState.currOpenSearch.searchValue);
+
+    // Check if the column header is a match of the search
+    const matchesSearch = props.uiState.currOpenSearch.matches.find((value) => {
```
naterush, line 87: I am a little worried about the performance here on large datasets. Can we test with 100+ columns with 1500 rows, searching for a common value?

mitosheet/src/mito/components/SearchBar.tsx
```diff
@@ -10,71 +10,225 @@ import Input from './elements/Input';
...
             }
-            const total = response.result;
-            if (total !== undefined && !isNaN(Number(total))) {
-                setTotalMatches(Number(total));
+            const new_total_number_matches = response.result.total_number_matches;
+            const new_matches = response.result.matches;
+            // Update the total matches. 
+            if (!isNaN(Number(new_total_number_matches))) {
+                setTotalMatches(Number(new_total_number_matches));
+            }
```
naterush, line 83: Why do you need to do this casting? I feel like if you get a number back, it should always be a number. I don't see this in the type system.

mitosheet/src/mito/components/SearchBar.tsx
```diff
@@ -10,71 +10,225 @@ import Input from './elements/Input';
...
+
+            if (new_matches?.length > 0) {
+                scrollMatchIntoView(new_matches[currentMatchIndex]);
             }
```
naterush, line 87: Why do we want to scroll the currentMatchIndex? Wouldn't we want to reset that?

mitosheet/css/elements/SearchBar.css
```diff
@@ -28,6 +28,19 @@
...
-.mito-close-search-button:hover {
+.mito-search-button:hover {
     background-color: var(--mito-highlight-light);
 }
+
+.mito-grid-cell-search-match {
+    background-color: var(--mito-highlight-light);
+    color: var(--mito-text);
+    font-weight: bold;
+}
```
naterush, line 39: I'm actually not sure we need this. I think just normal section in the sheet is enough to show that you moved to something new - just like in Excel!

In which case, can we remove this class, and remove the logic applying it to cells and headers?

naterush (changes requested):
Things are working really great. A few minor refactors, and then LGTM. Happy to review again if you want, just lemme know - but also merge if you want.

mitosheet/src/mito/components/SearchBar.tsx
marthacryan: Oh you're totally right, I assumed it'd be a string for some reason

mitosheet/src/mito/components/SearchBar.tsx
marthacryan: I added this so that it'd automatically scroll to the first match when the API call returns w/ the new matches.

marthacryan: Otherwise you would have to press enter to scroll to the new match - I think different applications handle this differently so up to you if you want to keep this behavior!

mitosheet/src/mito/components/endo/ColumnHeader.tsx
marthacryan: It works without any issues on the `RockArt_600Columns` dataset and on a dataset w/ 100k rows so I think it should be ok

mitosheet/src/mito/Mito.tsx
marthacryan: I moved it to `EndoGrid.tsx` so that I could access the container div for the calls to `ensureCellIsVisible`.

mitosheet/src/mito/components/SearchBar.tsx
```diff
@@ -10,71 +10,242 @@ import Input from './elements/Input';
...
-            const total = response.result;
-            if (total !== undefined && !isNaN(Number(total))) {
-                setTotalMatches(Number(total));
+            const new_total_number_matches = response.result.total_number_matches;
+            const new_matches = response.result.matches;
+            // Update the total matches. 
+            setTotalMatches(new_total_number_matches ?? 0);
+
+            if (new_matches.length > 0) {
+                scrollMatchIntoViewAndUpdateSelection(new_matches[currentMatchIndex]);
```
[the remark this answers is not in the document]
naterush: Ok -- I actually vote we remove the jump to first match. On large datasets, when I'm searching a larger string, it leads to me bouncing around the sheet in a way that confused me a lot! Let's just remove for now.

naterush (changes requested):
# Bug with total number of matches

I think this is related to the casting that was previous happening? But it seems like total number of matches is calculating wrong -- see the rock art dataset, able to scroll past the total number of displayed matches (or it at least appears like that).

Screen Shot 2023-09-21 at 2 00 27 PM

Once you fix that, LGTM

mitosheet/mitosheet/api/get_search_matches.py
```diff
@@ -0,0 +1,48 @@
...
+    unique_value_counts = df.value_counts()
+
+    # Use the same regex for all searching
+    search_regex = re.compile(search_value, re.IGNORECASE)
+
+    # Then, filter for the search value:
+    matches = unique_value_counts.filter(regex=search_regex, axis=0)
+
+    # Then, sum the matches:
+    total_number_matches = matches.sum()
```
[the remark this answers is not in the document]
naterush: (pandas number types are like lumpy numbers, which do all sorts of weird things)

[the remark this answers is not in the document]
marthacryan: Looks like that was actually an issue with `MultiIndex` - just pushed a fix!

PR #923, opened 2023-09-22
Add Replace

marthacryan:


mitosheet/mitosheet/code_chunks/replace_code_chunk.py
```diff
@@ -0,0 +1,62 @@
...
+        return 'Replace'
+    
+    def get_description_comment(self) -> str:
+        return f"Replace {self.search_value} with {self.replace_value} in {self.post_state.df_names[self.sheet_index]}"
+
+    def get_code(self) -> Tuple[List[str], List[str]]:
+        search_value = self.search_value
+        replace_value = self.replace_value
+        df = self.post_state.dfs[self.sheet_index]
+        df_name = self.post_state.df_names[self.sheet_index]
```
naterush, line 33: Can you save these in the __init__? Also only reference prev_state, which is another pattern we're trying to move to.

mitosheet/mitosheet/step_performers/replace.py
```diff
@@ -0,0 +1,109 @@
...
+                non_bool_columns = df.select_dtypes(exclude='bool')
+                df[bool_columns.columns] = bool_columns.astype(str).replace(f'(?i){search_value}', replace_value, regex=True).applymap(cast_string_to_bool).astype(bool)
+                df[non_bool_columns.columns] = non_bool_columns.astype(str).replace(f'(?i){search_value}', replace_value, regex=True).astype(non_bool_columns.dtypes.to_dict())
+            else:
+                df = df.astype(str).replace(f'(?i){search_value}', replace_value, regex=True).astype(df.dtypes.to_dict())
+
+            # Then, we replace the search_value inside the column headers
+            column_matches = [column for column in df.columns if re.search(re.compile(search_value, re.IGNORECASE), column)]
+            # We replace in the dataframe *and* in the state column_ids object
+            df.columns = df.columns.str.replace(f'(?i){search_value}', replace_value, regex=True)
```
naterush, line 74: You do the replace here. I think in the next line you can zip the old columns and new columns together, and that way you don't have to do the replace multiple times.

mitosheet/mitosheet/step_performers/replace.py
```diff
@@ -0,0 +1,109 @@
...
+                post_state, 
+                get_param(params, 'sheet_index'),
+                get_param(params, 'search_value'),
+                get_param(params, 'replace_value')
+            )
+        ]
+
+    @classmethod
+    def get_modified_dataframe_indexes(cls, params: Dict[str, Any]) -> Set[int]:
+        return {get_param(params, 'sheet_index')} # TODO: add the modified indexes here!
```
naterush, line 108: Remove TODO

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1483,9 +1483,21 @@ export const createActions = (
             searchTerms: ['import', 'custom import', 'user defined import'],
             tooltip: "Import data using the custom defined imports function "
         },
+        [ActionEnum.REPLACE]: {
+            type: ActionEnum.REPLACE,
+            shortTitle: 'Replace',
+            longTitle: 'Replace',
+            actionFunction: () => {
+                // TODO
```
naterush, line 1491: I think you can remove this action, as you already have a search one. You'll have to delete it from the ActionEnum

mitosheet/src/mito/components/SearchBar.tsx
```diff
@@ -189,56 +191,89 @@ export const SearchBar: React.FC<SearchBarProps> = (props) => {
...
+                        if (e.key === 'Enter') {
+                            void mitoAPI.editReplace(uiState.selectedSheetIndex, searchValue ?? '', replaceValue ?? '')
+                        }
+                    }}
+                    className='mito-input'
+                    placeholder='Replace...'
+                    autoFocus
+                />
+                <button className='mito-search-button' onClick={() => {
+                    void mitoAPI.editReplace(uiState.selectedSheetIndex, searchValue ?? '', replaceValue ?? '')
```
naterush, line 271: Should we maybe bail if they aren't defined?

mitosheet/mitosheet/tests/step_performers/test_replace.py
```diff
@@ -0,0 +1,519 @@
...
+            pd.DataFrame({'A': ['apple', 'Banana', 'Cherry']}),
+        ],
+        0,
+        "b",
+        "",
+        [
+            pd.DataFrame({'A': ['apple', 'anana', 'Cherry']}),
+        ]
+    )
+]
```
naterush, line 348: Very thorough and comprehensive set of tests, that give me a huge amount of confidence in replace! Very nice!

naterush (changes requested):
Overall, this is working really great. A few polish things:

## Icon Sizing
Screen Shot 2023-09-25 at 3 45 18 PM
Screen Shot 2023-09-25 at 3 47 01 PM

check how much smaller the expand icon is on VS Code. Let's shrink ours down a little bit (I think it's just a different icon, we can add a new one). And also make the spacing around it smaller. 

I'd also be down to make the other icons the UpArrow and DownArrow (I think we already have them) - that way they don't look the same as the expand...

## Clear matches after replacing

In VS Code, if you replace, then it then refreshes the matches. This is important as otherwise you're left scrolling through out of date results. 

I think this makes sense in our case as well. Let's just refactor the getting matches code into a function, and then call it post calling replace. 

## Index Issues, becoming -1

Screen Shot 2023-09-25 at 3 50 31 PM

How does the match index become -1? I am not sure -- can we investigate this?

Ping me whenever these polish things are fixed up, and I'll do a quick final review and then merge!

Great work!

mitosheet/src/mito/components/SearchBar.tsx
marthacryan: I think if the user wants to use empty string / undefined as the replacement value that should be allowed - I use this sometimes to remove strings from larger strings

marthacryan:
> I'd also be down to make the other icons the UpArrow and DownArrow (I think we already have them) - that way they don't look the same as the expand...

Done! How does it look? 
image
image

> In VS Code, if you replace, then it then refreshes the matches. This is important as otherwise you're left scrolling through out of date results.
> 
> I think this makes sense in our case as well. Let's just refactor the getting matches code into a function, and then call it post calling replace.

Done

> How does the match index become -1? I am not sure -- can we investigate this?

I'm not able to reproduce this - could you add more detailed steps to reproduce?

mitosheet/src/mito/components/SearchBar.tsx
```diff
@@ -188,57 +193,95 @@ export const SearchBar: React.FC<SearchBarProps> = (props) => {
         setTotalMatches(e.target.value === '' ? 0 : undefined);
     }
 
+    const handleReplace = () => {
+        void mitoAPI.editReplace(uiState.selectedSheetIndex, searchValue ?? '', replaceValue ?? '');
+        getMatches();
```
naterush, line 198: I think you might have to make handleReplace async, and then change the void -> async. Otherwise, you might have race conditions!

naterush:
For -1, maybe wrap the display in a `max(0, {whatever it is currently})` just so we don't see it?

PR #927, opened 2023-09-26
Add ability to replace only in selected columns

marthacryan:
Allows for users to replace values only within the columns that are currently selected. 
image

When no columns are selected, the button is disabled:
image

Todo:
- [x] Fix style
- [x] Add more backend tests


mitosheet/mitosheet/tests/step_performers/test_replace.py
```diff
@@ -350,13 +350,125 @@
...
+            pd.DataFrame({
+                'test1': [1, 2, 3],
+                'test2': [1.0, 2.0, 3.0], 
+                'test3': [True, False, True], 
+                'test4': ["string", "with spaces", "and/!other@characters3"], 
+                'test5': pd.to_datetime(['12-22-1997', '12-23-1997', '12-24-1997']), 
+            })
+        ],
+        0,
+        [ "test3", "test4" ],
```
naterush, line 398: Can you try with non-string column names?

mitosheet/mitosheet/step_performers/replace.py
```diff
@@ -40,48 +40,57 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
...
         pandas_start_time = perf_counter()
-        
         df = post_state.dfs[sheet_index]
 
+        # If no column_ids are specified, then we replace the values in all columns
+        if column_ids is None or len(column_ids) == 0:
+            column_ids = df.columns
+
+        # Selected columns is the dataframe with only the columns we want to replace values in
+        selected_columns = df[column_ids]
```
naterush, line 56: We shouldn't be accessing the dataframe by the column ids. I think we need to use the column ids map to get column headers, and then access by column headers here. Def make sure to rename variables through this step performer (as well as the code chunk) approproatiely as well.

Can you write one test where you create a mitosheet, rename a column, then replace in that column? Should catch this bug.

naterush (changes requested):
The column ids bug means this doesn't work if you rename columns and then try and replace in them. 

Once you fix that up (and add a test), LGTM.

mitosheet/mitosheet/step_performers/replace.py
```diff
@@ -40,48 +40,60 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
...
 
             # Then, we replace the column headers in the state column_ids object
-            new_columns = df.columns.str.replace(f'(?i){search_value}', replace_value, regex=True)
-
+            # We convert the column headers to strings because the column headers can be any type
+            # Then we call astype because we need to convert them back to their original type. 
+            new_columns = pd.Index([re.sub(search_value_regex, replace_value, str(column)) for column in selected_columns.columns]).astype(selected_columns.columns.dtype).to_list()
             # Update the column headers in the state column_ids object
-            for old_column_name, new_column_name in zip(df.columns, new_columns):
+            for old_column_name, new_column_name in zip(selected_columns.columns, new_columns):
```
naterush, line 85: selected_columns.columns illustrates that this isn't a great variable name choice here either.

mitosheet/mitosheet/tests/step_performers/test_replace.py
```diff
@@ -350,13 +350,233 @@
...
+                5: pd.to_datetime(['12-22-1997', '12-23-1997', '12-24-1997']), 
+            })
+        ],
+        0,
+        [ "4" ],
+        "4", 
+        "6", 
+        [
+            pd.DataFrame({
+                1: [1, 2, 3],
```
naterush, line 453: Check out these tests -- all of the dataframes have column headers of the same type. Think about the dataframes you've seen in the wild -- rarely do they have only numbers or only strings as headers.

Let's add some tests for this, which I think would catch the dtype bug with the index.

Try and think about each critical line of code you write - what test can you add that will stress test if it's correct?

mitosheet/mitosheet/step_performers/replace.py
```diff
@@ -40,48 +40,60 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
...
             else:
-                df = df.astype(str).replace(f'(?i){search_value}', replace_value, regex=True).astype(df.dtypes.to_dict())
+                selected_columns = selected_columns.astype(str).replace(search_value_regex, replace_value, regex=True).astype(selected_columns.dtypes.to_dict())
 
             # Then, we replace the column headers in the state column_ids object
-            new_columns = df.columns.str.replace(f'(?i){search_value}', replace_value, regex=True)
-
+            # We convert the column headers to strings because the column headers can be any type
+            # Then we call astype because we need to convert them back to their original type. 
+            new_columns = pd.Index([re.sub(search_value_regex, replace_value, str(column)) for column in selected_columns.columns]).astype(selected_columns.columns.dtype).to_list()
```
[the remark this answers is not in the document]
naterush: Something along these lines maybe?
```
new_columns = pd.Index([re.sub(search_value_regex, replace_value, str(column)) for column in selected_columns.columns]).map({h: type(h) for h in selected_columns.columns}).to_list()
```

naterush (changes requested):
Check out this simple failing test case when the column headers have mixed types.

Screen Shot 2023-09-28 at 9 58 17 AM

Ping me when it's fixed and I'll do a final review

naterush (commented):
Minor cleanup, let's get tests passing, and then LGTM!

PR #928, opened 2023-09-27
Consecutive empty columns

naterush:
# Description

Adds consecutive number of empty columns.

# Testing

Use the feature on a good Excel test file you make for yourself. 

# Documentation

No.


marthacryan (commented):
This is great! One thing I noticed is that we're leaving the empty rows in with the imported data, and I'm guessing that most users don't want those empty rows to be included. Other than that, LGTM!

PR #929, opened 2023-09-28
[WIP] Dash fixes

naterush:
# Description

- Changes to the callback. These changes are questionably better - but checkout new usage (you use Input and Output as you normally do).
- Fixes import folder
- Fixes error messages across the app.
- Adds missing parameters like df_names, sheet_functions, etc

# Testing

See the demo app here: https://github.com/mito-ds/dash-automation-demo

# Documentation

No.


mitosheet/mitosheet/tests/dash/test_dash.py
```diff
@@ -0,0 +1,99 @@
...
+        dcc.Input(id='input', value='initial value'),
+        html.Div(id='output'),
+        dcc.Store(id='state'),
+    ])
+    
+    @mito_callback(
+        Output('output', 'value'),
+        Input('input', 'value'),
+        State('state', 'value'),
+    )
```
aarondr77, line 30: Can you point me in the right direction of where I should learn about this stuff. Is there any reason you would want this without Mito?

mitosheet/mitosheet/mito_dash/v1/spreadsheet.py
```diff
@@ -149,74 +201,78 @@ def get_result(self):
...
-    if len(components) > 0:
-        callback_component = components[0]
+def mito_callback(*args, **kwargs):
+    # First, check if there are any args that are Inputs that contain a mito_id
+    indexes = get_spreadsheet_id_and_indexes_in_callback_args(*args)
 
-    if callback_component is None:
-        raise ValueError(f'Could not find Spreadsheet with id {input_id}. Make sure the input_id matches the id of a spreadsheet component')
+    # If there are no spreadsheet input components, then we just call the regular callback
+    if len(indexes) == 0:
```
aarondr77, line 251: If there is no spreadsheet should we even be supporting this case? Should we just through an error instead? 

Maybe the most common reason someone is going to be here is just because they don't understand how to create a mitosheet and the error message could be useful.

aarondr77:
I'm still getting this error when trying to use find and replace:
1. Open find and replace
2. Type anything 
3. Get this error

Screenshot 2023-10-02 at 1 35 21 PM

aarondr77:
Styling notes:
1. The unselected dataframe name should be dark against the light background
2. The grid seems a bit mis-sized: I think it needs to be taller

Screenshot 2023-10-02 at 1 36 51 PM

aarondr77 (commented):
Okay I think the changes made here are working well. There is still more Dash cleanup in general todo that I can do a bigger review on whenever you want? I left a couple of comments here. 

Once the CI is passing, LGTM into dev

mitosheet/mitosheet/mito_dash/v1/spreadsheet.py
naterush: They might just be iteratively developing, and not currently have a Mitosheet in their callback. This is probably pretty frequent!

mitosheet/mitosheet/tests/dash/test_dash.py
naterush: See the dash docs, as well as the app I asked you to build in Zulip.

naterush:
We'll address styling in a big styling pass.

PR #937, opened 2023-10-03
mitoinstaller: add better messaging

aarondr77:
# Description

Adds a more helpful and action oriented error message when the Mito installer cannot automatically upgrade from an old version of JupyterLab. 

# Testing

Try it. 

# Documentation

No.


naterush:
LGTM when tests pass.

PR #939, opened 2023-10-05
Further Dash Fixes

naterush:
# Description

This is a reasonably sprawling PR that fixes a bunch of issues with our Dash integration:
1. Moves dropdowns from `absolute` positioning to `fixed` positioning, so that positioning works the same across JupyterLab and Streamlit as well as Dash. The issue was that JupyterLab never scrolls the actual window, nor does streamlit (as it's in an iframe) -- so we never previously had to deal with this case. 
2. Makes the `@mito_callback` work in more cases, by doing 0 manipulation of the Input and State arguments to the `@callback` function (which has a whole bunch of ways you can [pass these](https://dash.plotly.com/flexible-callback-signatures)). 
3. Makes `@mito_callback` give you an error message when you use `@callback` instead of it. Now, it actually tells you what is wrong -- so that you have a hope of fixing it without being hopelessly confused. We could write some more docs here, as well.
4. Makes it so selection changes don't also trigger callbacks for `spreadsheet_result` to run. This was the source of performance problems, and generally it makes a ton of sense to be able to subscribe to each separately. 
5. Fixes some CSS issues: sheet tab coloring in Dash.
6. Adds the ability to pass a theme to a Dash component -- so our demo apps can look better. This is a bit buggy with multiple mitosheets on one page, when one is themed and another is not, but I think it's ok.

# Testing

1. Test dropdowns place themselves well on JupyterLab, Streamlit, and Dash.
2. Try the new `@mito_callback` decorator -- see if you can pass lists of Inputs and States (and make sure that it runs properly in this cases)
3. Use `@callback` and see the error message you get when using it with a Spreadsheet input - see if it's helpful.
4. Check the new dash usage, and check that selection changes trigger selection change callbacks, and that spreadsheet result changes trigger new spreadsheet result callbacks
5. Check sheet coloring. 
6. Check dash usage as well, see passed theme.

# Documentation

Once all this is merged in, I'm going to take another pass at the Dash documentation!

### DIFF
--- a/mitosheet/dash_usage.py
+++ b/mitosheet/dash_usage.py
@@ -1,35 +1,39 @@
-from unittest.mock import patch
-from dash import Dash, dcc, html, Input, Output, callback, State
+from dash import Dash, callback, Input, Output, html, dcc
 from mitosheet.mito_dash.v1 import Spreadsheet, mito_callback
+import pandas as pd
 
 app = Dash(__name__)
 
-app.layout = html.Div([
-    html.H6("Change the value in the text box to see callbacks in action!"),
-    html.Div([
-        "Input: ",
-        dcc.Input(id='my-input', value='initial value', type='text')
-    ]),
-    Spreadsheet(id='my-spreadsheet'),
-    html.Br(),
-    html.Div(id='my-output'),
+df = pd.DataFrame({'A': [1, 2, 3]})
 
+app.layout = html.Div([
+    html.H1("Stock Analysis", style={'color': 'white'}),
+    Spreadsheet(df, id='sheet'),
+    html.Div(id='output-code'),
+    html.Div(id='output-selection'),
 ])
 
-
-
-
+@mito_callback(
+    Output('output-code', 'children'),
+    Input('sheet', 'spreadsheet_result'),
+)
+def update_code(spreadsheet_result):
+    print("NEW SPREADSHEET RESULT")
+    return html.Div([
+        html.Code(spreadsheet_result.code(), style={'color': 'white'})
+    ])
+    
 
 @mito_callback(
-    Output(component_id='my-output', component_property='children'),
-    Input(component_id='my-spreadsheet', component_property='return_value'),   
-    prevent_initial_call=True
+    Output('output-selection', 'children'),
+    Input('sheet', 'spreadsheet_selection'),
 )
-def update_output_div(result):
-    print("HERE!")
-    result = result.selection()
-    return f'Output: {result}'
-
+def update_selection(new_selection):
+    print("NEW SPREADSHEET SELECTION")
+    return html.Div([
+        html.Code(str(new_selection), style={'color': 'white'})
+    ])
+    
 
 if __name__ == '__main__':
-    app.run(debug=True)
+    app.run_server(debug=True)
--- a/mitosheet/dash_usage.py
+++ b/mitosheet/dash_usage.py
@@ -7,10 +7,15 @@ app = Dash(__name__)
 df = pd.DataFrame({'A': [1, 2, 3]})
 
 app.layout = html.Div([
-    html.H1("Stock Analysis", style={'color': 'white'}),
-    Spreadsheet(df, id='sheet'),
-    html.Div(id='output-code'),
-    html.Div(id='output-selection'),
+    html.H1("Stock Analysis", style={'color': 'black'}),
+    Spreadsheet(df, id='sheet', theme={
+        'primaryColor': '#00FF00',
+        'backgroundColor': '#111111',
+        'secondaryBackgroundColor': '#222222',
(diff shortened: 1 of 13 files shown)


mitosheet/dash_usage.py
```diff
@@ -1,35 +1,39 @@
...
-        dcc.Input(id='my-input', value='initial value', type='text')
-    ]),
-    Spreadsheet(id='my-spreadsheet'),
-    html.Br(),
-    html.Div(id='my-output'),
+df = pd.DataFrame({'A': [1, 2, 3]})
 
+app.layout = html.Div([
+    html.H1("Stock Analysis", style={'color': 'white'}),
+    Spreadsheet(df, id='sheet'),
```
naterush, line 11: TODO: pass a theme for easy testing

mitosheet/mitosheet/mito_dash/v1/mito_callback.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def mito_callback(*args, **kwargs):
+        """
+        This is a replacement for the @callback decorator that improves the developer experience
+        with using the Mito Spreadsheet component. 
+
+        To see the documentation for this decorator, see the @mito_callback decorator documentation:
+        https://docs.trymito.io/mito-for-dash/api-reference#callback-props-and-types
+
+        In practice, this function replaces the default spreadsheet_result error string with the
```
naterush, line 35: Update comment, and move up. Do more explaining here.

mitosheet/mitosheet/mito_dash/v1/mito_callback.py
```diff
@@ -0,0 +1,77 @@
...
+
+                # TODO: do we need to handle kwargs? Probably. But we will do that in the future...
+
+                for index, arg in enumerate(new_args):
+                    # We go through all the arguments passed to the function, and:
+                    # 1) find those that are Mito Spreadsheet Result error strings
+                    # 2) Replace them with the the actual spreadsheet results
+                    # As this is a dramatically better Spreadsheet experience
+
+                    if isinstance(arg, str) and arg.startswith(WRONG_CALLBACK_ERROR_MESSAGE_FIRST_LINE):
```
naterush, line 51: Explain this better

mitosheet/mitosheet/mito_dash/v1/spreadsheet.py
```diff
@@ -163,7 +207,11 @@ def send(response):
                         },
                     }
                 )
-            
+
+            # TODO: document this - why we need to do this
```
naterush, line 211: TODO: explain these, and why they need to change. Also the `num_messages` thing.

mitosheet/mitosheet/types.py
```diff
@@ -356,6 +356,22 @@ class CodeOptions(TypedDict):
 
     UserDefinedImporterParamType = Literal['any', 'str', 'int', 'float', 'bool']
 
+    class MitoTheme(TypedDict):
+        primaryColor: str
+        backgroundColor: str
+        secondaryBackgroundColor: str
+        textColor: str
+
+    class MitoFrontendSelection(TypedDict):
```
naterush, line 365: Add these types to the selection stuff!

mitosheet/src/dash/MitoDashWrapper.tsx
```diff
@@ -5,11 +5,13 @@ import { getAnalysisDataFromString, getSheetDataArrayFromString, getUserProfileF
 
 export const DELAY_BETWEEN_SET_DASH_PROPS = 25;
 
+// TODO: Update
```
naterush, line 8: Document this!

mitosheet/src/mito/types.tsx
```diff
@@ -782,11 +782,11 @@ export interface AnalysisData {
     codeOptions: CodeOptions;
     userDefinedFunctions: FunctionDocumentationObject[];
     userDefinedImporters: UserDefinedImporter[];
-
     importFolderData: {
         path: string,
         pathParts: string[],
-    } | null
+    } | null;
+    theme: MitoTheme | null;
```
naterush, line 789: TODO: add this to Streamlit, and make it so that if you pass one to Streamlit, it overrules the Streamlit theme. Also add this to the PR description.

naterush (commented):
Cleanup and documentation, mostly.

aarondr77 (commented):
These improvements are working pretty well! Just a few small things and then LGTM. 

Improving the docs around this stuff seems like a great next step

aarondr77:
The pivot table dropdowns on lab are broken

https://github.com/mito-ds/mito/assets/18709905/226cba4f-0995-4702-b36b-20e01387e33a

naterush:
Merging when tests pass.

PR #942, opened 2023-10-10
Add function to backend to get fully parameterized function

marthacryan:
# Description
Adds a function to the backend to get the transpiled code as a function (that isn't called) with all of the current parameterization added. This includes: 
```
[
                'import_dataframe',
                'file_name_export_excel',
                'file_name_export_csv',
                'file_name_import_excel',
                'file_name_import_csv',
]
```

# Testing
Added tests in `test_transpile.py` to check that the output is what we expect.

### DIFF
--- a/mitosheet/mitosheet/mito_backend.py
+++ b/mitosheet/mitosheet/mito_backend.py
@@ -121,8 +121,9 @@ class MitoBackend():
     @property
     def fully_parameterized_function(self) -> List[str]:
         """
-        Returns the fully parameterized function string, which is the string
-        that is used to call the function in the code.
+        Returns the fully parameterized function string. This is used for
+        cases where we want to get the function string regardless of the 
+        code options the user provided. 
         """
         return get_script_as_function(
             self.steps_manager,
--- a/mitosheet/mitosheet/mito_backend.py
+++ b/mitosheet/mitosheet/mito_backend.py
@@ -40,6 +40,8 @@ from mitosheet.user.schemas import (UJ_MITOSHEET_LAST_FIFTY_USAGES,
                                     UJ_USER_EMAIL, UJ_AI_PRIVACY_POLICY)
 from mitosheet.user.utils import get_pandas_version, is_enterprise, is_pro, is_running_test
 from mitosheet.utils import get_new_id
+from mitosheet.transpiler.transpile_utils import get_script_as_function
+from mitosheet.transpiler.transpile import transpile
 from mitosheet.step_performers.utils.user_defined_functionality import get_functions_from_path, get_non_validated_custom_sheet_functions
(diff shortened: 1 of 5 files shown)


mitosheet/mitosheet/mito_backend.py
```diff
@@ -116,6 +118,28 @@ def __init__(
...
+            [],
+            transpile(self.steps_manager, add_comments=False),
+            self.steps_manager.code_options.get('function_name'),
+            [
+                'import_dataframe',
+                'file_name_export_excel',
+                'file_name_export_csv',
+                'file_name_import_excel',
+                'file_name_import_csv',
+            ],
```
naterush, line 139: My feeling is that we will most likely forget to update this code when we add new parameterization options. Do you wanna add a new valid param called `all` - which just turns all parametrizable params into a param for the function?

mitosheet/mitosheet/tests/test_transpile.py
```diff
@@ -550,6 +550,39 @@ def test_transpile_as_function_single_param_multiple_times(tmp_path):
...
+def test_transpile_fully_parameterized_function_string(tmp_path):
+    tmp_file1 = str(tmp_path / 'txt.csv')
+    tmp_file2 = str(tmp_path / 'file.csv')
+    df1 = pd.DataFrame({'A': [1], 'B': [2]})
+    df1.to_csv(tmp_file1, index=False)
+    df1.to_csv(tmp_file2, index=False)
+
+    mito = create_mito_wrapper()
+    mito.simple_import([tmp_file1])
+    mito.simple_import([tmp_file2])
```
naterush: deffo want more tests, namely one that tests all of the parameterization options.

mitosheet/mitosheet/mito_backend.py
```diff
@@ -116,6 +118,28 @@ def __init__(
 
         self.theme = theme
 
+    @property
+    def fully_parameterized_function(self) -> List[str]:
```
naterush, line 122: I think we can just return this as one string

mitosheet/mitosheet/tests/test_transpile.py
```diff
@@ -550,6 +550,75 @@ def test_transpile_as_function_single_param_multiple_times(tmp_path):
         f"txt, txt_1 = function(var_name)"
     ]
 
+@pandas_post_1_2_only
+@python_post_3_6_only
+def test_transpile_fully_parameterized_function_string(tmp_path):
```
naterush, line 555: This test is almost perf - I think we just need to pass a dataframe + df_names to the `create_mito_wrapper` - to test the dataframe parameterization.

naterush (changes requested):
Looks legit! Minor cleanup and then LGTM

PR #943, opened 2023-10-10
Add ParamData type

marthacryan:
# Description
Adds a `ParamData` type. Adds a `mito_backend.param_metadata` property. 

# Testing
Adds a test with the `fully_parameterized_function_string` tests to check that the expected params show up in the `param_metadata` as well.

### DIFF
--- a/mitosheet/mitosheet/api/get_parameterizable_params.py
+++ b/mitosheet/mitosheet/api/get_parameterizable_params.py
@@ -47,6 +47,8 @@ def get_parameterizable_params_metadata(steps_manager: StepsManagerType) -> List
                         'initial_value': param[0],
                         'type': param[1],
                         'subtype': param[2],
+                        # If the param is a df_name, then it is required for the function
+                        # because we won't be able to store the initial value in the MitoAnalysis for streamlit.
                         'required': param[1] == 'df_name',
                         'name': param_name
                 }
--- a/mitosheet/mitosheet/api/get_parameterizable_params.py
+++ b/mitosheet/mitosheet/api/get_parameterizable_params.py
@@ -7,7 +7,7 @@
 import json
 from typing import Any, Dict, List, Tuple
 from mitosheet.code_chunks.code_chunk_utils import get_code_chunks
-from mitosheet.types import ParamSubtype, ParamType, ParamName, ParamValue, StepsManagerType, ParamData
+from mitosheet.types import ParamSubtype, ParamType, ParamName, ParamValue, StepsManagerType, ParamMetadata
 from mitosheet.updates.args_update import is_string_arg_to_mitosheet_call
 from mitosheet.transpiler.transpile_utils import get_final_function_params_with_subtypes_turned_to_parameters
 
@@ -29,13 +29,13 @@ def get_parameterizable_params(params: Dict[str, Any], steps_manager: StepsManag
 
         return all_parameterizable_params
 
-def get_parameterizable_params_metadata(steps_manager: StepsManagerType) -> List[ParamData]:
+def get_parameterizable_params_metadata(steps_manager: StepsManagerType) -> List[ParamMetadata]:
         # First, get all the parameterizable params and the function params
         all_parameterizable_params = get_parameterizable_params({}, steps_manager)
         function_params: Dict[ParamName, ParamValue] = get_final_function_params_with_subtypes_turned_to_parameters(steps_manager, 'all')
         param_names = list(function_params.keys())
 
-        # Now, we need to map the parameterizable params to the ParamData type. 
+        # Now, we need to map the parameterizable params to the ParamMetadata type. 
         def map_tuple_to_param_data(args):
                 param = args[0]
                 param_name = args[1] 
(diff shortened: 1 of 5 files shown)


mitosheet/mitosheet/tests/test_transpile.py
```diff
@@ -618,6 +618,33 @@ def test_transpile_fully_parameterized_function_string(tmp_path):
         f"{TAB}return import_dataframe_0, txt, Sheet1",
         "",
     ])
+    assert mito.mito_backend.param_metadata == [
+        {
+            'initial_value': 't',
+            'type': 'e',
+            'subtype': 's'
```
naterush, line 625: What are these? This is strange if these are the actual values, b/c they feel like they shouldn't be.

mitosheet/mitosheet/types.py
```diff
@@ -344,6 +344,11 @@ class AITransformFrontendResult(TypedDict):
     ]
     ParamValue = str
 
+    class ParamData(TypedDict):
```
naterush, line 347: Document this a bit -- what its' for, why it exists, etc. This is going to end up being an internal interface. 

Also, can we think of a better name here? `ParamData` might be used elsewhere, and if not it's very close to other ones.

mitosheet/mitosheet/tests/test_transpile.py
naterush: Can we also add so the `mito_test_wrapper` gets a `string` initial argument as well as a dataframe initial argument?

naterush: (I know we don't let you parameterize it, but it's nice to include that current in the test)

mitosheet/mitosheet/api/get_parameterizable_params.py
```diff
@@ -28,3 +27,27 @@ def get_parameterizable_params(params: Dict[str, Any], steps_manager: StepsManag
                 all_parameterizable_params.extend(parameterizable_params)
 
         return all_parameterizable_params
+
+def get_parameterizable_params_metadata(steps_manager: StepsManagerType) -> List[ParamData]:
```
naterush, line 31: Almost all the code here is duplicated. Can you just get the `parametrizable_params` using the above function, and then just apply you mapping function to those? I don't think they need to be different implementations, really...

naterush: ^ this would also fix the bug you have with originally passed dataframes generating strange data from this function.

naterush (changes requested):
Overall, looks pretty legit. A few refactoring thoughts that will fix the one minor bug I see here. 

The other thing that this made me realize is that, sometimes, the `run` function is going to _require_ arguments. Imagine you do

```
mito_analysis = spreadsheet(df)
mito_analysis.run()
```

This... obviously wouldn't work! Because `df` is not something the `mito_analysis` still has access to. Really, `df` is going to become a required argument to `run`!

I think we should just update this in the plan of attack. Original arguments to the `spreadsheet` call become required, positional arguments to the `run` function. If you pass `3` things to `spreadsheet`, you must pass at least `3` things to `run` or it will error!

mitosheet/mitosheet/api/get_parameterizable_params.py
```diff
@@ -28,3 +28,27 @@ def get_parameterizable_params(params: Dict[str, Any], steps_manager: StepsManag
...
+
+# This function is used to get the parameterizable params for the UI in streamlit. 
+# It returns a list of ParamMetadata objects, which are defined in mitosheet/types.py
+# The ParamMetadata object includes various information that a streamlit developer might
+# need to display the parameterizable params in the UI.
+def get_parameterizable_params_metadata(steps_manager: StepsManagerType) -> List[ParamMetadata]:
+        # First, get all the parameterizable params and the function params
+        all_parameterizable_params = get_parameterizable_params({}, steps_manager)
+        function_params: Dict[ParamName, ParamValue] = get_final_function_params_with_subtypes_turned_to_parameters(steps_manager, 'all')
+        param_names = list(function_params.keys())
```
naterush, line 40: Ah -- one thing here is that these may not be ordered (dicts, by default, are not). I think you can fix this just by changing the `get_final_function_params_with_subtypes_turned_to_parameters` function to return an ordered dict!

mitosheet/mitosheet/api/get_parameterizable_params.py
```diff
@@ -28,3 +28,27 @@ def get_parameterizable_params(params: Dict[str, Any], steps_manager: StepsManag
...
+
+        # Now, we need to map the parameterizable params to the ParamMetadata type. 
+        def map_tuple_to_param_data(args):
+                param = args[0]
+                param_name = args[1] 
+                return {
+                        'initial_value': param[0],
+                        'type': param[1],
+                        'subtype': param[2],
+                        'required': param[1] == 'df_name',
```
naterush, line 50: Document this methinks

naterush (changes requested):
Looks really great. Minor fix up then LGTM

PR #944, opened 2023-10-11
Add MitoAnalysis class

marthacryan:
# Description
Adds a class that can be used by streamlit developers to store / access information about an analysis. 

# Testing
Updated the `streamlit.py` file to show usage. 

# Documentation
We should add documentation once we add some more functionality into this class.

### DIFF
--- a/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
+++ b/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
@@ -115,8 +115,8 @@ class MitoAnalysis:
     code: str
     code_options: CodeOptions
     def __init__(self, code, code_options, fully_parameterized_code, param_metadata):
-        self.code = code
-        self.code_options = code_options
+        self.__code = code
+        self.__code_options = code_options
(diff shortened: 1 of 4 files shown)


mitosheet/mitosheet/streamlit/v1/spreadsheet.py
```diff
@@ -108,6 +108,17 @@ def get_function_from_code_unsafe(code: str) -> Optional[Callable]:
         
     raise ValueError(f'No functions defined in code: {code}')
 
+# This is the class that is returned when the user sets return_type='analysis'
+# It contains data that could be relevant to the streamlit developer, and is 
+# used for replaying analyses. 
+class MitoAnalysis:
+    code: str
+    code_options: CodeOptions
```
naterush, line 116: I'm actually like -- why don't we make all of this private for now? If we want to expose these things publicly, we can always make accessors. But I think keeping it private for now is sensible - so we don't have to worry about compatibility.

naterush (changes requested):
Can you add a test for this return type in test_streamlit.py? If you do that and make the other variables private, then LGTM!

PR #954, opened 2023-10-17
Fix transpile bug

marthacryan:
**Problem**: In streamlit, when you passed a dataframe as a positional arg and didn't specify the name of the dataframe, the `steps_manager` was setting the param "value" to the empty string. Then when transpiling as a function, the `get_script_as_function` function was replacing all param values with their new variable names and **replaced all empty strings** with the new variable name for the positional argument. 

**Solution**: Updates the `original_args_raw_string` initialization to be after the preprocessing of df_names so that the df_name can be included automatically.

### DIFF
--- a/mitosheet/mitosheet/steps_manager.py
+++ b/mitosheet/mitosheet/steps_manager.py
@@ -228,7 +228,7 @@ class StepsManager:
         # always have an args update, this is the best we can do at this point in time. Notably, 
         # these are required to be set for transpiling arguments
         self.original_args_raw_strings: List[str] = []
-        for index, arg in enumerate(args):
+        for index, arg in enumerate(self.original_args):
             if isinstance(arg, str):
                 self.original_args_raw_strings.append(f'"{arg}"')
             else:
--- a/mitosheet/mitosheet/steps_manager.py
+++ b/mitosheet/mitosheet/steps_manager.py
@@ -213,15 +213,6 @@ class StepsManager:
             for arg in args
         ]
 
-        # We set the original_args_raw_strings. If we later have an args update, then these
-        # are overwritten by the args update (and are actually correct). But since we don't 
-        # always have an args update, this is the best we can do at this point in time. Notably, 
-        # these are required to be set for transpiling arguments
-        self.original_args_raw_strings: List[str] = [
-            f'"{arg}"' if isinstance(arg, str) else ''
-            for arg in args
-        ]
-
         # Then, we go through the process of actually preprocessing the args
         # saving any data that we need to transpilate it later this
         self.preprocess_execution_data = {}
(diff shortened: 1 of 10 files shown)


mitosheet/mitosheet/transpiler/transpile_utils.py
```diff
@@ -276,7 +276,11 @@ def get_script_as_function(
 
         # Then, for any additional function params we defined, we relace the internal param value. Note that 
         # we only replace for
```
naterush, line 278: This comment could be updated...

mitosheet/mitosheet/transpiler/transpile_utils.py
```diff
@@ -276,7 +276,11 @@ def get_script_as_function(
 
         # Then, for any additional function params we defined, we relace the internal param value. Note that 
         # we only replace for 
-        for param_name, param_value in function_params.items():
+        for index, (param_name, param_value) in enumerate(function_params.items()):
+            # The param value is empty string when a dataframe was passed in as a positional argument without a name
+            if param_value == '' and index < len(steps_manager.curr_step.df_sources) and steps_manager.curr_step.df_sources[index] == 'passed':
+                param_value = steps_manager.curr_step.df_names[index]
+                function_params[param_name] = param_value
```
naterush, line 283: This is almost legit, I think. There might be an issue is what happens if the user renames a dataframe in the middle of their analysis -- not sure if we get out of date in this case.

mitosheet/mitosheet/tests/test_transpile.py
```diff
@@ -597,28 +598,28 @@ def test_transpile_fully_parameterized_function_string(tmp_path):
...
     with pd.ExcelWriter(file_name_export_excel_0, engine="openpyxl") as writer:
         txt.to_excel(writer, sheet_name="txt", index=False)
     
-    return import_dataframe_0, txt, Sheet1
+    return {expected_in_function}, txt, Sheet1
 """
     assert mito.mito_backend.param_metadata == [
         {
-            'initial_value': 'test_df_name',
+            'initial_value': 'test_df_name' if df_names else '',
```
naterush, line 618: Do we want to make it so we get the dataframe name in this case, rather than just returning an empty string? Likely, we'd want to handle this in the the OG get parametrizable params functions...

naterush (changes requested):
I think this is going in the right direction! 

Here's an idea for a simpler approach that I think might be nicer:
1. We move `self.original_args_raw_strings` definition in StepsManager to be _after_ the df_names are created by the Preprocessing steps. 
2. We use these df_names to set self.original_args_raw_strings to be the default dataframe name we come up with -- in the case where we currently set ''

I think just make sure this works with passing a mixture of strings, dataframes, and Nones (which get filtered out and might change indexes of where the DF names go). A test or two would be sweet. 

Wanna give it a try and see if it just works? It would be less code, as well as a simpler set of conditions to reason about (as it kinda just makes self.original_args_raw_strings _actually correct_).

mitosheet/mitosheet/steps_manager.py
```diff
@@ -232,6 +223,17 @@ def __init__(
                 preprocess_step_performers.preprocess_step_type()
             ] = execution_data       
 
+        # We set the original_args_raw_strings. If we later have an args update, then these
+        # are overwritten by the args update (and are actually correct). But since we don't 
+        # always have an args update, this is the best we can do at this point in time. Notably, 
+        # these are required to be set for transpiling arguments
+        self.original_args_raw_strings: List[str] = []
+        for index, arg in enumerate(args):
```
naterush, line 231: Ah -- here's our bug. Args is changed in the above lines by the preprocessing steps, as all the args get converted to dataframes by the preprocessing steps. I think you can just `enumerate(self.original_args)` - and then everything should work!  Or, maybe better would be renaming the `args` overwriting to make it so the preprocessing steps use a variable called `processed_args` or something. Your call.

I think there was one more test above I thought would be good, namely with some None value's that get filtered out, that we should test for (b/c the df_names might be at the wrong index otherwise). 

Otherwise, once we add that test and fix any resulting bugs, LGTM.

PR #956, opened 2023-10-17
Add to_json and from_json to MitoAnalysis

marthacryan:
# Description
Adds a `to_json` and `from_json` method to the `MitoAnalysis` class for easier storage of the class in files. 

# Testing
Added basic test in `test_mito_analysis.py`. 

# Documentation
This should be added with the documentation for streamlit.

### DIFF
--- a/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
+++ b/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
@@ -132,6 +132,10 @@ class MitoAnalysis:
     def param_metadata(self) -> List[ParamMetadata]:
         return self.__param_metadata
     
+    @property
+    def fully_parameterized_function(self) -> str:
+        return self.__fully_parameterized_function
+    
     def to_json(self) -> str:
         return json.dumps({
             'code': self.__code,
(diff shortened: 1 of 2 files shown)


mitosheet/mitosheet/streamlit/v1/spreadsheet.py
```diff
@@ -122,6 +122,28 @@ def __init__(self, code: str, code_options: Optional[CodeOptions], fully_paramet
     def param_metadata(self) -> List[ParamMetadata]:
         return self.__param_metadata
     
+    def to_json(self) -> str:
+        return json.dumps({
+            'code': self.__code,
```
naterush, line 127: The one thing we should def add here is a version - so that if we change the format. 

I think let's:
1. Add a `self.mito_analysis_version = 1` variable to the class
2. Include this in the JSON. 

This will protect us in case we need to upgrade / change this format over time.

naterush (changes requested):
This looks great. One minor comment about how we can protect ourselves for schema changes. 

Can you also add one test for when the code some strange characters in it? E.g. it has a quote, and a bracket, and both, and slashes, and newlines, etc? Just to make sure that's handled properly. 

Once you make these changes, we'll wait to merge until deployment is finished (working on it).

PR #959, opened 2023-10-18
Add get_param_metadata to MitoAnalysis class

marthacryan:
# Description
This makes the API clearer for streamlit developers to access the `param_metadata` by type. It also changes the `ParamType` to facilitate these changes to the API.

# Testing
Added a test into the existing `MitoAnalysis` tests that checks that:
* When no type is passed, the entire array of param metadata is returned
* When 'import' type is passed, the imports are returned. 
* When 'export' type is passed, the exports are returned. 
* When 'export' type is passed but there are no export param metadata, an empty array is passed. 

# Documentation
This should be added when we add the `MitoAnalysis` / streamlit docs.

### DIFF
--- a/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
+++ b/mitosheet/mitosheet/streamlit/v1/spreadsheet.py
@@ -118,16 +118,12 @@ class MitoAnalysis:
         self.__fully_parameterized_function = fully_parameterized_function
         self.__param_metadata = param_metadata
         
-    @property
-    def param_metadata(self) -> List[ParamMetadata]:
-        return self.__param_metadata
-    
-    def get_param_metadata(self, type: Optional[ParamType]=None):
-        if type is None:
+    def get_param_metadata(self, param_type: Optional[ParamType]=None) -> List[ParamMetadata]:
+        if param_type is None:
             return self.__param_metadata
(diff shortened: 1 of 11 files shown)


mitosheet/mitosheet/streamlit/v1/spreadsheet.py
```diff
@@ -122,6 +122,13 @@ def __init__(self, code: str, code_options: Optional[CodeOptions], fully_paramet
     def param_metadata(self) -> List[ParamMetadata]:
         return self.__param_metadata
     
+    def get_param_metadata(self, type: Optional[ParamType]=None):
```
naterush, line 125: we should call it `param_type`, so it doesn't overlap with the builtin python `type` function.

mitosheet/mitosheet/streamlit/v1/spreadsheet.py
```diff
@@ -122,6 +122,13 @@ def __init__(self, code: str, code_options: Optional[CodeOptions], fully_paramet
     def param_metadata(self) -> List[ParamMetadata]:
```
naterush, line 122: I think we should remove this accessor now.

naterush (changes requested):
Super minor change, and then LGTM. Looks really great!

PR #960, opened 2023-10-19
Testing improvements

naterush:
# Description

1. make jupyter frontend tests run 3 times as flakey
2. Add a bunch of caching across the GitHub actions
3. Make it `setup.py` just installs Python (and does no JS). Make it so `npm install` just installs and does no building. 

All in all, this dramatically reduces the amount of time that our tests take to run. 

**Pytest: 18 minutes -> 7 minutes
Frontend: 11 minutes -> 6 minutes**

# Testing

Ignore the Mitoinstaller tests -- whatever.

See the other ones. 

# Documentation


mitosheet/dev/macsetup.sh
```diff
@@ -20,8 +21,9 @@ source venv/bin/activate
 # Install Python dependencies
 pip install -e ".[test, deploy]"
 
-# Install the npm dependences
+# Install the npm dependences, and build the JS 
 npm install
+npm build
```
marthacryan, line 26: This should be `npm run build`! It throws an error when you run it locally

marthacryan (changes requested):
VERY exciting PR! Requesting changes because the `macsetup.sh` script is broken but it'll be a quick fix. Oh and I agree that mitoinstaller tests are ok to fail, but maybe we should comment them out or something so that we don't have failing CI all the time? After that's fixed LGTM

PR #961, opened 2023-10-19
Update flag to include custom python imports in MitoAnalysis

marthacryan:
# Description
Allows for custom python imports in the `fully_parameterized_function` of the `mito_backend` so that the `MitoAnalysis` class can use custom sheet functions. 

# Testing
See mito-ds/streamlit-recon-wizard#3.

### DIFF
--- a/mitosheet/mitosheet/mito_backend.py
+++ b/mitosheet/mitosheet/mito_backend.py
@@ -133,7 +133,7 @@ class MitoBackend():
             add_comments=False,
             optimize=True,
             code_options_override={
-                'import_custom_python_code': False,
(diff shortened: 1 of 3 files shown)


naterush (changes requested):
@marthacryan can you add a test for this?

naterush:
@marthacryan I think we should also actually have a test that _reruns_ one of these, and makes sure the import is done correctly. 

Can you add a test where the other `.run()` calls is pretty much just a copy of the current test, but it confirms that the `run` function works with these imports? That way we're actually testing the behavior, vs. the code string. 

Then, LGTM!

PR #962, opened 2023-10-19
Glossary test

aarondr77:
# Description

Adds the `excel-to-python/formulas/math/ABS` page. This page is the first of our Excel to Python glossary and is getting released for testing purposes to make sure that the page satisfies the SEO audit before making more. 

Before we make more, we'll refactor this page so that it's a template that receives a page_content prop to populate the page. 

This page is intended to perform well on the following keywords:
- How to ABS Pandas
- Excel ABS in Python
- ABS Pandas
- etc. 

TODO: @naterush I hate the final Mito CTA card. Any ideas?

# Testing

Check out the page. 

# Documentation

No.


trymito.io/components/CodeBlock/CodeBlock.module.css
```diff
@@ -0,0 +1,49 @@
+.container {
```
naterush, line 1: Maybe rename.

trymito.io/components/Glossary/HorizontalNav/HorizontalNav.module.css
```diff
@@ -0,0 +1,9 @@
+.container {
```
naterush, line 1: Although, maybe we don't have to, because they aren't global styles? Even then, I think good practice to make these more specific.

trymito.io/components/Glossary/PageTOC/PageTOC.module.css
```diff
@@ -0,0 +1,14 @@
+.container {
```
naterush, line 1: Same here

trymito.io/components/Glossary/PageTOC/PageTOC.tsx
```diff
@@ -0,0 +1,104 @@
+import React, { useEffect, useState } from 'react';
+import { classNames } from '../../../utils/classNames';
+import pageTOCStyles from './PageTOC.module.css';
+
+type Heading = {
+    level: string;
+    id: string;
+}
+
+const PageTOC = () => {
```
naterush, line 10: This is a really interesting approach to this component. I don't have strong feelings, but I think it's legit. 

The only question is if the on scroll listener is way laggier than it needs to be... there's a rerender of the entire page for every scroll, which seems super unnecessary. 

What if instead, we debounced it? Even checking every 25ms or something is probably good.

trymito.io/components/TextButton/TextButton.module.css
```diff
@@ -22,21 +19,47 @@
   display: inline-block;
 }
 
-.text_button:hover {
+.purple {
```
naterush, line 22: I think we want to ideally not hardcode colors anywhere, if possible. Can we call this something similar to --color-text-primary (e.g. .primary-with-highlight-background or something). Same for the rest of these.

naterush (changes requested):
Great start. This is just an intermediate review, but I think the biggest major factor change we should make is making the entire page parametric in a JSON object description of the content. 

E.g. something like {
    name: 'ABS'
    relatedFunctions: ['SUM', ...],
    description: "The ABS function in Excel takes a single parameters and returns its absolute value.",
    syntax: {parameter: string, description: string, type: string, optional?: bool}[]
    examples: {...}[]
    pandas_header: "Implementing the absolute value in Pandas",
    pandas_examples: {
        subheader: "Calculate the absolute value of every cell in a Pandas series",
        description: "...",
        code: string # this is kinda annoying. Maybe could do a string array?
     }[],
     common_mistakes: // similar to above
}

This way, we can much more quickly change the format of the pages, -- and also build all of them off static props super easy.

trymito.io/pages/excel-to-python/functions/math/ABS.tsx
```diff
@@ -0,0 +1,329 @@
...
+      <Header />
+
+      <div className={pageStyles.container}>
+        <main className={classNames(excelToPythonStyles.main)}>
+          <div className={excelToPythonStyles.content}>
+            <div className={excelToPythonStyles.blog_content}>
+              <section className={classNames(excelToPythonStyles.title_card, excelToPythonStyles.section)}>
+                <div className={excelToPythonStyles.horizontal_navbar_container}>
+                  <GlossayHorizontalNavbar>
+                    <HorizontalNavItem title={'Function'} href={'/spreadsheet-automation'} />
```
aarondr77, line 79: These Nav items will be updated once we build the `Function` and `Math` landing pages. But for testing purposes, this is fine

trymito.io/styles/prism.module.css
```diff
@@ -0,0 +1,101 @@
+/* styles.module.css */
```
aarondr77, line 1: Delete

trymito.io/pages/_app.tsx
```diff
@@ -1,12 +1,22 @@
...
 import Script from 'next/script';
+import Prism from 'prismjs';
+import 'prism-themes/themes/prism-coldark-dark.css'
+
+
+require('prismjs/components/prism-python');
 
 function MyApp({ Component, pageProps }: AppProps) {
 
+  useEffect(() => {
```
aarondr77, line 15: Add a comment here to explain what is going on.

trymito.io/pages/excel-to-python/functions/math/ABS.tsx
```diff
@@ -0,0 +1,329 @@
...
+        <title>{`Excel to Python: ${functionName} - A Complete Guide`}</title>
+        
+        {/* Meta Description */}
+        <meta
+          name="description"
+          content={`Learn how to convert Excel's ${functionName} function to Python using Pandas. This comprehensive guide provides step-by-step instructions and practical examples.`}
+        />
+        
+        {/* Canonical URL (if applicable) */}
+        {/* <link rel="canonical" href={`https://www.example.com/excel-to-python/${function-name}-guide`} /> */}
```
aarondr77, line 42: Decide if we want a canonical url

aarondr77: I don't think we want to use canonical urls to point back to excel-to-python page. 
1. Hypercontext does not use canonical urls to direct their traffic back to the glossary homepage
2. Google reccomends using it if you have the same (or very similar) page implemented multiple times (ie: different languages)

When users search `Pandas ABS` we actually want to surface that ABS page.

aarondr77: Semrush recommends using self-referencing cannoncials if you don't want to direct the page equity somewhere else. Hypercontext also does this, so we will too

trymito.io/components/Glossary/HorizontalNav/HorizontalNav.module.css
aarondr77: I'm going to leave these for now because that's what we use across this codebase.

trymito.io/README.md
```diff
@@ -39,3 +39,18 @@ Check out our [Next.js deployment documentation](https://nextjs.org/docs/deploym
 
 #### Mobile First
 The structure of the codebase is mobile first! Specifically, if you removed all of the css code inside of the media queries, you'd be left with a mobile-only website. This approach worked well because basic, non-formatted HTML is mobile friendly, so if we designed the desktop version first, actually, we are adding css that makes the website non-mobile friendly, and then in the media queries, we're remaking it mobile friendly. Its a waste of code. 
+
+## Developing on for mobile
```
naterush, line 43: Cool!

trymito.io/styles/ExcelToPython.module.css
```diff
@@ -0,0 +1,147 @@
...
+    display: flex;
+    flex-direction: row;
+    align-items: baseline;
+}
+
+.related_functions_card * {
+    margin-right: 10px;
+}
+
+.h3 {
```
naterush, line 32: This is probably going to become a super confusing class name, I think!

trymito.io/styles/ExcelToPython.module.css
```diff
@@ -0,0 +1,147 @@
...
+    font-family: Graphik;
+    font-style: normal;
+    font-weight: 400;
+    margin-top: 2rem;
+    font-size: 1.25rem;
+}
+
+.table {
```
naterush, line 40: Same with these. Class names the same as elements -- let's not do this to ourselves

trymito.io/styles/ExcelToPython.module.css
```diff
@@ -0,0 +1,147 @@
...
+
+.table th,
+.table td {
+  padding: 20px;
+  text-align: left;
+  border: none; 
+  border-bottom: 1.5px solid black; 
+}
+
+.table thead {
```
naterush, line 58: Like here, confusing

trymito.io/components/Glossary/PageTOC/PageTOC.tsx
```diff
@@ -0,0 +1,106 @@
...
+        const headingsArray: Heading[] = Array.from(headingElements).map((heading) => ({
+            level: heading.tagName.toLowerCase(),
+            id: heading.id,
+        })).filter((heading) => heading.id !== "");
+    
+        setHeadings(headingsArray);
+    }, []); // This useEffect only runs once on component mount
+    
+
+    useDebouncedEffect(() => {
```
naterush, line 29: A debounced effect doesn't do what we want here. A debounced effect denounces running the effect when the [header] array changes. We don't want that -- we want to debounce the handleScroll function running when the scroll changes. 

To do so:
1. Remove the useDebouncedEffect (from trymito.io)
2. Debounce the handleScroll function from 'scroll' events

Google probably will be helpful here!

trymito.io/components/Glossary/PageTOC/PageTOC.tsx
```diff
@@ -0,0 +1,106 @@
...
+        };
+    }, [headings], 200);
+        
+
+    const handleItemClick = (headingId: string) => {
+        const targetElement = document.getElementById(headingId);
+
+        if (targetElement) {
+            // Calculate the target position with a 200px offset from the top so it doesn't get blocked by the header
+            const offset = 200;
```
naterush, line 75: Can we use .bottom or something instead, so we don't need to do this manual adjustment?

trymito.io/components/Glossary/PageTOC/PageTOC.tsx
```diff
@@ -0,0 +1,106 @@
...
+                    key={heading.id}
+                    onClick={(e) => handleItemClick(heading.id)}
+                    className={classNames(
+                        pageTOCStyles.item, 
+                        {[pageTOCStyles.indent]: heading.level === 'h3'},
+                        {'text-secondary': activeHeading?.id !== heading.id},
+                        {'text-highlight': activeHeading?.id === heading.id}
+                    )}
+                >
+                    {heading.id}
```
naterush, line 99: Id feels like a weird name for this. heading.name -- since the ID is something different for most of HTML - not something you display.

naterush (changes requested):
Some more cleanup here.

trymito.io/pages/excel-to-python/functions/math/ABS.tsx
```diff
@@ -0,0 +1,353 @@
...
+                  </h3>
+                  <p>
+                    To use the absolute value as part of a more complex operation, you can use the `apply()` function to apply the operation to every element in an pandas dataframe column.
+                  </p>
+                  <CodeBlock code = {`# Define a function to calculate the absolute sum of a row
+def abs_sum(row):
+  return row.abs().sum()
+                    
+# Create a new column 'ABS_SUM' by applying the custom function 
+df['ABS_SUM'] = df.['ABS'].abs(), axis=1)`
```
naterush, line 249: This code is wrong, FWIW

trymito.io/pages/excel-to-python/functions/math/ABS.tsx
```diff
@@ -0,0 +1,353 @@
...
+                  </p>
+                  <CodeBlock code = {`# Calculate the absolute difference between Column1 and Column2
+df['Absolute_Difference'] = (df['Column1'] - df['Column2']).abs()`
+                  }/>
+                  <h3 
+                    id="Complex ABS operations" 
+                    className={classNames(excelToPythonStyles.h3, excelToPythonStyles.link)}
+                  >
+                    Using ABS as part of a more complex operation
+                    <a className={excelToPythonStyles.section_copy} href="#Complex ABS operations">#</a>
```
naterush, line 239: Clicking on these href's scrolls super weird - farther than you want -- on safari.

naterush (changes requested):
Overall, looks really good -- I think this is going to be a helpful resource to a lot of people. 

We don't have to change the plan of attack now, but I do think that getting the page reviewed after we moved to useStaticProps would be a good idea. Namely, this might change how SEO works (we don't fully understand server side page generation yet, but not doing useStaticProps with JSON would make the whole thing not appear on SEO). 

So, we kinda don't really address a major piece of uncertainty. In the future, plan of attack wise, I think we could do less visual polish and the useStaticProps first, and then get a review. This would move up the uncertainty being addressed quicker...

aarondr77:
Deploying now. Will address this review during refactor this week.

trymito.io/pages/excel-to-python/functions/math/ABS.tsx
aarondr77: I'm not sure exactly how to fix this one. So leaving for now.

PR #977, opened 2023-10-24
dash: add RunnableAnalysis to dash v1 interface

naterush:
# Description

Makes it easy for Dash users to access the RunnableAnalysis class, namely for the `from_json` method.

# Testing

Import it.

# Documentation

Doc updates!


marthacryan (approved):
LGTM

PR #979, opened 2023-10-24
mitosheet: make runnable analysis take more flexible types

naterush:
# Description

Makes it so you can pass a dataframe to the `run` function for file path argument, and the `.run` function automatically converts it to a string that can be read by the pd.read_csv call in the function. 

In general, I think should continue to improve the flexibility of this interface. Users should be able to create an analysis from one datasource, and then rerun it easily off of another datasource.

# Testing

See new test.

# Documentation

I think we can make this clear in the docs, just with a little section called **Input Flexibility**


marthacryan (approved):
One small comment about the test but otherwise LGTM!

PR #982, opened 2023-10-24
Glossary 2

aarondr77:
# Description

Move from hardcoded content to json so we can easily add more content that becomes helpful for users. 

# Testing

check it out

# Documentation

No


trymito.io/excel-to-python-page-contents/types.tsx
```diff
@@ -0,0 +1,41 @@
+export type PageContent = {
```
aarondr77, line 1: Could leave a comment that says `see example in folder`, but I think this is fine

trymito.io/utils/arrays.tsx
```diff
@@ -0,0 +1,28 @@
...
+        }
+    })
+
+    return finalArray;
+}   
+
+/*
+    Returns if both arrays contains all of the same values, ignoring order
+*/
+export function arraysContainSameValues<T>(array1: T[], array2: T[]): boolean {
```
aarondr77, line 25: We want to check that the array contain the same values and in the same order...

aarondr77 (commented):
small cleanup then LGTM

Issue #972, opened 2023-10-23
Investigate Dash performance issues


naterush:
**Describe the bug**

Dash is slow. This is likely due to the selection updates, or GC searching. Investigate both of these.

PR #983, opened 2023-10-25
Dash fixes

naterush:
# Description

1. Fixes https://github.com/mito-ds/mito/issues/972 (we need to make selection an opt-in thing)
2. Makes the input data to the mitosheet more flexible. We don't yet need to promote this (because there are other things we need to fix), but generally I think it's really nice to just be able to slap anything at all into a `mitosheet.sheet` call (or whatever else) and get a working Mitosheet.

# Testing

See new tests. 

# Documentation

We'll need to update the documentation for filtering.


aarondr77:
If I use this code:

```
csv_string = """Name,Age,Occupation
Alice,28,Engineer
Bob,34,Data Scientist
Charlie,45,Teacher
Daisy,31,Doctor
"""
mitosheet.sheet(csv_string)
```

It generates invalid code: 

Screenshot 2023-10-26 at 11 35 43 AM

aarondr77:
If I generate this JSON string with records orientation, 

```
import pandas as pd

# Create a sample DataFrame
data = {'Name': ['Alice', 'Bob', 'Charlie'],
        'Age': [25, 30, 35]}

df = pd.DataFrame(data)

# Convert the DataFrame to a JSON string
json_string = df.to_json(orient='records')

# Print the JSON string
print(json_string)

mitosheet.sheet(json_string)
```

It generates an incorrect dataframe: 

Screenshot 2023-10-26 at 11 38 04 AM

And the generated code is invalid: 

Screenshot 2023-10-26 at 11 41 42 AM

aarondr77:
# df.to_json (without (orient='records')) generates invalid code

```
import pandas as pd

# Create a sample DataFrame
data = {'Name': ['Alice', 'Bob', 'Charlie'],
        'Age': [25, 30, 35]}

df = pd.DataFrame(data)

# Convert the DataFrame to a JSON string
json_string = df.to_json()

# Print the JSON string
print(json_string)

mitosheet.sheet(json_string)

```

aarondr77 (commented):
This PR should probably be two PRs. 

1. PR 1 - track_selection updating: This one is working well and LGTM. Could we add a test for it?
2. New input types to the mitosheet.sheet() call. This one is pretty buggy atm. 

Want to break this into two separate PRs, merge the track_selection one which is higher priority and then fix the other one later or kill it? If we keep the selection one around, let's heavily test it to make sure we don't break the mitosheet.sheet() call stuff which you've called out before as a high risk area + add some good documentation to make it usable. 

My vote would be to kill the second PR for now, fwiw. Seems like a couple more hours of fixing it up and testing it heavily that is not worth it rn. Its not associated with a issue, right?

PR #988, opened 2023-10-26
Glossary optimizations

aarondr77:
# Description

Remove intercom from the website since we don't use it anymore. 

# Testing

# Documentation

### DIFF
--- a/trymito.io/pages/excel-to-python/[...slug].tsx
+++ b/trymito.io/pages/excel-to-python/[...slug].tsx
@@ -141,7 +141,7 @@ const ExcelToPythonGlossaryPage = (props: {pageContent: PageContent}) => {
                     id={`Excel's ${functionNameShort} Function`}
                     className={excelToPythonStyles.link}
                   >
-                    Understanding the {pageContent.functioNameLong} Function in Excel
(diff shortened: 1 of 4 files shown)


trymito.io/pages/excel-to-python/[...slug].tsx
```diff
@@ -138,11 +138,11 @@ const ExcelToPythonGlossaryPage = (props: {pageContent: PageContent}) => {
               {/* Understanding the Excel Function */}
               <section className={excelToPythonStyles.section}>
                   <h2 
-                    id={`Understanding Excel's ${functionNameShort} Function`}
+                    id={`Excel's ${functionNameShort} Function`}
                     className={excelToPythonStyles.link}
                   >
-                    Understanding Excel&apos;s {functionNameShort} Function
-                    <a className={excelToPythonStyles.section_copy} href={`#Understanding Excel's ${functionNameShort} Function`}>#</a>
+                    Understanding the {pageContent.functioNameLong} Function in Excel
```
Comment (naterush) [inline trymito.io/pages/excel-to-python/[...slug].tsx:144]: `is functioNameLong` the right name? Can we rename this to `functionNameLong` everywhere if so?

naterush (changes requested):
Minor cleanup, then LGTM

PR #990, opened 2023-10-26
Redesign toolbar

marthacryan:
# Description
Moves the existing toolbar into the "Home" tab in preparation for moving the main menu contents into tabs. 
image

# Testing
Open in a notebook and try out interacting with the tab bar / toolbar buttons! I added a "test" tab so that you can try out switching tabs.


mitosheet/css/toolbar.css
```diff
@@ -181,4 +181,21 @@
...
+/* Tab Bar Style */
+.mito-toolbar-tabbar-button {
+  border: none;
+  background: none;
+  color: var(--mito-text-medium);
+}
+
+.mito-toolbar-tabbar-button-selected span {
+  border-bottom: 3px solid var(--mito-highlight);
+  color: var(--mito-gray);
```
naterush, line 196: Can we pick a better color for this? One of the parametric ones.

mitosheet/css/toolbar.css
```diff
@@ -181,4 +181,21 @@
   padding: 0px 5px;
   color: var(--mito-background);
   height: 18px;
+}
+
+
+/* Tab Bar Style */
+.mito-toolbar-tabbar-button {
```
naterush, line 188: rather than overloading button, maybe we rename this to a `tabbar-tabname` (and below).

mitosheet/css/toolbar.css
```diff
@@ -181,4 +181,21 @@
...
+  border: none;
+  background: none;
+  color: var(--mito-text-medium);
+}
+
+.mito-toolbar-tabbar-button-selected span {
+  border-bottom: 3px solid var(--mito-highlight);
```
naterush, line 195: In the mockups, we round the corners of these. I think it's a nice bit of polish.

mitosheet/src/mito/components/toolbar/HomeTabContents.tsx
```diff
@@ -0,0 +1,326 @@
...
+export const HomeTabContents = (
+    props: {
+        mitoAPI: MitoAPI
+        currStepIdx: number;
+        lastStepIndex: number;
+        highlightPivotTableButton: boolean;
+        highlightAddColButton: boolean;
+        actions: Actions;
+        gridState: GridState;
+        setGridState: React.Dispatch<React.SetStateAction<GridState>>;
```
naterush, line 29: setGridState is not used

mitosheet/src/mito/components/toolbar/Toolbar.tsx
```diff
@@ -53,50 +46,16 @@ export const Toolbar = (
         closeOpenEditingPopups: () => void
     }): JSX.Element => {  
 
+    const [currentTab, setCurrentTab] = React.useState<string>('Home');
+    const tabs: {[ tab: string ]: React.ReactElement } = {
```
naterush, line 50: Can we make `type ToolbarTab = 'Home' | ...` ... and use it where proper?

mitosheet/src/mito/components/toolbar/Toolbar.tsx
```diff
@@ -53,50 +46,16 @@ export const Toolbar = (
         closeOpenEditingPopups: () => void
     }): JSX.Element => {  
 
+    const [currentTab, setCurrentTab] = React.useState<string>('Home');
```
naterush, line 49: if you make it <ToolbarTab | undefined>, then it's really natural to express the closing operation (it's just undefined)

naterush (changes requested):
Sweet, great start. Minor cleanup, then this is ready to go. 

**Let's not merge this until we get the release out for this week, which I'm working on! But you can branch off of this PR going forward :)**

aarondr77:
The unselected tab names look disabled to me, unlike in Excel. 
Screenshot 2023-11-07 at 5 55 53 PM

Screenshot 2023-11-07 at 5 57 42 PM

Can we use the same text color for selectable tab headers as we do for the selectable toolbar buttons?

aarondr77:
When you hover over the tab headers, it shifts the placement of the rest of the tab headers because the bold text takes up more width. 

I think an easy fix would be to put each of the sheet tab headers inside a div that has a fixed with so that even if the text width changes the divs don't change sizes

https://github.com/mito-ds/mito/assets/18709905/cf72ea8e-a9fe-4a39-971b-1e34810afe81

aarondr77:
Can we update the highlighting so that it as least more closely uniform around the icon. 

In the Import icon, the highlighting pretty much does not exist above icon. You could probably either adjust the icon or find a better padding size? 

Screenshot 2023-11-07 at 5 45 50 PM

aarondr77:
To make it easier to click these buttons, it would be nice if the user could hover over more of the toolbar to select the button. See in the video: 

https://github.com/mito-ds/mito/assets/18709905/cd4f683a-dbcb-4c2f-958e-de8de7e7145c

aarondr77:
Can we add tooltips when the elements are disabled that explains these are disabled because you don't have a number column selected. 

Screenshot 2023-11-07 at 6 48 28 PM

aarondr77:
The current implementation of these buttons is reverse from Excel. 

Screenshot 2023-11-07 at 6 52 12 PM

aarondr77:
The Undo and Redo keyboard shortcuts are broken, but others like Find and Replace are working.

aarondr77:
Instead of showing the conditional formatting icon here, I think it's much better for the user if we continue showing the text color and background color that the user sets. 

Screenshot 2023-11-07 at 6 55 50 PM

Previously, we showed the text color (red) and background color (blue) like this: 
Screenshot 2023-11-07 at 6 57 13 PM

aarondr77:
You've already called this out, and I agree. The vast majority of adding and deleting events in Mito are on `columns` instead of `rows`. Let's make an icon that looks very similar to these, but is oriented for columns

Screenshot 2023-11-07 at 7 02 04 PM

aarondr77:
Two updates here:

1. There should be vertical divider bar after the `Delete` icon
2.  In Excel, the order goes `Filter` then `Find and Replace`, so lets go `Filter` `Find and Replace` `Dtype`

Screenshot 2023-11-07 at 7 03 59 PM

aarondr77:
Instead of using this new graph icon in the sheet tab, let's keep the old one. Its a bit overwhelming in the bottom

Screenshot 2023-11-07 at 7 06 21 PM

aarondr77:
### Height 

We used to show 13 rows of data, but now we only show 11 because the height of the grid component itself decreased. I think this is not ideal for two reasons: 
1. Being able to see your data is one of our most prominent value props! That's why entering full screen mode is one of the most used features. 
2. There are lots of taskpanes (merge, pivot, graph, etc) that contain a lot of information vertically and already require a lot of scrolling to use. 

Probably as a result of height change and some hardcoding, if you look at the footer, the selected tab and divider line doesn't take up the full height. There's a little bit of cleanup to do here. 
Screenshot 2023-11-07 at 7 07 22 PM


I think we can win back at least some of the vertical space

This is what we currently have:
Screenshot 2023-11-07 at 7 18 06 PM

Just by reducing some of the margins/padding in the toolbar (which I think still looks pretty good) we can get this:
Screenshot 2023-11-07 at 7 17 50 PM

aarondr77:
I think these three icons need to be remade. 

1. In the Undo/Redo icons: when you zoom in you can sere the separation between the arrow tip and the arrow body
2. In the clear icon: I don't think we need the black X inside the white X and the right hand side of the icon is cut off. 

Screenshot 2023-11-07 at 7 21 28 PM

aarondr77:
Maybe this icon would look cool if the center of the mirror was black?

Screenshot 2023-11-07 at 7 23 30 PM

aarondr77:
I know you're already tracking this, but just adding here as well: 

Let's do `pivot` `merge` `graph` `ai`

Where the `merge` is a dropdown button with three options:
1. Horizontal Merge
2. Vertical Concat 
3. Anti Merge

We could probably come up with better names than those too!

Screenshot 2023-11-07 at 7 24 10 PM

aarondr77:
I think we could use a lighter background color for the shading. 

Screenshot 2023-11-07 at 7 26 38 PM

Something close to this: 
Screenshot 2023-11-07 at 7 27 00 PM

aarondr77:
These icons have the background included by accident. They should only appear on hover

Screenshot 2023-11-07 at 7 27 38 PM

aarondr77:
I think these icons are a little bit small, especially the lighter `A` are hard to read. Can we make them a tad bigger?

Screenshot 2023-11-07 at 7 31 33 PM

aarondr77:
This icon is the correct one to use for `Filter and Sort` on the Home tab, but not for the `Filter` icon on the `Data` tab. On the Data tab, it should just have the funnel, not the A Z part

Screenshot 2023-11-07 at 7 32 37 PM

aarondr77:
I think these two icons make it seem like these two features are related, when they are not. 

Can we think of a different icon for FillNaN values

Screenshot 2023-11-07 at 7 33 56 PM

aarondr77:
Something that just tripped me up was that there isn't an easy way to add a new column when you're writing a formula using this tab. 

Screenshot 2023-11-07 at 7 36 05 PM

In Excel, there is an `Insert Function` button here. 
Screenshot 2023-11-07 at 7 37 03 PM

Maybe we could add a button `Insert Column` that creates a new column and opens the cell editor to `=`

aarondr77:
I love this checkmark feedback! Nice work

https://github.com/mito-ds/mito/assets/18709905/3511f14a-8ca8-483b-bea2-c5a678162491

aarondr77:
The height of the toolbar is slightly different on different tabs. See here that when I switch between the `Formulas` and `Data` tabs the height changes slightly. 

https://github.com/mito-ds/mito/assets/18709905/74984c65-a9b5-4010-82d7-4eecd1bc8bd8

aarondr77:
Let's make this text unelectable so this doesn't happen. I think `user-select: none;` will do it. 

Screenshot 2023-11-07 at 7 40 56 PM

aarondr77:
If there is a column selected and then you select a function the formula gets entered as `=CLEAN(column_header` but if you just have a cell selected, then the formula is entered as `=CLEAN(`

I think it makes sense to always generate the formula  `=CLEAN(`


https://github.com/mito-ds/mito/assets/18709905/9a24ecd1-f71b-49f3-b09a-e693adbd83f6

aarondr77:
The dropdown should be wider to support longer formulas, ie `Start of business month`


Screenshot 2023-11-07 at 7 49 13 PM

aarondr77:
1. Open a cell editor
2. Use the toolbar to select a formula and see that its added
3. Try to continue typing and notice that the the cell editor no longer has focus in the cell editor. 

I think we should always add the focus to the open cell after we use the formulas toolbar to add a formula

https://github.com/mito-ds/mito/assets/18709905/7185932b-6df9-4ce6-a907-22f5f2b31c97

aarondr77:
As much as I love our Mito purple color, I don't love staring at a giant purple bar at the top of the my spreadsheet. I think something like this is much better, and then continuing to use the purple color as an accent. Maybe experiment with a few of the colors we already have in our color scheme and decide which is best? 

Screenshot 2023-11-07 at 8 01 22 PM

Hmm... Black is pretty cool too! 
Screenshot 2023-11-07 at 8 03 17 PM

aarondr77:
Can we add back the `step history` button to the toolbar? I think it could live directly to the left of the `find and replace` and `fullscreen` cluster. It's not the most used feature, but I know that some of our power users really like it.

aarondr77:
One hunch I have is that the difference between the `Insert` and `Data` tabs are going to be confusing to users. Let's pay attention to this!

Another structure could be:
1. `Data`: All import, export, change imported data
2. `Transform`:  pivot tables, merge, filter, sort, etc. 

Idk where graphs would go in that grouping though... 

Anyways, I think the tree tests are going to be useful for learning about this! Looking forward to checking them out.

aarondr77:
Ooooo I would be down to add the text `Auto Save` in the top toolbar. One question I often get is, "Is this saving automatically?" 

Excel has it here. We could add the text "Auto Save" (without the ability to turn it off), and then a tooltip that gives a little bit more context when you hover over it. 

Screenshot 2023-11-07 at 8 13 13 PM

aarondr77:
Noticed this bug while testing. https://github.com/mito-ds/mito/issues/1018 However, its probably not related to this PR, right?

marthacryan:
> Can we add tooltips when the elements are disabled that explains these are disabled because you don't have a number column selected.

I think there is? Are there specific buttons that aren't saying that?

marthacryan:
> Maybe this icon would look cool if the center of the mirror was black?
> 
> Screenshot 2023-11-07 at 7 23 30 PM

I might just remove the fill altogether?

marthacryan:
> You've already called this out, and I agree. The vast majority of adding and deleting events in Mito are on `columns` instead of `rows`. Let's make an icon that looks very similar to these, but is oriented for columns
> 
> Screenshot 2023-11-07 at 7 02 04 PM

How does this look? 
image

marthacryan:
> As much as I love our Mito purple color, I don't love staring at a giant purple bar at the top of the my spreadsheet. I think something like this is much better, and then continuing to use the purple color as an accent. Maybe experiment with a few of the colors we already have in our color scheme and decide which is best?

I agree that the 

> This icon is the correct one to use for `Filter and Sort` on the Home tab, but not for the `Filter` icon on the `Data` tab. On the Data tab, it should just have the funnel, not the A Z part
> 
> Screenshot 2023-11-07 at 7 32 37 PM

To be clear though, these two buttons are doing the same thing. Should they be doing something different?

marthacryan:
> Probably as a result of height change and some hardcoding, if you look at the footer, the selected tab and divider line doesn't take up the full height. Neither does the graph tabs. There's a little bit of cleanup to do here. Screenshot 2023-11-07 at 7 07 22 PM

I'm not able to reproduce this - what browser are you in? Are you in JL or streamlit?

PR #991, opened 2023-10-27
Add style to home tab

marthacryan:
# Description
Adds style to the "home" tab in the toolbar. 
* Adds a small icon to indicate that a button has a dropdown
* Moves the number formatting into its own section 
* Changes the number formatting from a button to a select so that the user can see what the current number formatting is

# Testing
Play around with the toolbar

# Documentation


mitosheet/css/toolbar.css
```diff
@@ -135,8 +135,8 @@
   font-family: Roboto;
 }
 
-.mito-toolbar-button-container-enabled:hover {
-  background-color: var(--mito-highlight-light);
+.mito-toolbar-button-container-enabled:hover .mito-toolbar-button-icon-container {
+  background-color: var(--mito-light-gray);
```
naterush, line 139: This should be `--mito-very-light-gray`. It's darker than the mockup.

mitosheet/src/mito/types.tsx
```diff
@@ -997,7 +997,7 @@ export interface BaseAction<Type, StaticType> {
     type: Type;
     staticType: StaticType;
     // The short title for the action. Should be title case, as you want to display it.
-    shortTitle: string
+    shortTitle?: string
```
naterush, line 1000: Why is this undefined? Let's just always have one of these, ya?

naterush (changes requested):
# A New Plan of Attack

1. a PR for Insert tab, just correct buttons (this one is pretty much good for the Home tab)
2. a PR for each other tab, one by one
3. a PR for the top purple bar thing (moving the undo, redo, and clear buttons up) among other things
4. then, a big ole styling PR. Where we just polish polish polish. 

This way, we can make sure that our styling approach is reasonable and works across the board, and we don't have to keep bouncing back and forth. 

That being said, here are a few small styling notes -- we don't need to fix them up now! Let's get the other "tab PRs" up first, and we can fix these all up at the end. 

## Styling Things

### Format Buttons Vertical Spacing 

See mockup vs. actual. In general, I think we could do with making the full-height icons _bigger_, and then in turn this would give us more space for the double row. 

If you're adjusting icon height, I think it's probably easiest to do in the Figma mockups to make sure things are consistent. 

Screen Shot 2023-10-27 at 11 30 36 AM

Screen Shot 2023-10-27 at 11 29 31 AM

### Search is too high

Screen Shot 2023-10-27 at 11 36 02 AM

### There should be an on-hover on the ToolbarTabName

When you hover over a tab name, the purple thing should appear below it -- so you know you can click it.

naterush:
I feel like Home could be the same sized text as Edit is currently. Ik we're getting rid of Edit, but font-sizing wise it feels appropriate for these tabs.
Screen Shot 2023-10-27 at 11 46 51 AM

naterush:
Some additions and removals just focusing on current functionality:
1. We can remove the steps button. 
2. Search should be moved over to where it is in the mockups.

naterush:
Screen Shot 2023-10-27 at 11 49 05 AM

Since we're setting up the tab name div structure now, I think now is a reasonable time to move these buttons down to be in-line with the tab names.

mitosheet/css/toolbar.css
marthacryan: Oh yeah I think `--mito-very-light-gray` is actually the current background color of the toolbar. Maybe I can make an in-between gray?

mitosheet/src/mito/types.tsx
marthacryan: For toolbar buttons that don't have text displayed (i.e. more/less precision)

marthacryan:
> I feel like Home could be the same sized text as Edit is currently. Ik we're getting rid of Edit, but font-sizing wise it feels appropriate for these tabs.
> 
> Like in Excel, where tab names are bigger than button text.
> 
> Screen Shot 2023-10-27 at 11 46 51 AM

Ah yeah looks different in chrome - I'll add explicit font size!

mitosheet/src/mito/types.tsx
naterush: Can you leave a comment here about that?

mitosheet/css/elements/SearchBar.css
```diff
@@ -1,6 +1,6 @@
 .mito-search-bar {
     position: absolute;
-    top: 72px;
+    top: 117px;
```
naterush, line 3: This height has to be dynamic somehow, I think. Cus when the toolbar is closed it's at the wrong height. Wanna at least open an issue so we can come back to it in the future?

mitosheet/css/toolbar.css
```diff
@@ -135,8 +134,21 @@
...
+}
+
+.mito-toolbar-tabbar-tabname span {
```
naterush, line 141: I think these are still missing the :hover selectors for making it clear they are clickable, when closed.

mitosheet/src/mito/types.tsx
```diff
@@ -997,7 +997,8 @@ export interface BaseAction<Type, StaticType> {
     type: Type;
     staticType: StaticType;
     // The short title for the action. Should be title case, as you want to display it.
-    shortTitle: string
+    // If undefined, the toolbar will not display a title in the toolbar.
```
naterush, line 1000: Can we rename shortTitle to just `toolbarTitle`? I think this removes the need for a comment actually.

mitosheet/css/toolbar.css
```diff
@@ -200,13 +219,59 @@
...
+  font-size: 11px;
+}
+
+.mito-toolbar-container .select-container {
+  display: flex;
+  flex-direction: row;
+  height: 22px;
+  box-sizing: border-box;
+  width: 150px;
+}
```
naterush, line 261: Can we actually modify the select component to take a `height` prop? That way, we're not modifying the CSS from an unexpected location, and the component is well contained.

naterush: This goes for all the select css modification in the file.

naterush (changes requested):
Minor cleanup, then LGTM! Looks pretty great.

mitosheet/css/toolbar.css
naterush: I think it's actually ok to leave as is for now.

mitosheet/css/elements/SearchBar.css
marthacryan: Ah ok yes

marthacryan: I'll add a note in the spec

mitosheet/src/mito/types.tsx
marthacryan: Updated the name! Leaving the comment in because I think it's expressing something slightly different

PR #992, opened 2023-10-27
Glossary functions page

aarondr77:
# Description

- Adds the home page for the Excel to Python glossary 
- Makes the related functions buttons works
- Adds this glossary to the `About` toolbar because it's pretty cool and useful

With this, I think we are finished all of the required infrastructure for these pages + we've implemented all of the most import page optimizations, so we are good to write more useful pages for functions that people are commonly looking to create! 

# Testing

# Documentation


naterush (commented):
Code looks perf.

PR #998, opened 2023-10-30
Move icons into the action type

marthacryan:
# Description
This adds an optional field into the `BaseAction` type to prepare for adding icons for many of the actions that are currently in the main menu at the top. Defaults to the `StepsIcon` because it looks pretty generic. 

Also, adds the default behavior for `disabledTooltip` into the `ToolbarButton` class to avoid a bunch of extra lines of code for passing it as a prop. 

# Testing
Check that the toolbar has all of the icons you expect.

### DIFF
--- a/mitosheet/src/mito/components/toolbar/utils.tsx
+++ b/mitosheet/src/mito/components/toolbar/utils.tsx
@@ -3,25 +3,7 @@ import React from 'react';
 
 import { Action, UserProfile } from '../../types';
 import DropdownItem from '../elements/DropdownItem';
-import AIIcon from '../icons/AIIcon';
-import AddColumnIcon from '../icons/AddColumnIcon';
-import CatchUpIcon from '../icons/CatchUpIcon';
-import ClearIcon from '../icons/ClearIcon';
-import CodeSnippetIcon from '../icons/CodeSnippetIcon';
-import DeleteColumnIcon from '../icons/DeleteColumnIcon';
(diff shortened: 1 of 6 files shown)


mitosheet/src/mito/components/toolbar/utils.tsx
```diff
@@ -31,74 +13,12 @@ import UndoIcon from '../icons/UndoIcon';
...
-        case ToolbarButtonType.CODE_SNIPPETS: {return <CodeSnippetIcon />}
-
-        case ToolbarButtonType.CATCH_UP: {return <CatchUpIcon />}
-        case ToolbarButtonType.STEPS: {return <StepsIcon />}
-        case ToolbarButtonType.OPEN_FULLSCREEN: {return <OpenFullscreenIcon />}
-        case ToolbarButtonType.CLOSE_FULLSCREEN: {return <CloseFullscreenIcon />}
-        case ToolbarButtonType.OPEN_SEARCH: {return <SearchIcon />}
-    }
+export const getToolbarItemIcon = (action: Action): JSX.Element => {
+    return action.icon !== undefined ? <action.icon /> : <StepsIcon />
```
naterush, line 21: I think this could be updated to `getToolbarFallbackIcon` or something

naterush (changes requested):
Super minor cleanup, then LGTM. This is a great refactor!

PR #999, opened 2023-10-30
Add conditional format / format buttons to toolbar

marthacryan:
# Description
Adds buttons to the toolbar for conditional formatting, color formatting, copy, find&replace, and sort/filter. Also updates many of the icons to be consistent with the mockup. 

image

# Testing
See toolbar buttons. 

# Documentation


naterush:
# Dropdowns in toolbar are too far down

Screen Shot 2023-10-30 at 3 35 01 PM

Because the dropdowns place them right below the element they are the children of, I think this is likely just that the children of the dropdown button are at the wrong div level and we need to nest them a bit more.

naterush:
# Copy should be disabled if there is no selected data

Screen Shot 2023-10-30 at 3 36 59 PM

It might already exist, but it seems like we want a utility called `isAnySelectedData`

naterush:
# Increase/decrease precision should be deactivated if there is no data selected 

Screen Shot 2023-10-30 at 3 38 22 PM

Similar to above.

naterush:
Other buttons that should be properly disabled if there is no data:
1. Export
2. Conditional formatting
3. Dataframe formatting
4. Insert
5. Delete
6. Find and replace
...

I have a feeling this code is just broken in one place globally.

naterush:
# AI text is not centered

Screen Shot 2023-10-30 at 3 41 28 PM

naterush:
Per the spec, I think the delete button should delete a row if a row is selected. This requires changing the action.

mitosheet/src/mito/components/toolbar/HomeTabContents.tsx
```diff
@@ -146,6 +146,10 @@ export const HomeTabContents = (
                         }}/>
                 </Dropdown>
             </ToolbarButton>
+            <ToolbarButton
```
naterush, line 149: Nice, this file is very easy to edit, and the resulting file looks super clear + clean + easy to maintain going forward!

mitosheet/css/toolbar.css
```diff
@@ -203,6 +207,11 @@
   height: 18px;
 }
 
+p.mito-toolbar-button-label {
```
naterush, line 210: Do we need to select by p? Or can we just select the class.

naterush (changes requested):
Minor code changes, a few bugs around disabled / spacing, and then LGTM!

mitosheet/src/mito/components/icons/ConditionalFormatIcon.tsx
```diff
@@ -4,12 +4,19 @@
 import React from 'react';
 
 const ConditionalFormatIcon = (props: {backgroundColor: string, color: string}): JSX.Element => {
+    // TODO: use the props. Including this console log so that the linter doesn't complain
+    console.log(props)
```
naterush, line 8: We'd rather not spam the console -- I think you can just do `props` or if we don't use the props just remove them for now.

mitosheet/src/mito/components/toolbar/HomeTabContents.tsx
```diff
@@ -146,6 +146,10 @@ export const HomeTabContents = (
                         }}/>
```
naterush, line 146: In this file, the class `mito-toolbar-number-format-select` is reference but not used. I like using the `style` on the select in the way you do, though!

naterush (commented):
A few remaining things:
1. Find and replace should be disabled if there is no imported data
2. Missing the $ and % buttons -- let's add these now!
3. Overflow scroll is almost working perfectly (this is so sweet) -- overlap on undo/redo, but since these are moving soon we can just ignore.

This is looking freakin great!

PR #1003, opened 2023-10-31
Add insert tab to toolbar

marthacryan:
Adds a tab for "Insert" to the toolbar that looks like this: 
image

Todo:
- [x] Add special behavior for "Anti-Merge"
- [x] Add special behavior for 3 special graph types


mitosheet/src/mito/components/toolbar/InsertTabContents.tsx
```diff
@@ -0,0 +1,80 @@
...
+        sheetIndex: number,
+        closeOpenEditingPopups: () => void
+    }): JSX.Element => {
+
+    return (<div className='mito-toolbar-bottom'>
+        <ToolbarButton
+            action={props.actions.buildTimeActions[ActionEnum.Pivot]}
+            setEditorState={props.setEditorState}
+        />
+        <div className='mito-pivot-buttons'>
```
naterush, line 35: I don't see this class anywhere, and I missing it?

mitosheet/src/mito/types.tsx
```diff
@@ -964,6 +967,7 @@ export enum ActionEnum {
     Rename_Column = 'rename column',
     Rename_Dataframe = 'rename dataframe',
     Rename_Graph = 'rename graph',
+    Scatter_Plot = 'scatter graph',
```
naterush, line 970: I would be down to rename these actions to GRAPH_BAR, GRAPH_LINE, etc -- just so alphabetically they are close.

naterush: And moving in the actions.tsx file.

naterush (changes requested):
Minor code cleanup, but generally looks pretty good.

naterush:
Screen Shot 2023-11-01 at 9 58 13 AM

I think you maybe forgot to push some CSS!

marthacryan:
> Screen Shot 2023-11-01 at 9 58 13 AM
> I think you maybe forgot to push some CSS!

Oh weird I think this is a safari vs chrome thing. I'll fix it!

naterush:
I think we could have a bit more horizontal space between the transpose button and the divider. Let's make the divider just always handle spacing itself, with some margins?
Screen Shot 2023-11-01 at 10 57 35 AM

naterush:
Noting that icons might need some tweaks in the future (merge and concat aren't super legible), but we can fix that up late.

PR #1004, opened 2023-11-01
Add Data tab

marthacryan:
Adds a tab to the new toolbar for "Data"
image

Todo:
- [x] Add ascending / descending sort
- [x] Fix "fill nan" and "one hot encoding" to just use one line


mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1230,10 +1247,51 @@ export const getActions = (
...
+                return doesColumnExist(startingColumnID, sheetIndex, sheetDataArray) ? defaultActionDisabledMessage : 'There are no columns to sort in the selected sheet. Add data to the sheet.'
+            },
+            searchTerms: ['sort', 'ascending', 'arrange'],
+            tooltip: "Sort a column in ascending order."
+        },
+        [ActionEnum.SortDescending]: {
+            type: 'build-time',
+            staticType: ActionEnum.SortDescending,
+            icon: SortDescendingIcon,
+            longTitle: 'Sort column alphabetically reverse',
```
naterush, line 1273: Sort column reverse alphabetically

naterush (approved):
Code looks great.

naterush:
This clear button was actually meant to be a "clear filters" icon. I think that we can actually remove this for now, so let's just chop this one...

Screen Shot 2023-11-01 at 2 41 39 PM

naterush:
Screen Shot 2023-11-01 at 2 43 11 PM
 
Could we get Fill Missing on one one line?

naterush:
Screen Shot 2023-11-01 at 2 44 43 PM

Some spacing got screwed up on the Insert tab. **Now that I think about it, let's just remove the Bar graph icon (the top of the three vertical). It doesn't do anything unique, and I think we should try and keep things at most two high stacked -- we don't need to go over that.**

mitosheet/src/mito/components/toolbar/DataTabContents.tsx
```diff
@@ -0,0 +1,133 @@
...
+                        }
+
+                        return {
+                            ...prevUIState,
+                            toolbarDropdown: undefined
+                        }
+                    })
+                }
+                // If there are any custom importers, we want to make the dropdown wider
+                width={props.actions.runtimeImportActionsList.length > 0 ? 'large' : 'medium'}
```
naterush, line 116: I don't think is true for the reset index dropdown

mitosheet/src/mito/components/toolbar/DataTabContents.tsx
```diff
@@ -0,0 +1,133 @@
...
+        />
+        <ToolbarButton
+            action={props.actions.buildTimeActions[ActionEnum.Dataframe_Import]}
+            setEditorState={props.setEditorState}
+        />
+        <ToolbarButton
+            action={props.actions.buildTimeActions[ActionEnum.SNOWFLAKEIMPORT]}
+            setEditorState={props.setEditorState}
+        />
+
```
naterush, line 45: Ah -- one thing that's missing here that a) used to represented in the previous toolbar, and b) is in the mockup for the spec is that we want to get the runtime import actions and display them here. 

This should be pretty easy, you can just use the default import icon for all of them for now (and we can get better at this in the future).

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1576,6 +1637,30 @@ export const getActions = (
...
+
+                setUIState(prevUIState => {
+                    return {
+                        ...prevUIState,
+                        toolbarDropdown: 'reset-index'
+                    }
+                })
+            },
+            isDisabled: () => {
+                return doesAnySheetExist(sheetDataArray) ? defaultActionDisabledMessage : 'There are no dataframes to export. Import data.'
```
naterush, line 1659: This disabled message is wrong

naterush (changes requested):
Bit of cleanup, then LGTM -- no need for another review. Custom importers documentation can be found here: https://docs.trymito.io/how-to/importing-data-to-mito/import-generated-ui-from-any-python-function

PR #1010, opened 2023-11-02
Add Formula tab

marthacryan:
Adds a tab to select formulas:
image
image


mitosheet/src/mito/components/toolbar/FormulaTabContents.tsx
```diff
@@ -0,0 +1,111 @@
...
+        setEditorState: React.Dispatch<React.SetStateAction<EditorState | undefined>>;
+        analysisData: AnalysisData,
+        sheetIndex: number,
+        closeOpenEditingPopups: () => void
+    }): JSX.Element => {
+
+    /**
+     * Maps the formula categories to the corresponding action enum
+     */
+    const formulaCategories: { [category: string]: ActionEnum } = {
```
naterush, line 38: Can we make a `FuncitonCategory` type?

mitosheet/src/mito/components/toolbar/FormulaTabContents.tsx
```diff
@@ -0,0 +1,111 @@
...
+                            }
+
+                            return {
+                                ...prevUIState,
+                                toolbarDropdown: undefined
+                            }
+                        })
+                    }
+                    // If there are any custom importers, we want to make the dropdown wider
+                    width={props.actions.runtimeImportActionsList.length > 0 ? 'large' : 'medium'}
```
naterush, line 97: Don't think we want this fella

mitosheet/src/mito/components/toolbar/FormulaTabContents.tsx
naterush: - And then you could make `undefined` one valid type. 
- Then, you could add Formulas_Dropdown_More part of this map. 
- Then, you could avoid the special case for Formulas_Dropdown_More in render at the bottom.

mitosheet/src/mito/types.tsx
```diff
@@ -890,7 +890,7 @@ export interface UIState {
     selectedTabType: 'data' | 'graph';
     currOpenToolbarDropdown: undefined | ToolbarDropdowns;
     highlightedColumnIndex?: number;
-    toolbarDropdown: 'import' | 'format' | 'dtype' | 'export' | 'reset-index' | undefined;
+    toolbarDropdown: 'import' | 'format' | 'dtype' | 'export' | 'reset-index' | 'MATH' | 'LOGIC' | 'FINANCE' | 'DATE' | 'TEXT' | 'REFERENCE' | 'MORE' | undefined;
```
naterush, line 893: For the actual toolbar dropdown type, I'd be down to make these like `formula-math` or `formula-logic`. Just because `MORE` is not super informative...

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -629,6 +635,168 @@ export const getActions = (
...
         },
+        [ActionEnum.Formulas_Dropdown_Math]: {
+            type: 'build-time',
+            staticType: ActionEnum.Formulas_Dropdown_Math,
+            icon: MathFunctionsIcon,
+            longTitle: 'Math Formulas',
+            toolbarTitle: 'Math & Trig',
+            actionFunction: () => {
+                // We turn off editing mode, if it is on
+                setEditorState(undefined);
```
naterush, line 646: I think we _don't_ want to do this in these functions. We want the user to be able to click these dropdowns if they have an open cell editor, methinks....

mitosheet/src/mito/components/toolbar/FormulaTabContents.tsx
```diff
@@ -0,0 +1,111 @@
...
+                <DropdownItem
+                    title={functionObject.function}
+                    key={functionObject.function}
+                    onClick={(e) => {
+                        e?.stopPropagation();
+                        const rowIndex = props.gridState.selections[0].startingRowIndex;
+                        const columnIndex = props.gridState.selections[0].startingColumnIndex;
+                        const {startingColumnFormula, arrowKeysScrollInFormula, editingMode} = getStartingFormula(props.sheetData, props.editorState, rowIndex, columnIndex);
+                        const newFormula = `=${functionObject.function}(${startingColumnFormula.startsWith('=') ? startingColumnFormula.substring(1) : startingColumnFormula}`;
+                        props.setEditorState({
```
naterush, line 61: I feel like we should handle the case where the formula editor is already open? That way, the user can incrementally write a formula with this.

naterush (changes requested):
Some cleanup around closing and reopening the cell editor. Going to do some manual testing now.

naterush:
The biggest thing is changing what happens when the cell editor is open and you use one of these. In general, I think it should work like Excel (although excel does a bunch more than we need to). But mostly:
1. Don't close the cell editor if you click one of these. 
2. Handle if there is already a formula in the cell editor at the time of selecting a new formula

naterush:
If this isn't clear, feel free to write up a mini spec here and send it for review. In fact, maybe let's do that just to make sure we're on the same page about expected behavior...

mitosheet/mitosheet/public/v3/sheet_functions/number_functions.py
```diff
@@ -119,6 +121,7 @@ def CORR(s1: NumberRestrictedInputType, s2: NumberRestrictedInputType) -> Number
         "function": "CORR",
         "description": "Computes the correlation between two series, excluding missing values.",
         "search_terms": ["corr", "correlation", "r^2"],
+        "category": "MATH",
```
aarondr77, line 124: Quick callout that we should make sure to update these docs accordingly: https://docs.trymito.io/how-to/interacting-with-your-data/bring-your-own-spreadsheet-functions#displaying-syntax-hints-inside-of-mito

The dev could either set a category so their custom functions show up in the correct bucket, and/or we could create a `custom functions` category so the dev can have their functions all appear in one section. 

If they write a custom function and don't add a `category` then it should appear in the custom functions button.

mitosheet/docs/README.md
```diff
@@ -34,6 +35,7 @@ As a TypeScript type:
 interface FunctionDocumentationObject {
     function: string;
     description: string;
+    category?: 'MATH' | 'LOGIC' | 'FINANCE' | 'DATE' | 'TEXT' | 'REFERENCE';
```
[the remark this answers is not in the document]
marthacryan: Yes

mitosheet/src/mito/components/toolbar/FormulaTabContents.tsx
marthacryan: I don't think it's possible to make `undefined` a key in a dictionary but I added the type!

naterush (changes requested):
Super minor cleanup, then LGTM!

mitosheet/src/mito/Mito.tsx
```diff
@@ -129,6 +129,7 @@ export const Mito = (props: MitoProps): JSX.Element => {
         currOpenToolbarDropdown: undefined,
         toolbarDropdown: undefined,
         exportConfiguration: {exportType: 'csv'},
+        toolbarOpen: true,
```
naterush, line 132: This feels like duplicated state. What if we just store the `currentTab` in the UIState, and then use that to check if the toolbar is open?

naterush (changes requested):
Minor refactor

PR #1012, opened 2023-11-03
Add code tab

marthacryan:
Adds a tab for code related buttons to the toolbar:
image


mitosheet/src/mito/utils/actions.tsx
```diff
@@ -321,6 +322,28 @@ export const getActions = (
...
+                closeOpenEditingPopups();
+
+                const stringToCopy = analysisData.code.join('\n');
+
+                void writeTextToClipboard(stringToCopy);
+            },
+            isDisabled: () => {
+                if (!doesAnySheetExist(sheetDataArray)) {
+                    return 'There is no code to copy. Import data.';
+                }
```
naterush, line 341: Is this the right disabled? I think there could still be code in this case. I think we want to check that analysisData.code.length > 0

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -321,6 +322,28 @@ export const getActions = (
...
+        [ActionEnum.CopyCode]: {
+            type: 'build-time',
+            staticType: ActionEnum.CopyCode,
+            icon: CopyIcon,
+            toolbarTitle: 'Copy Code to Clipboard',
+            longTitle: 'Copy Code to Clipboard',
+            actionFunction: () => {
+                closeOpenEditingPopups();
+
+                const stringToCopy = analysisData.code.join('\n');
```
naterush, line 334: There used to be a util for building a code string. Does this have the imports and stuff? I think probably, as the refactor moved that to the backend.

naterush (commented):
Code is looking excellent. Going to review the toolbar now.

naterush:
Don't think the bug is in this PR, but no `Finance` formulas are appearing.

PR #1015, opened 2023-11-06
Add toolbar top and remove main menus

marthacryan:
Adds a top bar to the toolbar and remove the main menu:
image


naterush:
Screen Shot 2023-11-06 at 10 58 51 AM

Conditional formatting is always highlighted, by accident.

naterush:
Can we change all the icons on the top bar to be white? They are not super legible otherwise.

mitosheet/src/mito/components/toolbar/Toolbar.tsx
```diff
@@ -71,103 +60,18 @@ export const Toolbar = (
...
-                        />
-                    </ToolbarMenu>
-                    <ToolbarMenu type='View' uiState={props.uiState} setUIState={props.setUIState}>
-                        <ToolbarViewDropdown
-                            actions={props.actions}
-                            uiState={props.uiState}
-                            setUIState={props.setUIState}
-                            userProfile={props.userProfile}
-                        />
-                    </ToolbarMenu>
```
naterush, line 137: I think you can remove all these ToolbarMenus! There is a lot of them, and some infrastructure for managing them you can chop (e.g. open dropdown).

naterush (changes requested):
We can get rid of a couple files here!

PR #1017, opened 2023-11-08
mitosheet: add jupyterlab 4.0 support

naterush:
# Description

1. Adds JupyterLab 4.0 support in the hackiest of ways

TODO:
- [ ] Get all code-writing code working
- [ ] Then, figure out and fix the type issues rather than just overwriting with `any`
- [ ] Then, get a review on this -- and then we're done with this PR, just for investigation

Then, after this, we move to the supporting both at once:
1. Figure out how to get it building with versions of both dependencies. It's relatively easy with regular dependencies (see screenshot below), but might be more challenging with devDependency of JupyterLabBuilder
Screen Shot 2023-11-07 at 12 04 51 PM
2. For jupyterlab/builder -- I think we might have to figure something out here. We need to have different things in our package.json in either case... Some ideas:

1. Rewriting the build commands ourself.
2. Use a path to a file. Notably, this will keep the package.json the same, but requires choosing jupyterlab 3.0 or 4.0
3. Then, for the build:all we need some more work. We need to switch this builder? yes.

# Testing

All frontend tests should just pass!

# Documentation

No.


ktaletsk:
JupyterLab 3 reached EOL on May 15, 2024 https://jupyterlab.readthedocs.io/en/latest/getting_started/lifecycle.html
Are there plans to support 4.x?

PR #1019, opened 2023-11-08
Fix useLiveUpdating hook to be more robust

naterush:
# Description

Makes it so the hook doesn't have timing dependencies and race conditions, simplifies the logic, and allows task panes to stay open. 

Note: we might not want to actually keep task panes open, until we play later edits. Thoughts?

# Testing

Use the hook. Stress test it. 

# Documentation

No.


aarondr77:
> Note: we might not want to actually keep task panes open, until we play later edits. Thoughts?

Currently this error exists. So down to just keep them closed until we support this better. 

1. Merge two datasets together
2. Leaving the merge taskpane open, add a column to the new sheet
3. Then update the merge configuration and get this error 

```
  File "add_column.py", line 40, in execute
    if column_header_index < 0 or len(prev_state.dfs[sheet_index].columns) <= column_header_index:
IndexError: list index out of range
```

aarondr77:
1. Create a pivot table
2. Close the pivot table taskpane
3. Add a column to the original sheet
4. Click on the pivot table tab to reopen the taskpane
5. Click the Undo button
6. Update the pivot table configuration
7. Notice that a new pivot table is created

https://github.com/mito-ds/mito/assets/18709905/df9ee9b0-6e23-4fb9-9147-8b1bef1bae48

aarondr77:
Once you fix these up, will do a bit more testing then should be LGTM

aarondr77:
Just checking in: the pivot table bug described above is still there. Once you fix that up I'll do a final review

Issue #1006, opened 2023-11-02
Mito handles mixed date / string columns differently than pandas 


aarondr77:
1. Create this dataframe that has datetime values and string values in the same column
```
import pandas as pd
df = pd.DataFrame({'Date': [pd.to_datetime('1648784800000', unit='ms'), pd.to_datetime('1648784800000', unit='ms'), 'Q1 2023']})
```

2. Printing the dataframe in Jupyter gives: 
<img width="283" alt="Screenshot 2023-11-02 at 12 39 47 PM" src="https://github.com/mito-ds/mito/assets/18709905/62bc6438-9990-49fc-be80-196ab0deae36">

3. Displaying the data in Mito gives: 
<img width="173" alt="Screenshot 2023-11-02 at 12 40 47 PM" src="https://github.com/mito-ds/mito/assets/18709905/6b3c7847-21a3-468a-8270-1c4b0889813f">

4. Try converting the column to a datetime and get this error
```
Traceback (most recent call last):
  File "change_column_dtype.py", line 165, in execute
  File "utils.py", line 62, in get_to_datetime_params
    detected_format = get_datetime_format(string_series)
  File "utils.py", line 117, in get_datetime_format
    if seperator in sample_string_datetime:
TypeError: argument of type 'Timestamp' is not iterable
```

5. Clear analysis
6. Try writing the formula =FLOAT(Date0) and get this 
<img width="344" alt="Screenshot 2023-11-02 at 12 42 10 PM" src="https://github.com/mito-ds/mito/assets/18709905/04fd7a5d-8113-4fa4-b9af-49eb06066fe0">

PR #1021, opened 2023-11-08
Datetime fix

naterush:
# Description

Closes https://github.com/mito-ds/mito/issues/1006, by making it so you can convert mixed types to date times.

Note that this doesn't fix the frontend display issue, for two reasons:
1. It's hard to do so in a performant way
2. Leaving it makes it more clear to these that they have to convert to a date time

# Testing

See new tests.

# Documentation

N/A.


aarondr77:
LGTM once tests pass

PR #1027, opened 2023-11-13
Address review from #990

marthacryan:
Follow-up PR to #990 to address review. 

* Makes the conditional format card use the original icon
* Adds a loading indicator in the top right
* Updates the insert/delete icons to suggest columns
* Updates the color of the top toolbar
* Updates the background color of toolbar buttons on hover
* Fixes the filter icon in the column headers on hover
* Update the fill nan icon
* Add "Insert Function" and "Add column" to the formula tab
* Fix focus issues when clicking on the formula dropdowns


naterush (changes requested):
Code looks good. Using it now.

naterush:
This all looks excellent. Aaron prob could do with a review here as well.

aarondr77:
I think the AI Icon should be 1) larger and 2) darker. Maybe we can make use the dark purple as the primary color instead of just for the eyes. 

The button looks disabled to me. 

Screenshot 2023-11-14 at 4 22 55 PM

mitosheet/src/mito/components/toolbar/FormulaTabContents.tsx
```diff
@@ -101,13 +108,19 @@ export const FormulaTabContents = (
...
+        <div className="toolbar-vertical-line"/>
+
         {(Object.keys(formulaCategories) as FunctionCategory[]).map((category: FunctionCategory) => {
             // We don't want to display the finance category in the toolbar because we don't currently have any finance functions
             if (category === 'FINANCE') {
                 return null;
             }
             return getFormulaDropdown(formulaCategories[category], category)
         })}
+        {getFormulaDropdown(ActionEnum.Formulas_Dropdown_Custom, 'custom')}
```
naterush, line 123: I like that we display if there no formulas -- it's great for enterprise discoverability!

mitosheet/src/mito/components/toolbar/DataTabContents.tsx
```diff
@@ -18,31 +17,6 @@ export const DataTabContents = (
...
-                action={action}
-                iconOverride={<ImportIcon />}
-                key={action.staticType}
-            />)
-        })}
-
-        <div className='toolbar-vertical-line' />
-        
-        <ToolbarButton
-            action={props.actions.buildTimeActions[ActionEnum.UPDATEIMPORTS]}
```
naterush, line 41: In Excel, these are in the Data tab. I think we should break from that based on results testing, rather than just blind intuition.

That being said, we can keep this for now.

naterush (changes requested):
Code looks pretty good. Reviewing functionality now!

naterush:
Screen Shot 2023-11-15 at 9 52 40 AM

Tooltip on custom formulas is wrong. Also, I'd be down to disable this dropdown if you don't have any custom formulas!

naterush:
Screen Shot 2023-11-15 at 9 54 27 AM

Did we want to change these icons? I can't remember the decision here.

naterush:
Sorry to have you go back and forth (we're still refining the three way collab / review), but let's actually move all importing / changing import buttons back to Data.

In Excel, all importing from other sources is in the Data tab. So unless we have conclusive results of a test that say the Insert tab is better, let's not move them on a hunch.

We can try and sniff this out in our test!

naterush:
Screen Shot 2023-11-15 at 11 54 38 AM

Can we change this to: Can we remove the language about importing data first -- not relevant if I have data... Let's just have different disabled messages in each case.

naterush:
Can we also move the import buttons back to the Data tab? 

Then LGTM!

PR #1035, opened 2023-11-17
mitosheet: better guardrails

aarondr77:
# Description

- Provide a hint in filters to let users know its case sensitive 
- Give more helpful names to `is empty` and `is not empty` filter conditions. 

# Testing

Check it out. 

# Documentation


mitosheet/src/mito/components/taskpanes/ControlPanel/FilterAndSortTab/filter/FilterCard.tsx
```diff
@@ -175,6 +176,11 @@ function FilterCard (props: FilterCardProps): JSX.Element {
         <div>
             <div className='text-header-3 mt-15px'>
                 <p> Filter </p>
+                {isStringDtype(props.columnDtype) &&
+                    <p className='text-subtext-1'>
+                        Hint: Filters are case sensitive
+                    </p>
+                }
```
naterush, line 183: I think this sort of building encourages our interfaces to collect cruft -- and I'm not sure if it makes this better. For the 1/10 users this helps, this is extra language to parse for the 9/10 users. 

What does Excel do here? My vote is that we:
1. Make our default filter behavior match Excel's
2. Don't communicate about it, just rely on that standard

naterush (changes requested):

PR #1045, opened 2023-11-22
Notebook integration cleanup

naterush:
# Description

1. Cleanup up not used Python imports across all files
2. Removes DataTypeInMito notion, as we no longer have a notion of a single tutorial and so it was misleading anyways. 
3. Removes CSS and React components that we no longer use anywhere.
4. Removes some unused fields (dataTypeInMito, isLocalDeployment) in shared variables
5. Moves comm creation to a separate function, as usually `get_mito_backend` is not called inside of ipython -- so this is better factoring for when it is. 

# Testing

Make sure I didn't delete any CSS I shouldn't have. 

# Documentation

No.


naterush:
LGTM when tests pass.

Issue #1048, opened 2023-11-27
Investigate excel import + undo on streamlit


naterush:
**Describe the bug**

Saw a bug when:
1. Import an Xlsx file in streamlit
2. Click undo
3. Go and naviagte to another xlsx file
4. The loading indicator finishes (in Mito never appears, and in Streamlit the backend is up to date), but noting appears in the tab selection section (just kept loading there)5. 

Note that this might be because the file is open, and notably was on windows. Needs more investigation.

Also, if you change the number of skip lines and reimport, it should just overwrite... duh!

PR #1050, opened 2023-11-28
mitosheet: fix bug where use effect on undo and redo were triggered incorrectly

naterush:
…ncorrectly

# Description

Closes: #1048 

Previously, things would break in Streamlit you:
1. Imported an XLSX file
2. Undid the import
3. Tried to import another XLSX file (without closing the import taskpane)
4. The sheets in the file never show up

The bug fixed here would cause a getParam api call to be sent at the same time as the excel file metadata -- which would cause the excel file metadata to be dropped. 

# Testing

Test the series of steps above.

# Documentation

No.


marthacryan (approved):
LGTM

PR #1051, opened 2023-11-28
Dogfooding PR - toolbar tooltips and dash app update

### LINKED ISSUE #1024: Global Dataframe State Bug in Dash
Sorry I don't have time currently to create a minimal example. But see example here: https://drive.google.com/file/d/1xiJIKLmCTpeEhf_lt5fd-UmXuJViubzZ/view?usp=sharing

We should test if the same thing happens in Streamlit. 

In addition, you can see a temporary fix here: https://github.com/plotly/mito-dash-application/pull/2/commits/cdbdceeae5a93a6581cdc55fd360300305a9cd93

marthacryan:
* Updates the toolbar tooltips for:
  * import / export dropdowns
  * "More Functions" dropdown
* Updates the dash app to use the new spreadsheet API
* Removes underlining in the "Help" button at the top of the toolbar in notebook. 
* Fixes a focus issue where the cell editor didn't grab the focus when the user clicks on a formula dropdown item and the cell editor is already open. This was happening because the cell editor automatically grabs focus when it _opens_ but not when you change the formula. 
* Fixes #1024

### DIFF
--- a/mitosheet/src/mito/components/toolbar/FormulaTabContents.tsx
+++ b/mitosheet/src/mito/components/toolbar/FormulaTabContents.tsx
@@ -21,6 +21,7 @@ export const FormulaTabContents = (
         editorState: EditorState | undefined;
         setEditorState: React.Dispatch<React.SetStateAction<EditorState | undefined>>;
         analysisData: AnalysisData;
+        mitoContainerRef: React.RefObject<HTMLDivElement>;
     }): JSX.Element => {
 
     /**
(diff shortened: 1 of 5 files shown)


mitosheet/src/mito/components/toolbar/FormulaTabContents.tsx
```diff
@@ -59,6 +59,7 @@ export const FormulaTabContents = (
                                 ...props.editorState,
                                 formula: `=${functionObject.function}(${currentFormula.startsWith('=') ? currentFormula.substring(1) : currentFormula}`,
                             })
+                            document.getElementById('cell-editor-input')?.focus();
```
naterush, line 62: document is the wrong thing to use here, it may lead to bugs when there are multiple mitosheets in the same page.

naterush (commented):
Minor fixup, then LGTM

naterush:
Functionality working good!

mitosheet/src/mito/components/toolbar/FormulaTabContents.tsx
marthacryan: Ah good point - done

PR #1054, opened 2023-11-29
Add bulk column header transform to toolbar

marthacryan:
Adds a toolbar button for the existing bulk column header transform behavior. Adds into the data tab:
image


naterush (commented):
Could it be any easier to add to the toolbar? I think no. Really impressed with this -- testing now.

naterush:
LGTM when CI passes!

PR #1056, opened 2023-11-30
Update style and content of column header dropdown

marthacryan:
* Adds icons for the dropdown items
* Updates spacing
* Updates the content of the dropdown to be more consistent with Excel

image
image

Todo:
- [x] Fix right text spacing
- [x] Add icon for unique values


mitosheet/src/mito/components/endo/ColumnHeaderDropdown.tsx
```diff
@@ -39,188 +38,66 @@ export default function ColumnHeaderDropdown(props: {
...
             closeDropdown={() => props.setOpenColumnHeaderDropdown(false)}
-            width='medium'
         >
+            <DropdownItem {...getPropsForDropdownItem(ActionEnum.Copy)} />
+
+            <DropdownSectionSeperator isDropdownSectionSeperator={true}/>
+
             <DropdownItem
+                {...getPropsForDropdownItem(ActionEnum.Add_Column_Left)}
                 title='Insert Column Left'
```
naterush, line 65: I think these titles can be moved to the actions like the icons. So we have `toolbarTitle` and also `contextMenuTitle`. I would actually be down for `titleToolbar` and `titleContextMenu` -- like icons, this makes the org of the actions very nice and easy to parse and maintain.

mitosheet/src/mito/components/endo/ColumnHeaderDropdown.tsx
```diff
@@ -39,188 +38,66 @@ export default function ColumnHeaderDropdown(props: {
...
-    const columnIndex = Object.keys(props.sheetData.columnIDsMap).indexOf(props.columnID)
+    const getPropsForDropdownItem = (actionName: ActionEnum) => {
+        const action = props.actions.buildTimeActions[actionName];
+        return {
+            title: action.toolbarTitle ?? action.longTitle,
+            onClick: () => {
+                props.closeOpenEditingPopups();
+                void action.actionFunction();
+            },
+            icon: action.contextMenuIcon ? <action.contextMenuIcon /> : action.toolbarIcon ? <action.toolbarIcon/> : <StepsIcon/>,
```
naterush, line 49: Let's change these to iconContextMenu and iconToolbar, it makes them easier to parse, I think. 

TiP: you can just do a `fn + F2` on the actual Action type, and you'll be able to do this find and replace super duper easy!

mitosheet/src/mito/components/endo/ColumnHeaderDropdown.tsx
```diff
@@ -39,188 +38,66 @@ export default function ColumnHeaderDropdown(props: {
...
-                            currOpenTaskpane: {type: TaskpaneType.CONTROL_PANEL},
-                            selectedColumnControlPanelTab: ControlPanelTab.FilterSort
-                        }
-                    })
-                }}
-                disabled={!isNumberDtype(props.columnDtype)}
-                tooltip={!isNumberDtype(props.columnDtype) ? "Only number columns can be formatted currently" : undefined}
+            <DropdownItem {...getPropsForDropdownItem(ActionEnum.SortDescending)}
+                title='Sort A to Z'
+                icon={<SortDescendingIcon aColor='black'/>}
```
naterush, line 90: Move these into the actions if possible

naterush (commented):
Some code cleanup, mostly focused on moving move into the actions for easy maintainability!

Going to test functionality now.

naterush:
Screen Shot 2023-11-30 at 11 16 35 AM

Spacing of other dropdown items appears to have changed. Let's change it back.

naterush:
The rename action doesn't appear to work on my computer. My guess is that it's a similar bug to that strange "event propagating meant action called twice calling cell editors" thing -- see video attached.

https://github.com/mito-ds/mito/assets/21002431/9161e144-28b9-4889-ba4e-210f393356b4

naterush:
Opening the dataframe format task pane feels wrong to me. Maybe we can actually open the dropdown in the home toolbar -- so like change to the Home Tab, and then open the dropdown for the formats?

mitosheet/src/mito/components/endo/ColumnHeaderDropdown.tsx
marthacryan: I want to reuse this icon and change the colors by using this prop, so I'm keeping these as icon overrides for now.

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1375,12 +1392,16 @@ export const getActions = (
                 return doesColumnExist(startingColumnID, sheetIndex, sheetDataArray) ? defaultActionDisabledMessage : 'There are no columns in the dataframe to rename. Add data to the dataframe.'
             },
             searchTerms: ['rename', 'name', 'header'],
-            tooltip: "Rename the selected column."
+            tooltip: "Rename the selected column.",
+            displayKeyboardShortcuts: {
+                mac: 'Enter',
+                windows: 'Enter'
+            }
```
naterush, line 1399: I guess I was actually wrong about these -- oops. I think keyboard shortcuts are supposed to work from "anywhere" in the app. Enter only works if you have the column header selected. Sorry.

mitosheet/src/mito/components/endo/utils.tsx
```diff
@@ -191,4 +192,20 @@ export const getGraphIsSelected = (uiState: UIState): boolean => {
 */
 export const getDataframeIsSelected = (uiState: UIState, sheetDataArray: SheetData[]): boolean => {
     return uiState.selectedTabType === 'data' && sheetDataArray.length !== 0;
+}
+
+/** 
+ * Returns the props to give to a DropdownItem component, given an Action.
+ * Used by the context menus. 
+*/
+export const getPropsForDropdownItem = (action: Action, closeOpenEditingPopups: () => void) => {
```
naterush, line 201: I think a more clear name might be getPropsForContextMenuDropdownItem

naterush (commented):
Super minor cleanup.

naterush:
Clarifying what the above comment about formatting means: I think that the `Format` option for a column should open the dropdown in the toolbar that let's users select from `percentage`, etc. 

This is because right clicking on a column header and then clicking format most likely means you want to format the column... not the colors of the dataframe.

naterush:
Screen Shot 2023-11-30 at 4 22 04 PM

Alignment on dropdowns is still changed -- hence subtext is a bit off as well. I'm fine with new spacing, as long as it is applied for subtext too.

naterush:
Once you make these changes, LGTM -- no need for another review. 

This PR is excellent. Huge bang for buck!

PR #1057, opened 2023-11-30
Context menu scroll on hover

marthacryan:
Adds automatic scrolling on hover for the top and bottom of the dropdown. 
scrollcontext
scrollformulas


mitosheet/src/mito/components/elements/Dropdown.tsx
```diff
@@ -453,6 +453,26 @@ const Dropdown = (props: DropdownProps): JSX.Element => {
             return null;
         }
     });
+
+    const scrollRef = useRef<NodeJS.Timer | null>(null);
```
naterush, line 457: Can we rename scrollRef? Maybe `scrollTimerRef`?

mitosheet/src/mito/components/elements/Dropdown.tsx
```diff
@@ -453,6 +453,26 @@ const Dropdown = (props: DropdownProps): JSX.Element => {
...
+    const dropdownItemContainerRef = useRef<HTMLDivElement | null>(null);
```
naterush, line 458: `dropdownItemsContainerRef`

mitosheet/src/mito/components/elements/Dropdown.tsx
```diff
@@ -453,6 +453,26 @@ const Dropdown = (props: DropdownProps): JSX.Element => {
...
+    const [showScrollUp, setShowScrollUp] = useState(false);
+    const [showScrollDown, setShowScrollDown] = useState(childrenToDisplay.length > 6);
```
naterush, line 460: I think it's better to:
1. Not save these as state variables
2. Compute them on every render (just in the main body of the component - or with a helper function) - directly from the ref, as you do in the `updateScrollPosition` function. 
3. Not have to remember to call this function on scroll

As it is currently, there is a bug where if you scroll in a dropdown without using these hover buttons, they don't appear / disappear at the right times!

naterush (commented):
Two things:
1. When you hover over these things and we're scrolling, I could see there being some slight onHover effect - where we color them. Thoughts?
2. I first tried to click on these, like they were buttons. It closed the dropdown! I think you should be able to click, and if you do, it should just jump the scroll a reasonable amount (and then not keep scrolling). 

The state might get a little funky there, but I think that would be a nice bit of polish on the polish.

aarondr77:
Can we make sure that the icons for scroll up and scroll down are identical? 

Screenshot 2023-12-01 at 3 51 29 PM

aarondr77:
This subtext is now indented. Was this intended?

Screenshot 2023-12-01 at 4 37 39 PM

mitosheet/src/mito/components/elements/Dropdown.tsx
marthacryan: Ah that'll be a little more complicated than it seems I think. The component doesn't re-render on every scroll, so the reason we'd want this is **because** it's not going to automatically re-compute that on scroll. I added a scroll handler for the function to call this same `updateScrollPosition` function because we actually need a state function to make it re-render without the hover buttons when the scrolling reaches the bottom / top.

marthacryan:
> 1. When you hover over these things and we're scrolling, I could see there being some slight onHover effect - where we color them. Thoughts?

I think the standard behavior is to not change color - my expectation as a user would be for them not to change color. 

> 2. I first tried to click on these, like they were buttons. It closed the dropdown! I think you should be able to click, and if you do, it should just jump the scroll a reasonable amount (and then not keep scrolling).

Done!

mitosheet/src/mito/components/elements/Dropdown.tsx
```diff
@@ -216,6 +221,18 @@ const Dropdown = (props: DropdownProps): JSX.Element => {
             return;
         }
 
+        // If the user clicks on the scroll up or scroll down buttons,
+        // then we don't want to close the dropdown
```
naterush, line 225: I wonder if we might be forgetting to stop an event propagation here -- because clicking still sometimes closes the dropdown for me, at least when clicking the up arrow. See attached video. 


Uploading Scroll Bug.mov…

naterush (commented):
This polish is almost excellent, but I feel it needs a bit more polish before we merge it. The main thing is that when you scroll to the top, there is a jump as the top arrow disappears and no longer renders - which mean you get a confusing flash. 

https://github.com/mito-ds/mito/assets/21002431/52b8077d-d058-435b-ba80-bc5cfb599481

One option is to keep an empty white bar at the top of dropdowns that have >6 elements -- but I don't love this as it leaves an empty white space. 

Another option is moving the arrows beyond the bounds of the scrollable container? So like: the dropdown is always in the same position, but if there is an arrow at the top, it appears -5 px from the top of the dropdown div. 

This might not be too tricky to implement (I'm no CSS export though) -- but I feel like the behavior would make the dropdown less flickery. 

Thoughts?

naterush:
https://github.com/mito-ds/mito/assets/21002431/f51098f0-a61d-4be6-8755-a003cdb45bcd

Currently broken with pivot and the search bar. 

I'm going to vote we close this PR for now and make an attempt in the future. It's a bit more complex than we expect every time, and as a result it's turned into a time-sync!

naterush:
Speaking of which, before closing, wanna open an issue describing:
1. approaches we tried
2. What we learned
3. what we should try next?

PR #1058, opened 2023-11-30
Update style/content of index context menu

marthacryan:
* Adds icons to the index context menu
* Changes the options available in the index context menu
* Updates to using actions for the context menu. 

image


naterush:
Maybe the reset index icon can have a little X in the middle of it, to disambiguate it from the reset and keep?

PR #1062, opened 2023-12-04
Add cell context menu

marthacryan:
Adds a context menu for cells:
image

**Note on behavior**: This automatically selects the cell on right click. 

Todo:
- [x] Add "Filter to Cell value" action

### DIFF
--- a/mitosheet/src/mito/components/endo/CellDropdown.tsx
+++ b/mitosheet/src/mito/components/endo/CellDropdown.tsx
@@ -56,7 +56,7 @@ export default function CellDropdown(props: {
                 {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Filter], props.closeOpenEditingPopups)}
                 title='Create a Filter'
             />
-            <DropdownItem {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.RESET_AND_KEEP_INDEX], props.closeOpenEditingPopups)}/>
+            <DropdownItem {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.FilterToCellValue], props.closeOpenEditingPopups)}/>
         </Dropdown>
     )
 }
--- a/mitosheet/src/mito/components/endo/CellDropdown.tsx
+++ b/mitosheet/src/mito/components/endo/CellDropdown.tsx
@@ -1,0 +1,62 @@
+// Copyright (c) Saga Inc.
+
+import React from 'react';
+import { ActionEnum, UIState } from '../../types';
+import Dropdown from '../elements/Dropdown';
+import DropdownItem from '../elements/DropdownItem';
+import DropdownSectionSeperator from '../elements/DropdownSectionSeperator';
+import { TaskpaneType } from '../taskpanes/taskpanes';
+import { getPropsForContextMenuDropdownItem } from './utils';
+import { Actions } from '../../utils/actions';
+import { isCurrOpenDropdownForCell } from './visibilityUtils';
+
+/*
+    Displays a set of actions one can perform on a row
+*/
+export default function CellDropdown(props: {
+    display: boolean;
+    rowIndex: number;
+    colIndex: number;
+    setUiState: React.Dispatch<React.SetStateAction<UIState>>;
+    actions: Actions;
+    closeOpenEditingPopups: (taskpanesToKeepIfOpen?: TaskpaneType[]) => void;
+}): JSX.Element {
+    return (
+        <Dropdown
+            display={props.display}
+            closeDropdown={() => {
+                props.setUiState((prevUIState) => {
+                    const isCurrOpenDropdown = isCurrOpenDropdownForCell(prevUIState, props.rowIndex, props.colIndex);
+                    return {
+                        ...prevUIState,
+                        currOpenDropdown: isCurrOpenDropdown ? undefined : prevUIState.currOpenDropdown
+                    }
+                })
+            }}
+            width='medium-large'
+        >
+            <DropdownItem
+                {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Copy], props.closeOpenEditingPopups)}
+                title='Copy'
+            />
+
+            <DropdownSectionSeperator isDropdownSectionSeperator={true}/>
+
+            <DropdownItem {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Add_Column_Left], props.closeOpenEditingPopups)} />
+            <DropdownItem {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Add_Column_Right], props.closeOpenEditingPopups)} />
+
+            <DropdownSectionSeperator isDropdownSectionSeperator={true}/>
+
+            <DropdownItem {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Delete_Row], props.closeOpenEditingPopups)} />
+            <DropdownItem {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Delete_Col], props.closeOpenEditingPopups)} />
+
+            <DropdownSectionSeperator isDropdownSectionSeperator={true}/>
+
+            <DropdownItem
+                {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Filter], props.closeOpenEditingPopups)}
+                title='Create a Filter'
+            />
+            <DropdownItem {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.RESET_AND_KEEP_INDEX], props.closeOpenEditingPopups)}/>
+        </Dropdown>
+    )
+}
--- a/mitosheet/src/mito/components/endo/EndoGrid.tsx
+++ b/mitosheet/src/mito/components/endo/EndoGrid.tsx
@@ -741,6 +741,7 @@ function EndoGrid(props: {
                             transform: `translate(${gridState.scrollPosition.scrollLeft - translate.x}px, ${gridState.scrollPosition.scrollTop - translate.y}px)`,
                         }}
                         onContextMenu={() => {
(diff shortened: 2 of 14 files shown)


mitosheet/src/mito/utils/actions.tsx
```diff
@@ -482,6 +482,77 @@ export const getActions = (
...
+                }
+                
+                if (doesColumnExist(startingColumnID, sheetIndex, sheetDataArray)) {
+                    return defaultActionDisabledMessage
+                } else {
+                    return "There are no columns in the dataframe to delete. Add data to the sheet."
+                }
+            },
+            searchTerms: ['delete column', 'delete col', 'del col', 'del column', 'remove column', 'remove col',  'delete row', 'filter rows', 'rows', 'remove rows', 'hide rows'],
+            tooltip: "Delete all of the selected columns or rows from the sheet."
```
naterush, line 554: Tooltips need updating, as do the searchTerms

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -482,6 +482,77 @@ export const getActions = (
...
+
+                const rowsToDelete = getSelectedRowLabelsInSingleSelection(gridState.selections[0], sheetData);
+                if (rowsToDelete.length > 0) {
+                    return defaultActionDisabledMessage;
+                } else {
+                    return 'There are no rows selected.'
+                }
+            },
+            searchTerms: ['delete column', 'delete col', 'del col', 'del column', 'remove column', 'remove col',  'delete row', 'filter rows', 'rows', 'remove rows', 'hide rows'],
+            tooltip: "Delete all of the selected columns or rows from the sheet."
```
naterush, line 516: Same here

mitosheet/src/mito/components/endo/GridData.tsx
```diff
@@ -100,12 +111,46 @@ const GridData = (props: {
...
                                 >
                                     {displayCellData}
+                                    <CellDropdown
+                                        display={displayDropdown}
+                                        rowIndex={rowIndex}
+                                        colIndex={columnIndex}
+                                        setUiState={props.setUIState}
+                                        actions={props.actions}
+                                        closeOpenEditingPopups={props.closeOpenEditingPopups}
+                                    />
```
naterush, line 153: Are there performance impacts of these? It effectively adds `# of rendered cells` to what we render -- which might be a bit on larger screens.

mitosheet/src/mito/components/endo/EndoGrid.tsx
```diff
@@ -741,15 +741,21 @@ function EndoGrid(props: {
                             transform: `translate(${gridState.scrollPosition.scrollLeft - translate.x}px, ${gridState.scrollPosition.scrollTop - translate.y}px)`,
                         }}
                         onContextMenu={() => {
+                            console.log('right clicked on renderer container')
```
naterush, line 744: rm

mitosheet/src/mito/components/endo/CellDropdown.tsx
```diff
@@ -0,0 +1,62 @@
...
+import DropdownSectionSeperator from '../elements/DropdownSectionSeperator';
+import { TaskpaneType } from '../taskpanes/taskpanes';
+import { getPropsForContextMenuDropdownItem } from './utils';
+import { Actions } from '../../utils/actions';
+import { isCurrOpenDropdownForCell } from './visibilityUtils';
+
+/*
+    Displays a set of actions one can perform on a row
+*/
+export default function CellDropdown(props: {
```
naterush, line 16: Now might be a nice time to rename all of these to `ContextMenu` -- all three of em

mitosheet/src/mito/components/endo/IndexHeaderDropdown.tsx
```diff
@@ -15,25 +14,12 @@ import { Actions } from '../../utils/actions';
...
     actions: Actions;
     closeOpenEditingPopups: (taskpanesToKeepIfOpen?: TaskpaneType[]) => void;
 }): JSX.Element {
-
-    // Log opening this dropdown
-    useEffect(() => {
-        if (props.display) {
-            void props.mitoAPI.log('opened_index_header_dropdown')
-        }
-    }, [props.display])
```
naterush, line 35: Do you wanna remove the log for column headers too?

mitosheet/src/mito/components/endo/CellDropdown.tsx
```diff
@@ -0,0 +1,62 @@
...
+            <DropdownItem {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Delete_Col], props.closeOpenEditingPopups)} />
+
+            <DropdownSectionSeperator isDropdownSectionSeperator={true}/>
+
+            <DropdownItem
+                {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Filter], props.closeOpenEditingPopups)}
+                title='Create a Filter'
+            />
+            <DropdownItem {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.RESET_AND_KEEP_INDEX], props.closeOpenEditingPopups)}/>
+        </Dropdown>
```
naterush, line 60: I don't think we need the RESET_INDEX action in the cell

naterush (commented):
Working excellently. Minor code cleanup, and then LGTM.

mitosheet/src/mito/components/endo/GridData.tsx
marthacryan: I had this thought as well. What if we moved the display state into `GridData.tsx` and then only rendered it if it needs to be rendered?

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -714,6 +784,58 @@ export const getActions = (
...
+            titleContextMenu: 'Filter to Cell Value',
+            iconContextMenu: FilterIcon,
+            longTitle: 'Filter column',
+            actionFunction: async () => {
+                // We turn off editing mode, if it is on
+                setEditorState(undefined);
+                const {startingColumnIndex} = gridState.selections[0];
+                const columnID = sheetData.data[startingColumnIndex].columnID;
+                const { columnFilters, cellValue, columnDtype } = getCellDataFromCellIndexes(sheetData, gridState.selections[0].startingRowIndex, gridState.selections[0].startingColumnIndex);
+                const filters = columnFilters?.filters ?? [];
```
naterush, line 800: A lot of this data (e.g. the columnID) exists at the top of the action function, already accessible. Let's remove any duplicated variable definitions, and then just move the rest of it up (e.g. filters we can get up top) so we can use them in any action that needs them.

naterush: Since you disable this action when there is more than one cell selected, this should be safe to use the variables defined at the top of this function.

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -714,6 +784,58 @@ export const getActions = (
...
+                    condition = 'number_exactly';
+                } else if (columnDtype === 'boolean') {
+                    condition = cellValue ? 'boolean_is_true' : 'boolean_is_false';
+                } else if (columnDtype === 'datetime') {
+                    condition = 'datetime_exactly';
+                }
+                filters.push({
+                    condition: condition,
+                    value: cellValue
+                } as FilterType);
```
naterush, line 813: Check out the function `getExclusiveFilterData`. It implements this logic, while handling NaN values as well. 

I forgot about it until I saw this logic...

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -714,6 +784,58 @@ export const getActions = (
...
+                if (gridState.selections.length === 1 &&
+                    gridState.selections[0].startingRowIndex === gridState.selections[0].endingRowIndex &&
+                    gridState.selections[0].startingColumnIndex === gridState.selections[0].endingColumnIndex) {
+                    return defaultActionDisabledMessage;
+                } else {
+                    return 'This action can only be applied to a single cell.'
+                }
+            },
+            searchTerms: ['filter', 'remove', 'delete'],
+            tooltip: "Filter this dataframe based on the data in a column."
```
naterush, line 837: Update tooltip.

naterush (commented):
Not working with NaN values, but I think refactoring to use our existing utility will fix that. Then, LGTM - this is working great!

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -714,6 +787,61 @@ export const getActions = (
...
+                let condition = 'string_exactly';
+                if (columnDtype === 'number') {
+                    condition = 'number_exactly';
+                } else if (columnDtype === 'boolean') {
+                    condition = cellValue ? 'boolean_is_true' : 'boolean_is_false';
+                } else if (columnDtype === 'datetime') {
+                    condition = 'datetime_exactly';
+                }
+                if (isNaN) {
+                    condition = 'empty'
```
naterush, line 811: Wanna refactor this into a util like the exclusive one? I think it's the inclusive version?

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -714,6 +787,61 @@ export const getActions = (
...
+                }
+
+                filters.push({
+                    condition: condition,
+                    value: cellValue
+                } as FilterType);
```
naterush, line 817: Not sure we actually wanna push here, vs. just setting the filter to be this single filter. Thoughts?

naterush (commented):
A bit of minor cleanup, then LGTM.

mitosheet/src/mito/components/endo/GridData.tsx
naterush: That sounds reasonable to me! I wonder if the dropdown needs the anchor there (honestly, no idea) -- but why don't you give it a try?

mitosheet/src/mito/utils/actions.tsx
naterush: Given that we're changing the operator forcibly to `And` -- I think it makes sense to just clear the filters rather than pushing to them.

marthacryan: Oh ok

PR #1070, opened 2023-12-08
Add navigation

marthacryan:
Adds keyboard shortcuts for:
* opening previous / next sheet
* opening find and replace


mitosheet/src/mito/utils/keyboardShortcuts.tsx
```diff
@@ -52,6 +52,40 @@ export const keyboardShortcuts: KeyboardShorcut[] = [
...
         action: ActionEnum.Redo
+    },
+    {
+        macKeyCombo: {
+            altKey: true,
+            key: ['ArrowLeft']
+        },
+        winKeyCombo: {
+            ctrlKey: true,
+            key: ['PageUp']
```
naterush: I'm fine with arrow left and arrow right for windows, if PageUp doesn't work.

mitosheet/src/mito/types.tsx
```diff
@@ -1002,6 +1004,7 @@ export enum ActionEnum {
     Formulas_Dropdown_Custom = 'custom formulas dropdown',
     Formulas_Dropdown_More = 'more formulas dropdown',
     OpenSearch = 'open search',
+    OpenSearchAndReplace = 'open search and replace',
```
naterush, line 1007: Should these be open find, and open find and replace? And then we go update their titles in the actions as well

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1870,6 +1870,86 @@ export const getActions = (
...
+                setEditorState(undefined);
+                if (uiState.currOpenSearch.isOpen) {
+                    const searchInput = mitoContainerRef.current?.querySelector<HTMLInputElement>('#mito-search-bar-input');
+                    if (searchInput) {
+                        // If the search bar is already open, then we focus on the input and select all
+                        // to make it easier to search something new without removing the previous search
+                        searchInput.focus();
+                        searchInput.select();
+                    }
+                } else {
```
naterush, line 1888: This feels like it shouldn't be in an else. Otherwise, it's not opening replace if the search is already open, ya?

naterush (commented):
Some code cleanup

naterush:
If you use the navigate right/left in sheets, and there is no further sheet to navigate to, then it moves the selection in the sheet. 

Additionally, should we be checking (in the keyboard shortcut) if the action is disabled before running it? I think we should!

naterush:
Screen Shot 2023-12-08 at 9 40 16 AM

control + shift + f opens the find and replace, but if it was entirely closed before, it puts by cursor in replace. I feel like it should put the cursor in find.

unless there's something in find already, in which case it should go to replace. Yah?

mitosheet/src/mito/utils/actions.tsx
marthacryan: Ah yeah true

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1870,6 +1870,86 @@ export const getActions = (
...
+                        }
+                    })
+                }
+            },
+            isDisabled: () => {return (uiState.selectedSheetIndex < sheetDataArray.length - 1) ? defaultActionDisabledMessage : 'No next sheet.'},
+            searchTerms: ['sheet', 'index', 'next', 'forward'],
+            tooltip: "Go to the next sheet.",
+            displayKeyboardShortcuts: {
+                mac: 'Option+Right Arrow',
+                windows: 'Ctrl+Page Up'
```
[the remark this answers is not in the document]
marthacryan: We could also get rid of this field and instead just look at the keyboard shortcuts themselves

naterush (commented):
LGTM

PR #1074, opened 2023-12-11
Add selection keyboard shortcuts

marthacryan:
Adds the shortcuts for:
* Select columns for all cells in current selection
* Select rows for all rows in current selection
* Select all cells

Todo:
- [x] Fix select rows handling
- [x] Should the select all cells also select columns? 
- [x] test on windows
- [x] test in JL


mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1649,6 +1649,106 @@ export const getActions = (
...
+                // We close the editing taskpane if its open
+                closeOpenEditingPopups();
+
+                // We open the column selection taskpane
+                let newStartingColumnIndex = gridState.selections[0].startingColumnIndex;
+                let newEndingColumnIndex = gridState.selections[0].endingColumnIndex;
+                for (const selection of gridState.selections) {
+                    newStartingColumnIndex = Math.min(newStartingColumnIndex, selection.startingColumnIndex);
+                    newEndingColumnIndex = Math.max(newEndingColumnIndex, selection.endingColumnIndex);
+                }
```
naterush, line 1669: So this works differently than Excel on multiple selections. If you select multiple ranges in excels and then press control + space, it only selects the columns from the first selection. 

Would be down to match this behavior -- so pretty much just looking at `gridState.selections[0]`.

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1649,6 +1649,106 @@ export const getActions = (
...
+                // We close the editing taskpane if its open
+                closeOpenEditingPopups();
+
+                // We open the column selection taskpane
+                let newStartingRowIndex = gridState.selections[0].startingRowIndex;
+                let newEndingRowIndex = gridState.selections[0].endingRowIndex;
+                for (const selection of gridState.selections) {
+                    console.log(selection)
+                    newStartingRowIndex = Math.min(newStartingRowIndex, selection.startingRowIndex);
+                    newEndingRowIndex = Math.max(newEndingRowIndex, selection.endingRowIndex);
```
naterush, line 1704: Wanna try the same as above the other action here (e.g. just working with the first selection), and see if it fixes things?

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1649,6 +1649,106 @@ export const getActions = (
...
+                // We open the column selection taskpane
+                setGridState(prevGridState => {
+                    return {
+                        ...prevGridState,
+                        selections: [{
+                            startingRowIndex: 0,
+                            endingRowIndex: sheetData.numRows - 1,
+                            startingColumnIndex: 0,
+                            endingColumnIndex: sheetData.numColumns - 1,
+                            sheetIndex: sheetIndex
```
naterush, line 1743: Yeah, I think just setting rows to -1 is fine!

naterush: Aka just select column headers, which will highlight everything

naterush (commented):
Some code changes

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1649,6 +1649,109 @@ export const getActions = (
...
+                }
+                if (newEndingColumnIndex === -1) {
+                    newEndingColumnIndex = sheetData.numColumns - 1;
+                }
+                setGridState(prevGridState => {
+                    return {
+                        ...prevGridState,
+                        selections: [{
+                            startingColumnIndex: newStartingColumnIndex ?? 0,
+                            endingColumnIndex: newEndingColumnIndex ?? 0,
```
naterush, line 1677: It feels like ?? is actually a strange operator here, given that we're dealing with numbers. 

Let's make the undefined check explicit (e.g. `== undefined ? ...`) to avoid any unexpected behavior, and make this easier to reason about.

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1649,6 +1649,109 @@ export const getActions = (
...
+            staticType: ActionEnum.Select_Columns,
+            longTitle: 'Select column',
+            actionFunction: () => {
+                // We turn off editing mode, if it is on
+                setEditorState(undefined);
+
+                // We close the editing taskpane if its open
+                closeOpenEditingPopups();
+
+                // We open the column selection taskpane
```
naterush, line 1663: No we don't!

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1649,6 +1649,109 @@ export const getActions = (
...
+            staticType: ActionEnum.Select_Rows,
+            longTitle: 'Select row',
+            actionFunction: () => {
+                // We turn off editing mode, if it is on
+                setEditorState(undefined);
+
+                // We close the editing taskpane if its open
+                closeOpenEditingPopups();
+
+                // We open the column selection taskpane
```
naterush, line 1700: Nah!

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1649,6 +1649,109 @@ export const getActions = (
...
+                            startingColumnIndex: 0,
+                            endingColumnIndex: sheetData.numColumns - 1,
+                            sheetIndex: sheetIndex
+                        }]
+                    }
+                });
+            },
+            isDisabled: () => {return doesAnySheetExist(sheetDataArray) ? defaultActionDisabledMessage : 'There are no cells to select. Import data.'},
+            searchTerms: ['select', 'cells', 'select all columns'],
+            tooltip: "Select all columns in current sheet."
```
naterush, line 1753: maybe just select all data in the sheet

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1649,6 +1649,109 @@ export const getActions = (
...
+            staticType: ActionEnum.Select_All,
+            longTitle: 'Select all columns',
+            actionFunction: () => {
+                // We turn off editing mode, if it is on
+                setEditorState(undefined);
+
+                // We close the editing taskpane if its open
+                closeOpenEditingPopups();
+
+                // We open the column selection taskpane
```
naterush, line 1737: Nah

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1649,6 +1649,109 @@ export const getActions = (
...
+                let newStartingRowIndex = gridState.selections[0].startingRowIndex;
+                let newEndingRowIndex = gridState.selections[0].endingRowIndex;
+                if (newStartingRowIndex === -1) {
+                    newStartingRowIndex = 0;
+                }
+                if (newEndingRowIndex === -1) {
+                    newEndingRowIndex = sheetData.numRows - 1;
```
naterush, line 1707: I don't think these checks make a ton of sense to me. The main thing here is that starting index and ending index can be out of order. E.g. it's totally valid to have a startingRowIndex=5 and an endingRowIndex=-1, which means the user just first clicked row 5, and then selected upward to a column header. 

In this example case, it feels like this would end up selecting from row 5 to the end of the sheet -- which is not what we want. 

It feels like what we wanna do is:
1. Find the minRowIndex = min(startingRowIndex, endingRowIndex)
2. Find the maxRowIndex = ...
3. Set newStartingRowIndex = max(minRowIndex, 0)
4. Set newEndingRowIndex = min(maxRowIndex, sheetData.numRows - 1)

Going to check and verify this bug is real -- but I think the logic in this action (and the other one) is a bit weird.

mitosheet/src/mito/utils/keyboardShortcuts.tsx
```diff
@@ -53,6 +53,41 @@ export const keyboardShortcuts: KeyboardShorcut[] = [
         },
         preventDefaultAndStopPropagation: true,
         action: ActionEnum.Redo
+    },
+    {
```
naterush, line 57: This infrastructure is excellent.

mitosheet/src/mito/utils/actions.tsx
naterush: There is currently a bug if you have multiple rows selected and do Shift+" " -- it only takes the first or last row.

See here: 

https://github.com/mito-ds/mito/assets/21002431/0c478dfb-0d9d-436d-b8a8-d7718f94d793

mitosheet/src/mito/utils/actions.tsx
marthacryan: Oh yeah this is actually not even necessary anymore

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1649,6 +1649,109 @@ export const getActions = (
...
+
+                // We open the column selection taskpane
+                let newStartingColumnIndex = gridState.selections[0].startingColumnIndex;
+                let newEndingColumnIndex = gridState.selections[0].endingColumnIndex;
+                if (newStartingColumnIndex === -1) {
+                    newStartingColumnIndex = 0;
+                }
+                if (newEndingColumnIndex === -1) {
+                    newEndingColumnIndex = sheetData.numColumns - 1;
+                }
```
naterush, line 1671: Yeah, so there's currently a bug in this logic -- it doesn't appear to be broken in rows, but I think this check is wrong with columns. See attached video. 

https://github.com/mito-ds/mito/assets/21002431/f9dc5786-926a-4b85-93fb-df6a23b36a33

naterush (changes requested):

PR #1085, opened 2023-12-12
Add number format keyboard shortcuts

marthacryan:
Adds keyboard shortcuts for:
- Set default format (plain text)
- Set 2 decimal places (number)
- Set date-time format
- Set currency format
- Set percentage format
- Set scientific notation format

Todo:
- [x] Test in JL
- [x] Test in windows
- [x] Handle ctrl+shift+2 and ctrl+shift+3 differently (both now set datetime dtype)
- [x] Should these actions set the dtype to number?


mitosheet/src/mito/types.tsx
```diff
@@ -1009,8 +1009,12 @@ export enum ActionEnum {
     Pivot = 'pivot',
     Precision_Increase = 'precision increase',
     Precision_Decrease = 'precision decrease',
-    Currency_Format = 'set format to currency',
-    Percent_Format = 'set format to percent',
+    Set_Format_Currency = 'set format to currency',
+    Set_Format_Percent = 'set format to percent',
+    Set_Format_Default = 'set format to default',
+    Set_DateTime_Dtype = 'set datetime format',
```
naterush, line 1015: set datetime dtype

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1113,6 +1174,27 @@ export const getActions = (
...
+            isDisabled: () => {
+                if (!doesAnySheetExist(sheetDataArray)) {
+                    return 'There are no columns to format. Import data.'
+                }
+                
+                return getSelectedNumberSeriesColumnIDs(gridState.selections, sheetData).length > 0 ? defaultActionDisabledMessage : FORMAT_DISABLED_MESSAGE
+            },
+            searchTerms: ['percent', '%', 'number format', 'format'],
+            tooltip: 'Format all of the selected columns as percentage. This only changes the display of the data, and does not effect the underlying dataframe.'
+        },
```
naterush, line 1197: These are out of date

mitosheet/src/mito/utils/keyboardShortcuts.tsx
```diff
@@ -126,6 +126,97 @@ export const keyboardShortcuts: KeyboardShorcut[] = [
...
+            shiftKey: true,
+            keys: ['3']
+        },
+        winKeyCombo: {
+            ctrlKey: true,
+            shiftKey: true,
+            keys: ['3']
+        },
+        action: ActionEnum.Set_DateTime_Dtype
+    },
```
naterush, line 181: Why do we need to register both. Can you add some documentation?

mitosheet/src/mito/utils/keyboardShortcuts.tsx
```diff
@@ -126,6 +126,97 @@ export const keyboardShortcuts: KeyboardShorcut[] = [
...
+            ctrlKey: true,
+            shiftKey: true,
+            keys: ['^']
+        },
+        winKeyCombo: {
+            ctrlKey: true,
+            shiftKey: true,
+            keys: ['^']
+        },
+        action: ActionEnum.Set_Format_Scientific
```
naterush, line 219: I'm a little confused. We don't have `%` registered, and we don't have `6` registered. Why do we sometimes do both, but not always?

tests/streamlit_ui_tests/streamlit.spec.ts
```diff
@@ -376,4 +376,29 @@ test.describe('Keyboard Shortcuts', () => {
     await expect(mito.getByPlaceholder('Find...')).toBeVisible()
     await expect(mito.getByPlaceholder('Replace...')).toBeVisible()
   })
+
+  test('Set Number Format', async ({ page }) => {
```
naterush, line 380: Yeah a separate test for date time format is fine.

naterush (commented):
Some code questions

naterush:
The set date time dtype keyboard shortcut not working on dateimte columns -- I'm guessing because its' disabled in the wrong cases (e.g. only enabled on strings). 

Let's fix this, and add a test for it.

mitosheet/src/mito/utils/actions.tsx
marthacryan: Are they?

mitosheet/src/mito/utils/keyboardShortcuts.tsx
marthacryan: I'm not sure why the browser does this but I tested locally to find these.

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1113,6 +1174,27 @@ export const getActions = (
...
+
+                const selectedNumberSeriesColumnIDs = getSelectedNumberSeriesColumnIDs(gridState.selections, sheetData);
+                void mitoAPI.editChangeColumnDtype(sheetIndex, selectedNumberSeriesColumnIDs, ColumnDtypes.DATETIME, getRandomId())
+            },
+            isDisabled: () => {
+                if (!doesAnySheetExist(sheetDataArray)) {
+                    return 'There are no columns to format. Import data.'
+                }
+                
+                return getSelectedNumberSeriesColumnIDs(gridState.selections, sheetData).length > 0 ? defaultActionDisabledMessage : FORMAT_DISABLED_MESSAGE
```
naterush, line 1193: This is still the wrong disabled condition -- most often, users want to turn strings into date times.

mitosheet/src/mito/utils/keyboardShortcuts.tsx
```diff
@@ -126,6 +126,100 @@ export const keyboardShortcuts: KeyboardShorcut[] = [
             keys: ['h']
         },
         action: ActionEnum.OpenFindAndReplace
+    },
+    {
+        macKeyCombo: {
+            ctrlKey: true,
+            shiftKey: true,
+            keys: ['`', '~']
```
naterush, line 134: Hmm. So this triggers if any key is in the list, then? That makes sense, but it feels like a little weird -- e.g. what happens if it's a keyboard shortcut that requires two keys?

tests/streamlit_ui_tests/streamlit.spec.ts
```diff
@@ -376,4 +376,38 @@ test.describe('Keyboard Shortcuts', () => {
...
+    await expect(mito.getByText('1.00', { exact: true })).toBeVisible();
+  });
+
+  test('Set Datetime Dtype', async ({ page }) => {
+    const mito = await getMitoFrameWithTestCSV(page);
+    await mito.getByTitle('Column1').click();
+
+    await page.keyboard.press('Control+Shift+@');
+    await awaitResponse(page);
+    await expect(mito.getByText('1970-01-01')).toHaveCount(4);
```
naterush, line 411: This test isn't great, since it doesn't catch how the string column can't be turned to a date time. 

Since that is by far the most popular thing that'll happen here, can we:
1. Introduce a new test file called `types.csv` -- one of which is a string column that has date time strings on it.
2. Import this instead of `test.csv` (which we can keep as basic numbers)
3. Test we can change the date correctly for the common case

naterush (commented):
A bit of cleanup, then LGTM

mitosheet/src/mito/utils/keyboardShortcuts.tsx
marthacryan: Yeah hm. Maybe I should change the name of this field. or maybe I should figure out a way to convert these symbols to "lowercase"

PR #1086, opened 2023-12-12
Add action shortcuts

marthacryan:
Adds shortcuts for:
- Create a graph
- Open file import browser
- Create a filter
- Merge

Also, added handling for escape to close whatever taskpane is open.

Todo:
- [x] Test on windows
- [x] Test on JL
- [ ] Update pivot shortcut
- [x] Add tests


mitosheet/src/mito/utils/keyboardShortcuts.tsx
```diff
@@ -128,6 +128,67 @@ export const keyboardShortcuts: KeyboardShorcut[] = [
...
+            metaKey: true,
+            keys: ['o']
+        },
+        winKeyCombo: {
+            ctrlKey: true,
+            keys: ['o']
+        },
+        preventDefaultAndStopPropagation: true,
+        action: ActionEnum.Import_Files
+    },
```
naterush, line 152: On safari, I'm not sure this is working. Command + O opens the open file menu.

mitosheet/src/mito/utils/keyboardShortcuts.tsx
```diff
@@ -128,6 +128,67 @@ export const keyboardShortcuts: KeyboardShorcut[] = [
...
+        macKeyCombo: {
+            metaKey: true,
+            keys: ['m']
+        },
+        winKeyCombo: {
+            ctrlKey: true,
+            keys: ['m']
+        },
+        preventDefaultAndStopPropagation: true,
+        action: ActionEnum.Merge
```
naterush, line 175: Similarly, Streamlit, Safari, Mac -- this is minimizing the window. Thoughts?

naterush (commented):
Some questions about how Mac modifiers work on safari.

For Pivot, let me think on it. We can change it in a later PR I think.

mitosheet/src/mito/utils/keyboardShortcuts.tsx
marthacryan: Is the focus on the mitosheet?

naterush (commented):
Code looks good. On Chrome, I'm getting the following bug:
1. Import csv
3. Press fn + option + f1 (aka alt f1)
3. This opens the graphing
4. Press Command + Z, it does nothing -- which it should undo the graph creation

I think maybe spend 15 minutes trying to figure out what's up (my guess is we are sending a graphing messages twice, and don't need to) -- but otherwise feel free to just open an issue if it's not easy to solve. 

Otherwise, LGTM!

PR #1093, opened 2023-12-13
Add keyboard shortcut hint to toolbar button tooltip

marthacryan:
Adds keyboard shortcut in parentheses after the tooltip text for toolbar buttons and context menu items
image
image

### DIFF
--- a/mitosheet/src/mito/utils/keyboardShortcuts.tsx
+++ b/mitosheet/src/mito/utils/keyboardShortcuts.tsx
@@ -1,4 +1,4 @@
-import { Action, ActionEnum, KeyboardShorcut } from "../types"
+import { ActionEnum, KeyboardShorcut } from "../types"
 import { Actions } from "./actions"
 
 /**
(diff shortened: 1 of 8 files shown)


naterush:
Screen Shot 2023-12-14 at 11 48 48 AM

It's missing a space in the text. I also like it in () -- which it doesn't appear to have here.

mitosheet/src/mito/utils/keyboardShortcuts.tsx
```diff
@@ -286,15 +286,52 @@ export const keyboardShortcuts: KeyboardShorcut[] = [
...
+    return window.navigator.userAgent.toUpperCase().includes('MAC')
+        ? 'mac'
+        : 'windows';
+}
+
+/**
+ * @param action - The action to get the keyboard shortcut for.
+ * @returns A string describing the keyboard shortcut for the given action.
+ */
+export const getKeyboardShortcutString = (action: Action) => {
```
naterush, line 303: Wait, so we're notably not using the keyboard shortcuts defined in the action themselves. This is resulting in a O(actions * shortcuts) work when repeatedly calling this function. 

I know its a bit of duplicated data, but let's go through and just define the keyboard shortcuts in each action that has them. Then, let's just skip the .find step here, and use those.

naterush (changes requested):
Some cleanup for perf

naterush:
Code looks good -- still needs the extra space and some parens added to the tooltip.

naterush:
LGTM when tests pass!

PR #1097, opened 2023-12-14
Stop propagation of cmd+a when column header or tab editor is open

### LINKED ISSUE #1095: CMD+A keyboard doesn't work in column header cell editor
**Works in formulas**
1. Open a cell editor that has a formula in it
2. Press CMD + A to select the entire formula to edit. 
3. It works. Nice

**Does not work in column headers**
1. Open a column header 
2. Press CMD + A to select the entire column header to change
3. It closes the cell editor and selects the entire sheet.

https://github.com/mito-ds/mito/assets/18709905/6b76915a-e00b-4a3a-8adb-8b237954907e

marthacryan:
Fixes #1095. Adds a helper util to call in the `onKeyDown` event for the column header and tab name input divs to use.

### DIFF
--- a/mitosheet/src/mito/types.tsx
+++ b/mitosheet/src/mito/types.tsx
@@ -834,6 +834,12 @@ export interface KeyboardShortcut {
      */
     skipIfInTextInput?: boolean
 
+    /**
+     * Some keyboard shortcuts should only be triggered if the user does not have text selected.
+     * For example, cmd+c should copy the text in the input field, not copy the cells.
+     */
+    skipIfTextSelected?: boolean
+
     // For some keybindings (ex: cmd + y), we'd want to stop propagation
     // so that they don't open something in your browser
     preventDefaultAndStopPropagation?: boolean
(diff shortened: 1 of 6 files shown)


mitosheet/src/mito/types.tsx
```diff
@@ -828,6 +828,12 @@ export interface KeyboardShorcut {
     // the command registered to that ID in plugin.tsx.
     jupyterLabCommand?: string
 
+    /**
+     * Some keyboard shortcuts should only be triggered if the user is not in a text input. 
+     * For example. cmd+a
```
naterush, line 833: a little more here -- cmd + a should select the entire text input, not data in the sheet.

mitosheet/src/mito/types.tsx
```diff
@@ -818,7 +818,7 @@ export interface KeyBinding {
     metaKey?: boolean;
 }
 
-export interface KeyboardShorcut {
+export interface KeyboardShortcut {
```
naterush, line 821: Nice catch :)

naterush (commented):
We need to add to:
1. Cmd + C
2. Cmd + Z?
3. Cmd + Y?

Infrastructure looks good -- this is def the right approach here :)

naterush:
Oh, do you wanna add a little test for this one? You can just do Command + A for column headers, just so we have at least one sanity check on this.

PR #1099, opened 2023-12-14
Allow snowflake config variable names to be set for user auth

naterush:
# Description

This greatly improves the snowflake demos I can give - I always spend time fiddling around with these.

# Testing

Test it!

# Documentation

No.


mitosheet/mitosheet/api/get_validate_snowflake_credentials.py
```diff
@@ -25,9 +26,29 @@
 
 def get_cached_snowflake_credentials() -> Optional[SnowflakeCredentials]:
     global __cached_snowflake_credentials
+
+    # If the snowflake connector is importer, and the SNOWFLAKE_ACCOUNT environment variable is set,
+    # then we can use that to prepopulate the snowflake credentials.
+    if SNOWFLAKE_CONNECTOR_IMPORTED and 'SNOWFLAKE_ACCOUNT' in os.environ:
```
naterush, line 32: Should get for all of them.

naterush (commented):
Gotta update Github action.

aarondr77 (commented):
LGTM, but I think another great polish for demoing (another thing that comes up a lot for me when I show these features) is if there are env set, to use them in the generated code instead of exposing the username and password.

mitosheet/mitosheet/code_chunks/snowflake_import_code_chunk.py
```diff
@@ -36,13 +38,26 @@ def get_description_comment(self) -> str:
...
+                'cur = con.cursor()',
+            ]
+        else:
+            imports.append('import os')
+            snowflake_connection_code = [
+                'con = snowflake.connector.connect(',
+                f'    user=os.environ[\'SNOWFLAKE_USERNAME\'],',
+                f'    password=os.environ[\'SNOWFLAKE_PASSWORD\'],',
+                f'    account=os.environ[\'SNOWFLAKE_ACCOUNT\'],',
+                ')',
```
aarondr77, line 57: This should be the full connection string I think, not just the user, password, and account!

aarondr77:
# Importing twice is broken
1. Import one table using Snowflake
2. Import the same query again (or a different one)
3. Get an error

naterush:
Merging when tests pass

PR #1100, opened 2023-12-15
Add destination sheet index to merge

marthacryan:
Adds an optional param `destination_sheet_index` to the merge step performer / merge code chunk to enable editing existing merges. 

### Testing
Adds a test for the basic usage of the edit.

### DIFF
--- a/mitosheet/mitosheet/saved_analyses/step_upgraders/merge.py
+++ b/mitosheet/mitosheet/saved_analyses/step_upgraders/merge.py
@@ -136,3 +136,42 @@ def upgrade_merge_3_to_4(step: Dict[str, Any], later_steps: List[Dict[str, Any]]
         "step_type": "merge", 
         "params": params
     }] + later_steps
+
+
+def upgrade_merge_4_to_5(step: Dict[str, Any], later_steps: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    """
+    Support for destination_sheet_index to make edits to merges. 
+
+    OLD: {
+        "step_version": 4, 
+        "step_type": "merge", 
+        "params": {
+            sheet_index_one: 0,
+            sheet_index_two: 1,
+            merge_key_column_ids: [[old.merge_key_column_id_one, old.merge_key_column_id_two]],
+            selected_column_ids_one: List[_ids_],
+            selected_column_ids_two: List[_ids_],
+        }
+    }
+
+    NEW: {
+        "step_version": 5, 
+        "step_type": "merge", 
+        "params": {
+            sheet_index_one: 0,
+            sheet_index_two: 1,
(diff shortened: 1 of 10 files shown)


mitosheet/mitosheet/step_performers/merge.py
```diff
@@ -101,6 +109,10 @@ def transpile(
             )
         ]
     
+    
     @classmethod
     def get_modified_dataframe_indexes(cls, params: Dict[str, Any]) -> Set[int]:
+        destination_sheet_index = get_param(params, 'destination_sheet_index')
+        if destination_sheet_index: # If editing an existing sheet, that is what is changed
```
naterush, line 116: I think you have to do an explicit is not None check here -- otherwise if the merge sheet is the first (which is kinda rare, I guess, but can happen), we run into issues.

Do we need to fix this in pivot as well?

naterush (commented):
Code looking spot on! Great first PR.

A think you should aim for at least 5 tests of this fella -- there's probably a bunch of good examples in pivot overwriting!

mitosheet/mitosheet/step_performers/merge.py
marthacryan: Yeah I think we do

naterush (commented):
These tests look very good. I think one test to add would be:
1. Merge
2. Filter and rename and format a column in merge
3. Edit merge
4. Should just get the latter mrege

I only say it because filter and rename and format change state in a weird way sometimes, and so there are often bugs with them and other steps. 

Once you add this test, LGTM!

naterush:
Oh, something I totally missed -- you need to write a step upgrader, as we have a new prop now! There's documentation for that in the `mitosheet/saved_analyses/step_upgraders/__init__.py` -- but we need to bump the step version, and add in the new params, so we don't break old user analyses :)

mitosheet/mitosheet/saved_analyses/step_upgraders/merge.py
```diff
@@ -136,3 +136,42 @@ def upgrade_merge_3_to_4(step: Dict[str, Any], later_steps: List[Dict[str, Any]]
...
+    NEW: {
+        "step_version": 5, 
+        "step_type": "merge", 
+        "params": {
+            sheet_index_one: 0,
+            sheet_index_two: 1,
+            merge_key_column_ids: [[old.merge_key_column_id_one, old.merge_key_column_id_two]],
+            selected_column_ids_one: List[_ids_],
+            selected_column_ids_two: List[_ids_],
+            destination_sheet_index: 2
```
naterush, line 166: This should say : None

naterush (commented):
Tiny comment fix, then LGTM!

PR #1102, opened 2023-12-15
Only allow one sheet tab context menu to be open at a time

marthacryan:
Updates the context menus for sheet tabs to only be able to display one at a time.

### DIFF
--- a/tests/streamlit_ui_tests/streamlit.spec.ts
+++ b/tests/streamlit_ui_tests/streamlit.spec.ts
@@ -464,18 +464,21 @@ test.describe('Context Menus', () => {
     await expect(mito.locator('.mito-dropdown')).not.toBeVisible();
     await mito.getByTitle('Column1').click({ button: 'right' });
     await expect(mito.locator('.mito-dropdown')).toBeVisible();
+    await expect(mito.locator('.mito-dropdown')).toHaveText(/Sort A to Z/);
   })
 
   test('Index Header', async ({ page }) => {
     const mito = await getMitoFrameWithTestCSV(page);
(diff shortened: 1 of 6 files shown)


naterush:
I agree we should join this with the other dropdown state.

naterush:
If you:
1. Right click on index header
2. Click on import
3. Import dropdown opens correctly


But if you:
1. Right click on sheet tab
2. Click on import
3. Import dropdown opens and then closes quickly. 

I think likely just because the closing of the sheet tab dropdown needs to check if is in-fact the sheet tab that is currently open -- so should be an easy fix. See video below. 

https://github.com/mito-ds/mito/assets/21002431/9d10cbe3-9a4d-4e5e-aa20-5ed0f2eed317

naterush:
Noting we want to add a test or two here

tests/streamlit_ui_tests/streamlit.spec.ts
```diff
@@ -457,3 +457,51 @@ test.describe('Keyboard Shortcuts', () => {
     await expect(mito.locator('.endo-column-header-container-selected')).toHaveCount(1);
   });
 });
+
+test.describe('Context Menus', () => {
+  test('Column Header', async ({ page }) => {
+    const mito = await getMitoFrameWithTestCSV(page);
+    await expect(mito.locator('.mito-dropdown')).not.toBeVisible();
+    await mito.getByTitle('Column1').click({ button: 'right' });
+    await expect(mito.locator('.mito-dropdown')).toBeVisible();
```
naterush, line 466: Can you actually check the right header is open -- e.g. just checking that the first action is available? I think it makes sense to do this for each of these.

naterush (commented):
Minor test cleanup, then LGTM

PR #1109, opened 2023-12-20
mitosheet: don't open sheet tab action if shift is pressed

naterush:
# Description

See title.

# Testing

Test this.

# Documentation

No.


marthacryan (approved):
LGTM

marthacryan (changes requested):
Sorry just realized this code is duplicated in a few places. I think we should move this into the `Mito` component and stop propagation in the `onContextMenu` if `e.shiftKey`. Then we can remove this duplicated code in other places

naterush:
Opening an issue and merging for now.

PR #1111, opened 2023-12-22
Add ability to resize taskpane

marthacryan:
Adds a div between the `EndoGrid` and the `Taskpane` that allows the user to resize the taskpane on drag. 
resize-taskpane

### DIFF
--- a/mitosheet/css/taskpanes/DefaultTaskpane.css
+++ b/mitosheet/css/taskpanes/DefaultTaskpane.css
@@ -18,7 +18,6 @@
         The border around the entire Mito container and the footer 
         create the right and bottom borders of the taskpane.
     */ 
-    border-left: 1px solid var(--mito-text-light);
     border-top: 1px solid var(--mito-text-light);
 
     box-sizing: border-box;
--- a/mitosheet/css/taskpanes/DefaultTaskpane.css
+++ b/mitosheet/css/taskpanes/DefaultTaskpane.css
@@ -37,6 +37,18 @@
     justify-items: center;
 }
 
+.taskpane-resizer-container {
+    width: 5px;
+    cursor: col-resize;
+    border-top: 1px solid var(--mito-text-light);
+    display: flex;
+}
+.taskpane-resizer {
+    width: 1px;
+    cursor: col-resize;
+    background-color: var(--mito-text-light);
+}
+
 .default-taskpane-body-div {
     display: flex;
     flex-direction: column;
--- a/mitosheet/css/taskpanes/DefaultTaskpane.css
+++ b/mitosheet/css/taskpanes/DefaultTaskpane.css
@@ -42,6 +42,7 @@
     cursor: col-resize;
     border-top: 1px solid var(--mito-text-light);
     display: flex;
+    background-color: var(--mito-background);
 }
 .taskpane-resizer {
     width: 1px;
--- a/mitosheet/src/mito/Mito.tsx
+++ b/mitosheet/src/mito/Mito.tsx
@@ -1124,16 +1124,18 @@ export const Mito = (props: MitoProps): JSX.Element => {
                     </div>
                     {uiState.currOpenTaskpane.type !== TaskpaneType.NONE && 
                         <>
-                            <div
-                                className='taskpane-resizer-container'
-                                onMouseDown={(event) => {
-                                    event.preventDefault();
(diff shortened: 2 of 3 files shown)


mitosheet/src/mito/Mito.tsx
```diff
@@ -1003,12 +1003,35 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
             data-jp-suppress-context-menu 
             ref={mitoContainerRef} 
             tabIndex={0}
+            onMouseMove={(event) => {
+                if (resizingTaskpane) {
+                    event.preventDefault();
+                    event.stopPropagation();
+                    const { clientX } = event;
+                    const rawPanelWidth = document.body.clientWidth - clientX;
+                    const panelWidth = Math.max(
```
naterush, line 1020: 1. Can we name this taskpaneWidth? (and other variables above)
2. Also, I think let's clamp this with max and min to the current small and large taskpane sizes -- as this is what we have tested the UI to so far!

mitosheet/src/mito/Mito.tsx
```diff
@@ -1097,6 +1120,20 @@ export const Mito = (props: MitoProps): JSX.Element => {
                             actions={actions}
                         />
                     </div>
+                    <div
+                        className='taskpane-resizer'
```
naterush, line 1124: I don't see this CSS

naterush (commented):
Very fun polish here - I love it.

One this is that 2px is hard to hover over -- it's just so small. Can we make the valid hover area bigger? Visually, I think it's the right size -- it just would be nice if we have 4px to hover over.

Ping me and I can review again before merging

mitosheet/src/mito/Mito.tsx
marthacryan: Done!

mitosheet/src/mito/Mito.tsx
```diff
@@ -1003,12 +1003,35 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
 
+    const [resizingTaskpane, setResizingTaskpane] = useState(false);
+
     return (
         <div 
             className="mito-container" 
             data-jp-suppress-context-menu 
             ref={mitoContainerRef} 
             tabIndex={0}
+            onMouseMove={(event) => {
```
naterush, line 1014: I think we need to add more listeners here to cancel the drag. E.g. what happens if you:
1. Start resizing
2. Leave the Mito Container
3. Then, mouse up

When you go back inside the Mito Container, you're still resizing. 

I guess this is actually fine -- you just have to click to make it stop.

naterush (changes requested):
## Bugs in Resizing

https://github.com/mito-ds/mito/assets/21002431/d3f85b48-c2e4-4389-bacc-60000c42ed1e

1. Resizing appears to be getting the wrong width -- it's like 50px more than where my mouse is. Tested this on JupyterLab on safari and Chrome. **Perhaps you were developing on Streamlit, where because of iframe's your solution works?**
2. The max width limiting certain does not seem to be working - it goes way more than 350 px. 

## Potential Bug in Width Setting

Screen Shot 2024-01-04 at 8 00 25 AM

There is this code which cases on if the narrow taskpane is open before setting the width. I think this looks wrong now?

naterush:
Also, can we add a frontend test for this :)

mitosheet/src/mito/Mito.tsx
marthacryan: Yeah I think this is ok - fixing it might cause more trouble than good.

mitosheet/css/taskpanes/DefaultTaskpane.css
```diff
@@ -38,6 +37,18 @@
     justify-items: center;
 }
 
+.taskpane-resizer-container {
+    width: 5px;
+    cursor: col-resize;
+    border-top: 1px solid var(--mito-text-light);
+    display: flex;
+}
```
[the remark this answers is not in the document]
marthacryan: Works in streamlit dark mode already!

marthacryan:
> There is this code which cases on if the narrow taskpane is open before setting the width. I think this looks wrong now?

I think this is actually correct! The wide taskpane (as far as I understand) is only used for the graph, and we wouldn't want it to apply in that case.

PR #1115, opened 2024-01-02
Reorder code optimization

naterush:
# Description

This PR does a few cool things with code optimization:
1. Allows certain code chunks to reorder themselves, for better optimization. Currently, this is a simple greedy algorithm -- that performs reasonably well!
2. Improves our column insert code generation to be more standard in the case you add on in the end of the dataframe.

# Testing

I have a few more tests. Can you think of a few more, so we can be really secure in this change? 

# Documentation

No.


mitosheet/mitosheet/code_chunks/code_chunk.py
```diff
@@ -132,4 +146,27 @@ def get_parameterizable_params(self) -> List[Tuple[ParamValue, ParamType, ParamS
         """
         return []
     
+    def can_be_reordered_with(self, code_chunk: "CodeChunk") -> bool:
+        """
+        Returns true if the passed code chunk can be moved from
```
naterush, line 151: TODO: a bit better documentation.

mitosheet/mitosheet/code_chunks/step_performers/import_steps/excel_import_code_chunk.py
```diff
@@ -62,6 +62,8 @@ def get_created_sheet_indexes(self) -> List[int]:
     def get_parameterizable_params(self) -> List[Tuple[ParamValue, ParamType, ParamSubtype]]:
         return [(f'r{get_column_header_as_transpiled_code(self.file_name)}', 'import', 'file_name_import_excel')]
 
+    def get_source_sheet_indexes(self) -> List[int]:
+        return []
```
naterush, line 66: Note that if we have created sheet indexes, should define source as empty.

naterush (commented):
A bit more documentation

aarondr77 (commented):
LGTM. Code optimization is becoming a bigger and bigger black box to me, so having these good tests is even more important. I added a few more. Glad we have some many backend tests and that we do the run generated code and compare it to the state df by default!

aarondr77:
It seems like there is some non-determinism in our generated code, maybe specifically to pivot tables. 

https://github.com/mito-ds/mito/issues/1129

naterush:
@aarondr77 ignoring the non-determinism for now, since it's unrelated. If it cases further test issues, we can fix this up.

Issue #1105, opened 2023-12-18
Graphs generate in only black


aarondr77:
1. Create a graph
2. See that it is Black
3. Fix that

aarondr77:
In Jupyter on Chrome or Safari

naterush:
Note that this requires Streamlit to be installed. If Streamlit is installed, then the default plotly theme is set to `streamlit`, which does funky things with colors so they can get it to match the app. 

The fix for this is simple:
```
import plotly.io as pio
if pio.templates.default == 'streamlit':
    pio.templates.default = 'plotly'
```

Because streamlit updates the default theme when it's imported, you just have to make sure to run this code after importing streamlit -- aka, just before graph generation.

PR #1121, opened 2024-01-05
Fix graph colors

naterush:
# Description

Makes graph colors work correctly if streamlit is installed.

Closes https://github.com/mito-ds/mito/issues/1105

# Testing

```
import streamlit
import mitosheet
```

Then, make a graph.

Do this on this branch, and on dev, to confirm we fixed it. 

# Documentation

No.


marthacryan (approved):
LGTM

PR #1123, opened 2024-01-05
Add TODAY function

marthacryan:
* Adds the TODAY function. 
* Adds a couple small tests for the function
* Fixes a small issue with the script that generates the function documentation.


mitosheet/mitosheet/tests/public/v3/sheet_functions/date_functions/test_today.py
```diff
@@ -0,0 +1,23 @@
...
+import pandas as pd
+
+from mitosheet.public.v3.sheet_functions.date_functions import TODAY
+from mitosheet.tests.test_utils import create_mito_wrapper, create_mito_wrapper_with_data
+from datetime import date
+
+def test_today_valid():
+    assert TODAY() == pd.to_datetime(date.today())
+
+def test_today_in_sheet():
```
naterush, line 20: Can we write a test that creates an analysis and replays it that has the TODAY formula in it? Just to make sure it runs -- it should obvi return the same value between runs.

naterush (commented):
Working great. I think let's open that one issue on difference from excel, and otherwise LGTM.

PR #1127, opened 2024-01-05
Fix undo/redo bug in input elements

### LINKED ISSUE #1125: Undo and redo do not work in text inputs
**Describe the bug**

1. Type some text (e.g. in find and replace)
2. Press Command + Z
3. See that the text does not undo

I think we need to make our keyboard shortcut handling a bit more robust -- potentially adding to the keyboard shortcut type.

marthacryan:
Fixes #1125.


aarondr77:
This fix worked for Streamlit, but not for Jupyter Lab, Chrome. 

Let's not merge until:
1.  we fix for Jupyter Lab, Chrome since that is the most popular enviornment
2. add a frontend test 

No need to get this into the current release

naterush:
Closing this issue since we're not working on it, but we can prioritize it in the future if we want.

PR #1128, opened 2024-01-06
Mitosheet: replay edits on merge

naterush:
# Description

Replays edits on merge. 

Note that this PR is dramatically improved once we merge the PR https://github.com/mito-ds/mito/pull/1115 -- as the column insertion code at the end of the dataframe gets much more robust in this case. This is particularly useful in the case of merge, where editing is very likely to change the number of columns in the dataframe.

# Testing

See new tests. Maybe write a few. 

# Documentation

No.


aarondr77:
This is so good! And will be so much better once we test it with the reorder columns PR.

aarondr77:
Can we add a frontend test to go along with it?

PR #1132, opened 2024-01-09
Create graph from selection in dataframe

marthacryan:
Uses the selected columns to populate the axes in the graph configuration. Only covers the cases for bar charts, scatter plots, and line graphs because they are the only charts that can be created new. 
- Bar charts, line graphs : 
    - If the leftmost column is anything but a number, it is the x axis
    - If the y axes aren’t all the same type,
    - else only populate the y axis
- scatter plot:
    - one column: only populate y axis
    - multiple columns: first column is the x axis, rest of columns are the y axis

### DIFF
--- a/mitosheet/src/mito/utils/actions.tsx
+++ b/mitosheet/src/mito/utils/actions.tsx
@@ -1229,7 +1229,7 @@ export const getActions = (
             titleToolbar: 'Graph',
             longTitle: 'Create new graph',
             actionFunction: async () => {
-                const selectedColumnIds = getSelectedColumnIDsWithEntireSelectedColumn(gridState.selections, sheetDataArray[uiState.selectedSheetIndex])
+                const selectedColumnIds = getSelectedColumnIDsWithEntireSelectedColumn(gridState.selections, sheetDataArray[uiState.selectedSheetIndex], true)
                 await openGraphEditor(setEditorState, sheetDataArray, setUIState, sheetIndex, mitoAPI, undefined, selectedColumnIds);
             },
             isDisabled: () => {return doesAnySheetExist(sheetDataArray) ? defaultActionDisabledMessage : 'There are no dataframes to graph. Import data.'},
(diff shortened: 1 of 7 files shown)


mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1229,7 +1229,8 @@ export const getActions = (
             titleToolbar: 'Graph',
             longTitle: 'Create new graph',
             actionFunction: async () => {
-                await openGraphEditor(setEditorState, sheetDataArray, setUIState, sheetIndex, mitoAPI);
+                const selectedColumnIds = getSelectedColumnIDsWithEntireSelectedColumn(gridState.selections, sheetDataArray[uiState.selectedSheetIndex])
```
aarondr77, line 1232: In Excel, you get the same graph regardless of the order you select the columns. 

ie: if you have column A and B:
1. Select B
2. CMD+Select A
3. Build a graph and it will be the same as vice versa 

In Mito, that is not true because the `getSelectedColumnIDsWithEntireSelectedColumn` returns the columns in the order of the selections. 

Let's add an optional param to `getSelectedColumnIDsWithEntireSelectedColumn` to specify if you want to return the columns in sorted order and then use that here.

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -1241,7 +1242,8 @@ export const getActions = (
             iconToolbar: GraphIcon,
             longTitle: 'Create new bar chart',
             actionFunction: async () => {
-                await openGraphEditor(setEditorState, sheetDataArray, setUIState, sheetIndex, mitoAPI, GraphType.BAR);
+                const selectedColumnIds = getSelectedColumnIDsWithEntireSelectedColumn(gridState.selections, sheetDataArray[uiState.selectedSheetIndex])
```
aarondr77, line 1245: Update to use sorted columns also

tests/streamlit_ui_tests/streamlit.spec.ts
```diff
@@ -258,6 +258,31 @@ test.describe('Home Tab Buttons', () => {
...
+    await mito.getByTitle('Column1').click();
+    await mito.getByTitle('Column2').click({ modifiers: ['Shift']});
+    
+    await clickButtonAndAwaitResponse(page, mito, { name: 'Graph' })
+
+    await expect(mito.locator('p.select-text').getByText('Column1')).toBeVisible();
+    await expect(mito.locator('p.select-text').getByText('Column2')).toBeVisible();
+  })
+
+  test('Scatter plot from selection', async ({ page }) => {
```
aarondr77, line 273: Can you add a test for selecting the columns in the opposite order as well?

aarondr77 (commented):
Awesome job. You matched how Excel handles different selected column, graph types pairs super nicely. 

A little bit of code cleanup and then LGTM in to the long standing graph branch

PR #1139, opened 2024-01-11
mitosheet: support % string to float

aarondr77:
# Description

Let users cast a string, "82%" to a float.  A user reported this bug today

# Testing

See tests

# Documentation


naterush:
LGTM once tests pass.

PR #1143, opened 2024-01-12
Add ability to make multi-level dropdowns

marthacryan:
* Adds props to `DropdownItem`: `subMenu?: JSX.Element` and `onMouseEnter`. The creator of the dropdown is expected to keep track of the state for displaying submenus, because we want to make it easy to switch between submenus without bugginess. 
* Adds `position?: 'vertical' | 'horizontal'` prop to `Dropdown` so that the position can be determined correctly for submenus. We want the submenu to always appear directly to the right of the dropdown item, so the handling for `position === 'horizontal'` is just to set the bounding box to be directly to the right of the dropdown item, regardless of the position. 
* Adds a dummy example to `IndexHeaderContextMenu.tsx` for testing purposes. 
* **Note: also adds support for checkmarks in dropdowns.**
  * Adds `hasCheckmark` and `canHaveCheckmark` props to the `DropdownItem` class to ensure proper spacing

### DIFF
--- a/mitosheet/src/mito/components/endo/IndexHeaderContextMenu.tsx
+++ b/mitosheet/src/mito/components/endo/IndexHeaderContextMenu.tsx
@@ -20,14 +20,11 @@ export default function IndexHeaderContextMenu(props: {
     actions: Actions;
     closeOpenEditingPopups: (taskpanesToKeepIfOpen?: TaskpaneType[]) => void;
 }): JSX.Element {
-    const [ currentSubmenu, setCurrentSubmenu ] = React.useState<string | undefined>(undefined);
-
     return (
         <Dropdown
             display={props.display}
             closeDropdown={() => props.setUIState((prevUIState) => {
                 const isCurrOpenDropdown = isCurrOpenDropdownForCell(prevUIState, props.rowIndex, -1);
-                setCurrentSubmenu(undefined);
                 return {
                     ...prevUIState,
                     currOpenDropdown: isCurrOpenDropdown ? undefined : prevUIState.currOpenDropdown
@@ -45,17 +42,6 @@ export default function IndexHeaderContextMenu(props: {
             <DropdownItem
                 {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Delete], props.closeOpenEditingPopups)}
                 title='Delete Row'
-                onMouseEnter={() => setCurrentSubmenu('Delete Row')}
-                subMenu={<Dropdown display={currentSubmenu === 'Delete Row'} position='horizontal' closeDropdown={() => props.setUIState((prevUIState) => {
-                    const isCurrOpenDropdown = isCurrOpenDropdownForCell(prevUIState, props.rowIndex, -1);
-                    return {
-                        ...prevUIState,
-                        currOpenDropdown: isCurrOpenDropdown ? undefined : prevUIState.currOpenDropdown
-                    }
-                })}>
-                    <DropdownItem {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Copy], props.closeOpenEditingPopups)}/>
-                    <DropdownItem {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Copy], props.closeOpenEditingPopups)}/>
-                </Dropdown>}
             />
 
             <DropdownSectionSeperator isDropdownSectionSeperator={true}/>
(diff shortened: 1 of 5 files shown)


mitosheet/src/mito/components/endo/IndexHeaderContextMenu.tsx
```diff
@@ -20,11 +20,14 @@ export default function IndexHeaderContextMenu(props: {
     actions: Actions;
     closeOpenEditingPopups: (taskpanesToKeepIfOpen?: TaskpaneType[]) => void;
 }): JSX.Element {
+    const [ currentSubmenu, setCurrentSubmenu ] = React.useState<string | undefined>(undefined);
+
```
naterush, line 24: I think we additionally want to set the current submenu to undefined when the parent dropdown closes. 

Otherwise, if you:
1. Open a dropdown
2. Mouse over to open a submenu
3. Open another dropdown
4. Open the original dropdown
5. The original submenu is open

mitosheet/src/mito/components/endo/IndexHeaderContextMenu.tsx
```diff
@@ -20,11 +20,14 @@ export default function IndexHeaderContextMenu(props: {
...
     return (
         <Dropdown
             display={props.display}
             closeDropdown={() => props.setUIState((prevUIState) => {
                 const isCurrOpenDropdown = isCurrOpenDropdownForCell(prevUIState, props.rowIndex, -1);
+                setCurrentSubmenu(undefined);
```
naterush, line 30: Oh -- we do close it here?

I think perhaps this function doesn't get close if another dropdown is open -- no dropdown is closed, only a new one is open. 

Maybe we can remove this call, and just use an effect on props.display that sets the submenu as undefined when the dropdown is closed.

mitosheet/src/mito/components/endo/IndexHeaderContextMenu.tsx
```diff
@@ -50,7 +64,19 @@ export default function IndexHeaderContextMenu(props: {
...
 
-            <DropdownItem {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.RESET_AND_DROP_INDEX], props.closeOpenEditingPopups)}/>
+            <DropdownItem
+                onMouseEnter={() => setCurrentSubmenu('Copy Row')}
+                subMenu={<Dropdown display={currentSubmenu === 'Copy Row'} position='horizontal' closeDropdown={() => props.setUIState((prevUIState) => {
+                    const isCurrOpenDropdown = isCurrOpenDropdownForCell(prevUIState, props.rowIndex, -1);
+                    return {
+                        ...prevUIState,
+                        currOpenDropdown: isCurrOpenDropdown ? undefined : prevUIState.currOpenDropdown
+                    }
```
naterush, line 74: Is this code still right? I'm a little confused. Maybe this should be closing the subdropdown only?

naterush (commented):
Ok -- overall, this works really well. There's one API change I think we can make now that feels like it will make it much easier to work with these in the future -- but I don't think we need to make it now (let's just open an issue). 

It is a custom hook that looks like:
```
const useSubMenu<T> = (displayParentDropdown: bool): {openSubMenu: T | undefined, onMouseEnterSubMenu: (subMenu: T) => (() => void), getDropdownProps: (subMenu: T) => Partial<DropdownProps>} => {
     const [openSubMenu, setOpenSubmenu] = ...
     
     // that use effect here

     // Some function definitions for onMouseEnter and getDropdownProps
}
```

Then, when you use them in a component, it looks like this:
```
const {openSubMenu, onMouseEnterSubMenu, getDropdownProps} =  React.useState<'Delete Row' | 'Copy Row'>(props.display);

...

          <DropdownItem
                 onMouseEnter={onMouseEnterSubMenu('Copy Row')}
                 subMenu={<Dropdown {...getDropdownProps('Copy Row')}>
                     <DropdownItem hasCheckMark canHaveCheckMark {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Copy], props.closeOpenEditingPopups)}/>
                     <DropdownItem canHaveCheckMark {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.Copy], props.closeOpenEditingPopups)}/>
                 </Dropdown>}
                 {...getPropsForContextMenuDropdownItem(props.actions.buildTimeActions[ActionEnum.RESET_AND_DROP_INDEX], props.closeOpenEditingPopups)}/>

```

Once we have a few of these in real usage, we can attempt this refactor -- I think it might be nice.

PR #1147, opened 2024-01-16
Refactor graph code to use useLiveUpdatingParams and graphDataArray

marthacryan:
Updates the graph code to make upcoming changes easier to manage. 
* Uses `useLiveUpdatingParams` instead of having a bunch of code to do the same thing
* Moves to `graphDataArray` 
* Updates the `graphData` type to remove the `graphParams`. This will help separate the `graphParams` from the `graphOutput` and `graphTabName`, because upcoming changes are going to make separating them very useful

Todo:
- [x] Duplicate graph throws error -> for some reason this is called when you switch to the duplicate graph, and the second time it's called, `graphDataArray` is empty. 
- [x] Returning to a graph throws error: `get_params.py: 32` -> `for key, value in execution_data_to_match.items()`. Call to `items()` throws an error: `graph_id`.


naterush (commented):
Noting that switching from graphs to pivots caused a sheet crashing. 
Screen Shot 2024-01-16 at 10 37 18 AM

mitosheet/mitosheet/api/get_params.py
```diff
@@ -27,6 +28,14 @@ def get_params(params: Dict[str, Any], steps_manager: StepsManagerType) -> Optio
         if step.step_id == step_id_to_match:
             found_params = step.params
         
+        if step.params is not None and params_to_match is not None and len(params_to_match) > 0:
+            all_matched = True
+            for key, value in params_to_match.items():
+                if step.params[key] != value:
```
naterush, line 34: I think we could change this `[key]` access to a .get - both here and above in this function -- just to protect ourselves

mitosheet/mitosheet/step_performers/graph_steps/graph_duplicate.py
```diff
@@ -17,6 +18,11 @@
 class GraphDuplicateStepPerformer(StepPerformer):
     """
     This steps duplicates a graph of a given graphID. 
+    NOTE: WE ARE NOT USING THIS STEP IN THE CURRENT VERSION OF MITOSHEET. 
+    The reason is that we separated the graph params and the graph data, 
+    which caused graph duplication to not make sense as its own step. 
+    Now, on the frontend we just get the graph params and send them to the backend
+    as a new graph. We're keeping this around so that it doesn't break any analyses.
```
naterush, line 25: Nice comment

mitosheet/mitosheet/tests/step_performers/graph_steps/test_graph_duplicate.py
naterush, line 1: I think we can keep these tests -- this should still work!

mitosheet/src/mito/Mito.tsx
```diff
@@ -334,54 +332,6 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
-
-    // When we switch graphID's make sure that we keep the previousGraphIndex up to date
-    useEffect(() => {
-        previousGraphIndex.current = uiState.selectedGraphID !== undefined ?
-            Object.keys(analysisData.graphDataDict || {}).indexOf(uiState.selectedGraphID) : -1
-    }, [uiState.selectedGraphID])
-
-    // Update the selected sheet tab when the number of graphs change. 
-    useEffect(() => {
-        const graphIDs = Object.keys(analysisData.graphDataDict || {})
```
naterush, line 349: Do we still want this effect? It would be different behavior -- but when there's a graph, switching to it?

mitosheet/src/mito/Mito.tsx
```diff
@@ -334,54 +332,6 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
-                        graphID: newGraphID,
-                    },
-                }
-            })
-
-            // Update the previous graph index for next time
-            previousGraphIndex.current = graphIDs.indexOf(newGraphID)
-
-        // Handle graph removal
-        } else if (previousNumGraphs > newNumGraphs) {
```
naterush, line 372: Also, I wonder about this case -- what if there are graphs removed -- do we need to close the sidebar?

naterush: Honestly, both of these might be better if they were not in the effect -- but wondering if we handle them.

mitosheet/src/mito/components/footer/GraphSheetContextMenu.tsx
```diff
@@ -48,8 +52,17 @@ export default function GraphSheetTabContextMenu(props: {
...
-        const newGraphID = getRandomId()
-        await props.mitoAPI.editGraphDuplicate(props.graphID, newGraphID)
+        await openGraphEditor(
+            props.setEditorState,
+            props.sheetDataArray, 
+            props.uiState, 
+            props.setUIState,
+            props.mitoAPI,
+            props.graphDataArray,
+            props.graphID,
```
naterush, line 62: In duplicate, shouldn't we be opening with a new graph id?

naterush: Maybe the function params are just confusing

mitosheet/src/mito/components/taskpanes/Graph/GraphExportTab.tsx
```diff
@@ -52,7 +51,7 @@ function GraphExportTab(
                 <TextButton
                     variant='dark'
                     onClick={copyShowGraphCode}
-                    disabled={props.loading || props.graphOutput === undefined}
+                    disabled={props.graphOutput === undefined}
```
naterush, line 54: I think we need to keep loading here, yes? Or is it that loading is no longer accessible.

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -33,16 +25,68 @@ const GraphSidebar = (props: {
...
+        StepType.Graph,
+        props.mitoAPI,
+        props.analysisData,
+        50, // 50 ms debounce delay
+        {
+            getBackendFromFrontend: convertFrontendtoBackendGraphParams,
+            getFrontendFromBackend: convertBackendtoFrontendGraphParams,
+        },
+        {
+            // If we have a destination sheet index, we make sure to not overwrite the pivot
```
naterush, line 51: old comment

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -80,52 +114,13 @@ const GraphSidebar = (props: {
...
     useEffect(() => {
         if (!(window as any).Plotly) {
             void props.mitoAPI.log('plotly_define_failed');
         }
     }, [])
 
-    // Whenever the graph is resized, we update it (so it resizes as well)
-    useEffectOnResizeElement(() => {
-        setGraphUpdatedNumber(old => old + 1)
-    }, [], props.mitoContainerRef, '#mito-center-content-container')
```
naterush, line 102: I think we need to still run this effect -- it makes sure graphs are sized correctly. 

I think you can just change `setGraphUpdatedNumber(old => old + 1)` to like `setGraphParams` to a copy (make sure to test resizing).

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -142,44 +137,7 @@ const GraphSidebar = (props: {
...
-        const boundingRect: DOMRect | undefined = props.mitoContainerRef.current
-            ?.querySelector('#graph-div')
-            ?.getBoundingClientRect();
-
-        if (boundingRect !== undefined) {
-            await props.mitoAPI.editGraph(
-                graphID,
-                graphParams,
-                `${boundingRect?.height - 10}px`, // Subtract pixels from the height & width to account for padding
-                `${boundingRect?.width - 20}px`,
```
naterush, line 164: Do we need to still be getting the height? I feel like we might -- where does this happen now?

naterush: Oh -- I wonder if 100% just works in every case now. If it did, why would we be using the bounding rect above?

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -272,15 +228,6 @@ const GraphSidebar = (props: {
...
                 </div>
-                
-                {loading &&
-                    <div className='popup-div'>
-                        <LoadingSpinner />
-                        <p className='popup-text-div'>
-                            loading
-                        </p>
-                    </div>
-                }
```
naterush, line 283: We need loading indicator still!

naterush: Let's add this back to Export as well

mitosheet/src/mito/components/taskpanes/Graph/graphUtils.tsx
```diff
@@ -133,59 +134,6 @@ export const getDefaultSafetyFilter = (sheetDataArray: SheetData[], sheetIndex:
...
-        // Filter out column headers that no longer exist
-        const validColumnIDs = sheetDataArray[graphDataSourceSheetIndex] !== undefined ? sheetDataArray[graphDataSourceSheetIndex].data.map(c => c.columnID) : [];
-        const xAxisColumnIDs = intersection(
-            validColumnIDs,
-            graphParamsCopy.graphCreation.x_axis_column_ids
-        )
-        const yAxisColumnIDs = intersection(
-            validColumnIDs,
-            graphParamsCopy.graphCreation.y_axis_column_ids
-        )
```
naterush, line 170: This is a bug. We need to make sure to filter out invalid columns from the existing params. 

Check out how Pivot does it. It takes the existingParams in it's getDefaultPivotParams.

I think we can do the same for graph. If these existingParams are defined, then we can filter out invalid columns using this code.

mitosheet/src/mito/components/taskpanes/Graph/graphUtils.tsx
```diff
@@ -238,20 +186,36 @@ export const getGraphTypeFullName = (graphType: GraphType): string => {
...
+    }
+}
+
+export const convertFrontendtoBackendGraphData = (graphDataFrontend: GraphDataFrontend): GraphDataBackend => {
+    return {
+        graph_id: graphDataFrontend.graphID,
+        graph_output: graphDataFrontend.graphOutput,
+        graph_tab_name: graphDataFrontend.graphTabName,
+    }
+}
```
naterush, line 203: This feels like overkill. Let's just use the python variables on the frontend too - and skip this conversion.

mitosheet/src/mito/index.tsx
```diff
@@ -1,6 +1,6 @@
 export { Mito } from './Mito';
 export { 
-    AnalysisData, GraphDataBackend, GraphDataDict, GraphParamsBackend, PublicInterfaceVersion, SheetData, UserProfile,
+    AnalysisData, GraphDataBackend, GraphDataArray as graphDataArray, GraphParamsBackend, PublicInterfaceVersion, SheetData, UserProfile,
```
naterush, line 3: This seems wrong

naterush (changes requested):
Overall, this certainly moves this code in the right direction. A few pieces of behavior appear to be lost -- but going to test it now.

naterush:
# Issues with Graph Sizing

How we did it previous:
1. Got the size of the div that contains the graph
2. Tell Plotly to make it this big in PX
3. When resizing, use a hook to recreate the graph

You incidentally removed this when refactoring this code.

Ideally, we'd just be able to set the graph height and width as 100% -- and then resize it's container and it would all work. Can you experiment with this for a bit and see if you can get this working?

If this does not make sense let me know -- happy to further clarify.

naterush:
# Bug with Undo

1. Import data
2. Click Graph
3. Click undo
4. The graphing taskpane stays open, with some default params

This should close the graphing taskpane, presumably?

naterush:
# Bug with Escape

Escape closes the graphs. I think we should probably skip escape closing graphs when on the graph taskpane -- since there are other reasons to press escape (e.g. close a dropdown).

mitosheet/mitosheet/tests/step_performers/graph_steps/test_graph_duplicate.py
marthacryan: They don't work though, because the params aren't copied over

naterush: This is an issue for upgrading analyses, I'm realizing!

mitosheet/src/mito/components/endo/utils.tsx
```diff
@@ -185,7 +186,7 @@ export const doesSheetContainData = (sheetIndex: number, sheetDataArray: SheetDa
     Determines if a graphing tab is selected in the sheet
 */
 export const getGraphIsSelected = (uiState: UIState): boolean => {
-    return uiState.selectedTabType === 'graph' && uiState.selectedGraphID !== undefined;
+    return uiState.selectedTabType === 'graph' && uiState.currOpenTaskpane.type === TaskpaneType.GRAPH;
```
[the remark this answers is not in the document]
marthacryan: Are we even using this anymore?

mitosheet/src/mito/components/footer/GraphTabContextMenu.tsx
```diff
@@ -58,13 +71,22 @@ export default function GraphSheetTabContextMenu(props: {
...
         props.setUIState(prevUIState => {
             return {
                 ...prevUIState,
                 currOpenTaskpane: {
                     type: TaskpaneType.GRAPH, 
-                    graphID: props.graphID, 
-                    graphSidebarTab: GraphSidebarTab.Export
+                    graphSidebarTab: GraphSidebarTab.Export,
+                    openGraph: {
+                        type: 'existing_graph',
```
marthacryan, line 86: Are these types defined anywhere? I feel like they're long enough that it's worth making sure they're correct through typechecking.

mitosheet/src/mito/components/footer/SheetTabContextMenu.tsx
```diff
@@ -1,27 +1,21 @@
...
-import { getDefaultGraphParams } from '../taskpanes/Graph/graphUtils';
 import { TaskpaneType } from '../taskpanes/taskpanes';
 
 /*
     Helper function for finding all of the graph tab names
     that are created from a given sheet index
 */
-export const getGraphTabNamesAndIDsFromSheetIndex = (sheetIndex: number, graphDataDict: GraphDataDict): ({graphTabName: string, graphID: GraphID})[] => {
-    // Filter to only grapsh with the sheetIndex, and then get a list of the graph tab names
-    const filteredGraphDataJSON: GraphDataDict = Object.fromEntries(Object.entries(graphDataDict || {}).filter(([, graphData]) => {
```
marthacryan, line 19: I removed this code but I actually wanted to check in about what it originally did - I'm kind of confused by the comment?

mitosheet/src/mito/components/footer/SheetTabContextMenu.tsx
```diff
@@ -132,7 +112,6 @@ export default function SheetTabContextMenu(props: {
                 // doesn't compete updating the uiState to this sheet instead of
                 // the new graphID that we're creating
                 e?.stopPropagation()
-                void graphData()
```
marthacryan, line 135: Hmm I think I did this but I'm confused about why this still works. I checked and it looks like it still works?

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -143,148 +125,179 @@ const GraphSidebar = (props: {
...
+    // By switching either:
+    // 1. to a data sheet, if there are no graphs left
+    // 2. to the next graph, if we were on a graph that was deleted
+    // 3. to the last graph
+
+    const previousNumberOfGraphsRef = React.useRef<number>(props.graphDataArray.length);
+    const previouslySelectedGraphIndexRef = React.useRef<number>(props.graphDataArray.findIndex(graphData => graphData.graph_id === props.openGraph.graphID));
+    useEffect(() => {
+        const handleDeletedGraphs = async () => {
+            console.log("HANDLING", props.graphDataArray.length, previousNumberOfGraphsRef)
```
marthacryan, line 140: Console log

marthacryan: There are a few in this file

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -143,148 +125,179 @@ const GraphSidebar = (props: {
...
+    // If the number of graphs decreases, we make sure we're still on a graph that exists
+    // By switching either:
+    // 1. to a data sheet, if there are no graphs left
+    // 2. to the next graph, if we were on a graph that was deleted
+    // 3. to the last graph
+
+    const previousNumberOfGraphsRef = React.useRef<number>(props.graphDataArray.length);
+    const previouslySelectedGraphIndexRef = React.useRef<number>(props.graphDataArray.findIndex(graphData => graphData.graph_id === props.openGraph.graphID));
+    useEffect(() => {
+        const handleDeletedGraphs = async () => {
```
marthacryan, line 139: Wouldn't we want to just do all of this handling when a graph is deleted? Instead of listening for the change of the length of `graphDataArray`?

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -1,48 +1,66 @@
...
+import { TaskpaneType } from '../taskpanes';
 import GraphExportTab from './GraphExportTab';
-import { useEffectOnResizeElement } from '../../../hooks/useEffectOnElementResize';
+import GraphSetupTab, { GraphType } from './GraphSetupTab';
+import GraphSidebarTabs from './GraphSidebarTabs';
+import GraphStyleTab from './GraphStyleTab';
+import LoadingSpinner from './LoadingSpinner';
+import { convertBackendtoFrontendGraphParams, convertFrontendtoBackendGraphParams, getDefaultGraphParams, getGraphRenderingParams, getParamsForExistingGraph } from './graphUtils';
 
+export type OpenGraphType = {
```
marthacryan, line 20: Maybe we should move this to `types.tsx`

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -1,48 +1,66 @@
...
+    existingParams: GraphParamsFrontend
+} | {
+    type: 'new_graph'
+    graphID: GraphID,
+    graphType: GraphType
+    selectedColumnIds?: ColumnID[]
+} | {
+    type: 'new_graph_duplicated_from_existing',
+    graphID: GraphID,
+    graphIDOfDuplicated: GraphID,
```
marthacryan, line 32: newGraphId

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -1,48 +1,66 @@
...
+    type: 'existing_graph'
+    graphID: GraphID,
+    existingParams: GraphParamsFrontend
+} | {
+    type: 'new_graph'
+    graphID: GraphID,
+    graphType: GraphType
+    selectedColumnIds?: ColumnID[]
+} | {
+    type: 'new_graph_duplicated_from_existing',
```
marthacryan, line 30: maybe just `duplicate_graph`?

marthacryan: or `new_duplicate_graph`

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -56,38 +74,22 @@ const GraphSidebar = (props: {
...
-
-    /* 
-        When graphUpdatedNumber is updated, we send a new getGraphMessage with the current graphParams
-        in order to update the graphDataDict. We only increment graphUpdatedNumber when the user updates the params.
-
-        We use this method instead of using a useEffect on the graphParams because the graphParams update when we don't want
-        to sendGraphMessage, ie: during an Undo. 
-    */
-    const [graphUpdatedNumber, setGraphUpdatedNumber] = useState(0)
+        console.log("STARTING NEW STEP", props.openGraph);
```
marthacryan, line 88: console log

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -143,148 +125,179 @@ const GraphSidebar = (props: {
...
-        size.
-    */
-    const getGraphAsync = async () => {
-        // The reason that we use the mito-center-content-container instead of the graph-div is because the size of the graph div
-        // changes depending on the size of the graph. Specifically, when exiting fullscreen mode, the graph-div is wider
-        // than we actually have space for. 
-        const boundingRect: DOMRect | undefined = props.mitoContainerRef.current
-            ?.querySelector('#graph-div')
-            ?.getBoundingClientRect();
+    console.log("NEW SELECTED GRAPH", props.openGraph.graphID)
```
marthacryan, line 128: console.log

mitosheet/src/mito/components/taskpanes/Graph/graphUtils.tsx
```diff
@@ -56,18 +58,73 @@ const getAxisColumnIDs = (sheetData: SheetData, graphType?: GraphType, selectedC
...
+    if (openGraph.type === 'new_graph_duplicated_from_existing') {
+        const newValidParams = getValidParamsFromExistingParams(openGraph.existingParamsOfDuplicated, sheetDataArray);
+        return {
+            ...newValidParams,
+            graphID: openGraph.graphID,
+            graphRendering: graphRenderingParams
+        }
+    }
+
+    const newGraphType = openGraph.graphType;
```
marthacryan, line 115: Maybe just add a comment to explain this is the "new graph" case

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -604,8 +603,9 @@ export const getActions = (
             titleToolbar: 'Delete Graph',
             longTitle: 'Delete graph',
             actionFunction: async () => {
-                if (uiState.selectedGraphID) {
-                    await mitoAPI.editGraphDelete(uiState.selectedGraphID);
+                const selectedGraphID = uiState.currOpenTaskpane.type === TaskpaneType.GRAPH ? uiState.currOpenTaskpane.openGraph.graphID : undefined;
+                if (selectedGraphID) {
+                    await mitoAPI.editGraphDelete(selectedGraphID);
```
marthacryan, line 608: This is where I'd expect us to handle changing tabs and stuff

mitosheet/src/jupyter/jupyterUtils.tsx
```diff
@@ -116,20 +116,5 @@ export const getUserProfileFromString = (user_profile_json: string): UserProfile
...
-        const graphParamsFrontend = convertBackendtoFrontendGraphParams(graphParamsBackend)
-        graphDataDict[graphID] = {
-            ...graphDataBackendTyped,
-            graphParams: graphParamsFrontend
-        }
-    })
-
-    parsed['graphDataDict'] = graphDataDict;
-    return parsed;
+    return JSON.parse(analysis_data_json)
```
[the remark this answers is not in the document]
naterush: I removed the difference between graph data frontend and graph data backend -- see the types file.

mitosheet/src/mito/components/endo/utils.tsx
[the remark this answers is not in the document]
naterush: We are -- search it to see! It's badlt named, can you rename to `isAnyGraphSelected` or something

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
naterush: Sure, feel free to change these

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
naterush: graphID is good because then we can use graphID on the openGraph no matter what the type is -- making a lot of code simpler

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
naterush: Per my comment below, there's just a few cases we have to handle. 

If you want to try and make that change, I'm down for it. Maybe:
1. create a `async deleteGraph` function that deletes the graph using the API, then switches to the right new graph tab (or no graph tab, if there is none)
2. Call it in the graph context menu, in the action, and in the modal that deletes graphs that rely on a dataset that get deleted

mitosheet/src/mito/components/footer/SheetTabContextMenu.tsx
naterush: This is meant to be used to find 
```
        const dependantGraphTabNamesAndIDs = getGraphTabNamesAndIDsFromSheetIndex(props.sheetIndex, props.graphDataArray)
```

We def still need this. This is how graphs based on sheets get deleted when those sheets get deleted.

mitosheet/src/mito/components/footer/Footer.tsx
```diff
@@ -100,19 +99,19 @@ function Footer(props: FooterProps): JSX.Element {
...
                 })}
-                {Object.entries(props.graphDataDict || {}).map(([graphID, graphData], index) => {
+                {props.graphDataArray.map((graphData, index) => {
                     return (
                         <SheetTab
-                            key={graphID}
-                            tabName={graphData.graphTabName}
-                            tabIDObj={{tabType: 'graph', graphID: graphID}}
-                            isSelectedTab={selectedTabType === 'graph' && graphID === selectedGraphID}
+                            key={graphData.graph_id}
```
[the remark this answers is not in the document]
naterush: I think the `graph_id` is fine -- it's unoffensive to me at least

mitosheet/mitosheet/api/get_all_params_for_type.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    step_type = params['step_type']
+
+    # Loop over the steps backwards, so that we get the most recent one
+    found_params: List[Any] = []
+    for step in steps_manager.steps_including_skipped[:steps_manager.curr_step_idx + 1][::-1]:
+        if step.step_type == step_type:
+            found_params.append(step.params)
+
+    # Return nothing, if there is no step that meets this criteria
```
naterush, line 21: Comment is wrong

mitosheet/mitosheet/api/get_all_params_for_type.py
```diff
@@ -0,0 +1,22 @@
+from typing import Any, Dict, List
+from mitosheet.types import StepsManagerType
+
+
+def get_all_params_for_type(params: Dict[str, Any], steps_manager: StepsManagerType) -> List[Any]:
```
naterush, line 5: can we change this to `get_all_params_for_step_type`?

mitosheet/src/mito/api/api.tsx
```diff
@@ -358,7 +357,19 @@ export class MitoAPI {
                 return {result: result};
             }
         }
+    }
 
+    /*
+        Gets an array of all params matching a step type.
+    */
+    async getAllParamsForType<ParamType>(stepType: string): Promise<MitoAPIResult<ParamType[]>> {
```
naterush, line 365: I think there is a StepType type defined in `types.tsx`?

mitosheet/src/mito/components/footer/SheetTabContextMenu.tsx
```diff
@@ -1,27 +1,33 @@
...
-    }))
-
+    const response = await mitoAPI.getAllParamsForType<GraphParamsBackend>('graph')
+    const allGraphParams: GraphParamsBackend[] | undefined = 'error' in response ? undefined : response.result;
+    if (allGraphParams === undefined) {
+        return []
+    }
+    const filteredGraphDataJSON: GraphDataArray = graphDataArray.filter((graphData: GraphData, index: number) => {
+        return allGraphParams[index].graph_creation.sheet_index === sheetIndex
+    })
```
naterush, line 27: I think we only want the _most recent_ params on this graph, no? Otherwise, if you:
1. Make a graph off df1
2. Change it to df2
3. Delete df1

It will incorrectly identify this graph as worthy of deleting.

mitosheet/src/mito/components/taskpanes/Graph/graphUtils.tsx
```diff
@@ -56,18 +58,111 @@ const getAxisColumnIDs = (sheetData: SheetData, graphType?: GraphType, selectedC
     }
 }
 
+export const deleteGraphs = async (graphIDs: GraphID[], mitoAPI: MitoAPI, setUIState: React.Dispatch<React.SetStateAction<UIState>>, graphDataArray: GraphData[]) => {
```
naterush, line 61: This is much better than the hook. Great refactor here.

naterush (commented):
The `deleteGraphs` function is a huge, awesome, and excellent refactor. Great work on that. Took me from feeling mid about this refactor to feeling great about it!

Testing functionality now -- there's just a few minor things to fix up, then LGTM.

naterush:
Ah -- there's one bug with an non-existent graph being selected:
1. Create a graph
2. Duplicate it (selecting the new graph)
3. Click undo

This doesn't select the new correct graph. 

I think we can just handle this in the undo action -- post undo, just checking if a non-existent graph is selected, and selecting a good one. Perhaps there's a `ensureValidGraphSelected` util you can factor out of `deleteGraphs` and use here?

PR #1149, opened 2024-01-18
Add chart design tab

marthacryan:
Adds a tab that opens in the toolbar when a graph tab is open


mitosheet/mitosheet/step_performers/graph_steps/graph.py
```diff
@@ -97,8 +97,8 @@ def execute(cls, prev_state: State, params: Dict[str, Any]) -> Tuple[State, Opti
         graph_type = graph_creation["graph_type"]
         sheet_index = graph_creation["sheet_index"]
         safety_filter_turned_on_by_user = graph_preprocessing["safety_filter_turned_on_by_user"]
-        height = graph_rendering["height"] if graph_rendering is not None else "100%"
-        width = graph_rendering["width"] if graph_rendering is not None else "100%"
+        height = graph_rendering.get("height") if graph_rendering is not None else "100%"
+        width = graph_rendering.get("width") if graph_rendering is not None else "100%"
```
naterush: Oof -- lotsa conditions here. Can we just make these always defined?

mitosheet/mitosheet/steps_manager.py
```diff
@@ -503,6 +503,9 @@ def handle_edit_event(self, edit_event: Dict[str, Any]) -> None:
         if self.curr_step_idx != len(self.steps_including_skipped) - 1:
             return
 
+        if edit_event.get('refresh_use_live_updating_hooks'):
+            self.update_event_count += 1
```
naterush, line 507: Comments for sure!

naterush: 1. Why this exists (refreshData in useLiveUpdating hook for the toolbar etc)
2. Who it's used by
3. Other relevant potential solutions we considered, why they don't work -- why this is the easiest

mitosheet/src/mito/components/toolbar/GraphTabs/AddChartElementButton.tsx
```diff
@@ -0,0 +1,210 @@
...
+import AxisTitlesIcon from "../../icons/GraphToolbar/AxisTitlesIcon";
+import ChartTitleIcon from "../../icons/GraphToolbar/ChartTitleIcon";
+import EditIcon from "../../icons/EditIcon";
+import LegendIcon from "../../icons/GraphToolbar/LegendIcon";
+import GridlinesIcon from "../../icons/GraphToolbar/GridlinesIcon";
+import RangeSliderIcon from "../../icons/GraphToolbar/RangeSliderIcon";
+
+type AddChartElementSubMenus = 'axis-titles' | 'chart-title' | 'legend' | 'grid-lines';
+
+export const AddChartElementButton = (
```
naterush, line 17: I like that these buttons are broken into separate files.

mitosheet/src/mito/components/toolbar/GraphTabs/AddChartElementButton.tsx
```diff
@@ -0,0 +1,210 @@
...
+                                return props.updateGraphParam({graph_styling: { yaxis: { showgrid: !graphStylingParams?.yaxis.showgrid }}});
+                            }}
+                        />
+                        <DropdownItem
+                            title='Horizontal'
+                            icon={<GridlinesIcon orientation="horizontal"/>}
+                            canHaveCheckMark
+                            hasCheckMark={graphStylingParams?.xaxis.showgrid}
+                            onClick={() => {
+                                return props.updateGraphParam({graph_styling: { xaxis: { showgrid: !graphStylingParams?.xaxis.showgrid }}});
```
naterush: Overall, the usage of the submenus is not too bad!

mitosheet/src/mito/components/toolbar/GraphTabs/ChartDesignTabContents.tsx
```diff
@@ -0,0 +1,143 @@
...
+        selectedColumnsIds?: ColumnID[];
+        mitoContainerRef: React.RefObject<HTMLDivElement>;
+    }): JSX.Element => {
+    const currOpenTaskpane = props.uiState.currOpenTaskpane;
+    if (currOpenTaskpane.type !== TaskpaneType.GRAPH) {
+        throw new Error('ChartDesignTabContents should only be used when the graph taskpane is open.');
+    }
+    const openGraph = currOpenTaskpane.openGraph;
+    
+    // Get the step params and the step id of the currently open graph by getting the most recent step from this array
```
naterush, line 37: Some documentation to leave:
1. a note how this needs about the live updating hook updating when we make an edit here -- and the hack we use to make this happen
2. This requires the assumption that the open graph has send a message (e.g. this only works because graph automatically refresh)

mitosheet/src/mito/components/toolbar/GraphTabs/ChartDesignTabContents.tsx
```diff
@@ -0,0 +1,143 @@
...
+        />
+        <ChangeChartTypeButton
+            {...props}
+            params={params}
+            updateGraphParam={updateGraphParam}
+        />
+        
+        <div className='toolbar-vertical-line' />
+
+        {params === undefined ? <p> Loading... </p> : <GraphTypeConfigurations
```
naterush, line 99: Are there other cases where params could be undefined?

mitosheet/src/mito/components/toolbar/HomeTabContents.tsx
```diff
@@ -147,7 +147,6 @@ export const HomeTabContents = (
 
         <div className='mito-toolbar-number-format'>
             <Select
-                style={{ height: '22px', fontSize: '11px', width: '150px' }}
```
naterush, line 150: Why does this go away?

mitosheet/src/mito/components/toolbar/Toolbar.tsx
```diff
@@ -79,6 +82,9 @@ export const Toolbar = (
             })}
         </div>
     }
+    if (props.uiState.currOpenTaskpane.type === TaskpaneType.GRAPH) {
```
naterush, line 85: Another thing you could here is check that the open graph has the same graph id as the params from the steps list in the analysis data. So that we make sure we're accessing the right params.

mitosheet/src/mito/components/toolbar/Toolbar.tsx
```diff
@@ -168,7 +174,7 @@ export const Toolbar = (
                     />
                 </div>
             </div>
-            {currentTab !== undefined ? tabs[currentTab] ?? <div> No tab found </div> : undefined}
+            {currentTab !== undefined ? tabs[currentTab] ?? tabs['Home'] : undefined}
```
naterush, line 177: Ah -- so if you close a graph, this ends up being undefined. So we just default to the Home. 

This is reasonable -- but wanna leave a comment about why this happens (so that we can just not display a tab like something for charting and it will not break everything).

mitosheet/src/mito/utils/actions.tsx
```diff
@@ -728,6 +774,61 @@ export const getActions = (
...
+            iconToolbar: ExportIcon,
+            titleToolbar: 'Export',
+            longTitle: 'Open Export Dropdown',
+            actionFunction: () => {
+                setEditorState(undefined);
+                closeOpenEditingPopups();
+
+                setUIState(prevUIState => {
+                    return {
+                        ...prevUIState,
```
naterush, line 821: Do we need to set the currently selected toolbar tab in all of these? I think we do

mitosheet/css/toolbar.css
```diff
@@ -50,6 +50,49 @@
   cursor: pointer;
 }
 
+.mito-toolbar-bottom .select-container,
+.mito-toolbar-bottom .mito-input {
+  height: 22px;
+  font-size: 11px;
+  width: -webkit-fill-available;
+  margin-left: 4px;
+}
```
naterush, line 59: Hmm. We talked about this a bit during the toolbar refactor, but I don't love the lack of encapsulation that this imposes on the components. With this, you can't go to the .tsx and .css file for a component and see the CSS that drives it -- this is bad I think.

At the very least, I'd much rather this class be in `Input.css` so you can really understand what drives it. 

Also, does `-webkit-fill-available` work everywhere?

mitosheet/css/toolbar.css
```diff
@@ -50,6 +50,49 @@
...
+
+.mito-toolbar-bottom .mito-graph-configuration-container p {
+  text-wrap: nowrap;
+  padding-right: 4px;
+}
+
+.mito-toolbar-bottom .mito-graph-configuration-container .mito-input {
```
naterush, line 66: Move this too

mitosheet/css/toolbar.css
```diff
@@ -50,6 +50,49 @@
...
+  width: 50px;
+}
+
+.mito-graph-configuration-container .spacing-row {
+  width: auto;
+}
```
naterush, line 72: Some docs would be helpful here -- and move it

naterush (commented):
Some code cleanup. Overall, this is all very sensibly implemented and organized -- great work!

aarondr77:
Graph Configuration Options loading indicator persists:

1. Create a bar chart
2. Switch to a scatter plot or density heat map and see that it always says loading configuration options. 

If there are no configuration options, we can just not show anything 

https://github.com/mito-ds/mito/assets/18709905/b2b7a126-71ea-4f8c-babe-dc1ed09470c9

aarondr77:
When you hover over the `range slider` buttons, it still displays the grid lines secondary dropdown. 

Let's fix this by making the range slider have a secondary dropdown that has the `None` and `Horizontal` options for consistency with the other dropdown items. 


https://github.com/mito-ds/mito/assets/18709905/d83d3a76-8ae5-4685-b9a8-fe05203d6924

aarondr77:
1. In all of the icons besides the range slider, the element we are changing is in purple, can we update the range slider icon to the same. 
2. It looks like the range slider in the icon is cut off by the graph, can we fix that?

Screenshot 2024-01-18 at 4 17 43 PM

aarondr77:
I noticed this bug that is probably part of the graph refactor that we should fix. https://github.com/mito-ds/mito/issues/1151

aarondr77:
This should be an impossible state to get. Can't have `None` and grid lines

Screenshot 2024-01-18 at 4 43 45 PM

mitosheet/src/mito/components/toolbar/GraphTabs/ChartDesignTabContents.tsx
marthacryan: Not to my knowledge

mitosheet/src/mito/components/toolbar/HomeTabContents.tsx
marthacryan: Oh I added css into `toolbar.css` to handle this. Wanted it to be used for all toolbar selects.

mitosheet/src/mito/utils/actions.tsx
marthacryan: I think the assumption is that if you clicked on the button that triggers this, you're necessarily already in that tab.

marthacryan:
> 2. It looks like the range slider in the icon is cut off by the graph, can we fix that?

Oh that was actually my idea but I can move it lower! Thought that was maybe more accurate to what the range slider looked like

mitosheet/css/toolbar.css
marthacryan: Ok I'll move it in there! Also yeah good point I think I need to add a few different lines to cover the different browsers

mitosheet/src/mito/components/toolbar/Toolbar.tsx
marthacryan: Wouldn't we want to add that in the `ChartDesignTabContents`? As long as we have a graph tab open we'll want this tab to exist.

naterush: I think it's the same, but it's also unnecessary I think.

aarondr77:
A few small nitpicks here:

1. The font color for the number of bins is different than the rest of the config options 
2. The text for the other config options are cut off. 

Screenshot 2024-01-18 at 6 59 29 PM

I'm hoping that before the release we are going to have time to convert these into this interface because its about 10x more usable, but let's just do these small fixes now incase we run out of time. 

Screenshot 2024-01-18 at 7 03 21 PM

mitosheet/src/mito/components/taskpanes/Graph/GraphSetupTab.tsx
```diff
@@ -145,7 +145,22 @@ function GraphSetupTab(
                     sheetIndex={graphSheetIndex}
                     onChange={(newSheetIndex) => {
                         // Reset the graph params for the new sheet, but keep the graph type!
-                        const newSheetGraphParams = getDefaultGraphParams(props.mitoContainerRef, props.sheetDataArray, newSheetIndex, props.openGraph)
+                        const newSheetGraphParams = getDefaultGraphParams(
```
aarondr77, line 148: I'm still getting this bug https://github.com/mito-ds/mito/issues/1151 .But it looks like you caught a similar bug here, so very nice! 

The bug I saw is this:

1. Import a dataframe with columns A and B
2. Create a graph with A on the x axis and B on the Y axis
3. Switch back to the data tab
4. Delete column A
5. Switch back to the graph tab
6. Get this error

```
  File "column_headers.py", line 233, in get_column_headers_by_ids
    return [
  File "column_headers.py", line 234, in <listcomp>
    self.column_id_to_column_header[sheet_index][column_id]
KeyError: 'annual_income'
```


Instead of erroring, we want to just return the same graph, just without column A.

aarondr77:
1. Import a dataset
2. Create a graph with columns A and B
3. Duplicate the source dataframe
4. Delete column A from the duplicated dataframe
5. Reopen the graph and get an error because A does not exist. This shouldn't happen because A still exists in the source dataframe!


https://github.com/mito-ds/mito/assets/18709905/e87d4b32-1c85-4a9f-9d59-41a81aaef447

aarondr77:
[Probably a related bug to above]

1. Import a dataset with columns A, B, C, D, E
2. Create a graph with columns A and B
3. Delete columns C, D, E from the source dataframe
4. Reopen the graph and get an error because one of the deleted columns does not exist. This is strange because they aren't used in the graph. 

```
  File "column_headers.py", line 233, in get_column_headers_by_ids
    return [
  File "column_headers.py", line 234, in <listcomp>
    self.column_id_to_column_header[sheet_index][column_id]
KeyError: 'loan_amount'
```

https://github.com/mito-ds/mito/assets/18709905/1621a277-d1a3-4c36-baa1-c971e5e1505c

aarondr77:
On Mac/Chrome
1. Single, left click on the graph tab dropdown icon and see that the dropdown flashes. 
2. Do the same on a data tab and see that the dropdown opens. 

This is a new issue since the last release

https://github.com/mito-ds/mito/assets/18709905/a398e662-5ccc-436a-a845-061731633c92

naterush:
> Let's fix and add a regression frontend test for it! woohooo more robust software

@aarondr77 let's wait on tests for now. I think doing them in a comprehensive and single, planned move towards the end of the product cycle will lead to same quality tests for less total effort.

aarondr77:
Sounds good. I'll open a issue that keeps track of some of these edge cases we want to test so we don't foget

marthacryan:
> 1. The font color for the number of bins is different than the rest of the config options

Oh that's actually placeholder text, so it's supposed to be a different color!

mitosheet/src/mito/components/taskpanes/Graph/GraphSetupTab.tsx
marthacryan: Oh interesting, I'm not getting that error

marthacryan: Are you sure you have the most recent code?

marthacryan: Oh no I am in JupyterLab, interesting!

PR #1150, opened 2024-01-18
Enterprise logging

aarondr77:
# Description

- Creates environment variable for mito enterprise users to configure end point to understand mito usage. 
- Small refactor of the mito_config to make it a little bit more pleasant

# Testing

- see tests
- add a print statement and see the generated logs

# Documentation

We need to create docs!


aarondr77:
TODO: We only want to run do this logging if we are in Jupyter

mitosheet/mitosheet/enterprise/mito_config.py
```diff
@@ -62,26 +94,14 @@ def upgrade_mec_1_to_2(mec: Dict[str, Any]) -> Dict[str, Any]:
...
-        MITO_CONFIG_FEATURE_TELEMETRY: None,
-        MITO_CONFIG_PRO: None,
-        MITO_CONFIG_ENTERPRISE: None,
-        MITO_CONFIG_ENTERPRISE_TEMP_LICENSE: None,
-        MITO_CONFIG_CUSTOM_SHEET_FUNCTIONS_PATH: None,
-        MITO_CONFIG_CUSTOM_IMPORTERS_PATH: None
-    }
+
+    # Find the new keys in version 2 that are not in version 1
+    new_keys = set(MEC_VERSION_KEYS['2']) - set(mec.keys())
```
naterush, line 99: Sweet, improvement here!

mitosheet/mitosheet/enterprise/mito_log_uploader.py
```diff
@@ -0,0 +1,114 @@
+import json
```
naterush, line 1: Can we put this in `enterprise/telemetry/mito_log_uploader.py`

mitosheet/mitosheet/enterprise/mito_log_uploader.py
```diff
@@ -0,0 +1,114 @@
...
+import pprint
+import time
+from typing import Any, Optional, List, Dict
+import requests
+
+def preprocess_log_for_upload(log_event: str, log_params: Dict[str, Any]) -> Optional[Dict[str, Any]]:
```
naterush, line 7: This approach is currently:
1. Take our current logs that we generate
2. Filter them to make them into the format that we want

This means that our enterprise logging approach is like sorta dependent on our current logging approach (which we have iterated in the past). 

Another way to do this would be to go farther up-the-chain to the parameters to the `log` call.

No need to change anything, just calling this out for info/thought sharing reasons.

mitosheet/mitosheet/enterprise/mito_log_uploader.py
```diff
@@ -0,0 +1,114 @@
...
+    enterprises collecting logs from thousands of users do not 
+    break their logging server. 
+    """
+    def __init__(
+        self, 
+        log_url: str,
+        log_interval: Optional[int]
+    ):
+        self.log_url = log_url
+        self.log_interval = log_interval if log_interval is not None else 0
```
naterush, line 78: Let's pick a more reasonable default here -- maybe 10 seconds.

mitosheet/mitosheet/mito_backend.py
```diff
@@ -68,12 +69,18 @@ def __init__(
         # Setup the MitoConfig class
         self.mito_config = MitoConfig() # type: ignore
 
-        custom_sheet_functions_path = self.mito_config.get_mito_config()[MITO_CONFIG_CUSTOM_SHEET_FUNCTIONS_PATH]
+        # Set up the Mito Logger class
+        jupyter_log_server_url = self.mito_config.jupyter_log_server_url
+        if jupyter_log_server_url:
+            jupyter_log_server_batch_interval = self.mito_config.jupyter_log_server_batch_interval
+            self.mito_log_uploader = MitoLogUploader(jupyter_log_server_url, jupyter_log_server_batch_interval)
```
naterush, line 76: This conditional setting up the uploader isn't great practice -- we should have an else case (setting it to None) -- so you can always access it

mitosheet/mitosheet/tests/test_mito_analytics.py
```diff
@@ -8,8 +8,64 @@
...
+except:
+    version_python = 'no python'
 
 
 def test_not_printing_logs():
-    assert PRINT_LOGS is False
\ No newline at end of file
+    assert PRINT_LOGS is False
+
+def test_log_uploader():
```
naterush, line 35: Can we test:
1. A long upload interval does not call .post at all
2. Multiple edit events all get posted
3. Errors
4. Mitosheet rendered (you can mock this one somehow)

naterush (commented):
Some thoughts and ideas

mitosheet/src/mito/types.tsx
```diff
@@ -805,6 +808,9 @@ export interface MitoConfig {
     [MitoEnterpriseConfigKey.DISPLAY_SCHEDULING]: boolean,
     [MitoEnterpriseConfigKey.CUSTOM_SHEET_FUNCTIONS_PATH]: string,
     [MitoEnterpriseConfigKey.CUSTOM_IMPORTERS_PATH]: string,
+    [MitoEnterpriseConfigKey.JUPYTER_LOG_SERVER_URL]: string,
+    [MitoEnterpriseConfigKey.JUPYTER_LOG_SERVER_BATCH_INTERVAL]: string,
```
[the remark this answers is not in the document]
aarondr77: We decided that we were going to expose this so that we didn't need to configure for them. I'd rather expose this to remove back and forths with enterprises trying to set this. They could iterate themselves in 5 minutes to find the right one or we could go back and forth for weeks.

mitosheet/src/mito/types.tsx
```diff
@@ -717,8 +717,11 @@ export enum MitoEnterpriseConfigKey {
     DISPLAY_SCHEDULING = 'MITO_CONFIG_FEATURE_DISPLAY_SCHEDULING',
     LLM_URL = 'MITO_CONFIG_LLM_URL',
     ANALYTICS_URL = 'MITO_CONFIG_ANALYTICS_URL',
+    JUPYTER_LOG_SERVER_URL = 'MITO_CONFIG_JUPYTER_LOG_SERVER_URL',
+    JUPYTER_LOG_SERVER_BATCH_INTERVAL = 'MITO_CONFIG_JUPYTER_LOG_SERVER_BATCH_INTERVAL',
```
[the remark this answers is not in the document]
aarondr77: As per the spec, we wanted to provide different logging destinations depending on if they are in Streamlit or Jupyter, so that's why I settled on these names.

aarondr77:
TODO: Mitosheet rendered (you can mock this one somehow)

naterush (commented):
Sweet, LGTM.

PR #1157, opened 2024-01-22
Add infrastructure for vertical dropdown items

marthacryan:
Adds a dummy example of usage for "vertical dropdown items", which we'll be using to implement the submenus for the different chart types. 
vertical-submenus


mitosheet/css/elements/DropdownItem.css
```diff
@@ -84,5 +84,38 @@
     color: var(--mito-text-light) !important;
 }
 
+/* Special CSS for vertical dropdown items */
 
+.mito-dropdown-item-vertical .mito-dropdown-item-icon-and-title-container {
```
aarondr77, line 93: Since we are just adding the .mito-dropdown-item-vertical modifier infront of these other classes that already exist to overwrite classes styles that are set above, the order of this css file becomes important, which isn't the case anywhere else in the codebase. 

So I think it might be better long-term to add a `.mito-dropdown-item-horizontal` in for horizontal dropdowns by default. 

Then break this file into 3 sections: 

1. classes that get applied to all dropdown items
2. classes that only get applied to vertical dropdown items
3. classes that only get applied to horizontal dropdown items 

That way the order of the classes in this file don't matter going forward.

aarondr77 (commented):
This is awesome! Just one refactor note and then LGTM

mitosheet/css/elements/DropdownItem.css
marthacryan: Done!

PR #1158, opened 2024-01-23
Add chart type submenus

marthacryan:
Adds the actual submenus to the "Change Chart Type" dropdown. 
chart type submenus


aarondr77:
Bug: If you click the same button twice, it shouldn't change the graph. 

https://github.com/mito-ds/mito/assets/18709905/ed3bf166-8c1e-4ed9-8133-39e0f2337931

aarondr77:
Can we horizontally center the image here so its in the middle of the text?

Screenshot 2024-01-23 at 6 25 51 PM

aarondr77:
I think clicking on the top level dropdown item shouldn't trigger a change in the graph anymore. I think clicking on the top level dropdown items shouldn't do anything if it has a submenu. 

https://github.com/mito-ds/mito/assets/18709905/52dfadad-7a80-4387-8dae-5b4c8cea2a29

aarondr77:
It looks like the width of the dropdown item is set dynamically based on the width of the contents. Instead, can we ensure that all of them are the same width? 

Screenshot 2024-01-23 at 6 30 29 PM

aarondr77:
It looks like the spacing of whichever item has the check gets thrown off. Notice that the scatter icon + text is pulled to the left from the rest of the items. 

I think another good example of where we need to explicitly set the width of items to ensure consistency. 

Screenshot 2024-01-23 at 6 32 27 PM

aarondr77:
When you switch to a box plot, the `points` setting defaults to empty. Instead, it should default to `outliers`. 

https://github.com/mito-ds/mito/assets/18709905/3d8cc072-1861-426f-9e4c-f931fcc79b2b

aarondr77:
1. Make a line chart with a date on the x axis and a number on the y axis
2. Update the line chart to interpolated
3. Add another column to the Y axis 
4. Get this error

```
    raise ValueError(
ValueError: 
    Invalid value of type 'builtins.str' received for the 'shape' property of scattergl.line
        Received value: 'spline'

    The 'shape' property is an enumeration that may be specified as:
      - One of the following enumeration values:
            ['linear', 'hv', 'vh', 'hvh', 'vhv']
```


https://github.com/mito-ds/mito/assets/18709905/230c7134-915d-4a0e-aa8c-822927443fcd

marthacryan:
> This might just be a plotly limitation that you can only use `spline` if there is one value on the y axis? I'm not sure. Maybe check if the deployed version of graphing has the same issue?

I've been seeing that error but I'm having trouble reproducing it. Not sure what the issue is exactly

aarondr77:
Can we keep this a two rows of 4? I think seeing the pairs together vertically makes it so much more clear what the different options are. 

Ie: Both `Grouped` options are on top of each other, etc.

Screenshot 2024-01-25 at 2 46 46 PM

aarondr77:
For spacing let's just call this one `contour` since Density is the top level dropdown label

Screenshot 2024-01-25 at 2 48 26 PM

aarondr77:
### Bug 

It looks like we might have ran into some unexpected complexity with histogram orientations. Notice below the inconsistencies between clicking in the `vertical` histogram button and ending up with a vertical histogram sometimes and a horizontal histogram other times. 

In this graph, I have a one string column and two number columns and get a horizontal histogram. 

https://github.com/mito-ds/mito/assets/18709905/d4a63074-c5c8-41ad-b631-8598cd1fb00a

In this graph, I have one string column and one number column and get a vertical histogram. 

https://github.com/mito-ds/mito/assets/18709905/bdaf95ce-e277-4ca5-85a0-e9ff463ba22d

### Solution Approach

This definitely sets off the `there is some complexity here` alarm. So before we go and implement a solution, let's make a mini spec that: 1) shows our intended behavior, figures out the problem, and proposes a solution. I'll review that and then we can implement it.

mitosheet/src/mito/components/toolbar/GraphTabs/ChangeChartTypeButton.tsx
```diff
@@ -41,6 +45,30 @@ export const GRAPHS_WITH_UNIQUE_CONFIG_OPTIONS = [...new Set([
     ...GRAPHS_THAT_HAVE_LINE_SHAPE
 ])];
 
+/**
+ * This function is used to switch the x and y axes if the graph is being set to favor a certain axis.
+ * So, if there are multiple columns in the X axis and we set primaryAxis = 'y', then the x and y axes
+ * will be switched.
+ */
+const updatePrimaryAxis = (primaryAxis: 'x' | 'y', x_axis_column_ids?: ColumnID[], y_axis_column_ids?: ColumnID[]): [ColumnID[] | undefined, ColumnID[] | undefined] => {
```
aarondr77, line 53: Can we call this function `getXAndYColumnIDs()` or something like that just so it's clear that this function is not actually doing the updating of the params but instead just returning the column ids

aarondr77:
> Re: Spline bug above. 

Here is a reproducible error. Notably, if you change the .head(1000) to .head(5) then the bug goes away. 

We've massively struggled in the past to figure out good data size cutoffs for when to support plotly features and when not to because it is dependent on number of traces, data types, etc. 

So let's just get rid of `spline` support all together for now to keep our focus on the most important features. We can add it back if a user ever asks for it. 

Here is the file for the below code:

[new_york_listings_2024.csv](https://github.com/mito-ds/mito/files/14057414/new_york_listings_2024.csv)


```
import plotly.express as px

new_york_listings_2024 = pd.read_csv(r'/Users/aarondiamond-reivich/Downloads/new_york_listings_2024.csv')

# Filter the dataframe so that it does not crash the browser
new_york_listings_2024_filtered = new_york_listings_2024.head(1000)

# Construct the graph and style it. Further customize your graph by editing this code.
# See Plotly Documentation for help: https://plotly.com/python/plotly-express/
fig = px.line(new_york_listings_2024_filtered, x='neighbourhood_group', y=['latitude', 'longitude'], line_shape='spline')
fig.update_layout(
        title='neighbourhood_group, latitude (first 1000 rows) line', 
        xaxis={
            "showgrid": True, 
            "rangeslider": {
                "visible": True, 
                "thickness": 0.05
            }
        }, 
        yaxis={
            "showgrid": True
        }, 
        legend={
            "orientation": 'v'
        }, 
        paper_bgcolor='#FFFFFF'
    )
fig.show(renderer="iframe")
```

marthacryan:
> Can we keep this a two rows of 4? I think seeing the pairs together vertically makes it so much more clear what the different options are.
> 
> Ie: Both `Grouped` options are on top of each other, etc.
> 
> Screenshot 2024-01-25 at 2 46 46 PM

Oh interesting, what browser are you in? I set the width for that specific submenu to make it have 4 entries per row. Maybe there's some different handling of css in your browser?

aarondr77:
> Oh interesting, what browser are you in? I set the width for that specific submenu to make it have 4 entries per row. Maybe there's some different handling of css in your browser?

I'm on Mac,  Chrome

marthacryan:
Ohh wait I see - I adjusted the with of the submenu items so I just needed to fix the custom width.

PR #1159, opened 2024-01-23
Add interactive graph titles

marthacryan:
Allows the user to click on the chart title and the x/ y axis titles to
1. select the title and delete
2. edit the title


mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -99,6 +160,76 @@ const GraphSidebar = (props: {
...
+
+            /**
+             * Open popup when double clicked
+             */
+            gtitle.addEventListener('dblclick', () => {
+                setSelectedGraphElement({
+                    element: 'gtitle',
+                    xPosition: gtitle.getBoundingClientRect().left,
+                    yPosition: gtitle.getBoundingClientRect().top + 30,
+                    displayPopup: true
```
naterush, line 211: Feels like there's duplicated state here. Perhaps the better state is `editingPopupPosition : {xPosition, yPosition} | undefined` -- above, you set it to undefined, and below you have x and y.

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -99,6 +160,76 @@ const GraphSidebar = (props: {
             }
             const executeScript = new Function(graphOutput.graphScript);
             executeScript()
+            const div: any = document.getElementById(graphOutput.graphHTML.split('id="')[1].split('"')[0])
```
naterush, line 163: Factoring this into a helper seems nice.

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -99,6 +160,76 @@ const GraphSidebar = (props: {
...
+            });
+
+            ytitle.addEventListener('dblclick', () => {
+                setSelectedGraphElement({
+                    element: 'ytitle',
+                    xPosition: ytitle.getBoundingClientRect().left - 10,
+                    yPosition: ytitle.getBoundingClientRect().top - 40,
+                    displayPopup: true
+                })
+            });
```
naterush, line 232: Do we need to do any cleanup on these event listeners? Or are they cleaned up when the svg disappears? I have a feeling we need to do some cleanup.

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -4,17 +4,71 @@ import '../../../../../css/taskpanes/Graph/LoadingSpinner.css';
...
+                left: props.xPosition,
+                top: props.yPosition,
+            }}
+        >
+            <Input
+                className='popup-input'
+                value={newValue}
+                onKeyDown={(e) => {
+                    if (e.key === 'Backspace') {
+                        e.stopPropagation();
```
naterush, line 58: Document?

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -4,17 +4,71 @@ import '../../../../../css/taskpanes/Graph/LoadingSpinner.css';
...
+    
+    React.useEffect(() => {
+        setNewValue(props.value);
+    }, [props.value]);
+
+    React.useEffect(() => {
+        // Grab focus when we switch the location of the popup
+        const input = document.getElementsByClassName('popup-input')[0] as HTMLInputElement;
+        input.focus();
+    }, [props.xPosition, props.yPosition])
```
naterush, line 42: Why?

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
```diff
@@ -4,17 +4,71 @@ import '../../../../../css/taskpanes/Graph/LoadingSpinner.css';
...
+            <Input
+                className='popup-input'
+                value={newValue}
+                onKeyDown={(e) => {
+                    if (e.key === 'Backspace') {
+                        e.stopPropagation();
+                    }
+                    if (e.key === 'Enter') {
+                        props.setValue(newValue);
+                    }
```
naterush, line 62: Can we rename `props.setValue` to `props.updateGraphTitle` -- and update `setNewValue` to `setEditedTitle`?

Or something to make it more clear that the state in this component is _temporary and could be ditched_ and hasn't been sent to the backend -- while the `setValue` is actually updating the params.

naterush (changes requested):
Some minor code thoughts, overall this looks excellent!

mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
marthacryan: Ok so I'm not totally sure on the answer to this but it does help that we're using a unique ID for the graph. I did a quick test just now and it doesn't look like it calls the event handlers any extra times after it re-registers a new event handler.

aarondr77:
On JupyterLab, Chrome: there are some placement issues of the input.

### X axis 

Can we update the input so that the arrow is pointing towards the title and the input not covering up part of the title 

Screenshot 2024-01-25 at 1 35 46 PM

aarondr77:
On JupyterLab, Chrome: there are some placement issues of the input.

### Y axis 

Again would be nice if the arrow pointed towards the axis title and was centered on the axis title

Screenshot 2024-01-25 at 1 38 46 PM

aarondr77:
My intuition says that this delay between clicking Enter in the input and the loading indicator appearing is going to be pretty confusing to users. They are going to think that it didn't work and they need to submit the title again. 

We've seen things like this before when editing a column header in your dataframe would leave the old column header there until the new title is returned from the backend. 

This might be a bit challenging to fix though. The solution I have in mind is to not debounce the graph param edit if the user is changing a x axis,  y axis, or chart title. But I'm not sure the best way to conditionally do that. 

**Since this is just polish lets time box fixing it to 30 mins! If we don't get it in that time, its nbd**

https://github.com/mito-ds/mito/assets/18709905/9444689f-75ff-4851-bf81-c24c4d25b39b

aarondr77:
After interacting with the legend, the X and Y axis labels are no longer clickable, but the Y axis is. 

https://github.com/mito-ds/mito/assets/18709905/e19a026b-9412-443f-8644-52b3fa9d74c7

aarondr77:
Clicking on the `Edit Title` button in the Add Chart Element dropdown makes the title disappear instead of opening the text input. 

https://github.com/mito-ds/mito/assets/18709905/15c9642c-e101-4716-91c5-cc983afded04

aarondr77:
Now that there are separate locations for editing the axis labels, maybe we should break this into two different buttons so we can open the correct input field. 

Screenshot 2024-01-25 at 1 57 10 PM

mitosheet/css/taskpanes/Graph/GraphSidebar.css
```diff
@@ -93,6 +110,45 @@
     text-align: center;
 }
 
+.graph-element-popup-div.popup-div:before {
+    content: "";
+    display: block;
+    height: 10px;
+    width: 10px;
```
aarondr77, line 117: Looking at the graph titles of the graphs in the shaping document, they are usually longer than can fit in this input currently. 

I think we should make two updates:

1. When we open the title editor, have it long enough to display the current title (obviously up to certain extent if the column header is super long)
2. If the user adds more text to the label, have the text input expand (we do this for the cell editor in the grid so there is probably a utility you can use)

mitosheet/src/mito/components/taskpanes/Graph/graphUtils.tsx
```diff
@@ -487,4 +495,82 @@ export const openGraphSidebar = async (
             }
         })
     }
+}
+
+export const registerClickEventsForGraphElements = (graphOutput: GraphOutput, setSelectedGraphElement: React.Dispatch<React.SetStateAction<GraphElementType | null>>) => {
+    console.log('here')
```
aarondr77, line 501: remove

mitosheet/src/mito/components/taskpanes/Graph/graphUtils.tsx
```diff
@@ -487,4 +495,82 @@ export const openGraphSidebar = async (
...
+    const div: any = document.getElementById(graphOutput.graphHTML.split('id="')[1].split('"')[0])
+    if (div === null) {
+        return;
+    }
+    div.on('plotly_click', () => {
+        setSelectedGraphElement(null)
+    });
+    
+    // Main Title
+    const gtitle = div.getElementsByClassName('g-gtitle')[0]
```
aarondr77, line 514: Ooo since we are already getting this object, can we use the default title in the x axis, y axis, and chart titles when you open the editor for the first time. 

Currently, if you have not yet manually set the titles, then the input just opens to a blank input.

mitosheet/src/mito/components/toolbar/GraphTabs/GraphTypeConfigurations.tsx
```diff
@@ -184,7 +184,7 @@ export const GraphTypeConfigurations = (
                         subtext='only display sample points outside the whiskers'
                     />
                     <DropdownItem
-                        title={'supsected outliers'}
+                        title={'suspected outliers'}
```
aarondr77, line 187: excellent

mitosheet/css/taskpanes/Graph/GraphSidebar.css
```diff
@@ -52,6 +52,25 @@
     box-sizing: border-box;
 }
 
+.gtitle-highlighted .g-gtitle {
+    outline: 1px solid var(--mito-text-light);
+}
+
+.xtitle-highlighted .g-xtitle {
+    outline: 1px solid var(--mito-text-light);
```
aarondr77, line 60: Can we try to make this look a bit nicer? Probably:

1. Some padding around the text?
2. Darker lines? 

Maybe make a little mockup for it and then implement, I trust your design eye

Screenshot 2024-01-25 at 2 32 54 PM

mitosheet/css/taskpanes/Graph/GraphSidebar.css
```diff
@@ -93,6 +110,45 @@
...
+
+.graph-element-popup-div-caret-above.graph-element-popup-div.popup-div:before {
+    top: 0;
+}
+
+.graph-element-popup-div.popup-div {
+    padding: 3px;
+    background: var(--mito-highlight-medium);
+    border-radius: 3px;
+}
```
aarondr77, line 150: I think we can make this look a little bit more like the mockups if we change to the darker purple `var(--mito-highlight)`

Screenshot 2024-01-25 at 2 35 22 PM

aarondr77 (commented):
Nice work. The functionality is working well! I think there is an hour or so of polish that would also really make this feel great!

mitosheet/css/taskpanes/Graph/GraphSidebar.css
marthacryan: Agh yeah this is tricky. For some reason any css that I added to this element was being ignored. We can adjust the color of the line but it might be a bigger thing to add padding / border radius unfortunately

mitosheet/src/mito/components/taskpanes/Graph/graphUtils.tsx
marthacryan: Ah yeah definitely

mitosheet/css/taskpanes/Graph/GraphSidebar.css
marthacryan: Just updated! I refactored / moved the util you were talking about to be more general and it's working well! Lmk if you want me to adjust any of the defaults (max or min length)

mitosheet/css/taskpanes/Graph/GraphSidebar.css
aarondr77: Okay good callout. Let's leave for now then.

mitosheet/css/taskpanes/Graph/GraphSidebar.css
aarondr77: Looks good!

PR #1161, opened 2024-01-24
add code-options env var

aarondr77:
# Description

- Adds environment variable for turning off code options feature
- Fixes remaining broken keyboard shortcuts:
-- The function keys on Mac trigger Mac settings. ie: Control + F11 changes the volume. 
-- Uses control + F11 on windows to create a pivot table because F11 is used on windows Excel
-- Adds `shift` as required key for opening find and replace on windows because on chrome not using Ctrl + H opens chrome history

# Testing

See tests + try it out. 

# Documentation

No.


mitosheet/src/mito/utils/keyboardShortcuts.tsx
```diff
@@ -129,14 +129,15 @@ export const keyboardShortcuts: KeyboardShortcut[] = [
...
             keys: ['h']
         },
         action: ActionEnum.OpenFindAndReplace
     },
     {
         macKeyCombo: {
-            altKey: true,
-            keys: ['F1']
+            ctrlKey: true,
+            keys: ['G']
```
naterush, line 140: I don't see this change documented in the PR description?

naterush (commented):
LGTM when CI passes

PR #1164, opened 2024-01-25
dash: use a threading lock to avoid potential race conditions

naterush:
# Description

Avoids a potential race condition in processing messages from multiple threads. If you have a deployed version of Dash that uses multiple threads:
1. Do a pivot 1, this adds a single message to the queue. This enters it's critical section before taking the lock.
2. Click to make a param edit, starting pivot 2, this adds another message to the queue, which due to race condition happens to start its critical section as well
3. Pivot 1 finishes. But pivot 2 started off the state from before pivot 1. So there is no pivot remaining? 

# Testing

Dash should work as normal locally.

# Documentation

No.


naterush:
Closing this as it's not our bug source.

PR #1178, opened 2024-02-05
Add select data button to open taskpane in graph editor

marthacryan:
Adds a "Select Data" button to allow for opening and closing the graph "Select Data" taskpane. 
select data


mitosheet/src/mito/components/footer/GraphTabContextMenu.tsx
```diff
@@ -79,9 +81,10 @@ export default function GraphTabContextMenu(props: {
         props.setUIState(prevUIState => {
             return {
                 ...prevUIState,
+                selectedTabType: 'graph',
                 currOpenTaskpane: {
                     type: TaskpaneType.GRAPH, 
-                    graphSidebarTab: GraphSidebarTab.Export,
+                    graphSidebarOpen: true,
```
aarondr77, line 87: Nice. This is fine for now. In the future, we should do the refactor of moving the graph to being a type of tab instead a type of taskpane, and then we could use our taskpane infrastructure to handle this. But we did a lot of graph code cleanup already so this is fine.

aarondr77: Defaulting this to `true` when you are creating the graph for the first time is perfect. But if you are reopening a graph that already exists, let's default this to false.

mitosheet/src/mito/components/toolbar/GraphTabs/ChartDesignTabContents.tsx
```diff
@@ -138,6 +149,7 @@ export const ChartDesignTabContents = (
                 />
             </Dropdown>
         </ToolbarButton>
+        <ToolbarButton action={props.actions.buildTimeActions[ActionEnum.Open_Graph_Select_Data_Taskpane]}/>
```
aarondr77, line 152: It seems like the most sensible order would be: 

1. Select Data
2. Change Chart Type
3. Add Chart Element
4. Export 

What do you think?

Screenshot 2024-02-05 at 4 21 35 PM

aarondr77 (commented):
Small cleanup, then LGTM.

mitosheet/src/mito/components/toolbar/GraphTabs/ChartDesignTabContents.tsx
marthacryan: Good point! Done

PR #1179, opened 2024-02-05
Update style for chart type configs

marthacryan:
Updates the chart type configs style to be more aligned with the mockups. Also removes some code that was included to accommodate having more than 2 config options for a chart type. 
image


aarondr77:
Similar with the other PR, the height of the toolbar changes if you switch from a graph type that has additional config options to one without

aarondr77:
It looks like these are different fonts. One is gray and one is black. Let use the `Number of Bins` font for both

Screenshot 2024-02-06 at 10 11 51 AM

aarondr77:
Let's increase the width of the dropdown items 

Screenshot 2024-02-06 at 10 14 06 AM

mitosheet/src/mito/components/toolbar/GraphTabs/GraphTypeConfigurations.tsx
```diff
@@ -27,22 +27,22 @@ export const GraphTypeConfigurations = (
...
                         Aggregation Type
                 </LabelAndTooltip>
                 <Select
                     value={props.graphParams.graph_creation?.histfunc || 'count'}
                     onChange={(newHistfunc: string) => {
                         props.updateGraphParam({graph_creation: {histfunc: newHistfunc}})
                     }}
-                    width='medium'
-                    dropdownWidth='medium'
+                    style={{ width: '70px' }}
```
aarondr77, line 44: Can we make this a class, just so we can easily keep the width consistent across the different graph configs

aarondr77 (commented):
A few minor cleanups then LGTM

mitosheet/src/mito/components/toolbar/GraphTabs/GraphTypeConfigurations.tsx
marthacryan: Ok so I tried this out and the width of the other config option (points) was better as-is because it actually needs to be wider than these. I vote we keep these hard-coded here.

aarondr77: Sounds good!

PR #1180, opened 2024-02-06
Add dark mode for graph toolbar

marthacryan:
Replaces the hard-coded colors in the toolbar icons with the variables.


aarondr77 (commented):
A few of these icons got a little bit messed up. See that in dark mode there are more stripes. 

It might be just that you need to update some more fills in the icons, or I might have made bad icons that need a bit of updating in Figma. 

Once you fix these up LGTM. 


Screenshot 2024-02-06 at 10 22 35 AM

Screenshot 2024-02-06 at 10 22 46 AM

PR #1183, opened 2024-02-06
Add graph toolbar tests

marthacryan:
Adds playwright tests for making graphs using the new toolbar. Uses snapshots to make sure that the generated graph looks right. 

Adds tests for:
- Change chart type to:
  - linear line graph
  - horizontal line graph
  - Scatter graph
- Update titles through double clicking and through the toolbar:
  - graph title
  - x title
  - y title
- Show and hide:
  - gridlines (x and y)
  - titles
  - range slider

Found a couple of things while testing (yay already helping!):
- Throws an error when changing the chart type before the initial render
- Dropdown items are divs when they are better suited to be buttons (accessibility _and_ it helps with them working with the tests)


marthacryan:
update playwright snapshots

marthacryan:
update playwright snapshots

marthacryan:
update playwright snapshots

tests/streamlit_ui_tests/streamlit.spec.ts
```diff
@@ -300,6 +380,220 @@ test.describe('Home Tab Buttons', () => {
...
+
+    await expect(mito.locator('.popup-input')).toBeVisible();
+    await mito.locator('.popup-input').fill('Y axis Title');
+    await mito.locator('.popup-input').press('Enter');
+
+    await awaitResponse(page);
+    await expect(mito.locator('.g-ytitle', { hasText: 'Y axis Title' })).toBeVisible();
+  });
+
+  test('Hide X axis title', async ({ page }) => {
```
aarondr77, line 469: Can we set similar tests that clicking on the X axis title, Y axis title, and chart title and then pressing the delete key deletes it?

tests/streamlit_ui_tests/streamlit.spec.ts
```diff
@@ -300,6 +380,220 @@ test.describe('Home Tab Buttons', () => {
...
+
+    await expect(mito.locator('.popup-input')).toBeVisible();
+    await mito.locator('.popup-input').fill('X axis Title');
+    await mito.locator('.popup-input').press('Enter');
+
+    await awaitResponse(page);
+    await expect(mito.locator('.g-xtitle', { hasText: 'X axis Title' })).toBeVisible();
+  });
+
+  test('Update Y axis Title with double click', async ({ page }) => {
```
aarondr77, line 415: Can we add a test for clicking on the legend and then updating the Y axis title by double clicking on the axis label works. Since we saw that as an issue earlier?

tests/streamlit_ui_tests/streamlit.spec.ts
```diff
@@ -300,6 +380,220 @@ test.describe('Home Tab Buttons', () => {
...
+    await mito.getByRole('button', { name: 'Grid Lines' }).hover();
+    await mito.getByRole('button', { name: 'Vertical' }).click();
+    await awaitResponse(page);
+
+    // Check that the x gridlines are visible and the y gridlines are not affected
+    await expect(mito.locator('.ygrid.crisp')).toHaveCount(4);
+    await expect(mito.locator('.xgrid.crisp')).toHaveCount(5);
+  });
+
+  test('Escape key closes graph title input', async ({ page }) => {
```
aarondr77, line 581: Some other test ideas:

- [ ] Setting the format of the graph
- [ ] Is it possible to click the copy export button and then check that the code is copied to the clipboard?
- [ ] Create a graph, close the select data taskpane, reopen it, add a second trace and make sure graph renders. Maybe we can check that the legend is created
- [ ] Make a histogram and then change the aggregation function and check that the y axis label updates 
- [ ] Make a graph, switch to the data tab, switch back to the graph tab and the graph should re-render 
- [ ] Make a graph with column A on the X axis, switch back to the data tab, apply a filter to remove some of the strings from column A, reopen the graph and the filtered out data should no longer be in the graph
- [ ] Open the select sheet taskpane, open full screen mode, close full screen mode, make sure you can still click on the X button to close the taskpane

aarondr77 (commented):
Awesome start to testing. There is so much functionality in graphing that not having to test this stuff manually during releases is going to be a huge W. 

I thought of a few more tests that I think are worth adding too!

PR #1185, opened 2024-02-07
Glossary page improvements 2 7 24

aarondr77:
# Description

- Adds information about Mito into main glossary content 
- Adds Mito exported functions into glossary (I probably should've tried putting more of this in the template instead of the page contents, but I was anticipating making the Mito value prop more specific for each function and ran out of time)

# Testing

Try them out. 

# Documentation

No.


trymito.io/excel-to-python-page-contents/YEAR.json
```diff
@@ -54,6 +54,21 @@
...
+                "paragraphs": [
+                    "Mito is a Python library that implements Excel's spreadsheet formulas in pandas. Mito's YEAR function works exactly like the Excel YEAR function.",
+                    "To use it, `pip install mitosheet` and then use the following code:"
+                ],
+                "codeLines": [
+                    "# Import the mitosheet Excel functions",
+                    "from mitosheet.public.v3 import *;",
+                    "",
+                    "# Use YEAR function",
+                    "df1['year'] = YEAR(df['Datetime_Column'])"
```
naterush, line 69: Doesn't need to be date time.

trymito.io/public/excel-to-python/formula_writing.mp4
naterush, line 1: This file is massive - I do not think we should include it on any page, as it will nuke page load speed!

naterush (commented):
A little bit of cleanup.

naterush:
I think we should reorg a little bit. So:

Screen Shot 2024-02-08 at 6 04 27 PM

this CTA should have the code that shows you how to do this in Mito (since it's just so excellent). And then beneath that it has a button for install - I think this is good. 

Thoughts?

naterush:
I think ^ could be meaningful to how many users discover Mito's useful functionality with functions, so I vote we do it. Once we do, LGTM -- no need for another review.

trymito.io/excel-to-python-page-contents/HOUR.json
```diff
@@ -129,5 +129,10 @@
                 ]
             }
         ]
+    },
+    "mitoCTA": {
+        "codeLines": [
+            "df['Hour'] = HOUR(df['Datetime_column'])"
```
aarondr77, line 135: 1. Add comment that you don't need to type cast
2. Change column name

trymito.io/excel-to-python-page-contents/LOWER.json
```diff
@@ -143,5 +143,12 @@
                 ]
             }
         ]
+    },
+    "mitoCTA": {
+        "codeLines": [
+            "# Note: We don't need to convert the Text column to a string",
+            "# befor eusing Mito's LOWER function b/c Mito does it automatically",
```
aarondr77, line 150: before using

trymito.io/styles/ExcelToPython.module.css
```diff
@@ -135,7 +135,11 @@
     margin-left: 5px;
 }
 
-
+.in_content_cta_video {
```
aarondr77, line 138: remove

trymito.io/styles/ExcelToPython.module.css
```diff
@@ -202,6 +206,33 @@
...
+        flex-direction: column;
+        justify-content: space-evenly;
+        margin-right: 3rem;
+    }
+
+    .in_content_cta_text > p {
+        margin-top: 0;
+    }
+
+    .in_content_cta_video {
```
aarondr77, line 231: remove this and other ones that are not used

PR #1188, opened 2024-02-12
Fix graph title editor popup bug

### LINKED ISSUE #1181: Graph rename Y axis input cutoff
This happens in the cases below:
1. The x axis labels are very long and make the graph component very small (video below)
2. The y axis title is long and takes up the vertical space

I think we should be able to fix this by giving the popup a higher Z-component so that it sits above the toolbar. Or detecting that it is cutoff and having it move to a different location (ie: directly to the right of the center of the y axis title instead of the top of it)

https://github.com/mito-ds/mito/assets/18709905/d49682dd-7d8f-432c-b39d-74724d095e8c

marthacryan:
Fixes #1181. 
* Adds `ReactDOM.createPortal` to the popup so that it can be on top of the toolbar. 
* Adds calculation to keep the position of the popup from going to far up if the title is very long.

### DIFF
--- a/mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
+++ b/mitosheet/src/mito/components/taskpanes/Graph/GraphSidebar.tsx
@@ -59,6 +59,10 @@ export const Popup = (props: {
         input.focus();
     }, [props.position])
 
+    if (!props.mitoContainerRef) {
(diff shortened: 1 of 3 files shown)


marthacryan:
Closed and re-opened to kick the CI

PR #1190, opened 2024-02-12
Fix height inconsistencies between format tab options

### LINKED ISSUE #1176: Toolbar changes height on graph format tab
Switch from the Gridlines chart format tab to Chart Area format tab and notice that the toolbar grows vertically. 

The toolbar should always be the same size. 

Discovered while reviewing: https://github.com/mito-ds/mito/pull/1159

https://github.com/mito-ds/mito/assets/18709905/f10dd722-ff89-4238-80e2-1c9814b7a2f7

marthacryan:
Fixes #1176.

### DIFF
--- a/mitosheet/css/toolbar.css
+++ b/mitosheet/css/toolbar.css
@@ -64,6 +64,7 @@
 }
 
 .mito-toolbar-bottom p {
+  line-height: 11px;
   font-size: 11px;
 }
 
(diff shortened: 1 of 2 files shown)


naterush:
Screen Shot 2024-02-13 at 1 56 42 PM

Only a related comment (feel free to fix in a different PR), but I think we should update the spacing here so that the text is clearly associated with the input. Can we align things better?

marthacryan:
Agreed about this alignment. I think it'd be great to fix that up but maybe not super high priority?

PR #1194, opened 2024-02-15
Update look and feel of blog

marthacryan:
* Adds a more eye-catching banner for the title of blogs
* Adds a table of contents
* Adds a CTA to the right of the blog
* Adds a section at the bottom to see suggested blogs
* Updates font color / family / size of the fonts
* Adds "share" buttons for email, twitter, and linkedin
* Adds fade-in for dropdowns


trymito.io/styles/[slug].module.css
```diff
@@ -102,11 +124,75 @@
...
+    display: flex;
+    flex-wrap: wrap;
+}
+
+.suggested_blog_card {
+    width: 250px;
+    margin: 30px;
+}
+
+@media only screen and (max-width: 1400px) {
```
Comment (aarondr77) [inline trymito.io/styles/[slug].module.css:176]: Instead of using px here, I think it's best practice to use em. That's because things like the font on our website are based on em which means when the user zooms in on their browser, the font can update correctly. But to keep everything in sync, we should use em here too.

Clearly, not the best answer in the world. I haven't touched up on my rem, em, px knowledge in a few years.

trymito.io/styles/[slug].module.css
```diff
@@ -102,11 +124,75 @@
...
+    margin-left: 30px;
+}
+
 .post_content table thead {
     background-color: black;
     border-radius: 1rem;
 }
 
+.post_cta {
+    width: 265px;
```
Comment (aarondr77) [inline trymito.io/styles/[slug].module.css:142]: These should also be in rem. Take a look at some of the other css files too. 

The other reason that this is important is because our font size changes between mobile and desktop. So basing it off of rem means the width and spacing will be proportional to the font still

trymito.io/pages/blog/[slug].tsx
```diff
@@ -28,61 +29,77 @@ const PostPage = (props: {post: PostOrPage}) => {
...
+              </p>
+              {publishedAt && <p style={{ marginLeft: '15px' }}> {publishedAt}</p>}
+              {readingTime && <p style={{ marginLeft: '15px'}}>{readingTime} min read</p>}
             </div>
-              <div className={postStyles.post_content}>
+            <h1 style={{marginTop: '0'}}>{props.post.title}</h1>
+          </div>
+
+          {/* Table of Contents, post contents, and CTA */}
+          <div style={{ display: 'flex', flexDirection: 'row', alignContent: 'space-between' }}>
```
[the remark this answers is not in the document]
Comment (marthacryan) [reply trymito.io/pages/blog/[slug].tsx:66]: Ah yeah this is part of the thing where we accomadate different sized screens. I'm still trying to figure out a good way to navigate the resizing of screens with removing the cta and the toc.

trymito.io/pages/blog/[slug].tsx
```diff
@@ -2,87 +2,146 @@
...
+                      <p>How to tell the difference between a good and bad Python automation target.</p>
+                    </div>
+                  </Link>
+                  <Link href='/blog/10-mistakes-to-look-out-for-when-transitioning-from-excel-to-python'>
+                    <div className={postStyles.suggested_post}>
+                      <h4>10 Mistakes To Look Out For When Transitioning from Excel To Python</h4>
+                      <p>10 Common Mistakes for new programmers transitioning from Excel to Python</p>
+                    </div>
+                  </Link>
+                  <Link href='/blog/quantifying-mitos-impact-on-analyst-python-productivity'>
```
Comment (aarondr77) [inline trymito.io/pages/blog/[slug].tsx:138]: Can we update the design of these a bit to have some box around each one. Just to make it clear that there are four things here that you can click on? 

Screenshot 2024-02-19 at 10 26 31 AM

trymito.io/pages/blog/[slug].tsx
```diff
@@ -2,87 +2,146 @@
...
+                <a className={postStyles.linkedin_button}
+                  href={`https://linkedin.com/shareArticle?title=${encodeURIComponent(props.post.title ?? '')}&url=${encodeURIComponent(currentURL ?? '')}`}>
+                    <LinkedinLogo />
+                  </a>
+              </div>
 
-          <Footer />
+              {/* Suggested Posts */}
+              <div className={postStyles.suggested_posts_section}>
+                <h1>More Like This</h1>
```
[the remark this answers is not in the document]
Comment (aarondr77) [reply trymito.io/pages/blog/[slug].tsx]: I updated this, but leaving here for knowledge share

trymito.io/styles/[slug].module.css
```diff
@@ -1,19 +1,71 @@
...
+}
+
+
+
+/* Post Content */
 .post_content {
-    max-width: min(720px, 90vw);
+    margin: 0 2rem;
+    max-width: 70rem;
+    width: calc(100vw - 40rem);
```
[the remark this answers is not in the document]
Comment (aarondr77) [reply trymito.io/styles/[slug].module.css]: I removed the width calc and it seems to be working exactly the same. Lmk your thoughts

trymito.io/styles/[slug].module.css
```diff
@@ -47,33 +108,48 @@
...
 .post_content a {
     color: var(--color-purple);
 }
-
 .post_content a:hover {
     /* Underline */
     text-decoration: underline;
 }
 
+.post_content blockquote {
```
Comment (aarondr77) [inline trymito.io/styles/[slug].module.css:136]: There's some wackiness going on here when the screen size gets smaller. This picture is on mobile, but it also happens if you reduce the browser size. 
IMG_2252

trymito.io/styles/[slug].module.css
```diff
@@ -1,19 +1,86 @@
...
 .post_title h1 {
     font-size: 3rem;
+    max-width: 70rem;
+}
+
+.post_title p, .post_title h1 {
+    margin-top: 0
 }
 
+.post_metadata {
```
Comment (aarondr77) [inline trymito.io/styles/[slug].module.css:57]: This needs a bit of love on mobile. Maybe on mobile we can just show the date left aligned with the title and call it a day

Screenshot 2024-02-19 at 10 33 27 AM

trymito.io/styles/[slug].module.css
```diff
@@ -1,19 +1,86 @@
...
+    display: flex;
+    flex-direction: row;
+    margin-top: 2rem;
+}
+
+.post_additional_info {
+    margin-left: 3em;
```
Comment (aarondr77) [inline trymito.io/styles/[slug].module.css:64]: Why `em` here instead of `rem`?

trymito.io/pages/blog/[slug].tsx
```diff
@@ -1,88 +1,171 @@
...
+                  href={`https://twitter.com/intent/tweet?text=${encodeURIComponent(props.post.title ?? '')}&url=${encodeURIComponent(currentURL ?? '')}`}>
+                  <TwitterLogo />
+                </a>
+                <a className={postStyles.email_button}
+                  href={`mailto:?subject=${encodeURIComponent(props.post.title ?? '')}&body=${encodeURIComponent(currentURL ?? '')}`}>
+                    <EmailIcon />
+                  </a>
+                <a className={postStyles.linkedin_button}
+                  href={`https://linkedin.com/shareArticle?title=${encodeURIComponent(props.post.title ?? '')}&url=${encodeURIComponent(currentURL ?? '')}`}>
+                    <LinkedinLogo />
```
Comment (aarondr77) [inline trymito.io/pages/blog/[slug].tsx:118]: This button isn't working for me: 

Screenshot 2024-02-19 at 10 37 43 AM

trymito.io/pages/blog/[slug].tsx
```diff
@@ -1,88 +1,171 @@
...
+                      <h4>Research shows Mito speeds up by 400%</h4>
+                      <p>We&apos;re always on the hunt for tools that improve our efficiency at work. Tools that let us accomplish more with less time, money, and resources.</p>
+                    </div>
+                  </Link>
+                  <Link href='/blog/choosing-between-sql-and-python-best-practices-for-data-analytics-workflows'>
+                    <div className={postStyles.suggested_post}>
+                      <h4>3 Rules for Choosing Between SQL and Python</h4>
+                      <p>Analysts at the world&apos;s top banks are automating their manual Excel work so they can spend less time creating baseline reports, and more time building new analyses that push the company forward.</p>
+                    </div>
+                  </Link>
```
Comment (aarondr77) [inline trymito.io/pages/blog/[slug].tsx:149]: For these `Link` tags to work from an SEO perspective, they need to also have a <a> tag inside of them. Check this out -- after hovering over a link in Chrome, look in the bottom left hand corner to see if Chrome recognizes it as a link 


https://github.com/mito-ds/mito/assets/18709905/9ce760ce-afca-4997-8477-709beb047ac8

In the case of these, it does not! 


So we should update this to be implemented like this: 
Screenshot 2024-02-19 at 10 47 16 AM

trymito.io/styles/[slug].module.css
```diff
@@ -107,12 +183,189 @@
...
+    margin-right: 2rem;
+}
+
+.post_cta {
+    width: 16rem;
+    min-width: 16rem;
+    height: fit-content;
+    position: sticky;
+    top: 6rem;
+    background-color: #F2C94C;
```
Comment (aarondr77) [inline trymito.io/styles/[slug].module.css:218]: Want to make this a global color variable that we can reference elsewhere? It should be in globals.css

trymito.io/styles/[slug].module.css
```diff
@@ -107,12 +183,189 @@
...
+
+.suggested_post {
+    display: flex;
+    flex-direction: column;
+    justify-content: flex-start;
+    align-items: flex-start;
+    margin-right: 1rem;
+    text-decoration: none;
+    color: var(--color-text-primary);
+    width: calc(50% - 1rem);
```
Comment (aarondr77) [inline trymito.io/styles/[slug].module.css:280]: I think again we can just let the flex box handle this. Instead of specifying that you want the suggest post to be half the width and leave space for a margin, just set the margin you want and let the flex box figure out the width -- that's the beauty of flex boxes!

aarondr77 (commented):
The blogs are looking greatly improved! Excited to have this on our website. 

There some cleanup we can do to polish this off (I left comments below) + I updated some spacing. Mostly, I reduced the size of the h2 font and added more spacing in between everything. 

Once this is fixed up LGTM

trymito.io/pages/blog/[slug].tsx
Comment (marthacryan) [reply trymito.io/pages/blog/[slug].tsx:118]: Ah yeah that's not going to work locally because it uses the "current" url - I tested it with vercel and it worked (but used the vercel url)

trymito.io/styles/[slug].module.css
```diff
@@ -1,19 +1,71 @@
+.decorative_line {
+    position: absolute;
+    width: 1px;
+    height: calc(100% - 42rem);
+    border: 0.5px dashed rgba(208, 208, 208, 0.25);
+    left: max(19.5rem, calc(((100vw - 70rem) / 2) - 6px));
```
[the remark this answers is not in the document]
Comment (marthacryan) [reply trymito.io/styles/[slug].module.css:6]: Oh lol you're totally right I'll add a comment

trymito.io/styles/[slug].module.css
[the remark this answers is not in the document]
Comment (marthacryan) [reply trymito.io/styles/[slug].module.css]: Ah yeah it's actually not working the same - I added this because for some reason this happens when the content should be less than 70rem:
image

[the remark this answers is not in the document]
Comment (marthacryan) [reply trymito.io/styles/[slug].module.css]: I really wanted the blog to be responsive to the page width and for some reason flexbox was just not doing it as expected here. I'm not sure what was throwing it off but setting the explicit width fixed it. Let me know if you know of any potential solutions to this using the flexbox style options

trymito.io/styles/[slug].module.css
Comment (marthacryan) [reply trymito.io/styles/[slug].module.css:71]: How does this look? 
image

aarondr77:
On small screens, the title should probably become left aligned 

Screenshot 2024-02-20 at 9 39 09 PM

marthacryan:
> Can we move the footer outside of the width constraint so it takes up the entire width like the rest of the pages?
> 
> Screenshot 2024-02-20 at 9 35 48 PM

Oh huh not sure why it's doing that cause it's not in the post div. But now that I'm looking at the main site's footer, is it supposed to look like this? 
image

aarondr77:
One final idea here, since the big cta on the right hand side is only visible on pretty wide screens, what if we add the basic CTA that is on the glossary pages, in the  [...slug].tsx, file to the bottom of the table of contents once the screen gets small enough to not display it on the right? 

Screenshot 2024-02-20 at 9 41 10 PM

trymito.io/styles/[slug].module.css
```diff
@@ -1,19 +1,102 @@
...
+}
+
+
+
+/* Post Content */
 .post_content {
-    max-width: min(720px, 90vw);
+    margin: 0 2rem;
+    max-width: 70rem;
+    /* Without this, there is a horizontal scrollbar on the page */
```
Comment (aarondr77) [inline trymito.io/styles/[slug].module.css:98]: This is helpful, but can you take it a step further and say why there is a horizontal scrollbar on the page?

Just based on my experience working on the website if we need to hardcode things like this to prevent a scrollbar from appearing, its often because we're hardcoding things elsewhere when they could otherwise be handled with a flex box that has proper margins and padding set. 

But I could be totally wrong!

trymito.io/styles/[slug].module.css
Comment (aarondr77) [reply trymito.io/styles/[slug].module.css:71]: Ooh I think this looks good for the meta data, but can we just left align all of the set so that its not floating in the middle?

aarondr77 (commented):
Okay, just a tiny bit more polish than LGTM without another review! Nice work on this!

trymito.io/styles/[slug].module.css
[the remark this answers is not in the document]
Comment (aarondr77) [reply trymito.io/styles/[slug].module.css]: Makes sense! I appreciate your care for the polish here!

trymito.io/pages/blog/[slug].tsx
Comment (aarondr77) [reply trymito.io/pages/blog/[slug].tsx]: Yeah I think that does a good job of making them look separate from each other. Maybe when we take another pass at the `trymito.io/blog` page we can update this with a similar style. But that's for later!

trymito.io/pages/blog/[slug].tsx
Comment (aarondr77) [reply trymito.io/pages/blog/[slug].tsx:140]: Ooo make sense!

PR #1198, opened 2024-02-20
Enterprise logging fixes

naterush:
# Description

1. Upload on an interval, no matter if the user generates a log or not (handling race conditions with a lock)
2. Make sure we save old logs (added a test)
3. Use an exponential backoff in the case of failure (and adds tests for this)
4. Print a debug message with the errors on upload

# Testing

See new tests. Maybe write a few more.

# Documentation

I updated the documentation.


aarondr77 (commented):
LGTM once tests pass

PR #1202, opened 2024-02-21
Don't overwrite generated code

marthacryan:
* Pulls in some changes to analysis json writing from #1175 
* Adds a `code` field to the analysis JSON which is the direct output of `StepsManager.code()` - an array of strings that represents the most recent code generated with each line of code as an item in the array.


naterush (commented):
The one area I think we actually need to update is `read_and_upgrade_analysis` function, right? Or at least, let's keep this in mind in future PRs...

marthacryan:
Ah you were totally right - actually writing a test helped me catch that! Just pushed a commit that fixes both

mitosheet/mitosheet/saved_analyses/upgrade.py
```diff
@@ -358,6 +358,16 @@ def upgrade_saved_analysis_to_have_up_to_date_args(saved_analysis: Optional[Dict
...
+def upgrade_saved_analysis_to_have_code(saved_analysis: Optional[Dict[str, Any]], code: Optional[str]) -> Optional[Dict[str, Any]]:
+
+    if saved_analysis is None:
+        return None
+    
+    if code is not None:
+        saved_analysis['code'] = code
+    
+    return saved_analysis
+
```
naterush, line 370: This function doesn't actually add the code key when it doesn't exist. I think you need an else statement -- and we can just set it to empty for now? Also, can we make sure there's a test for this.

mitosheet/mitosheet/tests/saved_analyses/test_save_utils.py
```diff
@@ -281,6 +281,23 @@ def test_upgrades_old_analysis_before_replaying_it():
         pd.DataFrame({'A': [123], 'B_sum': [123]})
     )
 
+def test_save_analysis_has_code():
```
naterush, line 284: Yeah -- we just need a test that an old saved analysis (you can get this from your `.mito/saved_analyses` folder) upgrades to have a code key -- so we don't access it somewhere by accident.

mitosheet/src/jupyter/code.tsx
```diff
@@ -132,12 +132,20 @@ export function containsMitosheetCallWithAnyAnalysisToReplay(codeText: string):
...
-    Returns true if the cell contains the code generated for a specific analysis name
-*/
-export function containsGeneratedCodeOfAnalysis(codeText: string, analysisName: string): boolean {
-    return isMitoAnalysisCode(codeText) && codeText.includes(analysisName);
+/**
+ * This function is used to identify if the user has changed the contents of the code
+ * cell that Mito is using to store the generated code. We need to know this to avoid
+ * overwriting the user's code with the generated code.
+ * @param oldCode - The last analysisData code that was written to the cell
+ * @param codeCell - The cell that contains the code
```
naterush, line 140: This param changed

mitosheet/src/mito/Mito.tsx
```diff
@@ -516,6 +540,33 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
+                                        currOpenModal: {
+                                            type: ModalEnum.UserEditedCode,
+                                        }
+                                    }
+                                })
+                            },
+                            oldCodeRef?.current,
+                            overwriteUserEdits,
+                        )
+                    }}
```
naterush, line 567: This feels over abstracted to me. Why not just put this logic in the component itself? It's not like that component is ever going to do something different. Let's move it in by just making the `jupyterUtils` the prop to this component, and making this call in the component.

mitosheet/src/mito/components/modals/UserEditedCodeModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+                    </div>
+                </Fragment>
+            }
+            buttons={
+                <>
+                    <TextButton 
+                        variant='light'
+                        width='hug-contents'
+                        onClick={() => {
+                            props.onClickButton(true);
```
naterush, line 40: Why I think it's overabstracted: it's totally unclear what this function is doing when you read this component. I missed it on my first time through.

naterush: On top of moving this function into this component, I think we should rename it to something a bit more communicative.

mitosheet/src/plugin.tsx
```diff
@@ -160,12 +165,23 @@ function activateMitosheetExtension(
...
+            // trigger the dialog to ask them. 
+            if (overwriteIfUserEditedCode === undefined && !isEmptyCell(codeCell) && hasCodeCellBeenEditedByUser(oldCode, codeCellText)) {
+                triggerUserEditedCodeDialog();
+                return;
+            // Only write to the cell if either of the following are true:
+            // 1. The user has authorized overwriting the cell
+            // 2. The cell hasn't been edited by the user
+            // AND the cell exists. If the cell doesn't exist we can't write to it!
+            } else if (codeCell !== undefined && (overwriteIfUserEditedCode || !hasCodeCellBeenEditedByUser(oldCode, codeCellText))) {
                 writeToCell(codeCell, code)
```
naterush, line 181: What happened to the previous check here around containsGeneratedCodeOfAnalysis? Are we just missing this now / do we not need it?

I think we need to be really careful around this logic, as it's so complex. You did a good job simplifying since the last PR, but it's still not _obviously_ correct to me -- it take some reading.

mitosheet/src/mito/components/modals/UserEditedCodeModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+const UserEditedCodeModal = (
+    props: {
+        setUIState: React.Dispatch<React.SetStateAction<UIState>>;
+        mitoAPI: MitoAPI,
+        userProfile: UserProfile,
+        onClickButton: (overwriteUserEdits: boolean) => void,
+    }): JSX.Element => {
+    return (
+        <DefaultModal
+            header='Insert New Cell?'
```
naterush, line 24: I think we could change the header to something less action oriented and more communicative about _why_ the user is there. Let's change this to `Edit to Code Detected`?

naterush (commented):
A few bits of code cleanup and some questions. Overall, well designed and reasonable to understand.

naterush:
Working excellently in Lab and Notebook

mitosheet/src/plugin.tsx
marthacryan: Oh yeah I left a comment in that PR about this - I don't think we need that anymore because we're already checking that it matches the code we generate. After looking into what that function was doing, it seemed like it might interfere with our checks (i.e. if the user made some change that made it look like it wasn't our code, it would just automatically create another cell). Unless you have another case that would explain a need to put that back in, I think the current check is sufficient. Lmk!

aarondr77:
This implementation breaks our ability to change what code gets generated by code chunks. 

To see this: 
1. Use Mito to add a column
2. Edit the `get_description_comment` function of the `add_column_code_chunk`
3. Restart your kernel + refresh browser
4. Rerun the mitosheet.sheet() call
5. Get the code changed modal

naterush:
Great bug catch, Aaron. 

I think this is where the `code` saved in the analysis json can be used. Just an idea here, but what if the first time the sheet was rendered, we sent over the code in the analysis json. 

Because the code generated Mito will only change with an kernel restart and thus a re-rendering of the Mito sheet, we'd always be sending over the right code in this case. 

Thoughts?

aarondr77:
So specifically, the proposal is: 

1. On the first render (render count is 0), use the code saved in the analysis.json to compare against the generated code cell. This works because the code saved in the analysis.json won't include the changes to how Mito generates code. 
2. Every time else use the code that Mito generated for the previous generated code. 

This seems like it would work!

aarondr77:
### Scenario 1:
It's your first time rendering the mitosheet after restarting your kernel. 

If the code in the code cell is the **same** as the code stored in the analysis.json then, automatically overwrite the generated code with the code created from the mito backend.

And save, the code generated from the mito backend in the oldCodeRef for the next time the user makes an edit. 

To get this, we can make an API call. You can get the `analysis_name` from the steps_manager. Then, in the API call you can use the `read_analysis` utility. 

### Scenario 2: 
It's your first time rendering the mitosheet after restarting your kernel. 

If the code in the code cell is the **different** than the code stored in the analysis.json then this means that the user must have changed the generated code, so we should show them the dialog box. 

### Scenario 3: 
If it's not the first time rendering the mitosheet, then we compare the code in the code cell with the oldCodeRef code.

marthacryan:
Update on this:
* the above commit adds an API call to the backend to get the code field from the saved analysis file, which we then use as the "old code" when doing the initial `writeGeneratedCodeToCell` call.
* This requires an asynchronous call to the backend, **then** writing to the code cell. 
* This is creating a race condition with calling `execute` on the command registry of JupyterLab
* I attempted to add a state variable to keep track of whether this initial write had happened, but the race condition didn't appear to be fixed by that.

mitosheet/mitosheet/saved_analyses/upgrade.py
```diff
@@ -358,6 +358,18 @@ def upgrade_saved_analysis_to_have_up_to_date_args(saved_analysis: Optional[Dict
...
 
+def upgrade_saved_analysis_to_have_code(saved_analysis: Optional[Dict[str, Any]], code: Optional[List[str]]) -> Optional[Dict[str, Any]]:
+
+    if saved_analysis is None:
+        return None
+    
+    if code is not None:
+        saved_analysis['code'] = code
+    else:
+        saved_analysis['code'] = []
```
naterush, line 369: Feels like we need to handle this case in the code that gets form the backend as well. If the old code is empty, and there's code there, we shouldn't detect that they edited things (or write a new cell). Maybe we want to tag this empty case with 'empty' or something?

mitosheet/src/mito/Mito.tsx
```diff
@@ -278,15 +274,60 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
-        if (analysisData.renderCount >= 1) {
-            // Finially, we can go and write the code!
-            props.jupyterUtils?.writeGeneratedCodeToCell(analysisData.analysisName, analysisData.code, userProfile.telemetryEnabled, analysisData.publicInterfaceVersion);
+        void mitoAPI.getSavedAnalysisCode().then((response: MitoAPIResult<string[]>) => {
+            if ('error' in response) {
+                console.error(response.error);
+                return;
+            }
+            oldCodeRef.current = response.result;
+        });
```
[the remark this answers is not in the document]
naterush: Can you document what was happening with render count? It feels like if we could, we would ideally use that solution for simplicity -- but I don't understand why it's not working.

naterush (commented):
Some clarification as to this approaches motivation would be helpful.

mitosheet/mitosheet/saved_analyses/upgrade.py
```diff
@@ -358,6 +358,16 @@ def upgrade_saved_analysis_to_have_up_to_date_args(saved_analysis: Optional[Dict
     
     return saved_analysis
 
+def upgrade_saved_analysis_to_have_code(saved_analysis: Optional[Dict[str, Any]], code: Optional[str]) -> Optional[Dict[str, Any]]:
+
+    if saved_analysis is None:
+        return None
+    
+    if code is not None:
+        saved_analysis['code'] = code
```
naterush, line 367: Can we set code to None? To avoid key errors

mitosheet/src/mito/Mito.tsx
```diff
@@ -278,15 +274,62 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
+        if (oldCodeRef.current === undefined) {
+            void mitoAPI.getSavedAnalysisCode().then((response: MitoAPIResult<string[]>) => {
+                if ('error' in response) {
+                    console.error(response.error);
+                    return;
+                }
+                oldCodeRef.current = response.result;
+            });
+            return;
         }
```
naterush, line 298: Can we make the effect run an async function, and remove the return? So we just await the getSavedAnalsis, set the old code ref, and then continue with the rest of the code block

naterush (commented):
Minor cleanup, then I'll do a final review

naterush:
I think failing tests are actually catching an issue. For me, code gen doesn't work unless there's already an empty cell below the Mito sheet. 

1. Open a new notebook
2. Click the new Mitosheet button
3. Import some data
4. This ends up in the following state (notice no code is there)

Screen Shot 2024-03-08 at 3 19 01 PM

I have to manually add a code cell for things to work.

naterush:
Code gen is not working on Jupyter Notebook on first render at all -- can you dig into this? My guess is that this is an issue with render count, potentially? It works when you run the mitosheet.sheet call again...

Also, how are our tests passing here?

mitosheet/src/mito/Mito.tsx
```diff
@@ -279,15 +275,64 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
+                    return;
+                }
+                oldCodeRef.current = response.result;
+            });
+            // After setting the oldCodeRef, we return because this is supposed to represent
+            // the code from the last time analysisData changed
+            return;
         }
+    
+        void props.jupyterUtils?.writeGeneratedCodeToCell(
```
naterush, line 303: Do we need this function to be async anymore?

mitosheet/src/mito/Mito.tsx
```diff
@@ -279,15 +275,64 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
+                            codeWithUserEdits: codeWithUserEdits
+                        }
+                    }
+                })
+                void mitoAPI.log(
+                    'user_edited_code_cell', 
+                    {
+                        length_of_code_with_user_edits: codeWithUserEdits.length,
+                        length_of_code_without_user_edits: codeWithoutUserEdits.length
+                    }
```
naterush, line 324: Would be down to move this into a useEffect in the modal

naterush (commented):
Some minor cleanup. I'm ok with where we ended up -- it's a bit confusing, but the documentation is clarifying and I don't think we'll have to touch this too often.

Going to test it now!

naterush:
There's a minor bug where if you delete the code cell below Mito, it detects this as an edit to the generated code. I think it would be more sensible to insert a new empty code cell, but also fine if we jut open an issue for this one.

tests/jupyterlab_ui_tests/jupyterlab.mitosheet.spec.ts
```diff
@@ -10,7 +10,7 @@ test.describe('Mitosheet JupyterLab integration', () => {
     expect(await cellOuput?.innerHTML()).toContain('Insert');
   });
 
-  test.skip('Can run the generated code', async ({ page, tmpPath }) => {
```
naterush, line 13: Nice, good catch!

.github/workflows/test-mitosheet-frontend.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        python-version: ['3.6', '3.8', '3.10']
+        python-version: ['3.7', '3.8', '3.10']
```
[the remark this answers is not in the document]
marthacryan: Let's find out!

naterush (commented):
Ok. One more minor bug (I'd be fine with a test for this one one), the cleanup above, and then LGTM!

mitosheet/src/mito/Mito.tsx
naterush: with an empty dependency list, so runs on first render only

.github/workflows/test-mitosheet-frontend.yml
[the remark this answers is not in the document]
marthacryan: yep

mitosheet/src/mito/Mito.tsx
marthacryan: nope

PR #1204, opened 2024-02-22
Prevent overwriting code when the user has made changes to the generated code cell

marthacryan:
Updates the execute function of `writeGeneratedCodeToCell` to take `oldCode` as an input and prevent overwriting changes if the code in the cell is different from the Mito generated code. This required adding a `ref` to keep track of the last mito-generated code in the analysis json and updating it after the call to `writeGeneratedCodeToCell`.

### DIFF
--- a/mitosheet/src/jupyter/jupyterUtils.tsx
+++ b/mitosheet/src/jupyter/jupyterUtils.tsx
@@ -52,7 +52,7 @@ export const overwriteAnalysisToReplayToMitosheetCall = (oldAnalysisName: string
 }
 
 
-export const writeGeneratedCodeToCell = (analysisName: string, code: string[], telemetryEnabled: boolean, publicInterfaceVersion: PublicInterfaceVersion, oldCode?: string[]): void => {
+export const writeGeneratedCodeToCell = (analysisName: string, code: string[], telemetryEnabled: boolean, publicInterfaceVersion: PublicInterfaceVersion, oldCode: string[]): void => {
     if (isInJupyterLab()) {
         window.commands?.execute('mitosheet:write-generated-code-cell', {
             analysisName: analysisName,
--- a/mitosheet/src/jupyter/jupyterUtils.tsx
+++ b/mitosheet/src/jupyter/jupyterUtils.tsx
@@ -62,7 +62,7 @@ export const writeGeneratedCodeToCell = (analysisName: string, code: string[], t
             oldCode: oldCode
         });
     } else if (isInJupyterNotebook()) {
-        notebookWriteGeneratedCodeToCell(analysisName, code, telemetryEnabled, publicInterfaceVersion);
+        notebookWriteGeneratedCodeToCell(analysisName, code, telemetryEnabled, publicInterfaceVersion, oldCode);
     } else {
         console.error("Not detected as in Jupyter Notebook or JupyterLab")
     }
--- a/mitosheet/src/jupyter/jupyterUtils.tsx
+++ b/mitosheet/src/jupyter/jupyterUtils.tsx
@@ -52,13 +52,14 @@ export const overwriteAnalysisToReplayToMitosheetCall = (oldAnalysisName: string
 }
 
 
-export const writeGeneratedCodeToCell = (analysisName: string, code: string[], telemetryEnabled: boolean, publicInterfaceVersion: PublicInterfaceVersion): void => {
+export const writeGeneratedCodeToCell = (analysisName: string, code: string[], telemetryEnabled: boolean, publicInterfaceVersion: PublicInterfaceVersion, oldCode?: string[]): void => {
     if (isInJupyterLab()) {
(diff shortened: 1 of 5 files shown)


mitosheet/src/jupyter/jupyterUtils.tsx
```diff
@@ -52,16 +52,17 @@ export const overwriteAnalysisToReplayToMitosheetCall = (oldAnalysisName: string
 }
 
 
-export const writeGeneratedCodeToCell = (analysisName: string, code: string[], telemetryEnabled: boolean, publicInterfaceVersion: PublicInterfaceVersion): void => {
+export const writeGeneratedCodeToCell = (analysisName: string, code: string[], telemetryEnabled: boolean, publicInterfaceVersion: PublicInterfaceVersion, oldCode?: string[]): void => {
```
naterush, line 55: Ok -- so it's optional because upgraded analyses might not have it?

mitosheet/src/jupyter/notebook/extensionUtils.tsx
```diff
@@ -229,7 +229,16 @@ export const notebookWriteGeneratedCodeToCell = (analysisName: string, codeLines
 
     const codeCell = getCellAtIndex(mitosheetCallIndex + 1);
 
+
+    // We're removing the first line of the old code and the cell code because
+    // the cell code contains the analysis id and the old code does not
+    const oldCodeWithoutFirstLine = oldCode?.slice(1).join('\n');
+    const cellCodeWithoutFirstLine = getCellText(codeCell)?.split('\n').slice(1).join('\n');
```
naterush, line 236: A more robust check might be: 
```
const endsWithOldCode = getCellText(codeCell).endsWith(oldCode)
```

And a comment explaining this as well

mitosheet/src/plugin.tsx
```diff
@@ -161,7 +164,16 @@ function activateMitosheetExtension(
 
             const codeCell = getCellAtIndex(cells, mitosheetCallIndex + 1);
 
+            // We're removing the first line of the old code and the cell code because
+            // the cell code contains the analysis id and the old code does not
+            const oldCodeWithoutFirstLine = oldCode?.slice(1).join('\n');
+            const cellCodeWithoutFirstLine = getCellText(codeCell)?.split('\n').slice(1).join('\n');
+
```
naterush, line 171: Given that this logic requires the same documentation in two places, I feel like we should factor it into a helper function with a clear and descriptive name.

naterush: `src/jupyter/code.tsx` is probably a good file

mitosheet/src/mito/Mito.tsx
```diff
@@ -279,15 +279,21 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
          * We only write code after the render count has been incremented once, which
          * means that we have read in and replayed the updated analysis, etc. 
          */
         if (analysisData.renderCount >= 1) {
             // Finially, we can go and write the code!
-            props.jupyterUtils?.writeGeneratedCodeToCell(analysisData.analysisName, analysisData.code, userProfile.telemetryEnabled, analysisData.publicInterfaceVersion);
+            props.jupyterUtils?.writeGeneratedCodeToCell(analysisData.analysisName, analysisData.code, userProfile.telemetryEnabled, analysisData.publicInterfaceVersion, oldCodeRef?.current);
         }
+        // After using the ref to get the old code, we update it to the newest analysis.
+        oldCodeRef.current = analysisData.code;
```
naterush, line 296: Should this be up, inside the if statement? If no, why not.

mitosheet/src/jupyter/notebook/extensionUtils.tsx
naterush: You solution is better actually, given that it handles insertions more correctly. Def worth putting in a util in any case.

naterush (approved):
I'm wondering if the first PR (adding the code to the saved analysis) even makes sense? Like -- what is the `code` key doing in the analysis? Does it serve any purpose to keep it there, if we're just doing all this on the frontend? 

Btw, I think it's reasonable to do this all on the frontend -- this solution makes sense.... just wondering if I specified additional work for no reason.

Functionality is working great! Some code cleanup, then LGTM.

mitosheet/src/jupyter/jupyterUtils.tsx
marthacryan: Yes

mitosheet/src/mito/Mito.tsx
marthacryan: Oh yeah I think it's better to just update it any time the analysis data changes.

mitosheet/src/jupyter/jupyterUtils.tsx
marthacryan: I went ahead and removed that because we're using `analysisData`, which doesn't list `code` as optional

PR #1207, opened 2024-02-22
Pip install on main page

naterush:
# Description

1. Adds install instrucitons
2. Adds the `hello` command
3. does not add email collection in hello command, just to not get too fancy for now

# Testing

A few things here:
1. Test the new install instructions
2. Make sure the classes on buttons on the website are good
4. Make sure logging is good in the `hello` command, including adding failures

# Documentation

Should we update the docs as well/

### DIFF
--- a/mitosheet/mitosheet/hello.py
+++ b/mitosheet/mitosheet/hello.py
@@ -1,0 +1,99 @@
+import json
+import os
+import sys
+from mitosheet.telemetry.telemetry_utils import log
+import analytics
+
+MITO_STARTER_NOTEBOOK_PATH = './mito-starter-notebook.ipynb'
+MITO_STARTER_NOTEBOOK_CONTENTS = {
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": None,
+   "id": "c0957234-ba57-4199-adfa-9d17ede0b8f9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Run this cell to render a mitosheet\n",
+    "# See mito's documentation here: https://docs.trymito.io/how-to/creating-a-mitosheet\n",
+    "# Join our Discord for support here: https://discord.com/invite/XdJSZyejJU\n"
+    "\n",
+    "import mitosheet\n",
+    "mitosheet.sheet()"
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3 (ipykernel)",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.8.5"
+  },
+  "widgets": {
+   "application/vnd.jupyter.widget-state+json": {
+    "state": {},
+    "version_major": 2,
+    "version_minor": 0
+   }
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 5
+}
+
+def try_create_starter_notebook():
+    """
+    Creates MITO_STARTER_NOTEBOOK_CONTENTS at MITO_STARTER_NOTEBOOK_PATH,
+    which helps the user get started.
+
+    If the file already exists, will not replace it, as to not overwrite
+    users work in this area.
+    """
+    try:
+        if not os.path.exists(MITO_STARTER_NOTEBOOK_PATH):
+            with open(MITO_STARTER_NOTEBOOK_PATH, 'w+') as f:
+                f.write(json.dumps(MITO_STARTER_NOTEBOOK_CONTENTS))
+                
+        return True
+    except Exception as e:
+        log(
+            'hello_command_try_create_starter_notebook_error', 
+            failed=True,
+            error=e
+        )
+        # TODO: add a log
+        return False
+
+
+
+def run_hello():
+    log('hello_command')
+    created_starter_notebook = try_create_starter_notebook()
+
+    if not created_starter_notebook:
+        print(f"Did not have the permissions to create a starter notebook. Check our instructions for creating a new Mitosheet at: https://docs.trymito.io/how-to/creating-a-mitosheet")
+        exit(1)
+
+    log('hello_command_success')
+
+    # Flush any logs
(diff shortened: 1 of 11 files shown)


mitosheet/mitosheet/hello.py
```diff
@@ -0,0 +1,110 @@
...
+                f.write(json.dumps(MITO_STARTER_NOTEBOOK_CONTENTS))
+                
+        return True
+    except Exception as e:
+        log(
+            'hello_command_try_create_starter_notebook_error', 
+            failed=True,
+            error=e
+        )
+        # TODO: add a log
```
naterush, line 76: Remove

naterush (commented):
Add trymito.io code

trymito.io/components/InstallInstructions/InstallInstructions.tsx
```diff
@@ -15,16 +16,17 @@ const InstallInstructions = (props: {}): JSX.Element => {
...
-                <CodeBlock prefix='$ ' paddingRight='7rem'>
+                <CodeBlock prefix='$ ' paddingRight='7rem' className={PLAUSIBLE_COPIED_PIP_INSTALL_COMMAND}>
                     pip install mitosheet
                 </CodeBlock>
-                <CodeBlock prefix='$ ' paddingRight='7rem'>
+                <CodeBlock prefix='$ ' paddingRight='7rem' className={PLAUSIBLE_COPIED_MITOSHEET_HELLO_COMMAND}>
                     python -m mitosheet hello
                 </CodeBlock>
                 <div className={ctaButtons.cta_subbutton}>
                     <Link href={MITO_INSTALLATION_DOCS_LINK}>
```
aarondr77, line 26: 1. This should probably take you to the `create mitosheet` page instead of the install page?
2. Should we say something in addition like "If your install failed, try using these common installation issues fixes"
3. It would be good to link our Discord/Slack here so users can get support if they need since we don't have the the ability to show it if the pip install failed

mitosheet/mitosheet/hello.py
```diff
@@ -0,0 +1,109 @@
...
+        return False
+
+
+
+def run_hello() -> None:
+    log('hello_command')
+
+    print()
+    print()
+    print("\t 👋 Welcome to Mito!")
```
aarondr77, line 85: This text flies by so quickly I don't think anyone is going to see it. I wouldn't have seen it if I didn't know it was there. This text isn't particularly value-add so I think that's okay. 

Ht only other thing I could think to do is ask if they want us to launch Jupyter for them, but that seems worse.

mitosheet/mitosheet/hello.py
```diff
@@ -0,0 +1,109 @@
...
+    print("\tThis command launches Mito in JupyterLab, but Mito is also available in Jupyter Notebook Classic, Streamlit, and Dash.")
+    print()
+    print("\tSee our docs here: https://docs.trymito.io")
+    print()
+    print()
+
+    created_starter_notebook = try_create_starter_notebook()
+
+    if not created_starter_notebook:
+        print(f"Did not have the permissions to create a starter notebook. Check our instructions for creating a new Mitosheet at: https://docs.trymito.io/how-to/creating-a-mitosheet")
```
aarondr77, line 96: The bigger thing than not creating the starter notebook is not launching Jupyter for them. I vote we:

1. Don't mention that we could not create a starter notebook for them (since they probably weren't even expecting we do that anyways)
2. Just run the `jupyter lab` command anyways to get them into Jupyter.

mitosheet/mitosheet/hello.py
```diff
@@ -0,0 +1,109 @@
...
+        exit(1)
+
+    log('hello_command_success')
+
+    # Flush any logs
+    analytics.flush()
+
+    # Launch JupyterLab
+    os.execl(sys.executable, 'python', '-m', 'jupyter', 'lab', MITO_STARTER_NOTEBOOK_PATH)
```
aarondr77, line 105: Should we put this in a `try except`? Feels like we shouldn't expect that this will always work and we should log it -- moving the success log down

trymito.io/utils/plausible.tsx
```diff
@@ -14,6 +14,11 @@ export const PLAUSIBLE_INSTALL_DOCS_CTA_LOCATION_EXCEL_TO_PYTHON_GLOSSARY_TOC_CT
 export const PLAUSIBLE_INSTALL_DOCS_CTA_LOCATION_EXCEL_TO_PYTHON_GLOSSARY_IN_CONTENT_CTA = 'plausible-event-name=install_docs_cta_pressed+location_excel_to_python_glossary_in_content_cta'
 export const PLAUSIBLE_INSTALL_DOCS_CTA_LOCATION_BLOG = 'plausible-event-name=install_docs_cta_pressed+location_blog'
 
+// Copy the install instructions directly
+export const PLAUSIBLE_SCROLLED_TO_INSTALL_INSTRUCTIONS = 'plausible-event-name=scrolled_to_install_instructions+location_title_card'
+export const PLAUSIBLE_COPIED_PIP_INSTALL_COMMAND = 'plausible-event-name=copied_pip_install_command+location_install_instructions'
+export const PLAUSIBLE_COPIED_MITOSHEET_HELLO_COMMAND = 'plausible-event-name=copied_mitosheet_hello_command+location_install_instructions'
```
aarondr77, line 20: Nice -- make sure you set these as plausible goals

trymito.io/components/InstallInstructions/InstallInstructions.tsx
```diff
@@ -15,16 +16,17 @@ const InstallInstructions = (props: {}): JSX.Element => {
                 in under a minute
```
aarondr77, line 16: Maybe we should say "In two lines of code" instead of "in under a minute" because we know on some user's computers this could take 45 minutes to run if they need to install everything.

aarondr77 (commented):
Sweet a little bit of cleanup and then LGTM -- your call whether to take this through standard release process or do a special release. 

Let's update the main docs too!

mitosheet/mitosheet/hello.py
naterush: Yeah updated. Now, no log means success -- as we don't have a way of actually detecting success.

trymito.io/components/InstallInstructions/InstallInstructions.tsx
naterush: 1. Yes
2. No for now. 
3. Added

naterush:
Merging when tests pass

PR #1212, opened 2024-02-22
Add logging for handling user edits

marthacryan:
Adds logging when:
* The user chooses to overwrite the code changes they made
* The user chooses to insert a new cell and keep their changes
* Mito detects that the user has made a change that could be overwritten


mitosheet/src/mito/Mito.tsx
```diff
@@ -306,6 +306,7 @@ export const Mito = (props: MitoProps): JSX.Element => {
                             }
                         }
                     })
+                    void mitoAPI.log('user_edited_code_cell')
```
naterush, line 309: Would be nice if this log would capture what was edited here - can we log the old code and the new code?

naterush (commented):
This code looks good. Did you test these logs get generated in Mixpanel?

marthacryan:
> This code looks good. Did you test these logs get generated in Mixpanel?

Oh I didn't realize that you could do that from a PR - I'll try that out!

marthacryan:
Closed and reopened to kick CI

mitosheet/src/plugin.tsx
```diff
@@ -178,8 +179,16 @@ function activateMitosheetExtension(
             // 2. The cell hasn't been edited by the user
             // AND the cell exists. If the cell doesn't exist we can't write to it!
             } else if (codeCell !== undefined && (overwriteIfUserEditedCode || !hasCodeCellBeenEditedByUser(oldCode, codeCellText))) {
+                // Logging
+                if (overwriteIfUserEditedCode) {
+                    void mitoAPI.log('overwrite_user_edited_code', { length_of_code_with_user_edits: codeCellText.split('\n').length, length_of_code_without_user_edits: oldCode.length });
```
naterush, line 184: These variable names with `_` are better

mitosheet/src/jupyter/notebook/extensionUtils.tsx
```diff
@@ -239,8 +239,16 @@ export const notebookWriteGeneratedCodeToCell = (analysisName: string, codeLines
...
     } else if (codeCell !== undefined && (overwriteIfUserEditedCode || !hasCodeCellBeenEditedByUser(oldCode, codeCellText))) {
+        // Logging
+        if (overwriteIfUserEditedCode) {
+            void mitoAPI.log('overwrite_user_edited_code', { lengthOfCodeWithUserEdits: codeCellText.split('\n').length, lengthOfCodeWithoutUserEdits: oldCode.length });        
+        }
         writeToCell(codeCell, code)
     } else {
+        // Logging
+        if (overwriteIfUserEditedCode === false) {
+            void mitoAPI.log('insert_new_cell_for_user_edited_code', { lengthOfCodeWithUserEdits: codeCellText.split('\n').length, lengthOfCodeWithoutUserEdits: oldCode.length })
```
naterush, line 250: Change these param to snake case as well

naterush (commented):
Let's move this logging into the Modal. This is a very similar refactor to the `jupyterUtils` as props that I've been pushing for, and it has the following benefits:
1. Logging happens in an single location, rather than two. This would have fixed the snake_case vs camelCase bug this PR has. 
2. it's more clear when something should happen, and doesn't complicate some already complicated if statements. 

I think this is a good lesson generally -- if you can put the the effects of a button click _in_ the button click event handler, rather than elsewhere in the codebase, it's generally a good idea to do so.

mitosheet/src/mito/Mito.tsx
```diff
@@ -305,7 +305,9 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
```
naterush, line 308: It would be really nice to get the params in this log call as well. Can we make the oldCode and newCode params to the callback function for opening the dialogue, and use them here as well?

mitosheet/src/plugin.tsx
naterush: Yeah, confirmed this is a bug that makes params not work on jupyter notebooks

mitosheet/src/mito/components/modals/UserEditedCodeModal.tsx
```diff
@@ -37,7 +40,33 @@ const UserEditedCodeModal = (
                         variant='light'
                         width='hug-contents'
                         onClick={() => {
-                            props.onClickButton(true);
+                            props.jupyterUtils?.writeGeneratedCodeToCell(
```
naterush, line 43: these two on clicks are the same could be a function probably -- just defined in the component. :)

naterush (commented):
Sweet. Tiny cleanup then LGTM

naterush:
wait I found bug in the number of lines of code -- if I don't add any, it still shows them as different numbers. Is this because we're removing the starting lines or something?

marthacryan:
> wait I found bug in the number of lines of code -- if I don't add any, it still shows them as different numbers. Is this because we're removing the starting lines or something?

I think this is resolved!

PR #1213, opened 2024-02-23
Test modal for user editing code cell behavior

marthacryan:
Adds a test to the JupyterLab tests that checks:
* When the user makes a change to the mito generated code in a notebook, the next time that they make an edit the modal pops up to ask whether they want to overwrite the code
* When they press "overwrite code" the code is overwritten
* When they press "insert new cell" the code is still present, but at the next index in the notebook
* The dialog goes away after either of those buttons are pressed


naterush:
This test looks excellent.

naterush:
Let's just skip this check on 3.6 for now -- link to this PR, and just leave a note that we're not sure why this happened.

PR #1214, opened 2024-02-23
mitosheet: handle taskpane crashes locally

naterush:
# Description

Handles taskpane crashing issues locally, allowing the users to just reopen taskpane and avoid crashing the entire sheet.

**Before merging, make sure to clean up the Values Tab.**

# Testing

Open the values tab. Check resulting logs.

**Before merging, make sure to clean up the Values Tab.**

# Documentation

No.


mitosheet/src/mito/components/taskpanes/DefaultTaskpane/DefaultTaskpaneErrorBoundry.tsx
```diff
@@ -0,0 +1,52 @@
...
+            return {
+                ...prevUIState,
+                currOpenTaskpane: {type: TaskpaneType.NONE}
+            }
+        })
+
+        void this.props.mitoAPI?.log('taskpane_render_failed', {
+            js_error: error,
+            js_error_info: errorInfo,
+            js_taskpane_header: this.props.taskpaneHeader
```
aarondr77, line 39: This often contains the column header that we would not want to log

aarondr77: Instead, what if we log the taskpane type. Probably more useful anyways

aarondr77 (commented):
LGTM after tiny cleanup + is it possible to add a frontend test for this? I can't exactly think how unless we did some inject code into the taskpane that errors??

mitosheet/src/mito/components/taskpanes/DefaultTaskpane/DefaultTaskpaneErrorBoundry.tsx
naterush: Really hard to get honestly, and we don't have the taskpane type in every case. Rather, I'm going to just remove this for now -- and we'll see if we need to add it back.

PR #1217, opened 2024-02-26
Add notebook tests

marthacryan:
Adds a very simple test for jupyter notebook support
* Running into the same issue with Python3 3.6 that we're running into w/ the Lab tests. I tried running locally w/ Python 3.6 and didn't reproduce
* I want put together some utils as well. Just a note


tests/jupyter_notebook_config.py
```diff
@@ -0,0 +1,19 @@
+# Configuration file for jupyter-notebook.
+
+from tempfile import mkdtemp
+
+
+c = get_config()  #noqa
```
naterush, line 6: What is this?

tests/jupyter_notebook_config.py
```diff
@@ -0,0 +1,19 @@
+# Configuration file for jupyter-notebook.
```
naterush, line 1: Is this file from somewhere? Can you document or link to it if so?

tests/jupyter_server_test_config.py
```diff
@@ -4,7 +4,7 @@
 c.ServerApp.port_retries = 0
 c.ServerApp.open_browser = False
 
-c.ServerApp.root_dir = mkdtemp(prefix='galata-test-')
+c.ServerApp.notebook_dir = mkdtemp(prefix='galata-test-')
```
naterush, line 7: What's this change doing?

tests/notebook_ui_tests/notebook.mitosheet.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { IJupyterLabPageFixture, expect, test } from '@jupyterlab/galata';
+import { awaitResponse } from '../streamlit_ui_tests/utils';
+import path from 'path';
+
+test.describe.configure({ mode: 'parallel' });
+test.use({ appPath: '', autoGoto: false })
```
naterush, line 6: What are we actually using Galata for in this file? It kinda feels like we could just eliminate it and pretty much nothing would change, no?

naterush (commented):
This is an awesome start! A few questions, then LGTM.

tests/jupyter_notebook_config.py
marthacryan: I got this from calling `jupyter notebook --generate-config` because I was running into issues with the config not being picked up by the notebook server.

tests/notebook_ui_tests/notebook.mitosheet.spec.ts
marthacryan: Actually we're using it for `page.contents.uploadFile` in the `beforeEach`.

PR #1222, opened 2024-02-28
Add more graph tests

marthacryan:
* Refactors some of the tests to use utils for clarity
* Adds tests for:
  * Update x axis columns
  * Update y axis columns
  * deleting the source data deletes the graph
* Fixes a bug with deleting source data

### DIFF
--- a/tests/streamlit_ui_tests/graph.spec.ts
+++ b/tests/streamlit_ui_tests/graph.spec.ts
@@ -1,7 +1,8 @@
-import { expect, test } from '@playwright/test';
+import { FrameLocator, Page, expect, test } from '@playwright/test';
 import { awaitResponse, clickButtonAndAwaitResponse, clickTab, getMitoFrameWithTestCSV, getMitoFrameWithTypeCSV } from './utils';
 
 test.describe.configure({ mode: 'parallel' });
+test.use({ permissions: ['clipboard-read'] })
 
 const openPopupAndEditTitle = async (mito: any, selector: string, newTitle: string) => {
     await mito.locator(selector).dblclick();
@@ -34,8 +35,7 @@ const testEditTitleThroughContextMenu = async (page, selector) => {
 const testDeleteTitleThroughContextMenu = async (page, selector) => {
   const mito = await getMitoFrameWithTypeCSV(page);
 
-    await clickButtonAndAwaitResponse(page, mito, { name: 'Graph' })
-    await expect(mito.getByText('Column1 bar chart')).toBeVisible();
+    await openGraphEditor(mito, page);
 
     await mito.locator(selector).click({ button: 'right' });
     await mito.getByRole('button', { name: 'Delete Title' }).click();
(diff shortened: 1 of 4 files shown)


tests/streamlit_ui_tests/graph.spec.ts
```diff
@@ -44,13 +44,41 @@ const testDeleteTitleThroughContextMenu = async (page, selector) => {
...
+  await mito.locator('.mito-dropdown-item', { hasText: columnName }).click();
+  await awaitResponse(page);
+}
+
+const openGraphEditor = async (mito: FrameLocator, page: Page) => {
+  await clickButtonAndAwaitResponse(page, mito, { name: 'Graph' })
+  await expect(mito.locator('.g-gtitle')).toBeVisible();
+}
+
+const changeChartType = async (mito: FrameLocator, page: Page, chartType: string, chartSubType?: string, nth?: number) => {
```
aarondr77, line 58: It was confusing to me reading this utility when I would want to use `nth` / what it is. 

I think we use it if you are looking at a grouped bar chart to select between the `horizontal` and `vertical` ones. 

Maybe instead we could set the argument to `orientation: 'horizontal' | 'vertical' and then inside of the function we can convert those strings to 0 or 1 to be used in the nth function?

That would be easier for us to consume going forward probably.

tests/streamlit_ui_tests/graph.spec.ts
```diff
@@ -44,13 +44,41 @@ const testDeleteTitleThroughContextMenu = async (page, selector) => {
...
 
+const addColumnToAxis = async (mito: FrameLocator, page: Page, axis: 'X' | 'Y', columnName: string) => {
+  await mito.locator('.spacing-row', { hasText: `${axis} axis` }).locator('.mito-dropdown-button').click();
+  await mito.locator('.mito-dropdown-item', { hasText: columnName }).click();
+  await awaitResponse(page);
+}
+
+const openGraphEditor = async (mito: FrameLocator, page: Page) => {
+  await clickButtonAndAwaitResponse(page, mito, { name: 'Graph' })
+  await expect(mito.locator('.g-gtitle')).toBeVisible();
```
aarondr77, line 55: This would only pass if the before calling `openGraphEditor` you have an entire column selected, right? Because otherwise we won't generate a graph and the g-gittle won't be visible. 

Maybe we should just remove the check and add an `await awaitResponse(page);`

tests/streamlit_ui_tests/graph.spec.ts
```diff
@@ -44,13 +44,41 @@ const testDeleteTitleThroughContextMenu = async (page, selector) => {
...
+    await mito.locator('.mito-dropdown-item-vertical').getByText(chartSubType, { exact: true }).nth(nth ?? 0).click();
+  } else {
+    await mito.getByText(chartType).click();
+  }
+  await awaitResponse(page);
+}
+
+const addChartElement = async (mito: FrameLocator, page: Page, chartElement: string, chartSubElement?: string) => {
+  await mito.getByRole('button', { name: '▾ Add Chart Element' }).click();
+  await mito.getByRole('button', { name: chartElement }).hover();
```
aarondr77, line 71: ooo nice. Hover then click!

tests/streamlit_ui_tests/graph.spec.ts
```diff
@@ -83,39 +111,27 @@ test.describe('Graph Functionality', () => {
...
 
-    await clickButtonAndAwaitResponse(page, mito, { name: 'Graph' })
-    await expect(mito.getByText('Column1 bar chart')).toBeVisible();
-
-    await mito.getByRole('button', { name: '▾ Change Chart Type' }).click();
-    await mito.getByText('Line').hover();
-    await clickButtonAndAwaitResponse(page, mito, { name: 'Linear' })
+    await openGraphEditor(mito, page);
 
+    await changeChartType(mito, page, 'Line', 'Linear');
```
aarondr77, line 116: wow so nice to have this utility

tests/streamlit_ui_tests/graph.spec.ts
```diff
@@ -167,41 +173,56 @@ test.describe('Graph Functionality', () => {
...
+    // Change the aggregation function
     await mito.locator('.mito-graph-configuration-container').getByText('count').click();
     await mito.locator('.mito-dropdown-item', { hasText: 'sum' }).click();
     await expect(mito.locator('.g-ytitle')).toHaveText('sum of None');
+
+    // Change the bin size
+    await expect(mito.locator('g.point')).toHaveCount(6);
+    await mito.locator('.mito-graph-configuration-container').locator('input[type="number"]').fill('1');
+    await awaitResponse(page);
+    await expect(mito.locator('g.point')).toHaveCount(4);
```
aarondr77, line 191: I think some comments here to describe how we know that there should be a count of 6 and 4 would be useful

tests/streamlit_ui_tests/graph.spec.ts
```diff
@@ -514,12 +485,97 @@ test.describe('Graph Functionality', () => {
...
+    await openGraphEditor(mito, page);
+
+    await mito.locator('.footer').getByText('test' ).click();
+    await mito.locator('.tab', { hasText: 'test'}).click({ button: 'right' });
+    await mito.locator('.mito-dropdown-item', { hasText: 'Delete' }).click();
+    await awaitResponse(page);
+    await expect(mito.getByText('Delete Sheet and Dependant Graphs')).toBeVisible();
+    await mito.getByText('Delete Sheet and Graphs').click();
+    await awaitResponse(page);
+    await expect(mito.getByRole('button', { name: 'Import Files' })).toBeVisible();
```
aarondr77, line 530: Maybe we should check that there is not a graph tab in the footer? Here, we are only really checking that we have an empty tab

tests/streamlit_ui_tests/graph.spec.ts
```diff
@@ -514,12 +485,97 @@ test.describe('Graph Functionality', () => {
...
+  //   await awaitResponse(page);
+  //   await expect(mito.locator('.plotly-graph-div')).toHaveCSS('width', '712px');
+
+  //   // Change the window size and check that the graph has been updated
+  //   await page.setViewportSize({ width: 1000, height: 800 });
+  //   await awaitResponse(page);
+  //   await expect(mito.locator('.plotly-graph-div')).toHaveCSS('width', '512px');
+  // });
+
+  test('Delete a column that is being used in a graph', async ({ page }) => {
```
aarondr77, line 549: Great test!

aarondr77 (commented):
Okay nice tests here! You're testing a lot of great edge case behavior that its going to be nice to not have to test manually in future releases! woohoo I'm excited about that.

tests/streamlit_ui_tests/graph.spec.ts
marthacryan: Ah yeah that's true, that's the only case that they would have the same `chartSubType` value.

tests/streamlit_ui_tests/graph.spec.ts
marthacryan: I tried that but the way that we open the graph displays stuff and doesn't include a visual loading indicator, so it registers as finished before it actually loads the graph. This causes inconsistent behavior in the tests. You're right that a column needs to be selected but I think given that the tests are generally doing that it's ok. I could put it in a wrapper with opening a specific csv so that it's guaranteed that a column is selected?

aarondr77: Okay makes sense. Want to just add this as a comment in the utility so if we ever run into an issue with it we have this context to help us debug?

PR #1226, opened 2024-02-28
Test infrastructure improvements

naterush:
# Description

- tests: upload test artifacts seperately so we can actually look at them
- 

# Testing

# Documentation


tests/streamlit_ui_tests/utils.ts
```diff
@@ -48,19 +48,17 @@ export const getMitoFrameWithTypeCSV = async (page: Page): Promise<FrameLocator>
...
-        test will continue running before the response is received, causing tests to fail.
-    */
-    await expect(page.getByText("Processing")).toHaveCount(0) 
-    await expect(page.getByText("Running")).toHaveCount(0) 
+
+    // Check if the message "Running" is visible, and wait if it is
+    if ((await page.locator('text=Running').count()) === 0) {
+        // Wait at least 25 ms for the message to send, as there is 
+        // a 25ms delay in the message sending in streamlit. We actually
+        // wait 500ms to be safe -- which is a bit of a cost, but
```
aarondr77, line 56: I think this is a good call. The biggest slow down by far is flakey tests, so I would rather wait a little bit extra each awaitResponse to avoid a bunch of reruns due to flakey tests

tests/streamlit_ui_tests/utils.ts
```diff
@@ -48,19 +48,17 @@ export const getMitoFrameWithTypeCSV = async (page: Page): Promise<FrameLocator>
...
-        2. Streamlit to finish executing the message -- the text "Running" not being visible in the Streamlit 
-        loading indicator in the top right of the streamlit app
-
-        Sometimes the loading indicator appears before the Running message, so only relying on "Running" means the 
-        test will continue running before the response is received, causing tests to fail.
-    */
-    await expect(page.getByText("Processing")).toHaveCount(0) 
-    await expect(page.getByText("Running")).toHaveCount(0) 
+
+    // Check if the message "Running" is visible, and wait if it is
```
aarondr77, line 52: This should say "and wait if it is not" right?

tests/streamlit_ui_tests/utils.ts
aarondr77: Want to just add this justification to the end of this comment in the code

tests/streamlit_ui_tests/toolbar/home.spec.ts
```diff
@@ -85,10 +85,11 @@ test.describe('Home Tab Buttons', () => {
   test('Delete Column', async ({ page }) => {
     const mito = await getMitoFrameWithTestCSV(page);
 
-    await mito.getByTitle('Column1').click();
+    await mito.getByTitle('Column2').click();
```
aarondr77, line 88: Why does moving it from Column1 to Column2 fix the test?

aarondr77 (commented):
Tiny comment cleanup and then LGTM

tests/streamlit_ui_tests/toolbar/home.spec.ts
marthacryan: Honestly I don't know why he changed this hahah

PR #1227, opened 2024-02-29
Separate tests in different CI

marthacryan:


.github/workflows/test-mitosheet-frontend.yml
```diff
@@ -119,16 +121,30 @@ jobs:
...
+        project: 
+          - test-name: 'chromium'
+            install-name: 'chromium'
+          - test-name: 'firefox'
+            install-name: 'firefox'
+          - test-name: 'webkit'
+            install-name: 'webkit'
+          - test-name: 'Google Chrome'
+            install-name: 'chrome'
+        testfiles: ['context_menus', 'features', 'toolbar']
```
aarondr77, line 133: Right now there are tests that would not get run as a result of this -- `set_column_formula_tests.spec.ts` because its not in the features folder (my bad on that!)

It seems like the vast majority of tests are going to be in the `features` test, so maybe we will eventually want to change the breakdown, but looks fine for now.

.github/workflows/test-mitosheet-frontend.yml
```diff
@@ -119,16 +121,30 @@ jobs:
...
         python-version: ['3.10']
-        project: ['chromium', 'firefox', 'webkit', 'Google Chrome', 'Microsoft Edge']
+        project: 
+          - test-name: 'chromium'
+            install-name: 'chromium'
+          - test-name: 'firefox'
+            install-name: 'firefox'
+          - test-name: 'webkit'
+            install-name: 'webkit'
+          - test-name: 'Google Chrome'
```
aarondr77, line 131: How come we don't have Miscrosoft Edge anymore?

tests/mac-setup.sh
```diff
@@ -7,7 +7,12 @@ pip install -r requirements.txt;
 
 # Install necessary node packages
 npm install
-npx playwright install chromium webkit firefox chrome
+if [ $# -eq 0 ]
```
aarondr77, line 10: Can you add a comment explaining what this is and why we're doing it

tests/windows-setup.bat
```diff
@@ -9,7 +9,8 @@ pip install -r requirements.txt
 
 REM Install Node.js dependencies and Playwright browsers
 npm install
-npx playwright install chromium firefox chrome msedge
+if [%1]==[] npx playwright install chromium firefox chrome msedge else npx playwright install %1
```
aarondr77, line 12: Let's add the comment that explains what / why here too

aarondr77 (commented):
Nice work! Just a little bit of cleanup and then LGTM. Looks like these improvements about 1/2 the amount of time it takes for the CI to run?

.github/workflows/test-mitosheet-frontend.yml
marthacryan: Agreed

.github/workflows/test-mitosheet-frontend.yml
marthacryan: Great catch

PR #1229, opened 2024-03-01
Set column formula tests 2

aarondr77:
# Description

- Adds tests for individual cell edits 

# Testing

# Documentation


tests/streamlit_ui_tests/set_column_formula_tests.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+
+test('Set constant formula to existing column using cell editor', async ({ page }) => {
+    const columnHeader = 'Column1';
+
+    const mito = await getMitoFrameWithTypeCSV(page);
+    await setFormulaUsingCellEditor(page, mito, columnHeader, '=5');
+
+    // Check that the value in columnHeader are '5'
+    const cellValues = await getValuesInColumn(mito, columnHeader);
+    expect(cellValues).toEqual(Array(cellValues.length).fill('5'));
```
marthacryan, line 47: Wait I'm confused about this one. What exactly is this line doing?

tests/streamlit_ui_tests/set_column_formula_tests.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+
+    await mito.locator('#cell-editor-input').press('Enter');
+    await awaitResponse(page);
+
+    const cellValues = await getValuesInColumn(mito, columnHeader);
+    expect(cellValues).toEqual(['5', '11', '17', '23']);
+});
+
+
+test('No formula around rolling range displays [object Object]', async ({ page }) => {
```
marthacryan, line 185: So this is something that we want? Seems not super intuitive

marthacryan (commented):
Love these utils!! Just a couple comments but LGTM

tests/streamlit_ui_tests/set_column_formula_tests.spec.ts
aarondr77: This is just checking that the values in column `columnHeader` are all 5. Which they should be because I set the formula to =5.

tests/streamlit_ui_tests/set_column_formula_tests.spec.ts
aarondr77: ehh, yeah not sure if we actually do want it. But I guess we want to know if we change something that breaks it. I should probably leave that as a comment in this test so we have context if this test ever fails. Good point!

tests/streamlit_ui_tests/set_column_formula_tests.spec.ts
marthacryan: Ohhh wow ok that's a syntax that I was not aware of

tests/streamlit_ui_tests/set_column_formula_tests.spec.ts
marthacryan: Ok yeah just making sure!

PR #1240, opened 2024-03-05
Add code config tests

marthacryan:
Adds tests for:
- Add name to generated function
- Add parameter to generated function

Updates the streamlit app to also include the generated code
Updates the code config taskpane to pick up the first toggle


aarondr77 (commented):
LGTM

PR #1244, opened 2024-03-06
Add cross sheet formula test

marthacryan:
Adds tests for:
- Adding a vlookup, navigating back to the original sheet, then pressing enter
- Adding a vlookup, pressing enter from the other sheet
- Trying to use cross-sheet formulas with non-vlookup


tests/streamlit_ui_tests/grid/set_column_formula_tests.spec.ts
```diff
@@ -286,3 +288,77 @@ test('Write spreadsheet formula applied to individual cell', async ({ page }) =>
     const cellValues = await getValuesInColumn(mito, columnHeader);
     expect(cellValues).toEqual(['1', '0', '0', '0']);
 });
+
+test('Cross-sheet formula with VLOOKUP', async ({ page }) => {
```
aarondr77, line 292: One more test to that since we had a bug in this before is: Start writing a cross sheet formula using the cell editor and then once you write `=VLOOKUP(column10,` switch to the formula bar to finish the formula

aarondr77 (commented):
One more test idea and then LGTM

PR #1245, opened 2024-03-06
Reorganize tests to run faster by moving graph tests into separate CI

marthacryan:
The graph tests were taking up around half of the time in the taskpanes tests, and they were taking almost double the time of the other tests (toolbar, grid), so I moved the graph tests to be separate from the others to make the overall runtime faster.


aarondr77:
LGTM once last test passes

PR #1247, opened 2024-03-08
mitosheet: make mito height into a variable

naterush:
# Description

Allow streamlit height config.

# Testing

See test app. Test in JupyterLab and elsewhere too.

# Documentation

None.


aarondr77:
LGTM. Waiting to merge so we can decide on release process for this since it's time sensitive. 

TODO: Write docs

PR #1248, opened 2024-03-08
Better default number formatting

naterush:
# Description

1. Make float columns have better default formatting, with commas and two decimals
2. Make int columns have commas

# Testing

Export some Excels. Not a great thing to write tests for, sadly.

# Documentation

Can update language in docs.


aarondr77:
Waiting for tests to pass + decision on how to release. LGTM otherwise.

PR #1260, opened 2024-03-11
Fix multiple vlookup bug

### LINKED ISSUE #1243: Cannot use two VLOOKUPS in the same formula
1. Launch the streamlit app on our homepage
2. Duplicate the dataframe 
3. Write this formula in the duplicated dataframe 
```
=CONCAT(VLOOKUP(date1,df1!date:close, 2), VLOOKUP(date1,df1!date:volume, 3))
```
4. Get an unhelpful execution error 

Screenshot 2024-03-06 at 1 09 44 PM

marthacryan:
Fixes #1243. The issue was that the parser was only using the first index that matched a given sheet name to identify cross-sheet formulas, so the resulting code wasn't identifying any sheets after the first.

### DIFF
--- a/mitosheet/mitosheet/parser.py
+++ b/mitosheet/mitosheet/parser.py
@@ -477,13 +477,14 @@ def get_raw_parser_matches(
 
     for other_sheet_index, sheet_name in enumerate(df_names):
         if safe_contains(formula, f'{sheet_name}!', column_headers):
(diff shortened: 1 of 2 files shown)


naterush:
Excellent, LGTM when tests pass :)

naterush:
Actually, can you add a test for the example from the issue and then close the issue?

PR #1266, opened 2024-03-12
Fix bug in parsing specific index labels

### LINKED ISSUE #1230: =SUM(A:A) does not work when applying formula to single cell
1. Create a new column
2. Open the cell editor and toggle `edit entire cell` to `OFF`
3. Write a formula like =SUM(A:A)
4. Get an execution error. 

Individual cells should be able to do this!

marthacryan:
Fixes #1230. Removes addition of `.loc` call in the parsing of cell formulas with specific index labels. Not sure why that was added but it appears to not be necessary.

### DIFF
--- a/mitosheet/mitosheet/parser.py
+++ b/mitosheet/mitosheet/parser.py
@@ -1045,12 +1045,7 @@ def parse_formula(
             if is_datetime_index(df.index) and is_prev_version(get_pandas_version(), '1.0.0'):
                 index_labels = pd.to_datetime(index_labels)
 
-            if len(column_header_dependencies) > 0:
-                final_set_code = f'({code_with_functions}).loc[{get_column_header_list_as_transpiled_code(index_labels)}]' # type: ignore
-            else:
-                final_set_code = f'{code_with_functions}'
-                
-            final_code = f'{df_name}.loc[{get_column_header_list_as_transpiled_code(index_labels)}, [{transpiled_column_header}]] = {final_set_code}' # type: ignore
+            final_code = f'{df_name}.loc[{get_column_header_list_as_transpiled_code(index_labels)}, [{transpiled_column_header}]] = {code_with_functions}' # type: ignore
 
     else:
         final_code = f'{code_with_functions}'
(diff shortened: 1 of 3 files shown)


mitosheet/mitosheet/parser.py
```diff
@@ -1045,12 +1045,7 @@ def parse_formula(
             if is_datetime_index(df.index) and is_prev_version(get_pandas_version(), '1.0.0'):
                 index_labels = pd.to_datetime(index_labels)
 
-            if len(column_header_dependencies) > 0:
```
naterush, line 1048: Let's make sure we write a test with multiple column head dependencies.

mitosheet/mitosheet/parser.py
```diff
@@ -1045,12 +1045,7 @@ def parse_formula(
...
-                final_set_code = f'({code_with_functions}).loc[{get_column_header_list_as_transpiled_code(index_labels)}]' # type: ignore
```
naterush, line 1049: yeah, you're right -- this second .loc code is not needed, as pandas is smart enough to figure out index overlap. nice!

mitosheet/mitosheet/tests/step_performers/column_steps/test_set_column_formula.py
```diff
@@ -535,6 +535,12 @@ def test_set_specific_index_labels_twice():
 
     assert mito.dfs[0].equals(pd.DataFrame({'A': [1, 2, 3], 'B': [2, 0, 0]}))
 
+def test_set_specific_index_labels_to_header_header():
+    mito = create_mito_wrapper(pd.DataFrame({'A': [1, 2, 3]}))
+    mito.add_column(0, 'B')
+    mito.set_formula('=SUM(A:A)', 0, 'B', index_labels=[0])
+
+    assert mito.dfs[0].equals(pd.DataFrame({'A': [1, 2, 3], 'B': [6, 0, 0]}))
```
naterush, line 543: Let's add one test with multiple column header dependencies in the SUM formula

naterush (approved):
One additional test, then LGTM!

PR #1267, opened 2024-03-12
Fix bug switching between cell editor and formula bar

### LINKED ISSUE #614: Cannot switch to formula bar when cell editor is open
**Describe the bug**
I used to be able to switch from the floating cell editor to the formula bar by double clicking on the formula bar. Now, when I click on the formula bar with the floating cell editor open nothing happens.

### LINKED ISSUE #1254: Cannot open formula bar editor if floating cell editor is open
1. Open the floating cell editor
2. Try to switch to the formula bar. Notice that you cannot open the formula bar unless you have first close the floating cell editor

Just like in Excel, users should be able to switch from the cell editor to the formula bar. This is important because users often switch when they are writing long formulas. We don't want users to have to close the cell editor, losing their formula, in the process.

marthacryan:
When the formula bar was empty, it became not clickable because the height became 0. This sets the height manually (and then adds correct adjustment to make sure that the formula is justified correctly). This css is copied from the `.formula-bar-column-header` style. 

Fixes #614. Fixes #1254.

### DIFF
--- a/mitosheet/css/FormulaBar.css
+++ b/mitosheet/css/FormulaBar.css
@@ -21,6 +21,10 @@
     outline: none;
 
     color: var(--mito-text);
+    height: 30px;
+    display: flex;
+    justify-content: center;
+    flex-direction: column;
 }
 
 .formula-bar-vertical-line {


mitosheet/css/FormulaBar.css
```diff
@@ -21,6 +21,10 @@
     outline: none;
 
     color: var(--mito-text);
+    height: 30px;
+    display: flex;
+    justify-content: center;
+    flex-direction: column;
```
naterush, line 27: Code looks legit. Looking now.

naterush (commented):
LGTM when tests pass! Don't think this is worth a test, but if you wanted to add to the test that switches between the cell editor and formula bar... that would be reasonable!

PR #1270, opened 2024-03-13
Add better handling for datetime filters

### LINKED ISSUE #581: Filtering on datetime column gives temporary error when not typing quickly enough
When a user tries to filter a date column and types the date in manually and not quickly enough, an execution error screen pops up even  though the filter is applied properly when the users completes typing in the date correctly.

**Steps to reproduce:**

- Open any dataframe in mito that has a datetime column.
- Add a filter to that datetime column and select >= from the dropdown menu.
- Begin typing in a date starting with month and day (e.g., 10, 30)
- When you get to the year, type 20, then pause for a moment and an error will appear.
- Continue typing the year 2022 and see that the filter is applied to the column correctly.

**Expected behavior:**
- Either the error should not appear at all provided a valid year is entered within a reasonable amount of time, or the error window should disappear when the user enters a valid date so they don't have to close the error screen and/or wonder if the filter was properly applied.

**Actual behavior:**
- Execution error window appears and has to be manually closed.
- Error message below

image

### **Bonus bug within a bug :-) **
Unable to copy the text from the full traceback (hence the screenshot of the partial message).  When highlighting the text of the traceback, it appears to actually be copying some of the text in the column it was filtering on.

marthacryan:
Fixes #581. 
* Adds a try/catch statement to check if the datetime is valid, and if not just uses the pandas minimum timestamp. 
* **Note: this behavior might not be desirable, because it automatically fills in the minimum timestamp, which could be confusing / incorrect.**


mitosheet/mitosheet/code_chunks/step_performers/filter_code_chunk.py
```diff
@@ -189,6 +191,12 @@ def get_single_filter_string(
     condition = filter_["condition"]
     value = filter_["value"]
 
+    if is_datetime_dtype(column_dtype):
+        try: 
+            pd.to_datetime(value)
+        except:
+            value = pd.Timestamp.min + pd.Timedelta(seconds=1)
```
naterush, line 198: Can you document this explaining the bug and the reason for it?

mitosheet/mitosheet/code_chunks/step_performers/pivot_code_chunk.py
```diff
@@ -155,7 +155,11 @@ def get_code(self) -> Tuple[List[str], List[str]]:
             for column_id, value in self.values_column_ids_map.items()
         }
         pivot_filters: List[ColumnHeaderWithFilter] = [
-            {'column_header': self.prev_state.column_ids.get_column_header_by_id(self.sheet_index, pf['column_id']), 'filter': pf['filter']}
+            {
+                'column_header': self.prev_state.column_ids.get_column_header_by_id(self.sheet_index, pf['column_id']), 
+                'column_dtype': str(self.prev_state.dfs[self.sheet_index][pf['column_id']].dtype),
```
naterush, line 160: What happens if the column changes dtype after this is created?

mitosheet/mitosheet/tests/step_performers/test_filter.py
```diff
@@ -322,6 +322,14 @@
...
         ),
     ),
+    (
+        pd.DataFrame(
+            data={"A": pd.to_datetime(pd.Series(data=["12-2-2020", "12-3-2020"]))}
+        ),
+        FC_DATETIME_LESS_THAN_OR_EQUAL,
+        "12-2-0020",
+        pd.DataFrame(data={"A": pd.to_datetime(pd.Series(data=[], dtype='float64'))}),
+    ),
```
naterush, line 332: Little comment to explain this test would be nice

naterush (commented):
Minor documentation, then LGTM

Issue #1274, opened 2024-03-14
Conditional formatting with no color text set breaks export formatting


naterush:
**Describe the bug**
1. Add a conditional format, but only set the background color -- not the text color
2. Export the formatting to an Excel sheet. Notice it generates this nonsense code:
```
add_formatting_to_excel_sheet(writer, "df", df, 
        conditional_formats=[
            {'columns': ['A'], 'filters': [{'condition': 'not_empty', 'value': ''}], 'font_color': 'var(--mito-text)', 'background_color': '#000000'}
        ]
    )
```

Can we make it so if the color is a variable, it doesn't export? Let's also add a backend test for this (making sure to handle both font color and background color properly).

PR #1276, opened 2024-03-14
mitosheet: fix bug with not supplying background_color

aarondr77:
# Description

Make sure the app works if you only have font_color set

# Testing

See the test

# Documentation

No -- docs already say this is allowed


naterush (commented):
Actually, this is an issue from our interface as well -- see issue here: https://github.com/mito-ds/mito/issues/1274

Can you ensure this is fixed as well, and then write a test that verifies the export formatting code as well? Then LGTM

aarondr77:
Sweet. Fixing the above in a separate PR, since they are unrelated bugs.

PR #1278, opened 2024-03-14
Fix scroll bar bug in windows

marthacryan:
Before:
image

After:
image


naterush:
I don't think this is a perfect fix, but it's an improvement. LGTM, add a screenshot of this, and keep the issue open I think!

PR #1279, opened 2024-03-14
mitosheet: handle unset colors in conditional formatting

aarondr77:
# Description

Fixes https://github.com/mito-ds/mito/issues/1274

# Testing

# Documentation

### DIFF
--- a/mitosheet/src/mito/pro/taskpanes/ConditionalFormatting/ConditionalFormattingTaskpane.tsx
+++ b/mitosheet/src/mito/pro/taskpanes/ConditionalFormatting/ConditionalFormattingTaskpane.tsx
@@ -125,6 +125,19 @@ const ConditionalFormattingTaskpane = (props: ConditionalFormattingTaskpaneProps
     )
 
     const updateDataframeFormatParams = (newParams: RecursivePartial<DataframeFormat>): void => {
+        // If the user does not change the color or background color, we want to treat them as undefined
+        // so it doesn't change the data in the Mito Spreadsheet or the created Styler or Excel file
+        newParams.conditional_formats = newParams?.conditional_formats?.map(conditionalFormat => {
+            const newConditionalFormat = {...conditionalFormat};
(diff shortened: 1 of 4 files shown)


mitosheet/src/mito/pro/taskpanes/SetDataframeFormat/SetDataframeFormatTaskpane.tsx
```diff
@@ -80,6 +80,7 @@ const SetDataframeFormatTaskpane = (props: SetDataframeFormatTaskpaneProps): JSX
     }
 
     const updateDataframeFormatParams = (newParams: RecursivePartial<DataframeFormat>): void => {
+        console.log(newParams)
```
aarondr77, line 83: remove

mitosheet/src/mito/pro/taskpanes/ConditionalFormatting/ConditionalFormattingTaskpane.tsx
```diff
@@ -125,6 +125,19 @@ const ConditionalFormattingTaskpane = (props: ConditionalFormattingTaskpaneProps
...
+        newParams.conditional_formats = newParams?.conditional_formats?.map(conditionalFormat => {
+            const newConditionalFormat = {...conditionalFormat};
+            if (newConditionalFormat.color === "var(--mito-text)") {
+                newConditionalFormat.color = undefined;
+            }
+            if (newConditionalFormat.backgroundColor === "var(--mito-background-off)") {
+                newConditionalFormat.backgroundColor = undefined;
+            }
+            return newConditionalFormat;
+        })
```
naterush, line 139: This is a weird way to do this. Why not just default them to undefined when we create them?

naterush (commented):
I think this approach could be more clean and maintainable, unless I'm misunderstanding something.

naterush:
This actually isn't a solution

PR #1283, opened 2024-03-14
Fix bug when opening graph editor then pressing undo

### LINKED ISSUE #1092: Graph creates 2 steps on opening graph editor
**Describe the bug**
Graph creates 2 steps on opening graph editor. 

**To Reproduce**
1. Click on the graph toolbar button
2. Wait approx. 2 seconds
3. Press undo
4. Notice that the graph editor is still open
5. Press undo again
6. The graph editor closes. 

**Expected behavior**
Should only require one undo to close the graph editor.

marthacryan:
Fixes #1092. The issue describes a different bug (which I assume was fixed somewhere?) but in trying to reproduce it I caught a different bug. The bug I saw was that pressing undo would get rid of the graph, but not close the graph editor. There would be no graph, but because the graph editor is designed to create a new graph when it sees the graph editor is open, it would just create a new graph. I chose to solve this by adding a `useEffect` that would check in the `Mito` component when the `graphDataArray` changed, and if the graphDataArray was empty but the graph taskpane was open, it just closes that taskpane.

### DIFF
--- a/mitosheet/src/mito/Mito.tsx
+++ b/mitosheet/src/mito/Mito.tsx
@@ -335,6 +335,23 @@ export const Mito = (props: MitoProps): JSX.Element => {
         loadPlotly()
     }, [])
 
+    /**
+     * This useEffect handles when the user creates a graph, then presses undo.
+     */
+    useEffect(() => {
+        if (uiState.currOpenTaskpane.type === TaskpaneType.GRAPH &&
+            analysisData.graphDataArray.length === 0) {
+            setUIState(prevUIState => {
+                return {
+                    ...prevUIState,
+                    currOpenTaskpane: {type: TaskpaneType.NONE},
+                    selectedTabType: 'data',
+                    selectedSheetIndex: 0
+                }
+            })
+        }
+    }, [analysisData.graphDataArray])
+
     /* 
         When the number of sheets increases, we make sure
         that the last sheet is highlighted. If it decreases,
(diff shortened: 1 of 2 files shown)


mitosheet/src/mito/Mito.tsx
```diff
@@ -335,6 +335,23 @@ export const Mito = (props: MitoProps): JSX.Element => {
...
+            setUIState(prevUIState => {
+                return {
+                    ...prevUIState,
+                    currOpenTaskpane: {type: TaskpaneType.NONE},
+                    selectedTabType: 'data',
+                    selectedSheetIndex: 0
+                }
+            })
+        }
+    }, [analysisData.graphDataArray])
```
naterush, line 353: Can we make this run when the array changes length? Feels like it might avoid potential race conditons

mitosheet/src/mito/Mito.tsx
```diff
@@ -335,6 +335,23 @@ export const Mito = (props: MitoProps): JSX.Element => {
         loadPlotly()
     }, [])
 
+    /**
+     * This useEffect handles when the user creates a graph, then presses undo.
+     */
+    useEffect(() => {
+        if (uiState.currOpenTaskpane.type === TaskpaneType.GRAPH &&
+            analysisData.graphDataArray.length === 0) {
```
naterush, line 343: How does this not get triggered when the graph taskpane I open but the graph has not been created yet?

naterush (commented):
Some questions about how this doesn't trigger race conditions.

mitosheet/src/mito/Mito.tsx
marthacryan: It'd be triggered once the `graphDataArray` changes, but at that point the length of the graphDataArray is actually >0.

PR #1289, opened 2024-03-15
mitosheet: set default_apply_formula_to_column

aarondr77:
# Description

Let Streamlit app developer set the default cell editor to individual cell. 

# Testing

See tests + existing test "'Edit individual cell only applied to edited cell and preserves formula and toggle on next open of cell editor'"

# Documentation

Yes -- we need to add documentation to the streamlit api

### DIFF
--- a/mitosheet/mitosheet/mito_backend.py
+++ b/mitosheet/mitosheet/mito_backend.py
@@ -12,9 +12,8 @@ import os
 import re
 import time
 from sysconfig import get_python_version
-from typing import Any, Dict, List, Optional, Union, Callable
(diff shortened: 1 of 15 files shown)


naterush (commented):
Code looks good.

naterush:
Actually, realized that this is a public API -- and it's actually really confusingly named. `default_apply_formula_to_column` and then a boolean variable is almost optimized to confuse the user. 

We don't have to refactor all the internals, but can we change the public api to: `default_editing_mode=Literal['cell', 'column']`` I think this is way more clear!

naterush:
If you edit through the formula bar, it doesn't pick up the default editing mode. Is this intended?

PR #1302, opened 2024-04-09
Python cheat sheet form

aarondr77:
# Description

Adds a popup to the excel to python glossary pages that offers a free Pandas Cheat Sheet if the user adds shares their email and company. 

# Testing

For testing, it is configured to appear 5 seconds after page load on each page load. If we're happy with it, I'll change the settings so each user only gets it once per session

# Documentation

No.

### DIFF
--- a/trymito.io/pages/excel-to-python/[...slug].tsx
+++ b/trymito.io/pages/excel-to-python/[...slug].tsx
@@ -122,6 +122,12 @@ const ExcelToPythonGlossaryPage = (props: {pageContent: PageContent, glossaryPag
       </Head>
       <Header />
 
+      {/* 
+        A popup typeform that asks the user if they want to download a Pandas Cheat Sheet.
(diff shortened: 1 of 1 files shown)


naterush:
@aarondr77 how do I trigger this? Going to the excel to python pages on the preview don't cause this to pop up

aarondr77:
Hmm it's working for me. Try going in incognito mode, clicking on one of those pages and then waiting 5 seconds. It should pop up

naterush:
https://github.com/mito-ds/mito/assets/21002431/843f4fdd-3120-435f-b9b9-25771494158c


Safari, private browsing. Having the same experience in Chrome as well - waiting 20+ seconds

aarondr77:
It's working for Jake and I. Here is what it looks like. 

https://github.com/mito-ds/mito/assets/18709905/b24d5205-ab73-4984-b66c-85f51f9b5101

PR #1306, opened 2024-06-10
Improve logging

aarondr77:
# Description

Improves enterprise logging: 

1. Creating a separate error code for Mito sheet crashes. These are the most disruptive errors that we should prioritize fixing. 
2. Adding the error name as a parameter so you can easily differentiate between them in your dashboard. For example: `add_column_failed` or `set_column_formula_failed`. 

# Testing

See the new tests

# Documentation

We need to update the enterprise logging docs


naterush:
LGTM when Python tests are passing

PR #1326, opened 2024-09-06
Mito ai variable aware

aarondr77:
# Description

Creates a new extension called the Variable Manager that is responsible for keeping track of the current state of all defined variables. This current state is then used by the Mito Ai extension to write code that works in the context of your analysis, using your existing variables. 

# Testing

Create some variables and then ask the AI to generate code. Don't specify the exact variables to use and let the AI figure it out. 

Test tomato sure that with lots of various types of variables the code works properly. 

# Documentation

No.


mito-ai/src/Extensions/AiChat/Chat.tsx
```diff
@@ -83,16 +83,19 @@ const Chat: React.FC<IChatProps> = ({notebookTracker, rendermime}) => {
...
         const updatedManager = new ChatHistoryManager(chatHistoryManager.getHistory());
-        updatedManager.addUserMessage(finalInput, activeCellCode)
+        updatedManager.addUserMessage(finalInput, activeCellCode, variables)
 
         setInput('');
         setLoadingAIResponse(true)
 
         try {
+
+            console.log(updatedManager.getAIOptimizedHistory()[updatedManager.getAIOptimizedHistory().length - 1])
```
aarondr77, line 98: remove

PR #1330, opened 2024-09-11
Mitosheet default df renderer

aarondr77:
# Description

# Testing

# Documentation


mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+            script.text = props.jsCode;
+            document.body.appendChild(script);
+            console.log("JavaScript code injected and executed.");
+
+            // Clean up the script after execution
+            return () => {
+                document.body.removeChild(script);
+            };
+        }
+    }, [props.jsCode]);
```
aarondr77, line 29: Add better documentation for this

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+        super();
+        this._app = app;
+        this.addClass(CLASS_NAME);
+        this._notebookTracker = notebookTracker;
+    }
+
+    async renderModel(model: IRenderMime.IMimeModel): Promise<void> {
+        this.node.innerHTML = '';  // Clear previous content
+        const originalRawData = model.data['text/html']?.toString();
+        const isDataframeOutput = originalRawData?.includes('class="dataframe"');
```
aarondr77, line 52: Build a better identifier for this

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+
+        const notebook = this._notebookTracker.currentWidget?.content;
+        const cells = notebook?.model?.cells;
+        const activeCellIndex = notebook?.activeCellIndex
+
+        let dataframeVariableName = undefined;
+        if (activeCellIndex) {
+            const previousCell = getCellAtIndex(cells, activeCellIndex - 1)
+            dataframeVariableName = getLastNonEmptyLine(getCellText(previousCell))
```
aarondr77, line 62: Test more cases: things like having two dataframes as the output, non dataframes as the output, etc. 

Add UI tests for these

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+
+        if (isDataframeOutput) {
+            console.log('Dataframe detected!!!!');
```
aarondr77, line 67: remove

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+            // Define the Python code to run
+            const pythonCode = `mitosheet.sheet(${dataframeVariableName || ''} )`;
```
aarondr77, line 70: Add `import mitosheet`

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+            try {
+                const notebookPanel = this._notebookTracker.currentWidget;
+                const kernel = notebookPanel?.context.sessionContext.session?.kernel;
+
+
+                if (!kernel) {
+                    console.error('No active kernel found.');
```
aarondr77, line 78: Add a better message that is useful for debugging in the console

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+                    return;
+                }
+
+                console.log("GOT A KERNEL")
```
aarondr77, line 82: remove

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+                // Execute the Python code
+                const future = kernel.requestExecute({ code: pythonCode });
+
+                future.onIOPub = (msg: any) => {
+                    if (msg.header.msg_type === 'execute_result' || msg.header.msg_type === 'display_data') {
```
aarondr77, line 88: Figure out the correct msg_type. There should probably only be one that I'm looking for. 

Maybe I can even set a custom field in the message so I can case on that?

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+                    console.error('No active kernel found.');
+                    return;
+                }
+
+                console.log("GOT A KERNEL")
+
+                // Execute the Python code
+                const future = kernel.requestExecute({ code: pythonCode });
+
+                future.onIOPub = (msg: any) => {
```
aarondr77, line 87: Give this a better type

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+                    if (msg.header.msg_type === 'execute_result' || msg.header.msg_type === 'display_data') {
+                        const htmlOutput = msg.content.data['text/html'];
+                        if (htmlOutput) {
+                            console.log('Received HTML output:', htmlOutput);
+                            // Extract the javascript code so we can execute it when we render the output.
+                            // Remember, that the javascript code is actually what creates the sheet interface!
+                            const scriptMatch = htmlOutput.match(/<script[^>]*>([\s\S]*?)<\/script>/);
+                            const jsCode = scriptMatch ? scriptMatch[1] : '';
```
aarondr77, line 95: Explore what the other elements in the scriptMatch are. Leave a comment here to explain

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+                            // Create a React widget to display the HTML output
+                            const reactWidget = ReactWidget.create(
```
aarondr77, line 98: Is creating a widget the right thing to be doing? Maybe it should just be a react component?

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+                            );
+
+                            // Clear previous content and append the React widget
+                            this.node.innerHTML = '';
+                            Widget.attach(reactWidget, this.node);
+                        }
+                    }
+                };
+
+                future.done.then(() => {
```
aarondr77, line 109: Do I need this future.done clause at all?

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+            } catch (error) {
+                console.error('Error executing Python code:', error);
+                this.node.innerHTML = `<div style="color: red;">Error rendering sheet</div>`;
+            }
+
+        } else {
+            console.log("Non-dataframe content detected");
+            console.log(this._app);
+            this.node.appendChild(document.createTextNode("Non-dataframe content"));
```
aarondr77, line 121: If it is not a dataframe, then just render the regular component

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+                    }
+                };
+
+                future.done.then(() => {
+                    console.log('Python code executed successfully.');
+                })
+
+            } catch (error) {
+                console.error('Error executing Python code:', error);
+                this.node.innerHTML = `<div style="color: red;">Error rendering sheet</div>`;
```
aarondr77, line 115: If it errors, just fall back to rendering the default dataframe ouput

mitosheet/src/DataFrameMimeRenderer.tsx
```diff
@@ -0,0 +1,132 @@
...
+            console.log(this._app);
+            this.node.appendChild(document.createTextNode("Non-dataframe content"));
+        }
+
+        return Promise.resolve();
+    }
+
+    dispose(): void {
+        super.dispose();
+    }
```
aarondr77, line 129: Do I need to implement this or is it automatically supported?

mitosheet/src/index.tsx
```diff
@@ -1,11 +1,13 @@
...
+
+
 
 /**
  * By exporting this plugin directly, JLab 4 can get access 
  * directly to the plugin and activate it. Since Notebook 7 is 
  * built on top of JLab 4, this also activates the plugin for Notebook 7.
+ * 
+ * 
+ * Apparently this file is completely unused?? Understand that.
```
aarondr77, line 12: Investigate this by looking for the main entry point to the extension in tsconfig and setup.py

mitosheet/src/plugin.tsx
```diff
@@ -411,12 +423,28 @@ function activateMitosheetExtension(
         }
     });
 
+    // Add a custom renderer for the stderr output
+
+    const dataframeMimeType = 'text/html'
```
aarondr77, line 428: Give this a better name and explain why we are looking at the text/html mime type.

aarondr77:
Make code gen work

aarondr77:
Don't register analysis.

aarondr77:
Don't have email popup

aarondr77:
Question: Each time we add a mitosheet.sheet() call the notebook size grows by about 2MB, but adding Mito as a default dataframe output only increases the size of the saved notebook by 2KB. 

I think this is because the mime renderer doesn't actually get saved to the notebook (at least right now). But if I were to look at the size of the notebook while being used it would still grow 2MB per mito spreadsheet default dataframe ouput

mitosheet/src/DataFrameMimeRenderer.tsx
aarondr77: I think this is okay for now

aarondr77:
### Approaches tried using the mime render, documenting for future reference


The challenging part is figuring out which dataframe to display in the mitosheet. To figure this out,
    we need to find the code cell that triggered this dataframe render and get the dataframe on its last line. 

    Finding the code cell is challenging however. Below describes a few options we tried and why they don't work. 

    1. Using the activeCellIndex: We cannot use the activeCellIndex to identify the cell ID because when running 
    a bunch of cells in a row (for example, using run all cells) or when the code cell takes a few seconds to execute,
    the active cell in the notebook tracker updates before we're able to save it. As a result, we end up thinking the 
    code cell that triggered the dataframe render is at the bottom of the notebook. 

    2. Using the execution count: We cannot use the execution count because the execution count will not update
    until the mime render is created. When we run the code cell `df`, that cell is responsible for creating the mime renderer. 
    As a result, when we search the cells for the execution count, of ie: 3, the closest execution count that we get is 2. 
    
    Instead of using those approaches, we instead use the dom to find the corresponding code cell ID. This works as follows: 

    1. Render the default renderer so that the we have a DOM element to start with. 
    2. Traverse up to find the Code Cell that triggered the dataframe render (the first code cell we find)
    3. Get the code cell ID from the code cell's model. 
    4. Use the cell ID to find the input cell and read the dataframe name from it. 
    
### Why using Mime Renderes did not work

There is still a race condition bug where if code cell 1 creates a dataframe renderer, and code cell 2 edits the dataframe, the mitosheet output will show the dataframe state after code cell 2 has run, instead of the state of the dataframe at code cell 1. 

This occurs because in order to create the mitosheet, we need to execute the mitosheet.sheet() function. To do so, we had to send a new kernel message from the mimerender with the code mitosheet.sheet(df). However, becasue the kernel message queue might have had additional messages already queued that edited the df, by the time the mitosheet was rendered, it might have displayed a dataframe that reflected future code cell edits instead of the current state of the dataframe at the time the code cell with the hanging df was executed. This is not what we want.

mitosheet/Test Notebook.ipynb
```diff
@@ -0,0 +1,85 @@
+{
```
aarondr77, line 1: Turn this into a test

mitosheet/src/jupyter/extensionUtils.tsx
```diff
@@ -29,12 +29,82 @@ export function getCellAtIndex(cells: CellList | undefined, index: number): ICel
     return cell
 }
 
+export function getCellIndexByID(cells: CellList | undefined, cellID: string | undefined): number | undefined {
```
aarondr77, line 32: Delete this. I don't think we use this function.

mitosheet/src/jupyter/extensionUtils.tsx
```diff
@@ -29,12 +29,82 @@ export function getCellAtIndex(cells: CellList | undefined, index: number): ICel
...
+        return undefined;
+    }
+
+    // In order to get the cell index, we need to iterate over the cells and call the `get` method
+    // to see the cells in order. Otherwise, the cells are returned in a random order.
+    for (let i = 0; i < cells.length; i++) {
+        const cell = cells.get(i)
+        if (cell.type === 'code') {
+            const nonTypeSafeCell = cell as any
+            const executionCountEntry = nonTypeSafeCell.sharedModel.ymodel._map.get('execution_count')['content']['arr'][0];
```
aarondr77, line 52: Try to find a better way of doing this that doesn't access private variables. Is there a getter for this?

mitosheet/src/jupyter/extensionUtils.tsx
```diff
@@ -29,12 +29,82 @@ export function getCellAtIndex(cells: CellList | undefined, index: number): ICel
...
+            if (executionCountEntry === executionCount) {
+                return i
+            }
+        }
+    }
+
+    return undefined
+}
+
+export function getCellByExecutionCount(cells: CellList | undefined, executionCount: number | undefined): ICellModel | undefined {
```
aarondr77, line 62: Do we use this function?

mitosheet/src/jupyter/extensionUtils.tsx
```diff
@@ -29,12 +29,82 @@ export function getCellAtIndex(cells: CellList | undefined, index: number): ICel
...
+    if (cells == undefined || executionCount == undefined) {
+        return undefined;
+    }
+
+    return Array.from(cells).find((cell: ICellModel) => {
+        if (cell.type === 'code') {
+            const nonTypeSafeCell = cell as any
+            const executionCountEntry = nonTypeSafeCell.sharedModel.ymodel._map.get('execution_count');
```
aarondr77, line 70: again, find better approach

mitosheet/src/jupyter/jupyterUtils.tsx
```diff
@@ -47,24 +47,33 @@ export const overwriteAnalysisToReplayToMitosheetCall = (oldAnalysisName: string
...
-            analysisName: analysisName,
-            code: code,
-            telemetryEnabled: telemetryEnabled,
-            publicInterfaceVersion: publicInterfaceVersion,
-            oldCode: oldCode,
-            overwriteIfUserEditedCode: overwriteIfUserEditedCode,
-            triggerUserEditedCodeDialog: triggerUserEditedCodeDialog,
-        });
+        if (inputCellExecutionCount) {
+            window.commands?.execute('mitosheet:write-generated-code-cell-by-id', {
```
aarondr77, line 53: rename -> `by execution number`

mitosheet/src/jupyter/jupyterUtils.tsx
```diff
@@ -77,12 +86,18 @@ export const writeCodeSnippetCell = (analysisName: string, code: string): void =
...
 
-export const getArgs = (analysisToReplayName: string | undefined): Promise<string[]> => {
+export const getArgs = (analysisToReplayName: string | undefined, inputCellExecutionCount: number | undefined): Promise<string[]> => {
     return new Promise((resolve) => {
         if (isInJupyterLabOrNotebook()) {
-            window.commands?.execute('mitosheet:get-args', {analysisToReplayName: analysisToReplayName}).then(async (args: string[]) => {
-                return resolve(args);
-            })
+            if (inputCellExecutionCount) {
+                window.commands?.execute('mitosheet:get-args-by-id', {inputCellExecutionCount: inputCellExecutionCount}).then(async (args: string[]) => {
```
aarondr77, line 93: by execution-number

mitosheet/src/plugin.tsx
```diff
@@ -198,6 +214,76 @@ function activateMitosheetExtension(
...
+            const newCode = getCodeString(analysisName, codeLines, telemetryEnabled, publicInterfaceVersion, true);
+            const notebook = notebookTracker.currentWidget?.content;
+            const cells = notebook?.model?.cells;
+
+            if (inputCellExecutionCount === undefined || notebook === undefined || cells === undefined) {
+                return;
+            }
+
+            const mimeRenderInputCellIndex = getCellIndexByExecutionCount(cells, inputCellExecutionCount);
+            console.log('mimeRenderInputCellIndex', mimeRenderInputCellIndex)
```
aarondr77, line 239: remove these consoles

mitosheet/src/plugin.tsx
```diff
@@ -198,6 +214,76 @@ function activateMitosheetExtension(
...
+                // If the code cell that created the mitosheet mime render does not exist, 
+                // just return. I don't think this should ever happen because you can't 
+                // have a mimerender for a code cell that does not exist anymore.
+                return;
+            }
+
+            let codeCell = getCellAtIndex(cells, mimeRenderInputCellIndex + 1)
+            const codeCellText = getCellText(codeCell);
+
+            if (codeCell === undefined) {
```
aarondr77, line 250: Turn this logic into a helper function and share it with the other write generated code

mitosheet/src/plugin.tsx
```diff
@@ -251,12 +257,13 @@ function activateMitosheetExtension(
             }
         }
     });
-    
 
     app.commands.addCommand('mitosheet:create-mitosheet-from-dataframe-output', {
         label: 'creates a new mitosheet from the dataframe that is printed',
         execute: async (): Promise<void> => {
 
+            console.log("creating mitosheet from dataframe ouput")
```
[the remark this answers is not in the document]
aarondr77: TOD ^

mitosheet/src/plugin.tsx
```diff
@@ -404,19 +528,66 @@ function activateMitosheetExtension(
...
         }
+    }
+  
+    // Listen for new notebooks
+    notebookTracker.widgetAdded.connect((sender, notebookPanel) => {
+        
+        // When the session changes we need to re-import the package. For example, if the user 
+        // restarts a kernel. Also, becuase the extension is loaded before the kernel is created
+        // this sessionChange approach is required to import mitosheet when first opening a notebook.
+        notebookPanel.sessionContext.sessionChanged.connect((sessionContext, sessionContextArgs) => {
```
aarondr77, line 562: remove sessionContextArgs

mitosheet/src/plugin.tsx
```diff
@@ -404,19 +528,66 @@ function activateMitosheetExtension(
...
+            const kernel = sessionContext.session?.kernel as KernelConnection | null;
```
aarondr77, line 563: Is there a better approach then type casting?

mitosheet/src/plugin.tsx
```diff
@@ -411,12 +423,28 @@ function activateMitosheetExtension(
...
+    }, -1);  // Giving this renderer a lower rank than the default renderer gives this default priority
+}
+
     window.commands = app.commands; // So we can write to it elsewhere
 }
 
 const mitosheetJupyterLabPlugin: JupyterFrontEndPlugin<void> = {
-    id: 'mitosheet:plugin',
-    requires: [INotebookTracker],
+    id: 'mitosheet:plugin-new',
```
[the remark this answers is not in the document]
aarondr77: TODO

mitosheet/src/plugin.tsx
```diff
@@ -404,19 +528,66 @@ function activateMitosheetExtension(
...
+        so we can write to it elsewhere
+    */
+    window.commands = app.commands;
 }
 
 const mitosheetJupyterLabPlugin: JupyterFrontEndPlugin<void> = {
-    id: 'mitosheet:plugin',
-    requires: [INotebookTracker],
+    id: 'mitosheet:plugin-new',
+    requires: [INotebookTracker, IRenderMimeRegistry],
```
aarondr77, line 590: remove IRenderMimeRegistry

aarondr77 (commented):
Another review

aarondr77:
Make sure `print(df)` still returns the correct result.

PR #1342, opened 2024-10-18
mito-ai: fix bug with registering command in UseEffect

aarondr77:
# Description

Fixes bug where the following would occur: 
1. Send message
2. Clear Chat
3. Click explain code
4. See first chat message reappear. 

This would happen because the COMMAND_MITO_AI_SEND_MESSAGE is registered in a useEffect on initial render, which will always use the initial state values. Now, we access the chat history manager via a function to ensure we always get the most recent chat history, even when the command is executed later.

# Testing

Add a test for this: 
1. Send message
2. Clear Chat
3. Click explain code
4. See first chat message reappear. **It should not have the original message**

# Documentation

Not yet

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
@@ -44,6 +44,7 @@ const getDefaultChatHistoryManager = (): ChatHistoryManager => {
         return new ChatHistoryManager(chatHistory)
 
     } else {
+        console.log("HERE")
         const chatHistoryManager = new ChatHistoryManager()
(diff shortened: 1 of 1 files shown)


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -44,6 +44,7 @@ const getDefaultChatHistoryManager = (): ChatHistoryManager => {
         return new ChatHistoryManager(chatHistory)
 
     } else {
+        console.log("HERE")
```
aarondr77, line 47:
```suggestion

```

aarondr77 (commented):
LGTM

PR #1347, opened 2024-10-25
mito-ai: create separate prompt for each type of interaction

aarondr77:
# Description

Creates a separate code path for each type of AI completion. Each time we send a message to the AI there are three steps. 
1. Create a message. Some types of messages (like generating code) are more context hungry than others (explain active code cell). This step is responsible for building the write prompt with the right context for the specific message type.
2. Send the message to the AI and handle errors properly. This is the same for all types of AI messages. 
3. Post process. Currently, the only post processing step we have is showing code diffs. We want to do this in all cases except explaining code which should not return new code. 

# Testing

Try it out + see tests. 

# Documentation

No.

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
@@ -83,6 +83,7 @@ ${variables?.map(variable => `${JSON.stringify(variable, null, 2)}\n`).join('')}
 Complete the task below. Decide what variables to use and what changes you need to make to the active code cell. Only return the full new active code cell and a concise explanation of the changes you made.
 
 <Reminders>
+
 Do not: 
 - Use the word "I"
 - Include multiple approaches in your response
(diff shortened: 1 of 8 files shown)


mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
```diff
@@ -1,9 +1,11 @@
 import OpenAI from "openai";
-import { Variable } from "../VariableManager/VariableInspector";
+import { IVariableManager } from "../VariableManager/VariableManagerPlugin";
+import { INotebookTracker } from '@jupyterlab/notebook';
+import { getActiveCellCode } from "../../utils/notebook";
 
 export interface IDisplayOptimizedChatHistory {
     message: OpenAI.Chat.ChatCompletionMessageParam
-    error: boolean
+    type: 'openai message' | 'connection error'
```
aarondr77, line 8: I should add some documentation about this. It's not great code and should be explained.

mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
```diff
@@ -107,30 +119,124 @@ Use the pd.to_datetime function to convert the issue_date column to datetime.
 
 </Example>
 
-<Important Jupyter Context for Error Handling>
+Code in the active code cell:
```
aarondr77, line 122: Make these a separate function so its easier to read this file

PR #1349, opened 2024-11-01
Moved mito-ai tests into separate workflow

ngafar:
# Description

test-mitosheet-frontend will only be triggered if changes are made to `mitosheet/` or `tests/`; while test-mito-ai will cover changes made to the `mito-ai/` dir. 

# Testing

# Documentation

N/A


aarondr77:
LGTM!

PR #1351, opened 2024-11-01
update font size and line wrap of chat code

aarondr77:
# Description

1. Updates the text wrapping in the AI chat taskpane so that one line of code does not wrap to a second line in the code preview. Instead, you can scroll to see the full line of code. I think this makes the code more readable. 
2. Uses the jupyter default code family so if the user changes their settings, its reflected in our code preview
3. Set the font size to 12px so it takes up less space. 

# Testing

Try it out. 

# Documentation

No.

### DIFF
--- a/mito-ai/style/PythonCode.css
+++ b/mito-ai/style/PythonCode.css
@@ -8,7 +8,7 @@
     overflow-x: auto;
 }
 
-code {
+.code-message-part-python-code code {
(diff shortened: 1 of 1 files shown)


ngafar (commented):
I added the styling for the `code` class. Without it, the text was still wrapping.

ngafar (approved):

PR #1352, opened 2024-11-04
mito-ai: fix small chat input bug

aarondr77:
# Description

Fixes bug where dynamic calculation of chat input height was sometimes resulting in too small of an input. 

Screenshot 2024-11-04 at 10 00 13 AM

# Testing

This bug would only some of the times, and only when first opening the chat input without any messages already sent.  Try opening the chat input a few times and making sure you never see this bug. 

# Documentation

No.


ngafar:
Tested a few times, and the text area was correctly sized every time. Merging.

PR #1354, opened 2024-11-04
[WIP] Ai taskpane msg edit

ngafar:
# Description

# Testing

# Documentation


tests/mitoai_ui_tests/mitoai.spec.ts
```diff
@@ -35,6 +35,21 @@ test.describe('Mito AI Chat', () => {
     expect(code?.trim()).toBe("")
   });
 
+  test('Edit Message', async ({ page }) => {
```
aarondr77, line 38: Can we update this test to make sure that when you edit a message it clears all future messages from the chat history also? I think to do that, we should update the helper function you created from sending and immiedietly editing the message, to a function called `editMitoAIMesage` that takes a message index and a new message. 

That way we can do things like:
1. Create a first message 
2. Create a second message
3. Use the helper function to edit the first message

mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
```diff
@@ -36,20 +37,72 @@ const ChatMessage: React.FC<IChatMessageProps> = ({
...
+        setEditedContent((message.content as string).replace(/```[\s\S]*?```/g, '').trim());
+        setIsEditing(false);
+    };
+
+    if (isEditing) {
+        return (
+            <div className={classNames(
+                "message",
+                { "message-user": message.role === 'user' },
+                { 'message-assistant': message.role === 'assistant' },
```
aarondr77, line 77: I think editing the AI edited message might be scope creep for this PR. If a user is going to edit the AI message, then they are probably wanting to edit the AI generated code, not the text part of the response. But we don't have a good way of letting them edit the code yet. So let's just cut this for simplicity.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
```diff
@@ -36,20 +37,72 @@ const ChatMessage: React.FC<IChatMessageProps> = ({
...
+            )}>
+                <textarea
+                    value={editedContent}
+                    onChange={(e) => setEditedContent(e.target.value)}
+                    onKeyDown={(e) => {
+                        // Enter key sends the message, but we still want to allow 
+                        // shift + enter to add a new line.
+                        if (e.key === 'Enter' && !e.shiftKey) {
```
aarondr77, line 85: Let's add support so when the user presses `escape` it cancels editing if in editing mode and loses focus.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
```diff
@@ -74,7 +127,23 @@ const ChatMessage: React.FC<IChatMessageProps> = ({
...
                     return (
-                        <p>{mitoAIConnectionError && <span style={{marginRight: '4px'}}><ErrorIcon /></span>}{messagePart}</p>
+                        <div style={{ position: 'relative' }}>
+                            <p onDoubleClick={() => setIsEditing(true)}>
+                                {mitoAIConnectionError && <span style={{ marginRight: '4px' }}><ErrorIcon /></span>}
+                                {messagePart}
+                            </p>
+                            {message.role === 'user' && (
+                                <div style={{ display: 'flex', justifyContent: 'flex-end', marginTop: '4px' }}>
+                                    <button
```
aarondr77, line 137: Let's add a title tag here too so when the user hovers over the icon it says "Edit message" or something like that.

mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
```diff
@@ -134,6 +134,72 @@ Your task: ${input}`};
         this.history.aiOptimizedChatHistory.push(aiOptimizedMessage);
     }
 
+    updateMessageAtIndex(index: number, newContent: string): void {
+        const variables = this.variableManager.variables
+        const activeCellCode = getActiveCellCode(this.notebookTracker)
+
+        const aiOptimizedMessage: OpenAI.Chat.ChatCompletionMessageParam = {
```
aarondr77, line 141: Let's turn this prompt into a function so we don't have it here and in the `addChatInputMessage` function.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
```diff
@@ -36,20 +37,72 @@ const ChatMessage: React.FC<IChatMessageProps> = ({
     operatingSystem,
     setDisplayCodeDiff,
     acceptAICode,
-    rejectAICode
+    rejectAICode,
+    onUpdateMessage
 }): JSX.Element | null => {
+    const [isEditing, setIsEditing] = useState(false);
```
aarondr77, line 43: Maybe to consolidate state we can just have `editedContent` that is either a string or undefined. If it's undefined we are not in editing mode. If it's a string we are in editing mode?

Then when, `handleEditClick` is called, we can do this 

```
(message.content as string).replace(/```[\s\S]*?```/g, '').trim()
```

And when `handleCancel` is pressed, all we need to do is set `editedContent` to `undefined` instead of recalculating the original editing message.

aarondr77 (commented):
Nice work on this PR. It's a big step forward in complexity! Hopefully the codebase is making a bit more sense (and maybe you have some ideas on how to improve it too!) 

A few things to cleanup before LGTM. The two categories are: 1) Just polishing up the UI to make sure it's a great UX. 2) Creating some functions so we don't have duplicated code across the codebase. 

Ping me when you're done and I'll do one more small review before merging.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
ngafar: I'm going to mark this as resolved for now. After moving `ChatInput` into it's own component, `editedContent` and `isEditing` are no longer in the same file. This removes the option to consolidate.

Feel free to unresolved if you feel otherwise.

aarondr77 (approved):
LGTM once merge conflicts are squashed. I'll fix up and merge now.

PR #1363, opened 2024-11-11
FIX: Chatbox markdown format rendering

vibhatha:
# Description

The current version of the chatbox pane doesn't render the non-code segments properly as expected. This PR includes a markdown renderer which fixes this issue. 

# Testing

TBD

# Documentation

Note if any new documentation needs to be addressed or reviewed.


vibhatha:
@aarondr77 I need to add some tests, I guess we could check markdown plain strings with a sample text input. 
Do you have any suggestions for this?

mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
```diff
@@ -109,9 +110,12 @@ const ChatMessage: React.FC<IChatMessageProps> = ({
                 } else {
                     return (
                         <div style={{ position: 'relative' }}>
-                            <p onDoubleClick={() => setIsEditing(true)}>
+                            <p key={index} onDoubleClick={() => setIsEditing(true)}>
```
aarondr77, line 113: I don't think that this key is unique enough. Notice in this video that if I:
1. Add a prompt to the chat taskpane and send the message
2. Then use the explain button to clear the chat and send another prompt
3. The user chat message has both the original user input and the new explain user input

I think we can fix this by making the key `key={index + messagePart}`

https://github.com/user-attachments/assets/79123c50-5c4b-4ee0-a123-68e995a6c38e

mito-ai/src/Extensions/AiChat/ChatMessage/MarkdownBlock.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+            const node = renderer.node;
+            setRenderedContent(<div ref={(el) => el && el.appendChild(node)} />);
+        };
+
+        renderMarkdown();
+    }, [markdown, rendermime]);
+
+    return (
+        <div>{renderedContent || <div>No content</div>} </div>
```
aarondr77, line 31: Instead of returning "No content", let's just return an empty div.

aarondr77 (commented):
Awesome work on this PR. It's a huge improvement in readability of the chat taskpane. 

I caught one small bug to fix up, then once we add some tests, we should be good to go. 

For tests, maybe something as simple as: Ask the AI to generate a message that includes backticks around a word like "`df`" and then in the rendered display, we can make sure that the backticks are not visible. In the future, we can use snapshot tests that actually compare the look of taskpane to a correct screenshot, but we don't have the infrastructure for that yet and I think it is overkill.

mito-ai/src/Extensions/AiChat/ChatMessage/MarkdownBlock.tsx
vibhatha: Definitely.

aarondr77:
@vibhatha since this is your first contribution to this codebase,  can you check out our CLA here: https://github.com/mito-ds/monorepo/blob/dev/CLA.md

If you agree, just hit me back with "I have read the CLA Document and I hereby sign the CLA" -- and we'll get this merged in as soon as this small review is addressed!

vibhatha:
> Awesome work on this PR. It's a huge improvement in readability of the chat taskpane. 
> 
> 
> 
> I caught one small bug to fix up, then once we add some tests, we should be good to go. 
> 
> 
> 
> For tests, maybe something as simple as: Ask the AI to generate a message that includes backticks around a word like "`df`" and then in the rendered display, we can make sure that the backticks are not visible. In the future, we can use snapshot tests that actually compare the look of taskpane to a correct screenshot, but we don't have the infrastructure for that yet and I think it is overkill. 

Definitely. Let me work on it.

vibhatha:
> @vibhatha since this is your first contribution to this codebase,  can you check out our CLA here: https://github.com/mito-ds/monorepo/blob/dev/CLA.md
> 
> 
> 
> If you agree, just hit me back with "I have read the CLA Document and I hereby sign the CLA" -- and we'll get this merged in as soon as this small review is addressed! 

Oh yes. Let me check.

vibhatha:
@aarondr77 I have read the CLA Document and I hereby sign the CLA.

vibhatha:
@aarondr77 many CIs were failing. I rebased since there was another conflict after the rebase earlier. 

Am I missing something here? Please advice.

aarondr77:
It looks like the tests are failing that require the GitHub action secrets. I can see in the playwright report that the tests are failing like this: 

Screenshot 2024-11-19 at 9 39 01 PM

My guess is that this is a result of this being a forked PR instead of a cloned PR. I wonder if GitHub doesn't give forked PR's access to GitHub action secrets for security purposes. Let me do a little investigation.

aarondr77:
Looking at the tests that are passing, they are all ones that could pass regardless of whether the AI returns the correct thing. So I'm pretty confident it has to do with not having access to the secrets. 

Screenshot 2024-11-19 at 9 46 11 PM

I confirmed this [with a conversation with chatgpt](https://chatgpt.com/share/e/673cf8b9-ef18-800e-9d42-32fa9f9f2b6e)

PR #1364, opened 2024-11-12
@ Mention Variables

ngafar:
# Description

This PR introduces the "mention" feature for referencing variables in the chat pane. By using the `@` keyword a list of variables will be displayed in a dropdown, for easy selection. 

# Testing

Several cases to tested:

1. Create an @ mention at the beginning of a sentence.
2. Create an @ mention mid sentence.
3. Create an @ mention at the end of a sentence.
4. Edit an @ mention. - This feature was removed to keep things simple. 

In all of the cases above, pay attention to the position of the cursor. It should appear after the @ mention.  

Additionally, an integration test was added to the `mitoai_it_tests` labeled, "Variable dropdown shows correct variables." This test creates a dataframe with two columns: Apples and Bananas. The test then searches for `@ap`. Finally, an assert statement looks to make sure that (1) the Apple column is displayed in the dropdown, and (2) the Banana column is not displayed.  

# Documentation

TODO: Would be good to call out this feature by having a small section on the [Mito-AI/Chat](https://docs.trymito.io/mito-ai/chat) page. Remember to include a screenshot with the feature in action.


mito-ai/src/Extensions/AiChat/ChatMessage/ChatDropdown.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { ExpandedVariable } from './ChatInput';
+
+interface ChatDropdownProps {
+    options: ExpandedVariable[];
+    selectedIndex: number;
+    onSelect: (variableName: string, parentDf?: string) => void;
+}
+
+const ChatDropdown: React.FC<ChatDropdownProps> = ({
```
aarondr77, line 10: I don't think we need to do this right now since we don't yet have another use for the dropdown. But just wanted to call this out.

We've had a lot of success building component libraries for ourselves in the past. In the mitosheet/src/mito/components folder you can see a bunch of components we've created. Having these general purpose components speeds up development time so so much. So I think we should treat this dropdown as a component that we might want to reuse across the mito-ai extension. 

Creating good components means we should: 
1. Have good scoping of what the component is responsible for and what the parent of the component is responsible for. The parent component should be responsible for passing in the data, placing the component in the right spot, etc. The component itself should be responsible for its own logic. 
2. Variable names that make sense in the context of the component regardless of where it is used in the codebase. So for example, instead of the options being typed `ExpandedVariable[]`, we could type it `DropdownOptions[]` and then make sure that the things we pass in fit that.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatDropdown.tsx
```diff
@@ -0,0 +1,58 @@
...
+    onSelect,
+}) => {
+    const getShortType = (type: string) => {
+        return type.includes("DataFrame") ? "df"
+            : type.includes("<class '") ? type.split("'")[1]
+                : type;
+    }
+
+    return (
+        <ul className="chat-dropdown-list">
```
aarondr77, line 22: I think there's two improvements we should make here: 

1. Instead of showing every option by default, maybe we should only show the first 5 or so. Having this giant list gets a little bit out of hand. 
2. We should position the dropdown above the taskpane using the `z-index`. Just like when you right click to inspect your browser, the dropdown appears above the browser instead of as part of the webpage. This is the same with @ mentions in Slack, Zulip, Cursor, etc.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -33,42 +75,166 @@ const ChatInput: React.FC<ChatInputProps> = ({
...
+            setFilteredOptions(filtered);
+            setDropdownVisible(filtered.length > 0);
+            setSelectedIndex(0);
+        } else {
+            setDropdownVisible(false);
+            setFilteredOptions([]);
+        }
+    };
+
+    const handleOptionSelect = (variableName: string, parentDf?: string) => {
```
aarondr77, line 102: There's a bug here where if you 

1. Have a message like `Hello`
2. Move the cursor to the start of the input, right before the H
3. Type @ then the dropdown appears
4. Select something from the dropdown
5. It inserts the new item and drops the word Hello

https://github.com/user-attachments/assets/42fcb9cf-f6a4-4ff9-82f2-69439e01d5fa

mito-ai/src/Extensions/AiChat/ChatMessage/ChatDropdown.tsx
aarondr77: For example, right now, I can scroll down the taskpane to see this entire dropdown and not even see my chat input anymore. 

Screenshot 2024-11-14 at 7 24 47 PM

tests/mitoai_ui_tests/mitoai.spec.ts
```diff
@@ -112,7 +112,7 @@ test.describe('Mito AI Chat', () => {
 
   });
 
-  test.only('Fix Error and Explain code buttons clear chat history', async ({ page }) => {
```
aarondr77, line 115: 🤦 good catch

mito-ai/src/Extensions/AiChat/ChatMessage/ChatDropdown.tsx
aarondr77: Another example, when I've been using the chat taskpane for a bit and therefore the chat is full, the @ mention dropdown is opened off the bottom of the visible taskpane so I didn't even notice it right away. This is another reason why having the dropdown hover will be nice because it can open up instead of down, like cursor (see below)
Screenshot 2024-11-14 at 7 27 41 PM

Screenshot 2024-11-14 at 7 31 30 PM

mito-ai/src/Extensions/AiChat/ChatMessage/ChatDropdown.tsx
```diff
@@ -0,0 +1,58 @@
...
+                                    : getShortType(option.type) === 'col' ? 'orange'
+                                        : "green"
+                            }}
+                        >
+                            {getShortType(option.type)}
+                        </span>
+                        <span className="chat-dropdown-item-name">
+                            {option.variable_name}
+                        </span>
+                        {option.parent_df && (
```
aarondr77, line 46: Sweet! displaying the dataframe name is so needed. Look at this example I got while testing
Screenshot 2024-11-14 at 7 33 24 PM

aarondr77 (commented):
Nice work @ngafar ! You've implemented this so much faster than we expected! The dropdown is working super nicely overall. 

I left a review below that has a few things to fix up. There's kinda two buckets:
1. Moving more of the logic into the dropdown component itself + only using state variables when we actually need to be re-rendering. 
2. Making the dropdown look more like the Figma mockups. I left a few specific comments down below (like making it have a z-index so it hovers), but overall, would be great if you could just pull up the mockups and try to get this to look as close to that as possible. We want to really aim for highly polished features that look and feel great. I think a good step here would be: 1) If you have ideas about how to improve the mockups, feel free to hop into Figma and make a new version and send for review so we can finalize it. (Feel free to read that Refactor UI if you think it would help) 2) Let's try to get the implementation to look as close to that as possible. 

Ping me when you're ready for another review or let me know if you want to chat!

mito-ai/src/Extensions/AiChat/ChatMessage/ChatDropdown.tsx
ngafar: > We should position the dropdown above the taskpane using the z-index.

Might need some more clarity on this. `.chat-dropdown-list` has a z-index of 9999, which should ensure it's always on top.

aarondr77 (commented):
Okay, much improved and still working really well. I have a few more small comments here but they are all small and no big refactors at all. Mostly, just sharing some best practices I've learned building these before + a bit more polish. 

Once you implement, LGTM!

aarondr77:
btw, for debugging tests, it can be helpful to run them with the `--headed` flag so you can see it running in live time.

ngafar:
Running the `mitoai` tests in headed mode it looks like the issues is with tests that use the `createAndRunNotebookWithCells` function. Half of the time, this function will create the cell, but it won't run the cell. This causes the tests to timeout. Need to look more into this, but most (if not all) of the test failures here are from flaky results.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -33,19 +44,126 @@ const ChatInput: React.FC<ChatInputProps> = ({
...
+    useEffect(() => {
+        if (isDropdownVisible) {
+            calculateDropdownPosition();
+        }
+    }, [isDropdownVisible]);
+
     return (
-        <>
+        <div style={{ position: 'relative' }}>
+            {isDropdownVisible && showDropdownAbove && (
```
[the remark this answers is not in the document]
ngafar: Remove this line, and the dropdown just does not show up. Another CSS mystery.

mito-ai/style/ChatTaskpane.css
```diff
@@ -103,3 +107,83 @@
 .message-edit-buttons button:hover {
     background: #f0f0f0;
 }
+
+.chat-dropdown {
+    position: absolute;
+    width: 100%;
+    z-index: 9999;
```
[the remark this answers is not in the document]
ngafar: Spent some time working with `createPortal`; it's got potential, but it is a bit of a challenge to get it working properly. This is due to the somewhat complex nature of the sidebar, and needing the dropdown to be positioned relative to the textare. Things especially fall apart when the dropdown needs to be displayed above the textarea.

I'm going to suggest we put up with this forced scroll behavior for now, since users only scroll under very specific circumstances. When a user first enters the chat taskpane they should be fine, when the textare is at the bottom, users should also be fine. It's when the sidebar hasn't been filled yet, that the problem arises. 

Long term we could take a page from Cursor where the textarea starts on top, then is locked to the bottom after the first message.

mito-ai/style/ChatTaskpane.css
```diff
@@ -103,3 +107,87 @@
...
+    border-bottom-right-radius: 5px;
+}
+
+.chat-dropdown-item-type {
+    color: var(--muted-text-color);
+    font-family: var(--jp-code-font-family);
+    min-width: 35px; /* Fixed width for type column */
+}
+
+.chat-dropdown-item-name {
```
aarondr77, line 176: This is getting too cut off when both the column name is long and the dataframe name is long. 
Screenshot 2024-11-18 at 11 18 55 PM

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -62,13 +186,21 @@ const ChatInput: React.FC<ChatInputProps> = ({
...
             }
-        </>
+            {isDropdownVisible && !showDropdownAbove && (
+                <ChatDropdown
+                    options={expandedVariables}
+                    onSelect={handleOptionSelect}
+                    filterText={dropdownFilter}
+                    position="below"
+                />
+            )}
```
aarondr77: I'm still pretty sure we only need one of these. I'm going to try removing it and seeing what happens.

PR #1365, opened 2024-11-14
Mito AI Server

aarondr77:
# Description

When the user does not have a `OPENAI_API_KEY` set, they can now use the Mito AI server. It's just a simple lambda function that makes an Open AI request and returns a dictionary of `{completion: <completion>}`. If the user does have `OPENAI_API_KEY` set, then we will use it instead. 

# Testing

1. Try using the taskpane with and without an  `OPENAI_API_KEY`
2. Checkout the updated GitHub action workflow. We now run our mito ai tests using an open ai key and using the mito ai server. Having some testing for both is good as they have slightly different code paths, but maybe running all of the tests is overkill.

# Documentation

Yes, we need to create some documentation for this.


aarondr77:
We should update our logging of the lambda function to show which extension `mito-ai` or `mitosheet` is making the request.

ngafar (approved):

PR #1366, opened 2024-11-14
cleanup, cleanup, everybody, everywhere

aarondr77:
# Description

Our codebase is expanding! This PR removes old stuff that we no longer have any need for:
- The gitbook documentation. We used to sync this to GitHub and now we don't. 
- The tools folder that had one utility for working with mix panel logs
- Adds an entry to the .gitignore so we stop committing scratch work .ipynb files that are in the mitosheet and mito-ai folders. 

```
Clean up, clean up, everybody everywhere,
Tidy up the codebase, bring some air.
Delete the files that clutter and slow,
Clear out the parts we no longer show.
```

# Testing

Give it a sanity check to make sure everything looks good. If the tests pass, then we should be good I think.

# Documentation

No.


ngafar:
Goodbye untitled notebooks 👋, you'll be missed. Everything looks good. One failed test (in the mitosheet frontend), but I doubt the cleanup was the culprit. Merging.

PR #1368, opened 2024-11-14
Display advice in empty editor to use Mito AI

fcollonval:
# Description

Display advice to trigger the chat panel in empty cells

# Testing

Work out of the box
Added an integration test for the feature

https://github.com/user-attachments/assets/a06ce49f-6784-4014-9d34-44f7f658fce5

Latest version looks like that:

image

# Documentation

N/A


mito-ai/src/Extensions/emptyCell/emptyCell.ts
aarondr77, line 1: A small nitpick, so if it's not an easy fix, let's ignore it for now. 

But it would be nice if in an empty cell the user's cursor was still at the very start of the cell instead after the placeholder text.

aarondr77 (commented):
This is super sweet! I love this. It's a great touch.

mito-ai/src/Extensions/emptyCell/emptyCell.ts
fcollonval: Reolved in f3cc190

mito-ai/src/Extensions/emptyCell/emptyCell.ts
```diff
@@ -0,0 +1,94 @@
...
+    borderStyle: 'solid',
+    borderWidth: '1px',
+    fontSize: 'calc(var(--jp-code-font-size) - 2px)',
+    padding: '3px 5px',
+    verticalAlign: 'middle'
+  }
+});
+
+/**
+ * A facet that stores the chat shortcut.
```
aarondr77, line 30: My understanding is that when there are multiple code mirror extensions in one code mirror editor, this `combine` function is used to figure out how to reconcile them. So here, we are saying something like: If there are multiple chatShorcut decorations set, use the last one. 

Is this needed in order to update the decorations displayed? ie: If instead of returning the last value, we instead returned the first one, would we always display the placeholder text?

mito-ai/ui-tests/tests/mitoAdvice.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { expect, test } from '@jupyterlab/galata';
```
aarondr77, line 1: Just leaving this as a note that we decided we're going to move this test into the `tests` folder at the root of the monorepo. This test will then automatically run whenever we edit the `mito-ai` folder.

mito-ai/ui-tests/tests/mitoAdvice.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  await expect
+    .soft((await page.notebook.getCellLocator(0))!.getByRole('textbox'))
+    .toHaveText(
+      'Press Ctrl + E to ask Mito AI to do something. Start typing to dismiss.'
+    );
+
+  await page.notebook.setCell(0, 'code', 'print("Hello, World!")');
+
+  await expect
+    .soft((await page.notebook.getCellLocator(0))!.getByRole('textbox'))
```
aarondr77, line 17: ooo cool! I didn't know this soft assertion was a thing. This is great for letting us test each of the different permutations in one test. I like it.

aarondr77 (commented):
Just a little bit of cleanup and then LGTM without another review. Nice work! The tests aren't going to pass on a forked PR, so we'll see them pass on the CI after merging. And in the future, now that you have write access, you can open cloned PRs and the tests should run + pass there :)

mito-ai/src/Extensions/emptyCell/emptyCell.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  eq(other: AdviceWidget) {
+    return false;
+  }
+
+  toDOM() {
+    const wrap = document.createElement('span');
+    wrap.className = 'cm-mito-advice';
+    // FIXME it should come from JupyterLab setting
+    wrap.innerHTML = `Press ${this.shortcut.map(s => `<kbd>${s}</kbd>`).join(' + ')} to ask Mito AI to do something. Start typing to dismiss.`;
```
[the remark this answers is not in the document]
fcollonval: First part was a bug due to this editor not being force to be loaded after the AI chat plugin. Therefore the shortcut was not yet defined and it fall back to the default 'Ctrl + E'.

I updated the code to for plugin load order via token dependency and changed the default to 'Accel + E'.

mito-ai/src/Extensions/emptyCell/emptyCell.ts
fcollonval: Actually this apply only for the `Facet` that is storing configuration information. Using the latest value is usually the best option if we don't know better.

> It is unlikely to happen in this case.

mito-ai/ui-tests/tests/mitoAdvice.spec.ts
fcollonval: Yep and it is recommended as best practice: https://playwright.dev/docs/best-practices#use-soft-assertions

aarondr77 (approved):

Issue #1360, opened 2024-11-06
Mito AI chat code diffs should revert to original state if not handled by user. 


aarondr77:
Here is the issue: 
1. Generate code with AI 
2. Notice that the code is wrong, but don't Reject the AI code. Since you did not reject the AI code, the diffed code is still in the active cell
3. Send another message asking the AI to update the code. 
4. Now instead of seeing the diff between the original code and the AI generated code, you are seeing the diff between the diffed code from step 2 and the new AI generated code. That is completely not helpful. 

The solution for the user right now is for them to reject the code that they don't want after step 2, reverting the code cell back to the original state. But this is not a reasonable expectation and means our current solution is too fragile. Of course some users are going to skip the rejection step and just reprompt the AI. 

Here is one idea: If the user has not accepted or rejected the code the AI generated code, when the user either 1) starts typing in the input field or 2) the Ai taskpane loses focus (ie: the user is going to edit the notebook or something else) remove the code diff display and revert the code cell back to its original state.

aarondr77:
This is pretty bad for users. If they don't accept or reject the code before creating a new message or clearing the chat, then the accept and reject buttons go away!

ngafar:
We should definitely give this workflow some consideration. I'm partial to not showing the diff until the user asks for it. This way a user could prompt the AI, and get a suggestion. The suggestion would be shown in the sidebar, and from there the user would have the option to click an "Preview" button. Once the preview button has been clicked, we then show the diff, and prompt the user to accept/decline. 

We'd probably still need to test this, but I would assume the extra clicks would force uses to consider this as a workflow, i.e.: prompt -> preview -> accept/deny.

PR #1378, opened 2024-11-25
Auto-Remove Diffs Before Sending Message

ngafar:
# Description

This PR ensures that any unapproved code is rejected before sending the next message to the AI. This clears any visible code diffs, then sends the new message. 

Meant to address the first part of https://github.com/mito-ds/mito/issues/1360

# Testing

For manual testing:
- Send a message using Mito AI.
- Do not approve the suggested code changes.
- Send a second message. -> At this point the changes suggested in step two should no longer be visible. 

The steps above should also work when editing a message.

Additionally, integration tests can be found in the `mitoai_ui_test/` dir. See "Code diffs are automatically rejected before new messages are sent."

# Documentation

Not needed. 

We should keep an eye out, and users bring it up, we could add a FAQ section to the Mito AI section of the docs.


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -126,6 +126,10 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
         Send whatever message is currently in the chat input
     */
     const sendChatInputMessage = async (input: string) => {
+        // Step 0: If code diffs are displayed, reject the AI code, then send the message.
+        if (unifiedDiffLines !== undefined) {
+            rejectAICode()
+        }
```
aarondr77, line 132: Okay, so now seeing this in action: In a lot of ways this is better because it stops users from having broken code in their notebook (ie: the diffed code that is a mash up of the original code and the AI generated code)... but there is also a big bug here.  In particular, it opens up a dangerous flow that you can see here. 

1. Make an AI edit and see the response. 
2. Then: A user continue to assume that their edit is applied and there is no more actions that they need to do to accept it (which is what we're seeing users do), so they move to the next code cell to make their next edit
3. They use the AI to prompt for a new update. 
4. The `rejectAICode` writes the original code to the correct active code cell, overwriting what was in the code cell. 

https://github.com/user-attachments/assets/23e4dc91-6ed4-42c6-a329-2f829279c6ea

This is now more likely to happen because before users had to hit the `reject` button to deny the changes. Now, they just have to keep working. 

Here's a proposal on how to fix this: 

1. Create a new type called 

```python
export interface IAIOptimizedChatHistory {
    codeCellID: str
}
```

2. When we add a new message to the chat history manager, get the active code cell id and add it to the chat. We can do this very similarly to how we are currently getting the correct active code cell content. 

3. When we call the `rejectAICode` button, instead of overwriting the code in the current active code cell, overwrite the code in the code cell id. 

While we're at it, we can also update it so when the user accepts an AI response, we also write the code based on the saved code cell id instead of the current active code cell.

tests/mitoai_ui_tests/mitoai.spec.ts
```diff
@@ -72,6 +72,18 @@ test.describe('Mito AI Chat', () => {
     await expect(page.locator('.cm-codeDiffInsertedStripe')).toBeVisible();
   });
 
+  test('Code diffs are automatically rejected before new messages are sent', async ({ page }) => {
```
aarondr77, line 75: Let's add a test to make sure we don't overwrite users's code (ie: the bug above)

aarondr77 (changes requested):
Nice work turning this around. There's one important bug that I noticed here that I think we need to fix before merging. Fixing this bug is a bit more work, but not too much. 

What do you think?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -223,16 +234,16 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
         _applyCode(aiGeneratedCode)
     }
 
-    const rejectAICode = () => {
+    const rejectAICode = (codeCellID?: string) => {
```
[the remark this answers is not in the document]
aarondr77: Let's add a test or update an existing one to make sure we don't break this in the future too

aarondr77:
@ngafar it looks like the tests are failing because of issues with the mito-ai yarn file, so I just rebuilt the file and pushed it here. Let's see if that fixes it. 
Screenshot 2024-12-02 at 1 56 55 PM

mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
```diff
@@ -63,7 +69,12 @@ export class ChatHistoryManager {
         return { ...this.history };
     }
 
-    getAIOptimizedHistory(): OpenAI.Chat.ChatCompletionMessageParam[] {
+    getCurrentCellId(): string | undefined {
```
aarondr77, line 72: I think this function should live the the utilities probably.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -41,8 +41,8 @@ const ChatInput: React.FC<ChatInputProps> = ({
         }
         textarea.style.minHeight = 'auto';
 
-        // The height should be 20 at minimum to support the placeholder
-        const minHeight = textarea.scrollHeight < 20 ? 20 : textarea.scrollHeight
+        // The height should be 40 at minimum to support the placeholder
```
aarondr77, line 44: Sicko. This looks so much better

mito-ai/src/Extensions/AiChat/ChatMessage/CodeBlock.tsx
```diff
@@ -43,6 +43,11 @@ const CodeBlock: React.FC<ICodeBlockProps> = ({
         navigator.clipboard.writeText(codeWithoutMarkdown)
     }
 
+    const getCodeCellID = () => {
```
aarondr77, line 46: Should this be using the active cell? I think this is supposed to use the cell id that is stored in the chat history manager.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -183,12 +188,36 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+    const rejectAICodeIfNeeded = (chatHistoryManager: ChatHistoryManager) => {
+        // If the user has sends a msg, and code diffs are displayed, 
+        // then automatically reject the AI code before sending a new message.
+        if (unifiedDiffLines !== undefined) {
+            const codeCellID = chatHistoryManager.getCodeCellIDOfMostRecentAIMessage() || ''
+            rejectAICode(codeCellID, false)
+        }
+    }
+
+    const getCodeCellIDWithDiff = (codeCellID: string): string => {
```
aarondr77, line 200: Do we need this function? Don't we always have the cell id stored in the chat history manager? We don't need to handle it differently if that is currently the active code cell or if it's not the active code cell. I might be missing something though.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -250,13 +288,17 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+                acceptAICode(codeCellID)
             }
         })
 
         app.commands.addCommand(COMMAND_MITO_AI_REJECT_LATEST_CODE, {
             execute: () => {
-                rejectAICode()
+                const newChatHistoryManager = getDuplicateChatHistoryManager()
+                const codeCellID = newChatHistoryManager.getCodeCellIDOfMostRecentAIMessage() || ''
+                rejectAICode(codeCellID)
```
aarondr77, line 301: This is good for now. I wonder if it would be better if the `rejectAICode` and `acceptAICode` functions lived in the chatHistoryManager instead of here?

tests/mitoai_ui_tests/mitoAdvice.spec.ts
```diff
@@ -10,10 +10,10 @@ test('should display an advice message in empty code cell', async ({
   await expect
     .soft((await page.notebook.getCellLocator(0))!.getByRole('textbox'))
     .toHaveText(
-      'Start writing python or Press Ctrl + E to ask Mito AI to write code for you.'
+      'Start writing python or Press ⌘ + E to ask Mito AI to write code for you.'
```
aarondr77, line 13: oo I think this is going to break on windows now. So Let's just shorten the thing to be like: toHaveText 'Start writing python or Press'

tests/mitoai_ui_tests/mitoai.spec.ts
```diff
@@ -72,6 +72,80 @@ test.describe('Mito AI Chat', () => {
...
+    await createAndRunNotebookWithCells(page, ['\nprint("cell 0")']);
```
aarondr77, line 76: I think we can remove \n here since we're handling it in the createAndRunNotebookWithCells function

aarondr77 (commented):
Small cleanup then LGTM once tests are passing

mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
```diff
@@ -87,31 +82,40 @@ export class ChatHistoryManager {
...
+        const activeCellCode = getCellCodeByID(this.notebookTracker, activeCellID)
 
         const aiOptimizedMessage: OpenAI.Chat.ChatCompletionMessageParam = {
             role: 'user',
             content: createBasicPrompt(variables, activeCellCode || '', newContent)
         };
 
         // Update the message at the specified index
-        this.history.aiOptimizedChatHistory[index] = {message: aiOptimizedMessage, codeCellID: this.getCurrentCellId() || ''};
+        this.history.aiOptimizedChatHistory[index] = {message: aiOptimizedMessage, codeCellID: activeCellID|| ''};
```
aarondr77, line 120: Allow this to be undefined

aarondr77 (commented):
final cleanup

PR #1380, opened 2024-11-26
Dogfood 11 26 24

aarondr77:
# Description

- Adds an error param for `mito_free_tier_limit_reached` so we can differentiate why errors occur. 
- Improves lambda function logging
- Prevents the variable manager from completely failing when 
- Lets users press `tab` to select an element from the @ mention dropdown. (I instinctually tried using tab a few times today and noticed that other tools like Zulip let you press tab or enter to accept).
- Removes `FUNCTIONS` from the @ mention dropdown. This is just exported from mitosheet and is a dict with all the mitosheet functions. The user has no need to interact with this. 

# Testing

Try out all of the above. 

# Documentation

No.

### DIFF
--- a/mito-ai/mito-ai/utils/telemetry_utils.py
+++ b/mito-ai/mito-ai/utils/telemetry_utils.py
@@ -101,6 +101,8 @@ def log(
     # Add the error if it exists
     if error is not None:
         final_params['error'] = str(error)
+        
+    print(final_params)
 
     # Finally, do the acutal logging. We do not log anything when tests are
     # running, or if telemetry is turned off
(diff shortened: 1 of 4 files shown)


mito-ai/mito-ai/utils/telemetry_utils.py
```diff
@@ -101,6 +101,8 @@ def log(
     # Add the error if it exists
     if error is not None:
         final_params['error'] = str(error)
+        
+    print(final_params)
```
aarondr77, line 105: delete

ngafar (approved):

PR #1387, opened 2024-12-02
sync jupyter lab version pins

aarondr77:
# Description

Small PR to fix JupyterLab version pins. Inconsistent pinning of dependencies was leading to type errors (see below).

This PR makes sure that all of JupyterLab-related packages are using the same version of @jupyterlab/cells by updating the version pinning in the package.json.

Screenshot 2024-12-02 at 10 25 20 AM

# Testing

Build the extension and see passing tests. 

# Documentation

No.

### DIFF
--- a/mito-ai/package.json
+++ b/mito-ai/package.json
@@ -60,7 +60,7 @@
     "@codemirror/state": "^6.2.0",
     "@codemirror/view": "^6.9.6",
     "@jupyterlab/application": "^4.1.0",
-    "@jupyterlab/apputils": "^4.2.0",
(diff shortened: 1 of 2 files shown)


fcollonval (approved):

fcollonval:
This looks good to me. FYI you can also run the following command:

```
jlpm dlx yarn-berry-deduplicate && jlpm
```

This will reduce the yarn.lock to avoid duplicate versions. This has a side effect with juputerlab packages to align their version and avoid conflicts.

aarondr77:
Ohh awesome. That's good to know. Just ran that command as well. Thanks for the tip!

PR #1393, opened 2024-12-03
Mito AI Apply Button

ngafar:
# Description

Part two of the new diff-ing workflow. This PR introduces a new paradigm when working with the AI. 

To start, AI generated code is no longer added by default. Users now have to click on an *Apply* button, then select accept/deny. 

UI changes:

- Accept/deny buttons now colored.
- Cell being edited by AI is briefly highlighted. This should help users to be more cognizant of where changes are being made.  

# Testing

Several tests were updated to take into account the new apply -> accept/deny workflow. 

Additionally, be sure to try it out.

# Documentation

N/A - Shouldn't need to call any special attention, the workflow should be self-explanatory. However, if this is not the case, we could do a quick write-up for the docs.


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -129,34 +131,25 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
-        // Step 1: Update the chat history manager
-        const newChatHistoryManager = getDuplicateChatHistoryManager()
-        newChatHistoryManager.updateMessageAtIndex(messageIndex, newContent)
-
-        // Step 2: Send the message to the AI
-        const aiMessage = await _sendMessageToOpenAI(newChatHistoryManager)
-
-        // Step 3: Update the code diff stripes
-        updateCodeDiffStripes(aiMessage)
+        sendChatInputMessage(newContent, messageIndex)
```
aarondr77, line 152: love love this cleanup

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -129,34 +131,25 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
         const newChatHistoryManager = getDuplicateChatHistoryManager()
-        newChatHistoryManager.addChatInputMessage(input)
+        if (messageIndex !== undefined) {
+            newChatHistoryManager.updateMessageAtIndex(messageIndex, input)
+        } else {
+            newChatHistoryManager.addChatInputMessage(input)
+        }
 
         // Step 2: Send the message to the AI
         const aiMessage = await _sendMessageToOpenAI(newChatHistoryManager)
```
aarondr77, line 147: I know I wrote this code, but now reading it, I think a better name for this function would be `_sendMessageAndSaveResponse` just so its clear that we're also saving it to the chat history manager. Thoughts?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -54,6 +54,8 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
     const [unifiedDiffLines, setUnifiedDiffLines] = useState<UnifiedDiffLine[] | undefined>(undefined)
     const originalCodeBeforeDiff = useRef<string | undefined>(undefined)
 
+    const [aiCodeSuggestion, setAiCodeSuggestion] = useState<OpenAI.ChatCompletionMessage | undefined>(undefined)
```
aarondr77, line 57: I don't think we need this, right? It looks like we're only using this to store the result so we know how to calculate the codeDiffs, but instead of using this, we should just be able to get the last message from the ChatHistoryManager.

mito-ai/src/Extensions/AiChat/ChatMessage/CodeBlock.tsx
```diff
@@ -65,12 +69,28 @@ const CodeBlock: React.FC<ICodeBlockProps> = ({
...
+                                        title={`Accept code (${operatingSystem === 'mac' ? 'CMD+Y' : 'CTRL+Y'})`}
+                                    >
+                                        Accept
+                                    </button>
+                                    <button
+                                        className='code-block-deny-button'
+                                        onClick={() => { rejectAICode(codeCellID); setIsApplyingCode(false) }}
+                                        title={`Deny code (${operatingSystem === 'mac' ? 'CMD+D' : 'CTRL+D'})`}
+                                    >
+                                        Deny
```
aarondr77, line 90: Thoughts on adding the keyboard shortcuts back, but making them a bit less verbose by using the icons instead of CMD or CTRL. Like this? I think otherwise people might not discover it. 
Screenshot 2024-12-04 at 3 10 38 PM

aarondr77 (commented):
Small cleanup and then LGTM! I love all the polish work you added here! It makes things pop and feel pretty great IMO.

aarondr77:
We don't need to do this now, but your highlighting of the code cell also made me realize:
1. When the user starts typing a message, we can highlight the code cell that is getting added as context. Hopefully helping avoid the UX flaw where users add the wrong code cell as context. In addition, we can add the code cell into the chat message right away instead of waiting for them to send the message. 
2. When the user gets a response back, we could highlight the code cell that it will be applied to ahead of time to continue making it easier for users to avoid overwriting their code.

tests/mitoai_ui_tests/mitoai.spec.ts
```diff
@@ -8,15 +8,18 @@ test.describe.configure({ mode: 'parallel' });
 
 test.describe('Mito AI Chat', () => {
 
-  test('Apply AI Generated Code', async ({ page }) => {
-    await createAndRunNotebookWithCells(page, ['import pandas as pd\ndf=pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})']);
+  test('Apply and Accept AI Generated Code', async ({ page }) => {
+    await createAndRunNotebookWithCells(page, ['\nimport pandas as pd\ndf=pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})']);
```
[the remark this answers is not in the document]
ngafar: Good catch. Old habits die hard, even ones started yesterday.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
ngafar: Done. Your right, this was more apt name when most of the steps lived in `sendChatInputMessage`, but things have been moving out of that function at a pretty fast pace.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -54,6 +54,9 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
     const [unifiedDiffLines, setUnifiedDiffLines] = useState<UnifiedDiffLine[] | undefined>(undefined)
     const originalCodeBeforeDiff = useRef<string | undefined>(undefined)
 
+    const [isApplyingCode, setIsApplyingCode] = useState<boolean>(false);
+    const [codeWasAccepted, setCodeWasAccepted] = useState<boolean>(false);
```
aarondr77, line 58: Let's combine these state variable into something more like a state machine. We could call the variable: `codeReviewStatus`. We're either in:
1.  `chat preview` state where the user has not yet pressed the `apply` button
2. `code cell preview` state where the user is seeing the code diffs and deciding how they want to respond
3. `applied` state where the user has applied the code to the code cell

mito-ai/src/icons/Check.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+const CheckIcon: React.FC<{
```
aarondr77, line 3: My vote would be to have the buttons `Accept ⌘Y` and `Deny ⌘D` on Mac and `Accept CTRL+Y` and `Deny ⌘D`. My philosophy is information over design when it comes to things like this. 

But up to you for this version and we can iterate with user feedback! That's the joy of watching folks!

aarondr77 (commented):
Okay tiny cleanup and then LGTM. Sorry for the double review here, I should have been able to get all of this into the first review and cut down on back and forth!

PR #1396, opened 2024-12-05
No longer updating diffs in sendDebugErrorMessage

ngafar:
# Description

Fixes an issue where using the Smart Debug feature would automatically add the code to the suggested fix to the cell. Instead, users are now expected to use the apply button. 

Also paper clipping on a change to the support icon svg. For whatever reason the updates made in https://github.com/mito-ds/mito/pull/1391 are not reflected when installing from test pypi. The icon still shows the old headset, so I also updated the icon. 

# Testing

Create a cell with an error, and use the Smart Debug button. 

# Documentation

N/A


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -113,10 +113,7 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
         setChatHistoryManager(newChatHistoryManager)
 
         // Step 2: Send the message to the AI
-        const aiMessage = await _sendMessageAndSaveResponse(newChatHistoryManager)
-
-        // Step 3: Update the code diff stripes
-        updateCodeDiffStripes(aiMessage)
+        await _sendMessageAndSaveResponse(newChatHistoryManager)
```
aarondr77, line 116: Can we update a test so that we catch this. Also, let's make sure we catch the same in chat input request flow

aarondr77 (commented):
Tiny fix up then feel free to merge

PR #1401, opened 2024-12-06
Code gen test case core refactor

aarondr77:
# Description

Refactors the test cases into this format which will (hopefully) make it easier to share test cases across the inline code completion and ai chat. This also lets us easily test multiple user inputs against the same notebook state and expected output. We do a little bit of that now, but we can try more when we add more test cases. 

```python

@dataclass
class CodeGenTestCaseCore:
    """
    The input and expected output components of a code gen test case. 
    Used for `AI Chat Input Message` and `Inline Code Completion` test cases.
    """
    notebook_state: NotebookState
    expected_code: str
    variables_to_compare: Optional[List[str]] = None
    workflow_tags: List[Literal[
		    'variable_declaration', 
		    'function_declaration',
		    'df_transformation',
		    'df_creation',
		    'pandas',
		    'misc',
		    'multistep'
		]]
  
@dataclass
class ChatMessageTestCase:
		"""
		Test case for the AI Chat Input, which just needs the active code cell
		that is defined in the TestCaseCore
		"""
    name: str
    test_case_core: CodeGenTestCaseCore
    user_input: str
```

# Testing

Try running the tests. 

# Documentation

No.

### DIFF
--- a/evals/test_cases/dataframe_transformation_tests.py
+++ b/evals/test_cases/dataframe_transformation_tests.py
@@ -1,315 +1,396 @@
-from evals.eval_types import TestCase
+from evals.eval_types import CodeGenTestCaseCore, TestCase
 from evals.notebook_states import *
 
-DATAFRAME_TRANSFORMATION_TESTS = [
-    TestCase(
-        name="single_column_numeric_filter",
-        notebook_state=LOANS_DF_NOTEBOOK,
-        user_input="Filter the annual income column to > 100k",
-        expected_code="loans_df = loans_df[loans_df['annual_income'] > 100000]",
-        tags=["df_transformation", "pandas"],
-    ),
-    TestCase(
-        name="two_column_filter",
-        notebook_state=LOANS_DF_NOTEBOOK,
-        user_input="Filter the annual income column to > 100k and the loan condition to only include 'Bad Loan'",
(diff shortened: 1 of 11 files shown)


evals/test_cases/dataframe_transformation_tests.py
ngafar, line 1: For a lot of these, it seems like each `CodeGenTestCaseCore` maps to a single `TestCase`. Looking at it, it does make it difficult to parse each test. You'd have to pick a test, and then go all the way up, to figure out what the core has. 

While I do like the reusability of it, now that I'm looking at it, I wonder if this is the best way forward.

ngafar (approved):

evals/test_cases/dataframe_transformation_tests.py
aarondr77: Yeah, tbh when I started I thought that more of them were going to map to multiple test cases... But I think once we start doing inline code completion tests, a lot more of these will get reused. I could imagine each `CodeGenTestCaseCore` is used by a 1) Chat Input Test, 2) In line code completion test that starts with a prefix 3) In line code completion test that starts with a comment. 

So I'm hoping that it becomes more reused. Let's see if that's the case and if not, it will be easy to just inline these again.

PR #1403, opened 2024-12-09
Smart Debugger Evals

aarondr77:
# Description

First 45 smart debugger tests

- [ ] When the user overwrites the print function, the response should give back the correct code and also contain the text 'restart" "kernel"

### Testing Package Installation

I tried for a while to support testing package installations and ran into a few problems. Primarily:
1. In order to run package installation commands like `!pip install matplotlib`, we need to be inside of a ipython kernel. This is doable and maybe we should move to this anyways because it more closely mirrors Jupyter. 
2. But, `!pip install` commands are not an in-memory operation—it’s a file system operation. Running it places files into the environment’s site-packages. That means, when we run a  `!pip install` command for the expected code, it ends up storing that package forever. This means successive runs of our test suite won't actually test if the Ai generated the right code without first resetting the environment. 

For now, let's not test package installs to avoid this added complexity. There's still a lot of smart debugging tests that we can write for other sorts of debugging! 

### Evals that we should create later. These are more complicated. 
1. Jupyter specific issues that arise from running cells in the wrong order. Often, this is a variable undefined error. 
2. Installing packages. 

# Testing

# Documentation

### DIFF
--- a/evals/eval_types.py
+++ b/evals/eval_types.py
@@ -41,7 +41,16 @@ class SmartDebugTestCase:
     notebook_state: NotebookState
     invalid_code: str
     correct_code: str
-    tags: List[Literal['simple', 'function', 'pandas', 'import', 'typo', 'type_error', 'logic_correction']] 
+    tags: List[Literal[
+        'simple', 
+        'function', 
+        'pandas', 
+        'import', 
+        'typo', 
+        'type_error', 
+        'logic_correction', 
+        'matplotlib'
+    ]] 
     variables_to_compare: Optional[List[str]] = None
 
 
--- a/evals/eval_types.py
+++ b/evals/eval_types.py
@@ -48,6 +48,8 @@ class SmartDebugTestCase:
         'import', 
         'typo', 
         'type_error', 
+        'value_error',
+        'argument_error',
         'logic_correction', 
         'matplotlib'
     ]] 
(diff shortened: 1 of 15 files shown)


aarondr77:
@ngafar when you review, feel free to add the last 5 tests.

ngafar (approved):

evals/test_cases/smart_debug_tests/import_tests.py
```diff
@@ -27,6 +27,19 @@
...
+        notebook_state=EMPTY_NOTEBOOK,
+        invalid_code="""
+import pandas as ps
+df = pd.DataFrame({'A': [1, 2, 3], 'B': [4, 5, 6]})
+""",
+        correct_code="""
+import pandas as pd
+df = pd.DataFrame({'A': [1, 2, 3], 'B': [4, 5, 6]})
+""",
+        tags=['import']
```
aarondr77, line 41: Let's mark this a typo

evals/eval_types.py
```diff
@@ -48,6 +48,8 @@ class SmartDebugTestCase:
         'import', 
         'typo', 
         'type_error', 
+        'value_error',
+        'argument_error',
```
aarondr77, line 52: We should make sure to apply these to the rest of the tests too

PR #1406, opened 2024-12-10
Improve smart debugger prompt

aarondr77:
# Description

- Improves the smart debugger prompt by 27.5% from 58% to 74%
- Makes sure that we're using the same AI config in the evals and mito-ai by switching to the correct model and adding a system message. 
- Adds some helpful utils
- Since this prompt asks the AI to think a little bit before responding, we need to parse out the thoughts before adding it to the chat history manager. We don't want to show it to the user, and for now, we don't need to store it in the chat. 

Screenshot 2024-12-10 at 4 39 26 PM

Screenshot 2024-12-10 at 3 19 55 PM

# Testing

Run the evals. 

# Documentation


evals/test_runners/utils.py
```diff
@@ -30,3 +31,12 @@ def exec_code_and_get_globals_and_output(code: str) -> Tuple[Dict[str, Any], str
         raise e
     
     return globals, actual_output_str
+
+
+def remove_process_pausing_code_lines(code: str) -> str:
```
ngafar, line 36: Love this. Feels so unnatural to make a plot and not do `plt.show()`.

ngafar:
Looks good. I got similar results locally:

- prod_prompt_v1: 56%
- prod_prompt_v2: 72%

evals/test_runners/utils.py
aarondr77: Yeah. I created this because the AI code was sometimes returning `plt.show()` in it which was causing the tests to stop running.

PR #1407, opened 2024-12-10
[WIP] Mito AI: Chat Q&A Mode

ngafar:
# Description

Updated the prompt so that it no longer returns a code block every time. Instead Mito AI will now determine if a question requires code to be written, or if info is all that is needed. 

# Testing

# Documentation

N/A


mito-ai/src/Extensions/AiChat/PromptManager.tsx
```diff
@@ -7,7 +7,9 @@ export function createBasicPrompt(
 ): string {
     const prompt = `You are an expert python programmer writing a script in a Jupyter notebook. You are given a set of variables, existing code, and a task.
 
-Respond with the updated active code cell and a short explanation of the changes you made.
+There are two possible types of responses you might give:
+1. Code Update: If the task requires modifying or extending the existing code, respond with the updated active code cell and a short explanation of the changes made. 
+2. Explanation/Information: If the task does not require a code update, provide an explanation, additional information about a package, method, or general programming question, without writing any code.
```
aarondr77, line 12: Maybe we want to throw the word "concise" into the second one. A few of the responses I'm getting a quite long. "Provide a concise explanation"

aarondr77 (approved):
This is awesome. A ton of value unlocked for a small change. 

I think the last step here is to move this prompt over to the Evals and just make sure that this doesn't lead to a huge regression on our eval performance. I doubt it will, but it's a good habit to get into! 

Feel free to do this and LGTM without another review.

mito-ai/src/Extensions/AiChat/PromptManager.tsx
ngafar: Good catch. Added a sentence to the end of the second instruction. Dropped the character count of the response by ~30%. 

Screenshot 2024-12-11 at 10 29 53 AM

PR #1411, opened 2024-12-11
Added CMD+E test

ngafar:
# Description

Test to make sure AI taskpane shortcut (CMD+E) is working.  

# Testing

Run `CMD+E opens the AI chat` test.

# Documentation

N/A

### DIFF
--- a/tests/mitoai_ui_tests/mitoai.spec.ts
+++ b/tests/mitoai_ui_tests/mitoai.spec.ts
@@ -274,6 +274,7 @@ test.describe('Mito AI Chat', () => {
   });
 
   test('CMD+E opens the AI chat', async ({ page }) => {
+    await page.keyboard.press('Meta+B'); // Close the sidebar
     await page.keyboard.press('Meta+E');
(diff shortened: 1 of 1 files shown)


tests/mitoai_ui_tests/mitoai.spec.ts
```diff
@@ -272,5 +272,10 @@ test.describe('Mito AI Chat', () => {
     await page.keyboard.type("@none_type_col_A");
     await expect(page.locator('.chat-dropdown-item-name').filter({ hasText: 'none_type_col_A' })).toBeVisible();
   });
+
+  test('CMD+E opens the AI chat', async ({ page }) => {
```
aarondr77, line 276: Let's just update this test to make sure that the chat taskpane is closed to begin with. If it is, we can close it first and then open it with CMD + E

aarondr77 (commented):
Tiny fixup and then LGTM. 

Nice add. Having all these little nice polish features for the user is nice, but also makes it harder to maintain. Tests are the only way!

PR #1416, opened 2024-12-12
[mito-ai] Use only websocket

fcollonval:
# Description

This simplify the code and API of mito-ai by dropping the REST endpoint in favor of the websocket connection for the chat and the inline completion.

Simplifications:
- Remove REST handler
- Simplify the websocket messaging by sending directly the OpenAI request messages from the front end - this will ease sharing element between the chat panel and the inline provider.
- Use websocket client also in Chat panel (it means that there is two separate websocket openend - one for the panel and one for the inline completion)

The consequence is a fallback to the mito server to request completion in case no `OPENAI_API_KEY` is set by the user.

If the user request the mito server for inline completion, the inline completion won't be streamed and only the final suggestion will be provided.

> I remove a `test.only` statement that is currently preventing all integration tests to be executed for mito-ai.

Add status bar item in which are reported error and configuration:

image

# Testing

I checked the tests are all passing. And tested manually with and without the personal key.

# Documentation

Add note in mito-ai README about how to use the traitlets to configure the extension.


mito-ai/src/utils/handler.ts
```diff
@@ -1,120 +0,0 @@
...
-        }
-    }
-    if (response.status === 500) {
-        // This 500 error is set by the OpenAICompletionHandler class in the mito-ai python package. It is a 
-        // generic error that is set when we haven't handled the error specifically.
-        return {
-            type: 'error',
-            errorMessage: "There was an error communicating with OpenAI. This might be due to an incorrect API key, a temporary OpenAI outage, or a problem with your internet connection. Please try again.",
-        }
-    }
```
aarondr77, line 87: These three error messages have been pretty important for communicating with users what they need to do to fix their Mito AI installation. Can we update the web socket to return similar errors so we can display the correct messages to the user?

For example, currently on this PR if the mito_ai server extension is not enabled, instead of showing the message in the chat that tells them how to fix the issue, it doesn't show anything. 

For now, we can just handle these errors in the chat and can ignore what happens when we receive one of these messages from the inline code completer to simplify things.

mito-ai/mito_ai/providers.py
```diff
@@ -1,70 +1,81 @@
...
+    @property
+    def can_stream(self) -> bool:
+        """Whether the provider supports streaming completions.
+
+        Streaming is only supported if an OpenAI API key is provided.
+        """
+        return bool(self.api_key)
+
     @property
     def client(self) -> AsyncOpenAI:
```
aarondr77, line 80: I think that this property makes less sense now that we support using the `mito-ai-server` if no open ai key is set. Maybe we should update the return type to be `Union[AsyncOpenAI, None]` and have it `None` when no OpenAI Key is set. Then in the `request_completions` function we can just case on the `self.client is not None` instead of checking for the open ai key again. 

And maybe we can update the name of the property to `user_key_client` or something like that to make it more clear when this client is useful. 

What do you think?

aarondr77 (changes requested):
Nice work unifying these two approaches. I think there's two things to fix up before merging: 

1. Just renaming/reorganizing a few files and types to make things a bit more clear
2. Adding back some important error handling that we lost when deleting the error codes.

mito-ai/src/utils/models.ts
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * The chat completion messages history.
+   */
+  messages: OpenAI.Chat.ChatCompletionMessageParam[];
+  /**
+   * Whether to stream the completion or not.
+   */
+  stream: boolean;
+}
+
```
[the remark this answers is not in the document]
aarondr77: Or maybe even better is `IWebsocketMessageRequest` because we also are heavily using the word `message`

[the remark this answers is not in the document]
fcollonval: I kept the word `completion` because it is the wording used by the OpenAI API we are calling. As in the future we may call other kind of API, it felt better than a too generic `MessageRequest`. Let's discuss that.

mito-ai/src/utils/models.ts
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Mito AI completion reply.
+ */
+export interface ICompletionReply {
+  /**
+   * The type of the message.
+   */
+  type: 'inline_completion';
```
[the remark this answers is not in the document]
fcollonval: Good catch - I went through all _type_s to improve the wording used; ICompletionReply -> reply, ICompletionStreamChunk -> chunk

mito-ai/mito_ai/logger.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def get_logger() -> logging.Logger:
+    global _LOGGER
+    if _LOGGER is None:
+        app = Application.instance()
+        _LOGGER = logging.getLogger("{!s}.mito_ai".format(app.log.name))
+        Application.clear_instance()
+
+    return _LOGGER
```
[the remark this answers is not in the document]
fcollonval: This is a trick to get context in jupyter-server logs:

```
[D 2024-12-16 15:44:17.991 ServerApp.mito_ai] Requesting completion from Mito server.
```

making filtering easier :wink:

mito-ai/mito_ai/providers.py
```diff
@@ -1,70 +1,81 @@
...
+        config=True,
+        help="An upper bound for the number of tokens that can be generated for a completion, including visible output tokens and reasoning tokens.",
+    )
+
     # FIXME add validate function to check if the model is valid
     model = Unicode(
         "gpt-4o-mini", config=True, help="OpenAI model to use for completions"
     )
 
+    temperature = CFloat(
```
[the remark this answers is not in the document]
fcollonval: Those are traitlets. So you can change them using the command line; e.g.

```sh
jupyter lab --OpenAIProvider.temperature=0.5
```

I did not yet add doc for this because I did not test them yet :wink:

mito-ai/src/utils/handler.ts
fcollonval: Yep that changes is the major reason I kept this PR as draft; I'll address that in priority.

mito-ai/mito_ai/providers.py
fcollonval: I would go for `_openAI_client` if it is ok for you - the reason behind that is that it is local to the provider.

mito-ai/src/utils/handler.ts
aarondr77, line 87: Similarly, the error message we are displaying to users for this is not great right now. If I turn off my wifi and send a message in the chat, I get this error. Can we move it back to what we had originally?

And let's also update it in the status item too

Screenshot 2024-12-20 at 12 04 51 PM

mito-ai/README.md
```diff
@@ -18,6 +18,26 @@ To install the extension, execute:
 pip install mito-ai
 ```
 
+## Configuration
```
aarondr77, line 21: Thanks for adding to the docs!

mito-ai/src/Extensions/InlineCompleter/provider.ts
```diff
@@ -218,11 +203,16 @@ export class MitoAIInlineCompleter
...
         );
       }
 
-      return result.list;
+      return {
+        items: result.items.map(item => ({
+          ...item,
+          insertText: this._cleanCompletion(item.content)
+        }))
+      };
```
aarondr77, line 215: I'm having this happen pretty often where it looks like the server is sending completions, but I'm not seeing any suggestions in the code cell. And then if I like delete a character and add it back or something like that, the completion will show up next time. 

Any thoughts on what's going on here?

https://github.com/user-attachments/assets/da723e71-f910-4fda-a926-8977ece321fb

aarondr77 (commented):
This is working pretty well. There's a bit of polish that I want to do to the error messages before we merge it into dev. I tried to be more clear in these comments about what I think the error messages should say. Let me know if you agree / have better wordings / questions. 

There's also a few issues that I documented here https://github.com/mito-ds/mito/issues/1433 but I think we should get this merged into dev and then we can handle these separately before we release.

tests/mitoai_ui_tests/statusItem.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    await page.goto(`tree/${tmpPath}`);
+
+    await page.getByRole("button", { name: "Mito AI Status" }).click();
+
+    await expect(page.locator(".mito-ai-status-popup")).toHaveText(
+      "Mito AI StatusProvider: Mito serverConfiguration:Model: a-powerful-modelMax tokens: 20Temperature: 0.7"
+    );
+  });
+
+  test("should show the latest error", async ({ page, tmpPath }) => {
```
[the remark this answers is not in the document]
fcollonval: No it is very similar but this one in addition sends a fake OpenAI authentication error. see line 70

mito-ai/src/Extensions/InlineCompleter/provider.ts
fcollonval: I'll need to investigate this - this gonna take time unfortunately. But I guess it has something to do with the interaction between Mito completer and the JupyterLab completer handler.

Just for my information, which version of JupyterLab were you using?

fcollonval:
Hey @aarondr77 every thing looks good on my side - I don't know why the mitosheet are failing.

mito-ai/src/Extensions/InlineCompleter/provider.ts
aarondr77: No worries, I'm on jupyterlab 4.3.3

PR #1418, opened 2024-12-12
First 5 Inline code completion evals

aarondr77:
# Description

- Adds the first 5 inline code completion evals
- Adds the production prompt v1 for benchmarking 

# Testing

Run it! 

# Documentation

No.


evals/test_cases/inline_code_completion_tests/variable_tests.py
```diff
@@ -0,0 +1,92 @@
...
+    InlineCodeCompletionTestCase(
+        name="empty_notebook_variable_declaration",
+        test_case_core=CodeGenTestCaseCore(
+            notebook_state=EMPTY_NOTEBOOK,
+            expected_code="x=1",
+            workflow_tags=["variable_declaration"],
+        ),
+        prefix="""
+# Create a variable x and set it equal to 1
+x =""",
```
ngafar, line 14: Probably not an issue, but if I were using an inline completer and I had the following line:

```python
# Create a variable x and set it equal to 1
```

I wouldn't even think to type `x = `, I'd just assume the completer would do it.

ngafar:
One small comment, but otherwise LGTM.

evals/test_cases/inline_code_completion_tests/variable_tests.py
aarondr77: Good point. I made a new eval for that one

PR #1420, opened 2024-12-13
Expand Logging: Add New Top-Level Logs

ngafar:
# Description

Expanded logging to consider input location. Each log now has context appropriate fields. For example, the smart debugger with log the error and error type, while the sidebar chat will log the users query.  

The main logging changes can be found here: https://github.com/mito-ds/mito/blob/4ad1a0432b3ce110cdf53965e66d31dad48f20f1/mito-ai/mito_ai/utils/telemetry_utils.py#L132

A few things to consider:

- Logging only works for key type is `USER_KEY`. Server logging will be implemented in the next PR. 
- Mixpanel logging is [quite limited](https://docs.mixpanel.com/docs/data-structure/property-reference/data-type#string). To say on track, timeline wise, I've opted to ignore these constraints. After the other PRs in the shape are completed, I'll make a second pass, and see what can be done. For now, the logs should be good enough. 

# Testing

Checkout this branch, use the sidebar chat, smart debug, and the explain features. Each should have a corresponding log. 

# Documentation

N/A

### DIFF
--- a/mito-ai/mito_ai/utils/open_ai_utils.py
+++ b/mito-ai/mito_ai/utils/open_ai_utils.py
@@ -4,6 +4,7 @@
 # Copyright (c) Saga Inc.
 
 import os
+import json
 import requests
 from typing import Any, Dict, List
 from .db import get_user_field, set_user_field
@@ -141,11 +142,55 @@ def get_open_ai_completion(
             # If they DO have an Open AI key, use it to get a completion
             response = _get_ai_completion_with_key(ai_completion_data, OPENAI_API_KEY)
 
-            # Log the successful completion
-            log(
-                MITO_AI_COMPLETION_SUCCESS,
-                params={KEY_TYPE_PARAM: USER_KEY, "input_location": input_location},
-            )
+            if input_location == "smartDebug":
+                log(
+                    "smart_debug_success",
+                    params={
+                        KEY_TYPE_PARAM: USER_KEY,
+                        "error_message": last_message_content,
+                        "response": response,
+                    },
+                )
+            elif input_location == "codeExplain":
+                log(
+                    "code_explain_success",
+                    params={
+                        KEY_TYPE_PARAM: USER_KEY,
+                        "error_message": last_message_content.split("Error Message: ")[
+                            -1
+                        ],
+                        "error_type": last_message_content.split("Error Type: ")[
+                            -1
+                        ].split(": ")[0],
+                        "response": response,
+                    },
+                )
+            elif input_location == "sidebar":
+                log(
+                    "sidebar_success",
+                    params={
+                        KEY_TYPE_PARAM: USER_KEY,
+                        "user_input": last_message_content.split("Your task: ")[-1],
+                        "code_cell_input": json.dumps(
+                            last_message_content.split("Code in the active code cell:")[
+                                -1
+                            ]
+                            .strip()
+                            .split("```python")[1]
+                            .strip()
+                            .split("```")[0]
+                        ),
+                        "response": response,
+                    },
+                )
+            else:
+                log(
+                    f"{input_location}_success",
+                    params={
+                        KEY_TYPE_PARAM: USER_KEY,
+                        "note": "This input_location has not been accounted for in open_ai_utils.py. Please add it to the codebase.",
+                    },
+                )
             return response
     except Exception as e:
         key_type = MITO_SERVER_KEY if OPENAI_API_KEY is None else USER_KEY
--- a/mito-ai/mito_ai/utils/open_ai_utils.py
+++ b/mito-ai/mito_ai/utils/open_ai_utils.py
@@ -104,9 +104,7 @@ def _get_ai_completion_from_mito_server(
     return res.json()
 
 
-def get_open_ai_completion(
-    messages: List[Dict[str, Any]], input_location: str
-) -> Dict[str, Any]:
+def get_open_ai_completion(messages: List[Dict[str, Any]], prompt_type: str) -> Dict[str, Any]:
 
     initialize_user()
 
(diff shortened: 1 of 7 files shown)


ngafar:
@aarondr77 when you review this, I've added a `.skip` to one of the tests in [tests/mitoai_ui_tests/aiInlineCompleter.spec.ts](https://github.com/mito-ds/mito/blob/71f562ed86a11ee78e0c4c63ed84e38981eeeb1f/tests/mitoai_ui_tests/aiInlineCompleter.spec.ts), which was breaking seemingly the entire CI pipeline. Would be great if you could take a look.

mito-ai/mito_ai/utils/open_ai_utils.py
```diff
@@ -10,125 +10,140 @@
...
+    MITO_SERVER_KEY,
+    USER_KEY,
+    MITO_AI_COMPLETION_SUCCESS,
+    MITO_AI_COMPLETION_ERROR,
+    MITO_SERVER_NUM_USAGES,
+    MITO_SERVER_FREE_TIER_LIMIT_REACHED,
+)
+
+OPEN_AI_URL = "https://api.openai.com/v1/chat/completions"
+MITO_AI_URL = "https://ogtzairktg.execute-api.us-east-1.amazonaws.com/Prod/completions/"
```
ngafar, line 26: This was mostly just formatting.

mito-ai/mito_ai/utils/telemetry_utils.py
```diff
@@ -126,3 +127,75 @@ def log(
...
+        error_type = error_message.split(": ")[0]
+
+        log(
+            "smart_debug_success",
+            params={
+                KEY_TYPE_PARAM: key_type,
+                "code_cell_input": code_cell_input,
+                "error_message": error_message,
+                "error_type": error_type,
+                "response": response,
```
aarondr77, line 170: Hmm so actually, it looks like Mixpanel has a [255 character limit for string ](https://docs.mixpanel.com/docs/data-structure/property-reference/data-type#string) and so I'm not actually seeing the code that we respond with in this response because in the smart debug, we respond with some thought tokens first. But even long code cells for the chat input would get cut off. 

That's a bummer. 

It looked into it, and Segment does not have the same limitation, so it's just Mixpanel that is bottlenecking us right now.

aarondr77: Maybe we will eventually move away from Mixpanel for our Ai logging... but for now to keep us on track. Maybe we can use a function like this for the prompt and response. 

This function just takes a response and breaks it into chunks of 250 characters with the fields response_part_1, response_part_2, etc.

Thanks Claude :) 

```python
def chunk_response(response):
    """
    Split a response string into chunks of 250 characters.
    
    Args:
        response: The string to be chunked
        
    Returns:
        dict: A dictionary with keys 'response_part_1', 'response_part_2', etc.
    """
    chunk_size = 250
    chunks = {}
    
    # Handle empty or None response
    if not response:
        return {}
    
    # Calculate number of chunks needed
    num_chunks = (len(response) + chunk_size - 1) // chunk_size
    
    # Create chunks
    for i in range(num_chunks):
        start = i * chunk_size
        end = min(start + chunk_size, len(response))
        chunks[f'response_part_{i + 1}'] = response[start:end]
    
    return chunks
```

aarondr77: And while we're doing this we can remove `{"completion":` from the logged response

tests/mitoai_ui_tests/aiInlineCompleter.spec.ts
```diff
@@ -4,7 +4,7 @@ import { PromiseDelegate } from "@lumino/coreutils";
 const GHOST_SELECTOR = ".jp-GhostText";
 
 test.describe("first time setup", () => {
-  test("should ask the user to activate the inline completion", async ({
+  test.skip("should ask the user to activate the inline completion", async ({
```
aarondr77, line 7: Sounds good. We'll fix this up in a separate PR soon.

aarondr77 (changes requested):

tests/mitoai_ui_tests/aiInlineCompleter.spec.ts
ngafar: Thanks! Created an issue so we don't loose track: https://github.com/mito-ds/mito/issues/1426

mito-ai/mito_ai/utils/telemetry_utils.py
```diff
@@ -126,3 +127,75 @@ def log(
...
+    elif input_location == "sidebar":
+        user_input = last_message_content.split("Your task: ")[-1]
+        log(
+            "sidebar_success",
+            params={
+                KEY_TYPE_PARAM: key_type,
+                "user_input": user_input,
+                "code_cell_input": code_cell_input,
+                "response": response,
+            },
```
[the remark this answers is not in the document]
ngafar: Any reason why we'd want the full prompt? We already have the code in the cell and the user's input, the onyl two things that change. The rest of the prompt would be the same standard template.

mito-ai/mito_ai/utils/telemetry_utils.py
```diff
@@ -126,3 +127,75 @@ def log(
...
+        response: The response received from the AI
+    """
+
+    code_cell_input = json.dumps(
+        last_message_content.split("Code in the active code cell:")[-1]
+        .strip()
+        .split("```python")[1]
+        .strip()
+        .split("```")[0]
+    )
```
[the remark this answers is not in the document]
ngafar: This is probably the best solution in the long run, but for now I've added the following types which should cover our bases:

```ts
type PromptType = 'chat' | 'smartDebug' | 'codeExplain' | 'system'

export interface IAIOptimizedChatHistory {
    promptType: PromptType
    codeCellID: string | undefined
}
```

mito-ai/mito_ai/utils/telemetry_utils.py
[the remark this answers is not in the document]
aarondr77: I think mostly just so we can associate with which prompt it was generated by so we can like match the Ai accuracy in the wild to the specific prompt that produced it and we can track that metric over time. So just the prompt id would be totally fine then too.

mito-ai/mito_ai/utils/telemetry_utils.py
```diff
@@ -80,6 +81,32 @@ def identify() -> None:
         except Exception as e:
             pass
 
+def chunk_param(param: str, param_name: str, chunk_size: int=250) -> Dict[str, str]:
```
aarondr77, line 84: lovely!

aarondr77 (approved):

PR #1427, opened 2024-12-17
Added Preview Code Shortcut

ngafar:
# Description

CMD+Y now triggers preview, when appropriate. Also, moved shortcut tests to their own file. 

Note that the way key bindings are registered have been changed. We are now disposing bindings every time `codeReviewStatus` is updated. Here's a quick Claude explanation of why this is best practice:

1. Memory Leaks: Without disposing of key bindings, each time the useEffect runs (whenever codeReviewStatus changes), new key bindings are added without removing the old ones. This can lead to memory leaks and multiple bindings for the same keys.
2. Duplicate Handlers: Multiple active bindings for the same key combination can cause unexpected behavior, with multiple handlers firing for a single keystroke.
3. Cleanup Best Practices: React's useEffect cleanup function (returned function) is the proper way to clean up side effects when a component unmounts or when dependencies change.

# Testing

Send some messages, try the shortcuts. 

# Documentation

Yes, not now, but after the next release we should update this page: https://docs.trymito.io/mito-ai/chat

### DIFF
--- a/tests/mitoai_ui_tests/shortcuts.spec.ts
+++ b/tests/mitoai_ui_tests/shortcuts.spec.ts
@@ -15,13 +15,13 @@ test.describe('Mito AI Shortcuts', () => {
         await waitForIdle(page);
     });
     
-    test.only('Open the AI chat pane', async ({ page }) => {
+    test('Open the AI chat pane', async ({ page }) => {
         await page.keyboard.press(`${modifierKey}+B`); // Close the sidebar
         await page.keyboard.press(`${modifierKey}+E`);
         await expect(page.locator('.chat-input')).toBeVisible();
     });
 
-    test.only('Preview AI generated code', async ({ page }) => {
+    test('Accept AI generated code', async ({ page }) => {
         await sendMessageToMitoAI(page, 'Write the code df["C"] = [7, 8, 9]');
 
         // Preview the code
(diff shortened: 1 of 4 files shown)


tests/mitoai_ui_tests/shortcuts.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+const modifierKey = process.platform === 'darwin' ? 'Meta' : 'Control';
+
+test.describe('Mito AI Shortcuts', () => {
+
+    test.beforeEach(async ({ page }) => {
+        await createAndRunNotebookWithCells(page, ['import pandas as pd\ndf=pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})']);
+        await waitForIdle(page);
+    });
+    
+    test.only('Open the AI chat pane', async ({ page }) => {
```
aarondr77, line 18: Remove `.only`

tests/mitoai_ui_tests/shortcuts.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    test.only('Preview AI generated code', async ({ page }) => {
+        await sendMessageToMitoAI(page, 'Write the code df["C"] = [7, 8, 9]');
+
+        // Preview the code
+        await page.keyboard.press(`${modifierKey}+Y`);
+
+        // Code diffs should be visible after the user clicks preview
+        await expect(page.locator('.cm-codeDiffRemovedStripe')).toBeVisible();
+        await expect(page.locator('.cm-codeDiffInsertedStripe')).toBeVisible();
+    });
```
aarondr77, line 33: Maybe we can combine this with the next test.

aarondr77 (commented):
Awesome. I'll do this tiny cleanup and then merge it.

PR #1430, opened 2024-12-19
Auto-Scrolling on New Message

ngafar:
# Description

Whenever a new message is sent the page is automatically scrolled down, to make sure that the last question is "in frame."  

Also, be sure to check the shorter message, for when the taskpane is narrow. This was introduced in the last PR. 

# Testing

Send enough messages, so that there is a scroll bar. Make sure that the page scrolls down when you send messages. 

# Documentation

N/A

PR #1431, opened 2024-12-19
Active cell preview

aarondr77:
# Description

- Adds the cell preview into the chat input. Now, as soon as the user starts typing in the chat input, they see the cell that is getting sent as context. 
- When the user switches active code cells, it is updated in the chat input. 
- When the user clicks away from the chat input and there is no message typed out, the preview goes away. 
- The variable dropdown now sits onto of the cell preview to keep it closer to the input. 

Just in a tiny bit of testing that I've done, having the active cell preview already helped me catch a bug in my message. Nice! 

@ngafar there is a bit of weirdness here with the the chat input covering the message behind it when you submit the prompt. But I think this is going to be fixed in your PR, right?

https://github.com/user-attachments/assets/e7af36b5-ad07-4589-8c46-9bda344bee7b

# Testing

- Try those workflows above. 

# Documentation

No


mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -23,14 +31,40 @@ const ChatInput: React.FC<ChatInputProps> = ({
...
+    variableManager,
+    notebookTracker,
+    renderMimeRegistry,
 }) => {
     const [input, setInput] = useState(initialContent);
     const [expandedVariables, setExpandedVariables] = useState<ExpandedVariable[]>([]);
+    const textAreaRef = React.useRef<HTMLTextAreaElement>(null);
+    const [isFocused, setIsFocused] = useState(false);
+
+    // Add state to track active cell
```
aarondr77, line 43: Useless comment

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -23,14 +31,40 @@ const ChatInput: React.FC<ChatInputProps> = ({
...
+        // many times. To avoid a bunch of rerenders, we disconnet the listener 
+        // each time we use it and then recconect when we're done updating the active cell ID
+        notebookTracker.activeCellChanged.connect(activeCellChangedListener);
+    
+        return () => {
+            notebookTracker.activeCellChanged.disconnect(activeCellChangedListener);
+        };
+    }, [notebookTracker, activeCellID]);  
+
+    // State for the variable dropdown
```
aarondr77, line 64: Move these state variables up

aarondr77 (commented):
Small cleanup

ngafar:
> there is a bit of weirdness here with the the chat input covering the message behind it when you submit the prompt. But I think this is going to be fixed in your PR, right?

@aarondr77 Yup, once https://github.com/mito-ds/mito/pull/1430 gets merged, we should auto-scroll so that the last message is always shown.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -142,55 +171,92 @@ const ChatInput: React.FC<ChatInputProps> = ({
...
-                }}
-            />
+                    }}
+                />
+                {isDropdownVisible && isFocused && (
+                    <ChatDropdown
+                        options={expandedVariables}
+                        onSelect={handleOptionSelect}
+                        filterText={dropdownFilter}
+                        position={showDropdownAbove ? "above" : "below"}
```
ngafar, line 248: Since ChatInput is now always fixed to the bottom of the page, I think we can simplify this file by removing the logic for rendering the variable dropdown. It should always be above. Thoughts?

mito-ai/src/Extensions/CellToolbarButtons/CellToolbarButtonsPlugin.tsx
```diff
@@ -38,6 +38,7 @@ const CellToolbarButtonsPlugin: JupyterFrontEndPlugin<void> = {
             },
             isVisible: () => notebookTracker.activeCell?.model.type === 'code' && app.commands.hasCommand(COMMAND_MITO_AI_SEND_EXPLAIN_CODE_MESSAGE)
         });
+        console.log("mito-ai: CellToolbarButtonsPlugin activated");
```
ngafar, line 41: Do we want to keep console logs here, or was this for debugging?

mito-ai/src/Extensions/ErrorMimeRenderer/ErrorMimeRendererPlugin.tsx
```diff
@@ -45,6 +45,7 @@ const ErrorMimeRendererPlugin: JupyterFrontEndPlugin<void> = {
                 }
             }, -1);  // Giving this renderer a lower rank than the default renderer gives this default priority
         }
+        console.log("mito-ai: ErrorMimeRendererPlugin activated");
```
ngafar, line 48: Same as previous comment.

ngafar (approved):

ngafar:
I noticed a small issue when working with two cells that both contain content. If I select the first cell and click into the ChatInput area, the preview displays as expected. However, when I switch to the second cell, the contents of the second cell briefly flash in the context preview area before disappearing.

mito-ai/src/Extensions/CellToolbarButtons/CellToolbarButtonsPlugin.tsx
aarondr77: I think it might be nice to have these, so when things aren't working right we can look at the console logs and see if all the extensions started up properly.

aarondr77:
@ngafar I'm running into a flickering issue where when I have code in the preview and then when I switch which is the active cell, the taskpane flickers. If you take a QuickTime video of your screen and watch it in slow motion, you can see that this happens because the active cell preview is rendered with no code in between before switching to the new code preview. But I'm having a hard time tracking down the source of this. Do you mind taking a look?

aarondr77:
@ngafar I'm still seeing the flickering here 

https://github.com/user-attachments/assets/58821349-9aee-471a-a78b-4ddf04e21ad4

PR #1436, opened 2024-12-23
New chat buttons

aarondr77:
# Description

New buttons in the chat taskpane that will grab the user's attention more. 

# Testing

Try it out + notice that the tests are passing. 

# Documentation

Yes, when we release we should update the images.


ngafar:
Looks good. My one nitpick (and this is the pickiest of nitpicks): in my mind, hovering always makes the button darker.

ngafar (approved):

PR #1437, opened 2024-12-23
Accept reject in cell

aarondr77:
# Description

Adds the accept and reject buttons into the cell toolbar. 

Screenshot 2024-12-26 at 6 06 14 PM

Note that the cell toolbar is only displayed when there is space for it. If it overlaps with the code too much, then the buttons are not displayed. 

Also does a few other things:
- Switches the reject keyboard shortcut to `cmd+U` because `cmd+D` is already a keyboard shortcut that is used by code mirror to delete a line of code. When in cell editing mode, if you press cmd+D instead of rejecting, it deleted the current line. 
- Removes the unified code diffs from the state. We don't really need to be storing this in state at all. Now, we just pass it to the update function when we want to update the display. Simplifying the rerenders triggered by use effects is helpful I think. It makes the code easier to follow. 
- Removes some other unused code. 

# Testing

Try it out! And see the tests! 

# Documentation

Yes, we should update when we release.


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -295,8 +299,10 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
     const displayOptimizedChatHistory = chatHistoryManager.getDisplayOptimizedHistory()
 
     const previewAICode = () => {
+        console.log('PREVIEWING CODE')
```
aarondr77, line 302: remove

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -408,11 +416,65 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
         };
     }, [codeReviewStatus]);
 
+    const updateCellToolbarButtons = (cellStateBeforeDiff: React.MutableRefObject<ICellStateBeforeDiff | undefined>) => {
+        console.log("IN HERE 0")
```
aarondr77, line 420: remove

aarondr77: all of these

tests/mitoai_ui_tests/utils.ts
```diff
@@ -58,12 +71,37 @@ export const clickPreviewButton = async (page: IJupyterLabPageFixture) => {
...
-    await page.getByRole('button', { name: 'Accept code' }).click();
+export const clickAcceptButton = async (
+    page: IJupyterLabPageFixture,
+    // Express the useCellToolbar option like this so that they are keyword arguments in the tests
+    // and are therefore easy to read!  
+    { useCellToolbar = false }: { useCellToolbar?: boolean } = {useCellToolbar: false}
+) => {
+    if (useCellToolbar) {
+        await closeMitoAIChat(page);
+        await page.getByLabel('notebook content').getByText('Accept code').click();
```
aarondr77, line 82: Update

tests/mitoai_ui_tests/utils.ts
```diff
@@ -58,12 +71,37 @@ export const clickPreviewButton = async (page: IJupyterLabPageFixture) => {
...
+    page: IJupyterLabPageFixture,
+    // Express the useCellToolbar option like this so that they are keyword arguments in the tests
+    // and are therefore easy to read!  
+    { useCellToolbar = false }: { useCellToolbar?: boolean } = {useCellToolbar: false}
+) => {
+    if (useCellToolbar) {
+        await closeMitoAIChat(page);
+        // For some reason, the same selector doesn't work here. It ends up just scrolling the page
+        // up and down a bunch of times until it times out. So instead, we use this selector 
+        await page.getByLabel('notebook content').getByText('Reject code').click();
```
aarondr77, line 100: Update

ngafar (approved):

Issue #1433, opened 2024-12-20
Inline code completion new line and indentation issues 


aarondr77:
### New Line 
Notice that the new line completion starts on the next line, which is invalid code. 
<img width="836" alt="Screenshot 2024-12-20 at 11 05 56 AM" src="https://github.com/user-attachments/assets/77ff600e-80c3-4508-932b-d9d9987e13d3" />

### Indentation 
Notice that the docstring is not correctly indented and it causes the code to error.  

https://github.com/user-attachments/assets/3c97bb33-dae0-4e12-b68f-f0762791c89b

aarondr77:
The first thing we should probably do is move over our new inline code completion prompt and see if that changes any of these issues. Then lets iterate on the prompt and the post processing handling to fix this.

PR #1438, opened 2024-12-23
Fix: Inline code completion new line and indentation issues

ngafar:
# Description

Fix for https://github.com/mito-ds/mito/issues/1433 

Inline completions should now be formatted properly.

# Testing

Try it out by having the inline completion finish code that should 

1. be on the same line (for example, `print("he`) 
2. as well as code that should be on a new line (`# create a function that...`)

# Documentation

N/A


mito-ai/src/Extensions/InlineCompleter/provider.ts
```diff
@@ -346,15 +346,18 @@ export class MitoAIInlineCompleter
     this._fullCompletionMap.set(this._currentStream, fullCompletion);
 
     // Clean suggestion
-    let cleanedCompletion = fullCompletion.slice(0);
+    let cleanedCompletion = fullCompletion
+    // Remove opening ```python
+    .replace(/^```python\n?/, '')
+    // Remove closing ```
+    .replace(/```$/, '')
```
ngafar, line 353: With the v3 prompt code comes back formatted in markdown codeblock notation.

mito-ai/src/Extensions/InlineCompleter/provider.ts
```diff
@@ -346,15 +346,18 @@ export class MitoAIInlineCompleter
...
+    
     if (this._currentPrefix) {
       if (
         cleanedCompletion.startsWith(this._currentPrefix) ||
         this._currentPrefix.startsWith(cleanedCompletion)
       ) {
         cleanedCompletion = cleanedCompletion.slice(this._currentPrefix.length);
-      } else if (!['\n', ' '].includes(cleanedCompletion[0])) {
-        cleanedCompletion = '\n' + cleanedCompletion;
```
ngafar, line 357: This forces a newline regardless of what the LLM returns, not optimal.

mito-ai/mito_ai/utils/inline_prompt.py
```diff
@@ -0,0 +1,85 @@
...
+
+Output:
+```python
+```
+</Example 3>
+
+IMPORTANT: Notice in Example 3 that the output starts with a newline because the cursor appears at the end of a comment line.
+
+Your Task:
+
```
ngafar, line 78: Still need to add the global vars here to be in parity with the v3 prompt found in `evals/` and ideas on how best to do this?

mito-ai/mito_ai/providers.py
```diff
@@ -84,36 +85,11 @@ def _get_messages(
...
-            },
-            {
-                "role": "user",
-                "content": """Complete the following code responding only with additional code, 
-code comments or docstrings, and with no markdown formatting.""",
-            },
-            {"role": "user", "content": inputs["prefix"]},
+        prompt = inline_prompt.generate_prompt(inputs["prefix"], inputs["suffix"])
+        messages: List[ChatCompletionMessageParam] = [
+            {"role": "system", "content": COMPLETION_SYSTEM_PROMPT},
```
ngafar, line 90: Is keeping the system prompt here the best idea?

mito-ai/mito_ai/utils/inline_prompt.py
aarondr77: oooo I think we should move the prompt creation to the frontend so that we can access all of our things like the variable manager. So instead of passing each of these prompt pieces one by one, we can just pass the entire prompt to the backend, just like we do with the rest of our prompts.

mito-ai/src/Extensions/InlineCompleter/provider.ts
```diff
@@ -181,16 +180,22 @@ export class MitoAIInlineCompleter
...
       }
       const messageId = ++this._counter;
-
+      
       const prefix = this._getPrefix(request);
+      const suffix = this._getSuffix(request);
+      const variables = this._variableManager.variables;
+      const prompt = createInlinePrompt(prefix, suffix, variables);
+      const openAIFormattedMessages: OpenAI.Chat.ChatCompletionMessageParam[] = [
+        // {"role": "system", "content": COMPLETION_SYSTEM_PROMPT},
```
aarondr77, line 189: Let's delete this if we're not using it

mito-ai/src/Extensions/InlineCompleter/provider.ts
```diff
@@ -414,18 +348,11 @@ interactive outputs.`
...
-        cleanedCompletion.startsWith(this._currentPrefix) ||
-        this._currentPrefix.startsWith(cleanedCompletion)
-      ) {
-        cleanedCompletion = cleanedCompletion.slice(this._currentPrefix.length);
-      } else if (!['\n', ' '].includes(cleanedCompletion[0])) {
-        cleanedCompletion = '\n' + cleanedCompletion;
-      }
-    }
-    return cleanedCompletion;
+    return rawCompletion
```
aarondr77, line 351: Can we double check that it's handling new line characters at the start of the completion correctly. 

Maybe the AI is just not returning a new line character here, or maybe we are ignoring it. 
Screenshot 2024-12-27 at 11 50 30 AM

aarondr77: In some cases, it does appear to be working though 
Screenshot 2024-12-27 at 11 53 14 AM

mito-ai/src/Extensions/InlineCompleter/provider.ts
```diff
@@ -1,8 +1,6 @@
 import { Notification, showErrorMessage } from '@jupyterlab/apputils';
-import { IEditorMimeTypeService } from '@jupyterlab/codeeditor';
 import {
-  IEditorLanguageRegistry,
-  type IEditorLanguage
+  // type IEditorLanguage
```
aarondr77, line 3: remove

aarondr77 (commented):
Sweet this is a ginormous improvement in results. Nice work! 

Just a little bit of cleanup and then LGTM. I found a fair bit of scenarios where I didn't get the code that I wanted, but that's not the goal of this PR, so I'm just going to add them to our list of evals that we want to create.

mito-ai/src/Extensions/InlineCompleter/provider.ts
ngafar: Fixed this in-prompt. Tried to squeeze it into an existing instruction:

> 1. If the cursor appears at the end of a complete line (especially after a comment), ALWAYS start your code with a newline character

However, it just would not take. So now there's a specific line on formatting functions:

> 2. If the cursor appears at the end of a function definition, ALWAYS start your code with a newline character

PR #1447, opened 2024-12-27
[Investigation] Mito theme light

aarondr77:
# Description

I investigated creating our own theme for Jupyter. This could include a new color scheme as well as new css styling for different elements. In this investigation I did a few things. 

1. Added our own color scheme
2. Removed extra borders that I don't think look good. Especially the big dark grey border around the main area widget. 
3. Added extra padding to the code cell and rounded its corners

Screenshot 2024-12-27 at 4 21 52 PM

I started this based on the cookie cutter here https://github.com/jupyterlab/extension-template


aarondr77:
Do not merge this

PR #1448, opened 2024-12-27
Always show accept reject buttons

aarondr77:
# Description

Updates the cell toolbar css strategy so that when we want to display the Accept and Reject buttons we always can. 

Read the css docstring for a more thorough explanation. 
Screenshot 2024-12-27 at 4 53 20 PM

# Testing

Try it out + see tests. 

# Documentation

No


ngafar (approved):
This looks good. 

One potential idea: replace the text with icons on narrower screens. This way the text isn't blocking too much of the user's code.

PR #1455, opened 2024-12-30
No focus on smart debug

aarondr77:
# Description

Do not focus on the chat input when using the Smart Debug button. We have already sent the message, so the user does not need to type anything. Without doing this, the active cell preview appears and is confusing. 

# Testing

See test + try it out. 

# Documentation

No.


ngafar (approved):
Love these qol updates

PR #1466, opened 2025-01-03
Separated Inline and Chat Usage Checks

ngafar:
# Description

Separated the usage checks for inline completions and chat interactions. This ensures that users who have completed their trial period for inline completions but still have remaining chat messages can continue using the chat feature.

# Testing

Test the following cases with a server key:

- chat messages < 500 & first inline date < 30 days -chat ✅, inline ✅
- chat messages > 500 & first inline date < 30 days - chat ❌, inline ✅
- chat messages < 500 & first inline date > 30 days - chat ✅, inline ❌
- chat messages > 500 & first inline date > 30 days - chat ❌, inline ❌

# Documentation

N/A


ngafar:
Shelving this for now.

PR #1470, opened 2025-01-07
Demo video on homepage

ngafar:
# Description

Added loom video to home page

# Testing

See preview

# Documentation

n/a

### DIFF
--- a/evals/prompts/inline_code_completion_prompts/prod_prompt_v4.py
+++ b/evals/prompts/inline_code_completion_prompts/prod_prompt_v4.py
@@ -1,0 +1,111 @@
+from evals.eval_types import InlineCodeCompletionPromptGenerator, NotebookState, ChatPromptGenerator
+
+__all__ = ['prod_prompt_v4']
+
+class _ProdPromptV4(InlineCodeCompletionPromptGenerator):
+    prompt_name = "prod_prompt_v4"
+
+    def get_prompt(self, prefix: str, suffix: str, notebook_state: NotebookState) -> str:
+    
+        return f"""You are a code completion assistant integrated into JupyterLab. Your role is to intelligently complete code based on the user's partial input and context.
+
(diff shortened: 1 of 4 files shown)


aarondr77:
Can we make the video larger? Its a bit too small on the home screen to be able to like read the code comfortably or read the Mito AI chat.

aarondr77:
Is it also possible to change the default back to 1x speed instead of like 1.2x?

evals/prompts/inline_code_completion_prompts/prod_prompt_v4.py
```diff
@@ -0,0 +1,111 @@
+from evals.eval_types import InlineCodeCompletionPromptGenerator, NotebookState, ChatPromptGenerator
```
aarondr77, line 1: Oops don't think this code was meant to be here.

aarondr77:
On Mobile, the box around the video doesn't really fit the aspect ratio. Would be nice if it looked a bit better. 
Best Python Spreadsheet Automation   Code Generation  Mito


The best way to develop on mobile is to connect your local host to your phone. You can see live updates each time you save the code just like you do on your computer. There's some instructions on how to set this up in the trymito.io README, but might need to ask ChatGPT for some help too

ngafar:
Video is now larger. One caveat is that I can't seem to make it less than 100% wide and responsive, without introducing black bars 🤮. But it is readable on most desktop screen sizes. 

> Is it also possible to change the default back to 1x speed instead of like 1.2x?

The only option is to hide the speed. However, this also hides the videos total length, and I think it's nice to give users the option to speed up if they want. [More info on Loom embedding params](https://support.loom.com/hc/en-us/articles/360002208317-How-to-embed-your-video-into-a-webpage).

Also, I took out the border. Super distracting with the larger video size.

PR #1471, opened 2025-01-07
Inline Prompt Update - Line rewrite

ngafar:
# Description

Update the inline prompt to address issue where it rewrites the entire line. 

For example, with the old prompt `pri<cursor>` would return `priprint()`.

This PR tried to fix this issue entirely in prompt. To do so we:

1. Fix some of the examples we were providing to the LLM.
2. Add a fifth clause to the *Critical Formatting Rules* section.
3. Simplified the "Your job is to..." paragraph. This also lets the LLM be a little more imaginative.

We have not fully solved the issue, but there is a noticeable improvement. Looking at the evals, accruacy has dropped from 48% to 46%. However, 2% does seem to be within the acceptable margin of error (yesterday the old prompt was returning 50%, so there is some flux).    

# Testing

Clone it, prompt it.

# Documentation

N/A - behind the scenes prompt update


ngafar:
This is outdated. Might be easier to get the LLM to write the entire line -- as opposed to a truncated version. 

See: https://github.com/mito-ds/mito/pull/1483

PR #1474, opened 2025-01-08
Updated quota checking logic

ngafar:
# Description

Check to see if a user key is present before checking for server quota overages. 

# Testing

Update the user file to simulate overages. 

# Documentation

N/A

### DIFF
--- a/mito-ai/mito_ai/providers.py
+++ b/mito-ai/mito_ai/providers.py
@@ -189,6 +189,17 @@ This attribute is observed by the websocket provider to push the error to the cl
         if self._models is None:
             self._validate_api_key({"value": self.api_key})
 
+        # If the user has an OpenAI API key, then we don't need to check the Mito server quota.
+        if self.api_key:
+            return AICapabilities(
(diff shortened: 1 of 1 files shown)


aarondr77 (commented):
Sweet good catch! LGTM. 

After deploying this, to make sure we don't break this important logic in the future, can we create a few simple pytests? I think can they just: 

1. Create a mock user.json 
2. Call some of these helper functions like (`capabilities` and `check_mito_server_quota`  and make sure they return the correct stuff for each scenario. 
3. Add the pytests to our CI so they always run. 

I think the scenarios we need are: 

1. OS user, Mito Server, Below Limit
2.  OS user, Mito Server, Over Limit
3. OS user, OpenAI key set, Below Limit 
4. OS user, OpenAI key set, Over Limit 
5. Pro user, Mito Server, Below Limit
6. Pro user, Mito Server, Over Limit
7. Pro user, OpenAI key set, Below Limit 
8. Pro user, OpenAI key set, Over Limit

PR #1478, opened 2025-01-10
Added Tests for Server Quota

ngafar:
# Description

Added the following tests:

1. OS user, Mito Server, Below Limit
2. OS user, Mito Server, Over Limit
3. OS user, OpenAI key set, Below Limit
4. OS user, OpenAI key set, Over Limit
5. Pro user, Mito Server, Below Limit
6. Pro user, Mito Server, Over Limit
7. Pro user, OpenAI key set, Below Limit
8. Pro user, OpenAI key set, Over Limit

Note that I am only running tests on Python 3.10 and 3.11. I had to remove 3.8, because you cant use patches within a `with` loop. However, this shouldn't be too much of an issue since the code in question is only used for testing.

Full explanation from ChatGPT:

> The issue lies in the way you are using patch.dict and patch together in a with statement. In Python 3.8, patch.dict and patch objects are not context managers when combined using parentheses and do not have the __enter__ method. This behavior is allowed in Python 3.10 and later due to improved support for nested context managers. 

# Testing

cd into `mito-ai` and run:

```
```

Then run `pytest`.

# Documentation

N/A

### DIFF
--- a/.github/workflows/test-mito-ai-backend.yml
+++ b/.github/workflows/test-mito-ai-backend.yml
@@ -44,7 +44,7 @@ jobs:
         cd mito-ai
         pytest
       env:
-        OPENAI_API_KEY: ${{ matrix.use-mito-ai-server && '' || secrets.OPENAI_API_KEY }}
+        OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
     - name: Upload test-results
       uses: actions/upload-artifact@v3
       if: failure()
--- a/.github/workflows/test-mito-ai-backend.yml
+++ b/.github/workflows/test-mito-ai-backend.yml
@@ -14,7 +14,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        python-version: ['3.8', '3.10', '3.11']
+        python-version: ['3.10', '3.11']
       fail-fast: false
 
     steps:
--- a/.github/workflows/test-mito-ai-backend.yml
+++ b/.github/workflows/test-mito-ai-backend.yml
@@ -35,6 +35,10 @@ jobs:
       run: |
         cd mito-ai
         pip install -e ".[test, deploy]"
+        jlpm install
+        jlpm build
+        jupyter labextension develop . --overwrite
+        jupyter server extension enable --py mito_ai
     - name: Run tests
       run: |
         cd mito-ai
--- a/.github/workflows/test-mito-ai-backend.yml
+++ b/.github/workflows/test-mito-ai-backend.yml
@@ -1,0 +1,50 @@
+name: Test - Mito AI Backend
+
+on:
+  push:
+    branches: [ dev ]
+    paths:
+      - 'mito-ai/**'
+  pull_request:
+    paths:
+      - 'mito-ai/**'
+jobs:
+  test-mitoai-backend:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        python-version: ['3.8', '3.10', '3.11']
+      fail-fast: false
+
+    steps:
+    - name: Cancel Previous Runs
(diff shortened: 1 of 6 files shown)


mito-ai/mito_ai/tests/providers_test.py
```diff
@@ -0,0 +1,167 @@
+import os
+from datetime import datetime
+from unittest.mock import patch
+from mito_ai.providers import OpenAIProvider
+
+CHAT_LIMIT = 500
```
aarondr77, line 6: Instead of recreating this, let's just import `OPEN_SOURCE_AI_COMPLETIONS_LIMIT` so we're making sure to use the correct number.

mito-ai/mito_ai/tests/providers_test.py
```diff
@@ -0,0 +1,167 @@
...
+
+    with (
+        patch.dict(os.environ, {"OPENAI_API_KEY": ""}),
+        patch("mito_ai.providers._num_usages", 1),
+        patch("mito_ai.providers._first_usage_date", today),
+    ):
+        capabilities = llm.capabilities
+
+        assert capabilities.provider == "Mito server"
+        assert llm.last_error is None
```
aarondr77, line 26: Hmmm maybe we should also be testing the `check_mito_server_quota` function directly? Cause that is the actual function that is stopping us from using the server, right? Not the capabilities display.

mito-ai/mito_ai/tests/providers_test.py
```diff
@@ -0,0 +1,167 @@
...
+    No error should be thrown, since the user is using their own key.
+    """
+
+    llm = OpenAIProvider()
+    today = datetime.now().strftime("%Y-%m-%d")
+
+    with (
+        patch("mito_ai.providers._num_usages", CHAT_LIMIT + 1),
+        patch("mito_ai.providers._first_usage_date", today),
+    ):
```
aarondr77, line 81: In these tests, where are we setting the OpenAI key? Is it relying on the GitHub action?

mito-ai/mito_ai/tests/providers_test.py
```diff
@@ -0,0 +1,167 @@
...
+        patch("mito_ai.providers._first_usage_date", today),
+        patch("mito_ai.utils.open_ai_utils.is_pro", return_value=True),
+    ):
+        capabilities = llm.capabilities
+
+        assert "user key" in capabilities.provider
+        assert llm.last_error is None
+
+
+def test_pro_user_openai_key_set_above_limit():
```
aarondr77, line 150: Should we add tests to make sure that the `inline_code_completion` limits work also? I don't think we're testing that right now, right?

aarondr77: Especially since you just caught a bug in this logic !

aarondr77 (changes requested):

mito-ai/mito_ai/tests/providers_test.py
ngafar: We're depending on the key set in the Github action.

.github/workflows/test-mito-ai-backend.yml
```diff
@@ -0,0 +1,54 @@
...
+        jlpm install
+        jlpm build
+        jupyter labextension develop . --overwrite
+        jupyter server extension enable --py mito_ai
+    - name: Run tests
+      run: |
+        cd mito-ai
+        pytest
+      env:
+        OPENAI_API_KEY: ${{ matrix.use-mito-ai-server && '' || secrets.OPENAI_API_KEY }}
```
aarondr77, line 47: ooo I see, `matrix.use-mito-ai-server` is never true here since we don't have it set up in the matrix above, so that's why we always have the OPENAI_API_KEY

Let's remove the conditional check here

aarondr77 (approved):
Awesome! Yay for good tests

Issue #1414, opened 2024-12-12
Evals we should create


aarondr77:
# Mito AI Chat

- [ ] **Very Large Active Code Cells**. Active code cell was very large, (it contained a mapping from each country to its continent, generated by the AI), then when he asked it to make a change to the final line of that code cell that actually applies the mapping, instead of returning the entire mapping with the edit, it just returned the new edit. 

# In Line Code Completer 
- [ ] 

# Smart Debugging 
- [ ]

jake-stack:
I asked Mito AI to tell me how to do simple machine learning. The first section of the response was 

"Choose a Machine Learning Library: Popular libraries include scikit-learn, TensorFlow, and PyTorch. For beginners, scikit-learn is often recommended due to its simplicity and ease of use."

When I did another prompt that said "import the libraries from section 1", it imported none of these libraries and instead imported mitosheet and pandas. 

It would be great if the AI could better recognize prompts that are referring to previous responses in the chat.

aarondr77:
Incorrect Autocomplete Response

```python
class Category(BaseModel):
    name: str

categories = [
    'dataframe transformation', 
    'importing data from file', 
    'visualization/graphing',
    '

]

for event in tqdm(chat_data):
    msg_body = categorize.generate_categorize_prompt(event["user_input"], categories) 
    messages = [
        {"role": "user", "content": msg_body},
    ]
    
    completion = client.beta.chat.completions.parse(
        model=model,
        messages=messages,
        temperature=temperature,
        response_format=Category,
    )

    category = completion.choices[0].message.parsed

    if category not in categories:
        categories.append(category)

    event["category"] = category.name
```

<img width="774" alt="Image" src="https://github.com/user-attachments/assets/1eb50006-8184-437e-b2f8-f76cd2f6a4cb" />

PR #1483, opened 2025-01-14
Do not repeat prefix and suffix in autocomplete

aarondr77:
# Description

Improves our autocomplete responses by changing what we're asking of the AI. Previously, we heavily relied on the Ai to only return the new code that the user should type. This wasn't working well. Now, we ask the AI to return the complete line of code and then deduplicate it with the end of the prefix and beginning of the suffix. 

To do so, autocomplete prompts in our evals can now optionally implement a post processing step. 

This PR also fixes a few broken evals and adds some new evals based on the prefix issues documented here https://github.com/mito-ds/mito/issues/1414

On the new set of evals our v3 prompt (with some of bug fixes suggested by @ngafar) scored an average of 63% across three runs and the new v4 prompt with this prefix/suffix stripping strategy scored an average 65.6% of across three runs. 

# Testing

Use it kinda heavily and see if you're getting prefix repetition issues

# Documentation

No.

### DIFF
--- a/mito-ai/src/Extensions/InlineCompleter/provider.ts
+++ b/mito-ai/src/Extensions/InlineCompleter/provider.ts
@@ -188,7 +188,7 @@ export class MitoAIInlineCompleter
       const result = await this._client.sendMessage({
         messages: openAIFormattedMessages,
         message_id: messageId.toString(),
-        stream: true,
+        stream: false,
         type: 'inline_completion',
       });
 
--- a/mito-ai/src/Extensions/InlineCompleter/provider.ts
+++ b/mito-ai/src/Extensions/InlineCompleter/provider.ts
@@ -179,6 +179,7 @@ export class MitoAIInlineCompleter
 
       const prefix = this._getPrefix(request);
       const suffix = this._getSuffix(request);
+
       const variables = this._variableManager.variables;
       const prompt = createInlinePrompt(prefix, suffix, variables);
       const openAIFormattedMessages: OpenAI.Chat.ChatCompletionMessageParam[] = [
@@ -210,7 +211,7 @@ export class MitoAIInlineCompleter
       return {
         items: result.items.map(item => ({
           ...item,
-          insertText: this._cleanCompletion(item.content)
+          insertText: this._cleanCompletion(item.content, prefix, suffix)
         }))
       };
     } finally {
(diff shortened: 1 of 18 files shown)


mito-ai/src/Extensions/InlineCompleter/provider.ts
```diff
@@ -342,14 +343,33 @@ export class MitoAIInlineCompleter
     });
   }
 
-  private _cleanCompletion(rawCompletion: string) {
-    return rawCompletion
+  private _cleanCompletion(rawCompletion: string, prefix?: string, suffix?: string) {
```
ngafar, line 346: Using this, but I'm still seeing the same issues. I added a console log to this function, and it seems like prefix is always null.

I am seeing some minor success with adding this line to the fetch function:

```
this._currentLine = prefix.split('\n').slice(-1)[0]
```

and then passing _currentLine as the prefix param in _cleanCompletion().

aarondr77: I just pushed some debugging code, can you use the inline completer and then check what is getting printed to your logs. I expected it to look like this. 

Screenshot 2025-01-15 at 5 14 39 PM

mito-ai/src/Extensions/InlineCompleter/provider.ts
```diff
@@ -187,7 +188,7 @@ export class MitoAIInlineCompleter
       const result = await this._client.sendMessage({
         messages: openAIFormattedMessages,
         message_id: messageId.toString(),
-        stream: true,
+        stream: false,
```
ngafar, line 191: Was this a temporary thing for me to test, or does streaming need to be re-enabled?

mito-ai/src/prompts/InlinePrompt.tsx
```diff
@@ -5,15 +5,14 @@ export function createInlinePrompt(
...
 
 You're given the current code cell, the user's cursor position, and the variables defined in the notebook. The user's cursor is signified by the symbol <cursor>.
-
+    
 CRITICAL FORMATTING RULES:
-1. If the cursor appears at the end of a complete line (especially after a comment), ALWAYS start your code with a newline character
-2. If the cursor appears at the end of a function definition, ALWAYS start your code with a newline character
-3. If the cursor appears in the middle of existing code or in an incomplete line of code, do NOT add any newline characters
-4. Your response must preserve correct Python indentation and spacing
+1. Include a new line character at the start of your response if you want the code you are writing to be added on the line after the cursor. For example, if the cursor is at the end of a comment, you should start your response with a newline character so that the code you write is not added to the comment.
```
ngafar, line 13: I'm noticing a lot more completions starting with a newline. For example: 
Screenshot 2025-01-16 at 4 50 30 PM

mito-ai/src/prompts/InlinePrompt.tsx
```diff
@@ -71,19 +70,66 @@ Defined Variables: {{}}
...
+\`\`\`
+</Example 5>
+
+IMPORTANT: Notice in Example 5 that the output is indented several times because the code must be executed as part of the else block.
 
 Your Task:
 
-Defined Variables: ${variables?.map(variable => `${JSON.stringify(variable, null, 2)}\n`).join('')}
+Defined Variables: 
+${variables?.map(variable => `${JSON.stringify(variable, null, 2)}\n`).join('')}
```
ngafar, line 132: Should this prompt also have a post_process_output function, like evals/prompts/inline_code_completion_prompts/prod_prompt_v4.py?

ngafar: nvm, this should be taken care of by provider.ts, inside the `fetch` function, with: 

```
insertText: this._cleanCompletion(item.content, prefix, suffix)
```

right?

mito-ai/src/Extensions/InlineCompleter/provider.ts
ngafar: Also, with `stream` set to true. i am still seeing the prefix issue. Any idea on why this is the case? `_cleanCompletion` should get called regardless.

ngafar: If we do decide to turn off stream moving forward, we should look at adding the blinking cursor, to indicate that the AI is "thinking" about the completion. Otherwise, there's no way to know if you should expect a response or not.

aarondr77: I think we can leave this disabled for now. In all cases for autocomplete, we'll use a single response instead of streaming. This is what cursor and GitHub copilot do I think. I think it's less distracting for users. But I agree that having some sort of "thinking" visual indicator could be useful. We'll save that for later.

mito-ai/src/prompts/InlinePrompt.tsx
aarondr77: Yeah for now its just in the cleanCompletion code

Issue #1479, opened 2025-01-13
Inline code completion keyboard shortcut not set to TAB


aarondr77:
Two users I watched did not have the inline code completion keyboard shortcut automatically set to tab. 

<img width="548" alt="Screenshot 2025-01-13 at 11 39 27 AM" src="https://github.com/user-attachments/assets/7841ed71-e186-4976-a031-8ab38538bb9f" />

ngafar:
According to Claude (via Cursor), this might be happening because the default key to accept inline completions is handled by the core @jupyterlab/completer package. 

We can probably overwrite this. Looking into solutions now.

PR #1485, opened 2025-01-15
Manually setting shortcut for inline-completer:accept

ngafar:
# Description

Fixes https://github.com/mito-ds/mito/issues/1479

This is an issue we've seen happen at least twice, where the default shortcut for accepting an inline completion was not set to `Tab`. To avoid this, we're simply setting the Tab key as the default for the `inline-completer:accept` command. This happens right after a user enables Mito AI for the first time. 

To keep this as nondestructive as possible, we're only using the `.jp-mod-inline-completer-active` selector, and setting the shortcut once. 

# Testing

Disable Mito AI by updating `.jupyter/serverconfig/mitoaiconfig.json` so that `settingsChecked` is false. Then start the server, and look for the "enable Mito AI notification." 

# Documentation

https://docs.trymito.io/mito-ai/autocomplete#updating-keyboard-shortcuts

### DIFF
--- a/mito-ai/src/Extensions/InlineCompleter/index.ts
+++ b/mito-ai/src/Extensions/InlineCompleter/index.ts
@@ -37,15 +37,6 @@ export const completionPlugin: JupyterFrontEndPlugin<void> = {
     settingRegistry: ISettingRegistry,
     variableManager: IVariableManager
   ) => {
-    // Explicitly add a keyboard binding for the inline-completer:accept command.
-    // This should be automatically set by the inline-completer extension,
-    // but we've seen some users not have the tab key set as the default key binding.
-    app.commands.addKeyBinding({
-      command: 'inline-completer:accept',
-      keys: ['Tab'],  
-      selector: '.jp-Notebook'  // Only active in notebooks
-    });
-
     if (typeof completionManager.registerInlineProvider === 'undefined') {
       // Gracefully short-circuit on JupyterLab 4.0 and Notebook 7.0
       console.warn(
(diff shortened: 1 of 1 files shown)


fcollonval (commented):
Hey @ngafar 

The code is ok. But I am not sure it is the right path.

Without this, if a user want deliberately to change the default inline completion keyboard shortcut or wants to use `Tab` for something else, it may do so through settings configuration. Adding that line will force the `Tab` shortcut to be bound without customization. Moreover `Tab` is highly sensitive for keyboard navigation for people with disabilities - so we should be very careful about it is usage; in this specific case, is there a reason to not use the same selector as the official one?
See https://github.com/jupyterlab/jupyterlab/blob/87cb79fe29c623f1e2f3d00e8d0d51c656ef12d4/packages/completer-extension/schema/inline-completer.json#L23

Regarding the trouble seen by users, it is probably due to the change of the shortcut at a later stage in https://github.com/jupyterlab/jupyterlab/pull/16565

All that said, I'm wondering if the best path is not to request the current inline-completion keyboard shortcut and prompt the user if it is not `Tab`; suggesting to _correct_ it and storing the answer to avoid noisy notifications. That would be similar to the validation step done later in this plugin.

fcollonval:
Here is a video showing default stock JupyterLab 4.3.4 (above) and JupyterLab with Mito AI  (below) #1485

https://github.com/user-attachments/assets/f0063da9-e102-46d9-b95c-274f361de970

You can see that capturing Tab at with the current selector breaks keyboard navigation.

ngafar:
Thanks for the advice @fcollonval, those are great ideas. I've updated my PR to only show a notification if the users default shortcut for `inline-completer:accept` isn't tab. I've also updated the testing directions above. 

However, I am curious how users can change this shortcut? Looking in settings > Keyboard Shortcuts, I don't see any results when I search for "inline-completer:accept," and a search for "complete" only shows the command to "Display the completion helper."

Screenshot 2025-01-16 at 3 12 36 PM

fcollonval:
This is weird. I have the expected result for both stock Lab and Lab + this PR:

image

We should probably look at what is going on in your case. Any error in the web console?

fcollonval (changes requested):

fcollonval:
Regarding testing, I encourage you to set up directly an integration test for galata; it is more worthy than patching the code that could result is unwanted changes pushed into the main branch.

To see an example of patching dynamically config and settings, you can have a look at: https://github.com/jupyterlab/jupyterlab/blob/main/galata/test/jupyterlab/announcements.test.ts

aarondr77:
Okay, just documenting the next steps we decided on. When we display the Enable notification for the inline completer, we will automatically set the Tab keyboard shortcut. Something like this. 
```
{
      "command": "inline-completer:accept",
      "keys": ["Tab"],
      "selector": ".jp-mod-inline-completer-active"
}
```

That way users can
1.  open the keyboard shortcut panel and overwrite it if they have a different preferences. 
2. we are only applying the shortcut to the ghost completer and not overwriting the default action of tab

PR #1488, opened 2025-01-22
Mito AI: Restore chat on refresh

ktaletsk:
# Description

- [x] Keep the full history (prompts, display messages, metadata) in the backend. 
- [x] Move backend storage from memory to the disk (JSON or sqlite). 
- [x] Create new websocket message types to receive the history. 
- [x] Restore history when the extension loads.

# Testing

1. Start a new chat or explain code or debug error
2. Refresh the browser page or close and reopen JupyterLab page

# Documentation

N/A


mito-ai/mito_ai/handlers.py
```diff
@@ -26,12 +28,77 @@
...
+                    self._display_history = data.get("display_history", [])
+            except Exception as e:
+                print(f"Error loading history file: {e}")
+    
+    def _save_to_disk(self):
+        """Save current history to disk."""
+        data = {
+            "llm_history": self._llm_history,
+            "display_history": self._display_history,
+        }
```
aarondr77, line 67: Another option here is to just store the metadata that we send from the frontend to the backend in order to create the prompt in the .json file. Then, we can easily recreate both the `llm_history` and the `display_history` from them. There's no reason for the backend to care about the `display_messages` for example. 

What do you think about refactoring this? Is there an advantage to storing everything in the .json file itself?

mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
```diff
@@ -242,32 +248,10 @@ export class ChatHistoryManager {
...
         role: 'user',
-        content: `\`\`\`python
+        content: activeCellCode ? `\`\`\`python
 ${activeCellCode}
 \`\`\`
 
-${input}`};
-}
-
-const removeInnerThoughtsFromMessage = (messageContent: string) => {
```
aarondr77, line 252: I think we're going to need this sort of post-processing code in the frontend for when we stream responses to the frontend.

mito-ai/mito_ai/handlers.py
ktaletsk: The reason for the backend to care about `display_messages` is to be able to restore them in the frontend when the extension loads.

I chose the format of what is stored in JSON to match what is stored in the memory, so it is easy to serialize/deserialize. That being said, we don't have to strore full text representations of either and can recalculate on the fly from context objects. 

As I see it, there is no compeling reason to store that metadata, becuase the only thing we do with it is to calculate LLM prompts and display messages.

mito-ai/mito_ai/handlers.py
```diff
@@ -26,12 +28,77 @@
...
+from .utils.schema import MITO_FOLDER
 
 __all__ = ["CompletionHandler"]
 
+# Global message history with thread-safe access
+class GlobalMessageHistory:
+    def __init__(self, save_file: str = os.path.join(MITO_FOLDER, "message_history.json")):
+        self._lock = Lock()
+        self._llm_history: List[Dict[str, str]] = []
+        self._display_history: List[Dict[str, str]] = []
```
aarondr77, line 45: Just to keep our language the same across the code base, can we call these `_ai_optimized_history` and `_display_optimized_history`

mito-ai/mito_ai/handlers.py
```diff
@@ -26,12 +28,77 @@
...
+            self._display_history = []
+            self._save_to_disk()
+
+    def append_message(self, llm_message: Dict[str, str], display_message: Dict[str, str]) -> None:
+        with self._lock:
+            self._llm_history.append(llm_message)
+            self._display_history.append(display_message)
+            self._save_to_disk()
+
+    def truncate_histories(self, index: int) -> None:
```
aarondr77, line 94: Two quick things here: 

1. It looks like we are not removing the original message that the user edits. Maybe just an off by one bug here? 
2. In the UI, when the user edits the message and presses submit, we are not clearing all future messages until the AI responds. That looks a bit whacky. Can we clear the messages right away. If this isn't a regression introduced by this PR, then feel free to just open an issue and we can fix independently. 


https://github.com/user-attachments/assets/e43c492b-33b3-4021-aedb-969f2e585b1e

aarondr77 (commented):
Awesome. This is working pretty well and feels so nice to be able to keep working on my chat after reloading the page. 

A few things to fix up below. Nothing major at all. 

The one additional thing is can we create a few tests in the new mito-ai/mito-ai/tests folder that just test the backend writing and reading functionality. Some tests for the following would be great: 
- Reads the file contents and creates the right data structure
- Overwrites the data correctly
- The truncate function works correctly
- Creates the file if it doesn't exist, etc.

mito-ai/mito_ai/handlers.py
```diff
@@ -26,12 +28,77 @@
...
+            try:
+                with open(self._save_file, "r", encoding="utf-8") as f:
+                    data = json.load(f)
+                    self._llm_history = data.get("llm_history", [])
+                    self._display_history = data.get("display_history", [])
+            except Exception as e:
+                print(f"Error loading history file: {e}")
+    
+    def _save_to_disk(self):
+        """Save current history to disk."""
```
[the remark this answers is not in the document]
ktaletsk: Yes, definitely. I will add version `1` for this PR, but it will be quickly obsolete

mito-ai/mito_ai/handlers.py
ktaletsk: 1. Yes, off by one bug. I re-added system message in the frontend, so the indexing starts from 1 on the frontend.
2. Looks like a regression introduced by the previous PR

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -79,10 +79,45 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+        await websocketClient.ready;
+        
+        const aiResponse = await websocketClient.sendMessage({
+            type: 'fetch_history',
+            message_id: UUID.uuid4(),
+            metadata: {},
+            stream: false
+        });
+
+        const items = (aiResponse as any).items ?? [];
```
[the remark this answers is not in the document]
ktaletsk: Actually, I did not need that casting as `any`. `ICompletionReply` has `.items`

tests/mitoai_ui_tests/shortcuts.spec.ts
```diff
@@ -22,6 +25,8 @@ test.describe('Mito AI Shortcuts', () => {
     });
 
     test('Accept AI generated code', async ({ page }) => {
+        await clearMitoAIChatHistory(page);
```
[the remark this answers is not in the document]
ktaletsk: Not sure if restarting the kernel is supposed to affect anything in the chat. Maybe reloading JupyterLab webpage?

PR #1492, opened 2025-01-28
.github: remove python 3.8 for end of life

aarondr77:
# Description

Drops Python 3.8 from our CI since it has reached End of Life. 

# Testing

See that tests pass. 

# Documentation

Yes. We now support Python 3.9 and higher.


ktaletsk (approved):
Clean break!

PR #1501, opened 2025-02-04
Agent Mode Backend

ngafar:
# Description

- Added an agent mode toggle button in the chat taskpane header. 
- Added agent prompt.
- Updated webhook endpoints to handle agent requests.  

For now the agent mode button is just the help icon. This will be updated in my next PR handling the frontend.

# Testing

Toggle agent mode, and send a request. You should see the response show up as a JSON object. 

# Documentation

No. Will document at the end.


aarondr77 (approved):
Okay great. This all looks reasonable to me! it would be nice to get this merged into dev just to avoid future merge conflicts, but since its not ready for release and we don't want to block releases, lets just remove the button to turn on Agent mode so users can't access it.

PR #1502, opened 2025-02-04
Model o3

aarondr77:
# Description

Adds o3-mini for pro users for chat messages. This leads to a significant increase in eval performance. With o3-mini, the average eval score is 68% in 240 seconds. With gpt-4o-mini, the average eval score is 57% in 174 seconds. The time difference per eval is 1.3 seconds. This is an acceptable difference for the increase in accuracy

Interesting, the smart debugger does not yet perform better with o3-mini as compared to gpt-4o-mini, so we continue using gpt-4o-mini for now. With o3-mini, the average eval score is 69% in 239 seconds. With gpt-4o-mini, the average eval score is 71% in 297 seconds. The time difference per eval is 1.3 seconds.

This also does a fair bit of cleaning up the backend to simplify it down to just the things we are using:
- Removes some configuration options from the provider
- Creates both a synchronous and asynchronous client so we can use the appropriate one for the job
- Simplifies error handling to make sure its consistent across completion approaches with the user key and the mito server

# Testing

Try it out with pro and open source plans and make sure that it works! 

# Documentation

We should update the docs to let users know that upgrading to pro gives them better models.


evals/ai_api_calls/get_open_ai_completion.py
```diff
@@ -2,17 +2,27 @@
...
-    )
+    
+    if model == "gpt-4o-mini":
+        response = client.chat.completions.create(
+            model=model,
+            messages=[
+                {"role": "system", "content": "You are an expert Python programmer."},
+                {"role": "user", "content": prompt}
+            ],
+            temperature=0.0
```
aarondr77, line 16: Add a comment

mito-ai/mito_ai/providers.py
```diff
@@ -368,18 +339,17 @@ async def stream_completions(
...
             parent_id=request.message_id,
         )
+        
+        # Validate that the model is supported. If not fall back to gpt-4o-mini
+        if model not in self.models:
+            model = "gpt-4o-mini"
 
         # Send the completion request to the OpenAI API and returns a stream of completion chunks
         try:
+            completion_function_params = get_open_ai_completion_function_params(model, self.max_completion_tokens, request.messages)
```
aarondr77, line 349: Check if self.max_completion_tokens is still something we set

evals/prompts/smart_debug_prompts/prod_prompt_v2.py
```diff
@@ -114,6 +114,7 @@ def parse_date(date_str):
 - Make the solution as simple as possible.
 - Reuse as much of the existing code as possible.
 - Do not add temporary comments like '# Fixed the typo here' or '# Added this line to fix the error'
+- The code in the SOLUTION section should be a python code block starting with ```python and ending with ```
```
ngafar, line 118: We've never had an issue with formatting, but [structured outputs](https://platform.openai.com/docs/guides/structured-outputs) could open up some possibilities in the future.

PR #1510, opened 2025-02-05
Fix/openai timeout

aarondr77:
# Description

Sets max_retries and timeout so that when the AI fails, we try again. This should prevent `timeout` errors. 

# Testing

**To test this, switch to the development lambda function URL** 

Test this code and make sure it works with both the OpenAI User Key and Mito Server. Once we deploy, we can see if it reduces the number of timeout errors. 

# Documentation

No.


mito-ai/mito_ai/providers.py
```diff
@@ -220,6 +235,7 @@ async def request_completions(
         Returns:
             The completion
         """
+        TIMEOUT = 100
```
aarondr77, line 238: Delete

mito-ai/mito_ai/providers.py
```diff
@@ -229,8 +245,12 @@ async def request_completions(
                 if model not in self.models:
                     model = "gpt-4o-mini"
 
-                completion_function_params = get_open_ai_completion_function_params(model, request.messages, False, response_format)
+                completion_function_params = get_open_ai_completion_function_params(model, request.messages, False, response_format, TIMEOUT)
```
aarondr77, line 248: Don't use TIMEOUT

mito-ai/mito_ai/providers.py
```diff
@@ -229,8 +245,12 @@ async def request_completions(
...
+                
+                print("Sending message")
```
aarondr77, line 250: remove

mito-ai/mito_ai/providers.py
```diff
@@ -292,6 +315,7 @@ async def request_completions(
                 )
 
         except BaseException as e:
+            print(e)
```
aarondr77, line 318: remove

ngafar (approved):

PR #1512, opened 2025-02-06
mito-ai: add mypy tests

aarondr77:
# Description

Adds type checking to the mito-ai backend! Yay. This will pay for itself in no time. 

I pushed a mypy.ini file that defines our mypy rules so it should perform the same on all of our computers. 

# Testing

1. See that the new test mito-ai mypy workflow passes. 
2. Also make sure that you can run these locally by running `mypy mito_ai`

# Documentation

No.


mito-ai/mito_ai/utils/open_ai_utils.py
```diff
@@ -17,14 +17,16 @@
...
-MITO_AI_URL = "https://ogtzairktg.execute-api.us-east-1.amazonaws.com/Prod/completions/"
 
-OPEN_SOURCE_AI_COMPLETIONS_LIMIT = 500
-OPEN_SOURCE_INLINE_COMPLETIONS_LIMIT = 30 # days
+MITO_AI_URL: Final[str] = "https://ogtzairktg.execute-api.us-east-1.amazonaws.com/Prod/completions/"
 
-__user_email = None
-__user_id = None
+OPEN_SOURCE_AI_COMPLETIONS_LIMIT: Final[int] = 500
+OPEN_SOURCE_INLINE_COMPLETIONS_LIMIT: Final[int] = 30 # days
```
ngafar, line 26: huh, the `Final` type is new to me. Very cool!

PR #1523, opened 2025-02-13
Notebook Variables in Agent Prompt

ngafar:
# Description

Passing the variable manager to the agent prompt. 

# Testing

Create a notebook with content. Then send a message in agent mode. 

If you want to reference a variable in agent mode, make sure to run the cells to register the variable with the variable manager.  

# Documentation

No

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
@@ -410,6 +410,16 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
         // Get the plan from the chat history
         const plan = chatHistoryManager.getDisplayOptimizedHistory().filter(message => message.type === 'openai message:agent:planning')
 
+        // Move to the last cell of the notebook
+        // We don't want to overwrite any code in the notebook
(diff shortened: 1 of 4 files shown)


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -410,6 +410,16 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
         // Get the plan from the chat history
         const plan = chatHistoryManager.getDisplayOptimizedHistory().filter(message => message.type === 'openai message:agent:planning')
 
+        // Move to the last cell of the notebook
+        // We don't want to overwrite any code in the notebook
+        const notebook = notebookTracker.currentWidget?.content;
+        if (notebook) {
+            notebook.activeCellIndex = notebook.widgets.length - 1;
+        }
```
aarondr77, line 418: This is a good simplification for now. 

In the near term we might want to let the agent just add new code cell in the place the user prompts the agent from. 

In the longer-term, I think we'll want to be smart about where the code we are generating goes. Agent will become not just a tool for creating new code, but also for editing existing code probably.

aarondr77 (approved):

PR #1525, opened 2025-02-14
Mutli chat history

aarondr77:
# Description

This PR introduces multiple chat threads in Mito AI (see screenshot below)

image

1. Clear history button is replaced with Start New Chat button.
2. History button opens a dropdown menu with the list of all chat threads.
    - Dot to the left of the thread name indicates the currently open chat
    - Trash icon on the right allows user to delete that thread (even if it is currently selected)
    - Clicking on a different thread loads it
3. Short name for each chat thread is generated with LLM provider (gpt-4o-mini at the moment) based on the first two messages sent to the chat (common practice for AI chat apps). If there hasn't been two messages yet, the chat is named `(New chat)`.

# API changes
1. Major refactor of `message_history.py`. 
    - Instead of a single history, multiple chat threads are stored in memory and backed up to the disk into separate.json files.
    - Chat history schema version is now being tracked in `CHAT_HISTORY_VERSION` and verified on the history loading from disk to avoid conflicts.
    - `GlobalMessageHistory.append_message` is now asynchronous and requires LLM provider to be passed (to be able to generate short chat thread name).
2. New dataclasses introduced to create non-chat replies: ChatThreadItem, StartNewChatReply, FetchThreadsReply, DeleteThreadReply. They use descriptive fields, i.e. `threads`, `thread_id`, `creation_ts` instead of coercing `items` field in `CompletionReply`. They are used in handlers `on_message` function.
3. In the frontend `ChatHistoryManager` has updated `PromptType`s: 
    - Removed `clear_history`
    - Added `start_new_chat`, `get_threads` and `delete_thread`
4. In `ChatTaskpane` removed `ResetIcon` and added `NewChatIcon`, `DeleteIcon`, `HistoryIcon`, `OpenIndicatorIcon` to be shown in chat selector UI
5. `chatThreads` and `activeThreadId` are added as new React states of `ChatTaskpane`.
6. `CompletionWebsocketClient.sendMessage` now uses generics (`message: T`, `Promise<R>`) to check against specified request and response types. 
7. Exapnded `CompleterMessage` with new types `IFetchHistoryReply`, `IStartNewChatReply`, `IFetchThreadsReply`, `IDeleteThreadReply`

# Testing

# Documentation


mito-ai/mito_ai/message_history.py
```diff
@@ -50,103 +96,266 @@ class GlobalMessageHistory:
...
+        """
+        Deletes a chat thread by its ID. Removes both the in-memory entry and the JSON file.
+        """
+        with self._lock:
+            if thread_id in self._chat_threads:
+                del self._chat_threads[thread_id]
+                path = os.path.join(self._chats_dir, f"{thread_id}.json")
+                if os.path.exists(path):
+                    try:
+                        os.remove(path)
```
aarondr77, line 336: To be extra safe, can we add a few more checks here to make sure we only delete the right thing. We should check: 

1. The path is to a file and not a directory
2. The path ends in .json 
3. The contents of the file are a thread? 

Claude also is recommending that we:

1. Check to ensure that the file path does not contain wild card characters like `*`
2. Check to ensure that the path does not contain a path traversal like `../`

mito-ai/mito_ai/message_history.py
```diff
@@ -50,103 +96,266 @@ class GlobalMessageHistory:
...
+                        return True
+                    except Exception as e:
+                        print(f"Error deleting thread {thread_id}: {e}")
+            return False
+
+    def get_threads(self) -> List[ChatThreadItem]:
```
aarondr77, line 342: I expect that the vast majority of chats are never going to be deleted. So maybe instead of sending all of the threads to the frontend, we should have some sort of cutoff? We only send the most recent 50 threads for example? 

To make this easier, what do you think about using the creation_ts as the as the ID? That way we can easily pull the most recent X chats without having to open every single chat file (of which I guess there could be 100's) and read the file contents.

mito-ai/mito_ai/models.py
```diff
@@ -237,3 +235,61 @@ class FetchHistoryReply:
     # Message type.
     type: Literal["reply"] = "reply"
 
+@dataclass(frozen=True)
+class ChatThreadItem:
```
aarondr77, line 239: Maybe we should call this `ChatThreadMetadata`? `Item` makes it sound to me like its the entire ChatThread

mito-ai/mito_ai/models.py
```diff
@@ -237,3 +235,61 @@ class FetchHistoryReply:
...
+    parent_id: str
+
+    # Chat thread item.
+    thread_id: str
+
+    # Message type.
+    type: Literal["reply"] = "reply"
+
+@dataclass(frozen=True)
+class FetchThreadsReply:
```
aarondr77, line 268: Nice ! I love that we're stronger typing the websocket

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -94,68 +115,143 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
-            role: item.role,
-            content: item.content
-        }));
+      // Update the state with the new ChatHistoryManager
+      setChatHistoryManager(newChatHistoryManager);
+      setActiveThreadId(threadId);
+    };
+    
+    const deleteThread = async (threadId: string) => {
+      await websocketClient.ready;
```
aarondr77, line 197: Do we need this at the start of each function? Instead can we move it into the sendMessage function so we don't have to remember to do it everywhere?

mito-ai/mito_ai/models.py
```diff
@@ -237,3 +235,61 @@ class FetchHistoryReply:
...
+class DeleteThreadReply:
+    """
+    Message sent from model to client after deleting a chat thread.
+    """
+
+    # Message UID.
+    parent_id: str
+
+    #Success message
+    created: bool
```
aarondr77, line 292: This should probably be `success` or `deleted` instead of `created`.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -94,68 +115,143 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+         metadata: metadata,
+         stream: false
+      });
+
+      if(response.created === true) {
+         const updatedThreads = chatThreads.filter(thread => thread.thread_id !== threadId);
+         setChatThreads(updatedThreads);
+         if(activeThreadId === threadId) {
+           if(updatedThreads.length > 0) {
+              fetchChatHistoryForThread(updatedThreads[0].thread_id);
```
aarondr77, line 219: I thought this was a bug at first because the active chat doesn't always need to be the first chat in the thread, but then I noticed that when we change to a different chat, we make that the active chat. 

So in that case, do we even need to separately keep track of the `activeThreadId`? Can't we just use the first chat always?

mito-ai/src/icons/DeleteIcon.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const DeleteIcon : React.FC = () => {
+    return (
+        <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 24 24">
+            <path d="M6 19c0 1.1.9 2 2 2h8c1.1 0 2-.9 2-2V7H6zM19 4h-3.5l-1-1h-5l-1 1H5v2h14z" fill="black" className="jp-icon3 jp-icon-selectable"/>
```
aarondr77, line 6: I think we should update this to `<g fill="currentColor">` instead of hardcoding to black? Although, it does already seem responsible to light and dark mode is that because of `className="jp-icon3 jp-icon-selectable"`?

mito-ai/src/icons/HistoryIcon.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const HistoryIcon : React.FC = () => {
+    return (
+      <svg xmlns="http://www.w3.org/2000/svg" width="16" viewBox="0 0 24 24">
+        <path d="M13.5 8H12v5l4.28 2.54.72-1.21-3.5-2.08zM13 3a9 9 0 0 0-9 9H1l3.96 4.03L9 12H6a7 7 0 0 1 7-7 7 7 0 0 1 7 7 7 7 0 0 1-7 7c-1.93 0-3.68-.79-4.94-2.06l-1.42 1.42A8.9 8.9 0 0 0 13 21a9 9 0 0 0 9-9 9 9 0 0 0-9-9" fill="black" className="jp-icon3 jp-icon-selectable"/>
```
aarondr77, line 6: maybe same thing here about updating to `<g fill="currentColor">`?

mito-ai/src/icons/NewChatIcon.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const NewChatIcon: React.FC = () => {
+    return (
+        <svg width="16" height="16" viewBox="0 0 56 56" fill="none" xmlns="http://www.w3.org/2000/svg">
+            <path d="M 27.9883 47.7344 C 29.1367 47.7344 30.0976 46.8203 30.0976 45.6953 L 30.0976 30.1094 L 45.2383 30.1094 C 46.3633 30.1094 47.3476 29.1484 47.3476 28 C 47.3476 26.8516 46.3633 25.9141 45.2383 25.9141 L 30.0976 25.9141 L 30.0976 10.3047 C 30.0976 9.1797 29.1367 8.2656 27.9883 8.2656 C 26.8398 8.2656 25.9024 9.1797 25.9024 10.3047 L 25.9024 25.9141 L 10.7617 25.9141 C 9.6367 25.9141 8.6524 26.8516 8.6524 28 C 8.6524 29.1484 9.6367 30.1094 10.7617 30.1094 L 25.9024 30.1094 L 25.9024 45.6953 C 25.9024 46.8203 26.8398 47.7344 27.9883 47.7344 Z" fill="black" className="jp-icon3 jp-icon-selectable"/>
```
aarondr77, line 6: And one more <g fill="currentColor">

mito-ai/src/icons/OpenIndicatorIcon.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const OpenIndicatorIcon: React.FC = () => {
+  return (
+    <svg xmlns="http://www.w3.org/2000/svg" width="5" viewBox="0 0 24 24">
+      <circle cx="12" cy="12" r="8" fill="green" className="jp-icon3 jp-icon-selectable"/>
```
aarondr77, line 6: Also here, the fill=green doesn't seem to be doing anything

mito-ai/src/utils/websocket/websocketClient.ts
```diff
@@ -129,11 +128,16 @@ export class CompletionWebsocketClient implements IDisposable {
    * Sends a message across the WebSocket. Promise resolves to the message ID
    * when the server sends the same message back, acknowledging receipt.
    */
-  sendMessage(message: ICompletionRequest): Promise<ICompletionReply> {
-    const pendingReply = new PromiseDelegate<ICompletionReply>();
+  sendMessage<T extends ICompletionRequest, R extends CompleterMessage>(
```
aarondr77, line 131: love it!

tests/mitoai_ui_tests/mitoai.spec.ts
```diff
@@ -29,7 +29,7 @@ test.describe('Mito AI Chat', () => {
     await createAndRunNotebookWithCells(page, ['import pandas as pd\ndf=pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})']);
     await waitForIdle(page);
 
-    await clearMitoAIChatHistory(page);
+    await startNewMitoAIChat(page);
```
aarondr77, line 32: Can we update our tests to use multiple workers again? They are pretty slow to run rn

aarondr77 (commented):
Nice work! This is working well! 

The only real comment here is whether it makes sense to count reading the history as an interaction. It seems like that might be causing a few bugs and reordering when you just open a chat, but don't actually edit it feels a little bit hard to follow as the user. Your call. 


The other thing is can we just create some backend purest for this the message_history.py file. We can do this in a separate PR so we can get this merged. But it would be nice to test some things like: 

1. When we create/delete a chat thread it gets both added/deleted from memory and disk
2. We can't delete things other than thread json files
3. etc

mito-ai/src/icons/OpenIndicatorIcon.tsx
ktaletsk: It seems to work after moving to LabIcon from of React.FC
image

mito-ai/mito_ai/handlers.py
```diff
@@ -125,21 +128,58 @@ async def on_message(self, message: str) -> None: # type: ignore
...
+            reply = FetchThreadsReply(
+                parent_id=parsed_message.get("message_id"),
+                threads=threads
+            )
+            self.reply(reply)
+            return
+
+        # Handle delete_thread: delete the specified thread
+        if type == MessageType.DELETE_THREAD:
+            thread_id_to_delete = metadata_dict.get('threadID')
```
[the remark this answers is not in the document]
ktaletsk: Roger that. What do you think about `promptType` name on the backend? Change to snake case too?

mito-ai/mito_ai/message_history.py
```diff
@@ -50,103 +96,266 @@ class GlobalMessageHistory:
...
+        For backward compatibility: returns the display history of the newest thread.
+        """
         with self._lock:
-            self._ai_optimized_history.append(ai_optimized_message)
-            self._display_history.append(display_message)
-            self._save_to_disk()
+            thread_id = self._get_newest_thread_id()
+            if not thread_id or thread_id not in self._chat_threads:
+                return []
+            # If history is requested, that is also considered an interaction
```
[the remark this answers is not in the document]
ktaletsk: I was considering the following scenario: user opens *not* the most recent chat, then refreshes the page. I wanted the user to not lose the currently viewed chat. However, it is clear that changing the timestamp is not a great solution. Maybe we can store the last viewed thread in the backend or frontend, separately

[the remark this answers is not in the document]
ktaletsk: Urgh... The bug you saw (and I confirmed) was due to selecting the oldest thread instead of the newest in the frontend. We are returning threads in the reverse chronological order from the backend, so I need to use the first one, not the last:

```
- const latestThread = chatThreadsResponse.threads[chatThreadsResponse.threads.length - 1];
+ const latestThread = chatThreadsResponse.threads[0];
```

mito-ai/mito_ai/message_history.py
```diff
@@ -321,17 +321,49 @@ def delete_thread(self, thread_id: str) -> bool:
...
+                print(f"Path traversal attempt detected: {path}")
+                return False
+            
+            # Ensure we're only deleting .json files
+            if not path.endswith('.json'):
+                print(f"Not a .json file: {path}")
+                return False
+            
+            # Check if the file exists and is actually a file (not directory)
+        if os.path.exists(path):
```
aarondr77, line 354: This should be indented inside the `lock` right?

aarondr77 (commented):
Just one small comment and then feel free to merge!

PR #1528, opened 2025-02-14
Agent Mode: v1 Polish

ngafar:
# Description

1. When you edit an agent:planning step, don't add the active cell as context.
2. Don't clear that chat when switching to agent planning mode
3. Scroll to the new message returned from the agent, the agent:execute code in the chat right away. Currently, we don't scroll to it until after we apply and execute the code I think.

# Testing

Test out the three scenarios above. 

# Documentation

N/A

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
@@ -123,7 +123,11 @@ export class ChatHistoryManager {
         }
         
         this.displayOptimizedChatHistory[index] = { 
-            message: getDisplayedOptimizedUserMessage(newContent, activeCellCode),
+            message: getDisplayedOptimizedUserMessage(
+                newContent, 
+                activeCellCode,
+                isAgentMessage
+            ),
             type: isAgentMessage ? 'openai message:agent:planning' : 'openai message',
             codeCellID: activeCellID,
             promptType: isAgentMessage ? 'agent:planning' : 'chat'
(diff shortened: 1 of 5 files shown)


aarondr77:
Let's add tests for: 
1. When you edit an agent:planning step, don't add the active cell as context.
2. Don't clear that chat when switching to agent planning mode

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -708,15 +709,13 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
         {
             label: 'Chat',
             onClick: () => {
-                clearChatHistory()
```
aarondr77, line 711: I think this actually introduces a bug

In this code, we find all of the agent messages that we want to handle by filtering down to just the agent:planning messages. But now that we are not clearing the chat, if you use the agent multiple times in one chat, it will get all previous messages, right?

```typescript
 const executeAgentPlan = async () => {
        setAgentModeEnabled(false)

        // Get the plan from the chat history
        const plan = chatHistoryManager.getDisplayOptimizedHistory().filter(message => message.type === 'openai message:agent:planning')
```

mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
```diff
@@ -288,10 +292,15 @@ export class ChatHistoryManager {
 }
 
 
-const getDisplayedOptimizedUserMessage = (input: string, activeCellCode?: string): OpenAI.Chat.ChatCompletionMessageParam => {
+const getDisplayedOptimizedUserMessage = (
```
aarondr77, line 295: Is this just a display optimized message thing that we need to fix?  We also need to make sure that we're not sending the active cell code as part of the params too.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
aarondr77: Another bug here is: 
1. If I start use agent mode
2. Execute the steps
3. Turn agent mode back on to send another agent message, it will go to the screen with the Let's go button instead of just letting me prompt the AI again for a new agent conversation. 

It seems like we might need to be keeping track of like an agent message queue.

ngafar: Yup. Going to revert back to clearing the chat history when the user enters agent mode, so we can avoid this.

mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
ngafar: Also updated `updateMessageAtIndex` to take care of this:

```
const activeCellCode = isAgentMessage ? undefined : getCellCodeByID(this.notebookTracker, activeCellID)
```

aarondr77 (approved):

PR #1531, opened 2025-02-17
Websocket api

aarondr77:
# Description

Stronger types for the web socket API

# Testing

Test it. Ideally, we get our tests to a place where they are so strong that refactors like this (which shouldn't change any behavior) don't need any manual testing. 

# Documentation

No.


mito-ai/src/utils/websocket/models.ts
ngafar, line 1: Adding metadata types closer to the frontend is a huge win. I don't think I've ever sent a message to the backend, and have it work the first time. Hopefully this changes that.

ngafar (approved):

PR #1533, opened 2025-02-17
Agent Mode: Pre-Release Polish

ngafar:
# Description

- Added `user_input` field when logging agent mode prompts. We no longer have to go though the `full_prompt` to see the user's question.
- Editing the original agent mode prompt will erase all plans, and generate a new plan. Previously, this would just add a new message to the chat, and the user would have to scroll to see the new plan. 
- Agent plan messages now default to `grey-900`. Using the Jupyter colors meant that plans were unreadable in dark mode.

While it is an extension of chat, we also want to make sure users understand that it is a slightly different paradigm. To this extent:
- Added purple border to agent mode input.
- Added glow animation when switching to agent mode. 

# Testing

Try it out.

# Documentation

N/A

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
@@ -144,7 +144,7 @@ export class ChatHistoryManager {
         }
     }
 
-    addAgentMessage(message: string): IOutgoingMessage {
+    addAgentMessage(message: string, index?: number): IOutgoingMessage {
         const variables = this.variableManager.variables
 
         const metadata: IChatMessageMetadata = {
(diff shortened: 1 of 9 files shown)


mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
```diff
@@ -161,6 +161,12 @@ export class ChatHistoryManager {
             }
         )
 
+        // If editing the first agent message, the user will want a new plan.
+        // So we drop all steps in the agent's previous plan.
+        if (index === 1) {
+            this.displayOptimizedChatHistory = this.displayOptimizedChatHistory.slice(0, index + 1);
+        }
+
```
aarondr77, line 169: Can you make sure we have a frontend test that will catch this if we break it?

aarondr77 (approved):

mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
ngafar: This should be covered in the "Edit original message" test. 

TL;DR we are checking to make sure the last message before edit isn't the same as the last message after edit.

PR #1537, opened 2025-02-18
Stop agent

aarondr77:
# Description

Let's the user stop the agent's execution! Since we don't to close the websocket, if we're in the process of getting a response back from the Ai, we just wait for it and then send no further messages. 

# Testing

Try it out + see two new tests! 

# Documentation

We should make when we release

### DIFF
--- a/mito-ai/src/utils/agentActions.tsx
+++ b/mito-ai/src/utils/agentActions.tsx
@@ -74,21 +74,3 @@ export const retryIfExecutionError = async (
 
     return 'success'
 }
-
-
-export const stopAgentExecution = (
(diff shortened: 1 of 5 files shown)


mito-ai/src/utils/agentActions.tsx
```diff
@@ -60,9 +68,27 @@ export const retryIfExecutionError = async (
...
     }
 
-    return true
-}
\ No newline at end of file
+    return 'success'
+}
+
+
+export const stopAgentExecution = (
```
aarondr77: Delete

ngafar (approved):

PR #1539, opened 2025-02-19
Agent Blacklisted Code

aarondr77:
# Description

Don't let the agent execute a list of potentially harmful commands like deleting files, folders or sql tables. If the user wants to execute these commands, they should copy the code and execute it manually. 

# Testing

Try it out, but also see the new set of jest tests for this function. 

# Documentation

Maybe we should update our docs to let users know that there is _some_ security built in, but that they should of course still be careful.


mito-ai/src/utils/blacklistedWords.tsx
```diff
@@ -0,0 +1,82 @@
+// Function to check for blacklisted words that could be dangerous for the agent to execute
+
+export const checkForBlacklistedWords = (code: string): { safe: boolean, reason?: string } => {
```
ngafar, line 3: Curious why you choose to do it this way. I would imagine it'd be a lot easier to have the LLM return a strcutred output with a binary field that flags potentially dangerous code.

tests/mitoai_ui_tests/agent.spec.ts
```diff
@@ -305,5 +305,47 @@ test.describe("Agent mode auto error fixup", () => {
...
+        await waitForIdle(page);
+
+        await clickOnMitoAIChatTab(page);
+        await waitForIdle(page);
+
+        // Switch to agent mode 
+        await clickAgentModeToggleButton(page);
+    });
+
+    test.only("Blacklisted command shows error and prevents execution", async ({ page }) => {
```
ngafar, line 328: Happens to the best of us.

ngafar (approved):

mito-ai/src/utils/blacklistedWords.tsx
aarondr77: Hmmm that's a good point. I think the only reason I did it this way was because it seemed easier and more deterministic. There were just a couple of specific things that I wanted to try to avoid. But I think you're totally right that in the future if we want to handle a lot more cases, your approach is better.

PR #1543, opened 2025-02-20
Agent Mode: Delete Steps

ngafar:
# Description

Added new delete icon. Note that this only works on the frontend.  

# Testing

Send a message to the Agent. Get a plan, delete a message, run the plan. The deleted message should not be executed as part of the plan.

# Documentation

N/A


tests/mitoai_ui_tests/agent.spec.ts
```diff
@@ -95,6 +95,20 @@ test.describe("Agent mode print hi", () => {
         expect(lastAgentMessageContent).toContain(newMessage);
     });
 
+    test.only("Delete message in agent's plan", async ({ page }) => {
```
aarondr77: aha! happens to the best of us :)

mito-ai/src/icons/GarbageIcon.tsx
```diff
@@ -0,0 +1,11 @@
...
....14504C9.85156 2.38434 10 2.71451 10 3.07878V3.33617C10 3.57259 9.80797 3.76461 9.57154 3.76461L0.428461 3.76502C0.192028 3.76502 0 3.573 0 3.33658V3.07919C0 2.71489 0.14844 2.38472 0.38775 2.14545C0.627059 1.90618 0.957247 1.75813 1.32153 1.75813L1.32141 1.75772ZM5.86918 0.857224H4.13065C4.00318 0.857224 3.88723 0.909443 3.80293 0.993728C3.71864 1.07802 3.66642 1.19397 3.66642 1.32143V1.75768H6.33382V1.32143C6.33382 1.19397 6.2816 1.07801 6.19731 0.993728C6.11302 0.909438 5.99706 0.857224 5.8696 0.857224H5.86918ZM3.67266 5.93643C3.67266 5.70001 3.48063 5.50799 3.2442 5.50799C3.00777 5.50799 2.81574 5.70001 2.81574 5.93643V8.13416C2.81574 8.37058 3.00777 8.5626 3.2442 8.5626C3.48105 8.5626 3.67266 8.37058 3.67266 8.13416V5.93643ZM5.42847 5.93643C5.42847 5.70001 5.23644 5.50799 5.00001 5.50799C4.76358 5.50799 4.57155 5.70001 4.57155 5.93643V8.13416C4.57155 8.37058 4.76358 8.5626 5.00001 8.5626C5.23685 8.5626 5.42847 8.37058 5.42847 8.13416V5.93643ZM7.18428 5.93643C7.18428 5.70001 6.99225 5.50799 6.75582 5.50799C6.51939 5.50799 6.32736 5.70001 6.32736 5.93643V8.13416C6.32736 8.37058 6.51897 8.5626 6.75582 8.5626C6.99266 8.5626 7.18428 8.37058 7.18428 8.13416V5.93643Z" fill="black" />
```
aarondr77, line 6: At the end of this <path there is still fill="black". I think since we are suing fill=currentColor, we should delete the fill=black

mito-ai/src/icons/Pencil.tsx
```diff
@@ -1,9 +1,9 @@
...
+            <path d="M9.88308 1.82456C9.68326 1.49798 9.43085 1.18192 9.1259 0.876387C8.8209 0.570854 8.5054 0.317998 8.18991 0.117862C7.88492 -0.0717913 7.48531 -0.0296413 7.22238 0.223215L1.15425 6.29155C1.12269 6.32317 1.10167 6.35476 1.08065 6.39691L0.0289843 9.38893C-0.0341088 9.55749 0.00796491 9.73662 0.134151 9.86302C0.218298 9.94732 0.333977 10 0.449651 10C0.502235 10 0.554818 9.98947 0.596894 9.97894L3.58363 8.92541C3.6257 8.91488 3.65723 8.88326 3.6888 8.85168L9.76744 2.78325C10.0304 2.51987 10.0724 2.13008 9.88311 1.82454L9.88308 1.82456ZM1.56439 6.80777C1.86938 6.98685 2.18488 7.2397 2.47934 7.53471C2.77382 7.82972 3.01568 8.14578 3.19449 8.45128L1.46976 9.06235C1.41717 8.957 1.34357 8.85164 1.25942 8.76734C1.16476 8.67252 1.0701 8.60931 0.964937 8.55664L1.56439 6.80777Z" fill="black" fill-opacity="0.87" />
```
aarondr77, line 6: Same with the fill=black here

mito-ai/src/icons/Pencil.tsx
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 
 const PencilIcon: React.FC = () => (
-    <svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <svg width="14" height="14" viewBox="0 0 10 10" fill="none" xmlns="http://www.w3.org/2000/svg">
```
aarondr77, line 4: Idk, but do we need to edit the width and height as well? or does it just scale?

mito-ai/style/ChatTaskpane.css
```diff
@@ -143,10 +143,59 @@
...
+    height: 28px;
+    position: relative;
+}
+
+.message-action-buttons button svg {
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+}
```
aarondr77, line 175: Do we need to be doing this at the svg level? Isn't it enough to just place the message-action-buttons in the right location which we are already doing?

mito-ai/style/ChatTaskpane.css
```diff
@@ -143,10 +143,59 @@
...
+
+.message-action-buttons button {
+    cursor: pointer;
+    opacity: 0;
+    transition: opacity 0.2s ease-in-out, background-color 0.2s ease-in-out;
+    background-color: var(--jp-layout-color3);
+    border: none;
+    padding: 0px;
+    border-radius: 0;
```
aarondr77, line 160: Should we move all of this to .message-action-buttons also?

mito-ai/style/ChatTaskpane.css
```diff
@@ -143,10 +143,59 @@
...
+
+.message-action-buttons button:first-child {
+    border-top-left-radius: 4px;
+    border-bottom-left-radius: 4px;
+}
+
+.message-action-buttons button:last-child {
+    border-top-right-radius: 4px;
+    border-bottom-right-radius: 4px;
+}
```
aarondr77, line 185: Yeah I think we can just set message-action-buttons instead of having to do this casing? 

Like 

```css
.message-action-buttons {
     border-radius: 4px
}
```

mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
```diff
@@ -205,13 +208,22 @@ const ChatMessage: React.FC<IChatMessageProps> = ({
...
+                    <button
+                        className="message-start-editing-button"
+                        onClick={handleEditClick}
+                        title="Edit message"
+                    >
+                        <PencilIcon />
+                    </button>
+                    <button
+                        className="message-delete-button"
+                        onClick={() => onDeleteMessage?.(messageIndex)}
```
aarondr77, line 221: In practice we're always setting an `onDeleteMessage`. So let's just make this non-optional in the component declaration and then rely on the logic we already have here to only display these buttons if it is `editable`

aarondr77: The only thing is that the `delete` is a bit weird on a regular user input message because it just removes the user input, but not the associated response. I don't think it's a huge deal for now, but maybe we can just open an issue to remind ourselves to fix this at some point.

aarondr77 (approved):
Nice work! The buttons look pretty sweet IMO. I think the main thing here is that we can clean up the css a bit just so its a bit less fragile moving forward

mito-ai/style/ChatTaskpane.css
ngafar: Unfortunately not. We want the buttons to look like a single connected unit, with rounded corners only on the outer. Having a single border radius has all four corners rounded.

mito-ai/src/icons/Pencil.tsx
ngafar: It scales. Changing them both to be 14 causes the svgs to be off centered. I think this was a call Composer made.

PR #1545, opened 2025-02-24
Agent: Fix for warnings + errors

ngafar:
# Description

This update fixes a bug in agent mode's smart debugging: if a cell produces both a warning and an error message, the debugging feature would fail. Now, smart debugging works correctly even when both types of messages are output.

# Testing

1. Enter agent mode
2. Create any plan. 
3. Edit the plan so that one of the steps reads, "Using python, create a warning and an error message."

When the plan is run, you should see that the cell with the warning and error will trigger the smart debug workflow.  

# Documentation

N/A

### DIFF
--- a/mito-ai/src/utils/agentActions.tsx
+++ b/mito-ai/src/utils/agentActions.tsx
@@ -45,8 +45,11 @@ export const retryIfExecutionError = async (
 
         // If the code cell has an error, we need to send the error to the AI
         // and get it to fix the error.
-        const errorTraceback = cell?.model.outputs?.toJSON()[0].traceback as string[]
-        const errorMessage = getFullErrorMessageFromTraceback(errorTraceback)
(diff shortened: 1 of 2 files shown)


mito-ai/src/utils/agentActions.tsx
```diff
@@ -45,8 +45,11 @@ export const retryIfExecutionError = async (
 
         // If the code cell has an error, we need to send the error to the AI
         // and get it to fix the error.
-        const errorTraceback = cell?.model.outputs?.toJSON()[0].traceback as string[]
-        const errorMessage = getFullErrorMessageFromTraceback(errorTraceback)
+        const errorOutput = cell?.model.outputs?.toJSON().find(output => output.output_type === "error");
```
aarondr77, line 48: Nice fix!

PR #1548, opened 2025-02-25
Agent find and update evals

aarondr77:
# Description

Builds 10 evals that pass the entire notebook to the AI in the format:

```
[{
cell_id: <>
cell_type: code | markdown
code: <>
}]
```

Then, asks the AI to return a CellUpdate which is a cell id + new code for that cell. 

In these (still somewhat simple evals), its able to get the change right 100% of the time, but notice that with o3-mini it is pretty slow!

Screenshot 2025-02-27 at 1 14 57 PM

Screenshot 2025-02-27 at 1 18 37 PM

# Testing

Check it out and maybe add one or two? 

# Documentation

No.


aarondr77 (commented):
cleanup

evals/prompts/agent_find_and_update_prompts/simple_prompt.py
```diff
@@ -0,0 +1,60 @@
...
+    {{
+        cell_type: Literal['code', 'markdown']
+        id: str
+        code: str
+    }},
+    {{
+        cell_type: Literal['code', 'markdown']
+        id: str
+        code: str
+    }},
```
ngafar, line 25: Probably nothing, but do we need to have this example object twice? I would guess the fact that it's in an array should make it obvious that there'd be multiple items. 
```suggestion
    {{
        cell_type: Literal['code', 'markdown']
        id: str
        code: str
    }},
```

Issue #1544, opened 2025-02-24
Agent: if step has no code, don’t create new cell


ngafar:
When executing an agent's plan, each step will create a new code cell. This should not happen when a step has no code associated with it.

PR #1549, opened 2025-02-25
Agent: No code, no cells

ngafar:
# Description

Fixes https://github.com/mito-ds/mito/issues/1544

# Testing

In Agent mode, send any prompt. Edit at least one of the steps to be something like "skip me." When the Agent executes this item in the plan it should not create a new cell. 

# Documentation

N/A

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
@@ -587,11 +587,17 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
                 break;
             }
 
-            // Insert a new cell for the next step
-            await app.commands.execute("notebook:insert-cell-below")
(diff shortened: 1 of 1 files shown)


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -587,11 +587,17 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
                 break;
             }
 
-            // Insert a new cell for the next step
-            await app.commands.execute("notebook:insert-cell-below")
+            // Only create a new cell if the AI generated code in its response
+            if (aiMessage?.message) {
```
aarondr77, line 591: Technically this is not the most recent aiMessage anymore because at this point, we will have already gone through the retry during execution loop. But I guess this works because if the original one was empty then there is no way we would error.

aarondr77 (approved):

Issue #1536, opened 2025-02-18
Websocket Connection Automatically Closed


ngafar:
If I am working on a notebook, and I leave it idle for a little while, when I come back none of the AI features work. 

Looking at the browser console:

```
WebSocket is already in CLOSING or CLOSED state.

sendMessage | @ | websocketClient.js:106
```

This is the file from the message above: https://github.com/mito-ds/mito/blob/6cf1dc6e165c46770a8067fecc549fe35fb6edbc/mito-ai/src/utils/websocket/websocketClient.ts#L157

ngafar:
Also, you never know when you hit this error. Chat just shows _Thinking..._ indefinitely.

PR #1558, opened 2025-03-06
Add reconnect logic

ngafar:
# Description

Fixes https://github.com/mito-ds/mito/issues/1536

# Testing

Create a notebook and open the chat taskpane. Additionally, open dev tools to monitor the connection state.

Next, close your laptop, or put it to sleep. Wait a few seconds, then check on your notebook. You should see that the connection has been disconnected.

Send a message, in chat, and confirm that the connection is reestablished, and wait for the response.  

# Documentation

N/A


mito-ai/src/utils/websocket/websocketClient.ts
```diff
@@ -268,6 +289,66 @@ export class CompletionWebsocketClient implements IDisposable {
...
+    
+    try {
+      // Reinitialize connection
+      await this._initialize();
+      // Reset reconnect attempt counter on success
+      this._reconnectAttempt = 0;
+    } catch (error) {
+      console.error(`Reconnection attempt ${this._reconnectAttempt} failed:`, error);
+      // Try again recursively with exponential backoff if we haven't exceeded the limit
+      if (this._reconnectAttempt < this._maxReconnectAttempts) {
```
aarondr77, line 344: We already have this check above on line 315, so I think we can just call the this.reconnect(false) again and let the function handle it

mito-ai/src/utils/websocket/websocketClient.ts
```diff
@@ -268,6 +289,66 @@ export class CompletionWebsocketClient implements IDisposable {
...
+
+  /**
+   * Current reconnection attempt
+   */
+  private _reconnectAttempt = 0;
+
+  /**
+   * Attempt to reconnect the websocket with exponential backoff
+   */
+  async reconnect(forceReset: boolean = true): Promise<void> {
```
aarondr77, line 305: What is a forceReset? Can you maybe document this arg in the docstring here?

mito-ai/src/utils/websocket/websocketClient.ts
```diff
@@ -134,17 +148,34 @@ export class CompletionWebsocketClient implements IDisposable {
...
+            console.log('Successfully reconnected, now sending message');
+          } catch (reconnectError) {
+            console.error('Failed to reconnect websocket:', reconnectError);
+            reject(new Error('Failed to reconnect websocket before sending message'));
+            return;
+          }
+        }
+
+        if (this._socket && this._socket.readyState === WebSocket.OPEN) {
+          const id = message.message_id ?? crypto.randomUUID();
```
aarondr77, line 169: It looks like previously we were just sending `this._socket.send(JSON.stringify(message))` . Do we want to be creating a new id here if it doesn't exist?

aarondr77 (commented):
I'm having a little bit of trouble testing this. I can't get the message that says my websocket is disconnecting to appear. But the code looks good to me. Any advice on testing?

aarondr77 (approved):
Just dropping the instructions you shared for testing here so we can find them easily: 


1. Run jupyter lab.
2. Open the Jupyter Lab instance in your browser, and open dev tools. KEEP THIS WINDOW OPEN.
3. From the same terminal as step 1, quit jupyter.
4. Re-run jupyter lab.
5. Ignore any new Jupyter page that pops up, and instead go to the page from step 2.

At this point you should notice that the window from step 2 has shown a disconnect message in dev tools. Send a new message and watch it reconnect.

PR #1564, opened 2025-03-11
Mito AI Speed Test

ngafar:
# Description

Added a pytest to benchmark server vs. user key latency. 

# Testing

```bash
python -m pytest mito_ai/tests/performance_test.py -v -s
```

Also, you need to edit the `is_running_test` condition in open_ai_utils.py. 

After theses are complete you might need to scroll up to find the results table. T

# Documentation

N/A


aarondr77 (approved):
Can we just add the commands to run this into the README before merging?

PR #1578, opened 2025-03-17
Citation backend

aarondr77:
# Description

Now the Ai is instructed to create citation objects when it shares an insight with you. 

# Testing

Try it out and then make sure that the citation it generates is reasonable.

Some things to pay attention to: 
1. The citation should not be included in a comment of the code. It should only appear in the chat. 
2. The citation line should go to the right line (I've noticed this incorrect a few times)

# Documentation

When we release


mito-ai/mito_ai/providers.py
```diff
@@ -64,7 +64,7 @@ class OpenAIProvider(LoggingConfigurable):
     # We will try a total of 3 times. Once on the initial request, 
     # and then twice more if the first request fails.
     # Note that max_retries cannot be set to None. If we want to disable it, set it to 0.
-    timeout = 45
+    timeout = 90
```
aarondr77, line 67: go back to 45

PR #1582, opened 2025-03-19
Use 3 workers for tests

aarondr77:
# Description

Writes to the active Chat Thread instead of the most recent chat thread and restores the parallel testing with 3 workers. 

My updated understanding of playwright tests on GitHub actions is:

> GitHub Actions workers (or runners) are isolated execution environments. When you specify multiple workers (using parallelism or matrix strategy), GitHub provisions separate virtual machines or containers for each worker. Each worker: 1) **Has its own isolated file system** 2) Runs independently of other workers 3) Gets its own copy of your repository. This means each worker has its own completely separate working directory with a fresh checkout of your code.

As a result, we can run the tests in parallel and not run into issues with conflicting updates to the Jupyter server's `active_thread_id` or writes to the `ai-chats` directory. 

Coincidentally, setup tools released a new version today that broke the CLI on Python 3.12, so I also pinned the version of setup tools temporarily until they fix it. 

# Testing

Try it out + see that these concurrent tests pass. 

# Documentation

None


mito-ai/mito_ai/message_history.py
```diff
@@ -133,6 +133,7 @@ def __init__(self) -> None:
 
         # Load existing threads from disk on startup
         self._load_all_threads_from_disk()
+        self._active_thread_id = self._get_newest_thread_id() or self.create_new_thread()
```
aarondr77, line 136: If no chats are created yet, create a new one.

mito-ai/mito_ai/message_history.py
```diff
@@ -218,60 +220,54 @@ def ai_optimized_history(self) -> List[ChatCompletionMessageParam]:
         For backward compatibility: returns the LLM history of the newest thread.
         """
         with self._lock:
-            thread_id = self._get_newest_thread_id()
-            if not thread_id or thread_id not in self._chat_threads:
+            if not self._active_thread_id or self._active_thread_id not in self._chat_threads:
```
aarondr77, line 223: I don't think we need this first part of the check

mito-ai/mito_ai/message_history.py
```diff
@@ -218,60 +220,54 @@ def ai_optimized_history(self) -> List[ChatCompletionMessageParam]:
...
     
     @property
     def display_history(self) -> List[ChatCompletionMessageParam]:
         """
         For backward compatibility: returns the display history of the newest thread.
         """
         with self._lock:
-            thread_id = self._get_newest_thread_id()
-            if not thread_id or thread_id not in self._chat_threads:
+            if not self._active_thread_id or self._active_thread_id not in self._chat_threads:
```
aarondr77, line 237: same

mito-ai/setup.py
```diff
@@ -91,19 +91,22 @@ def get_data_files_from_data_files_spec(
         "tornado>=6.2.0",
         "traitlets",
         "pydantic",
+        "packaging>=23.1,<24.0"
```
aarondr77, line 94: We might be able to remove `packaging`

PR #1583, opened 2025-03-20
Polish chat history

aarondr77:
# Description

Polishes the Chat History Dropdown:
- Makes sure it scrolls vertically if there are a bunch of chats
- Does not get cut off horizontally if its a small chat taskpane
- A bunch of tiny spacing, color, etc. polish

# Testing

I added a dev script that you can run to quickly generate a bunch of ai-chats. You can use this to fill your ai-chats folder if it's empty. 

Make sure it looks good. Feel free to fix up any tiny polish issues, but I think it's probably good enough! 

# Documentation

Yes, we need to add docs about Chat History!

### DIFF
--- a/mito-ai/dev/README.md
+++ b/mito-ai/dev/README.md
@@ -1,0 +1,18 @@
+# Mito AI Dev Scripts
+
+### Generate AI Chats
+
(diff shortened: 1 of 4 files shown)


aarondr77:
These tests are fixed in https://github.com/mito-ds/mito/pull/1582

PR #1586, opened 2025-03-20
Single error notification for inline completion

ngafar:
# Description

Added a flag to make sure we are only showing the inline completion error notification once per session. If the user closes it, they won't see it again, nor will it stack up as they keep typing. 

# Testing

Update `MITO_AI_PROD_URL` in `open_ai_utils.py`, and break the URL. Then start typing. 

# Documentation

N/A

### DIFF
--- a/mito-ai/src/Extensions/InlineCompleter/provider.ts
+++ b/mito-ai/src/Extensions/InlineCompleter/provider.ts
@@ -56,6 +56,8 @@ export class MitoAIInlineCompleter
   // We only want to display the free tier limit reached notification once 
   // per session to avoid spamming the user. 
   private _displayed_free_tier_limit_reached_notification = false;
+  // Similarly, we only want to show the general completion failure notification once
+  private _displayed_completion_failure_notification = false;
 
 
   constructor({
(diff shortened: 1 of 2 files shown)


mito-ai/src/Extensions/InlineCompleter/provider.ts
```diff
@@ -326,19 +328,24 @@ export class MitoAIInlineCompleter
...
-      actions: [
-        {
-          label: 'Show Traceback',
-          callback: async (): Promise<void> => {
-            await showErrorMessage('Inline completion failed on the server side', {
-              message: error.traceback ?? 'An unknown failure happened when requesting a completion.'
-            });
+    // Only show the notification if we haven't shown one already
+    if (!this._displayed_completion_failure_notification) {
+      Notification.emit(`Inline completion failed: ${error.error_type}`, 'error', {
```
aarondr77, line 333: Sweet! While we are at it, can we update this message so it's more understandable. And maybe we can add a `Get Help` button or something that just opens their email with the mailto link and fills in our emails + a subject 

Screenshot 2025-03-20 at 5 25 00 PM

aarondr77 (approved):

PR #1590, opened 2025-03-25
Response Streaming Backend - User Key

ngafar:
# Description

Added response streaming to the backend — only for "user key" users.  

# Testing

Use `wscat` or Python to make a request to the backend. Send a sufficiently complex prompt to ensure that you can visually confirm the response streaming. 

Right now this only works for chat. 

# Documentation

N/A


mito-ai/mito_ai/completion_handlers/chat_completion_handler.py
```diff
@@ -56,6 +56,59 @@ async def get_completion(
...
+        
+        # Stream the completions using the provider's stream method
+        async for chunk in provider.stream_and_save_completions(
+            message_type=MessageType.CHAT,
+            messages=message_history.ai_optimized_history,
+            model=MESSAGE_TYPE_TO_MODEL[MessageType.CHAT],
+            message_id=message_id,
+            message_history=message_history
+        ):
+            yield chunk
```
aarondr77, line 110: I think there's one important architectural change that can make here that will make this a lot less code and much more maintainable going forward. Since `stream_completion` and `get_completion` are pretty much the same except for this one part of code, let's refactor it so that the shared code is in a helper function that we can reuse. We've already had bugs where we forget to do things like handle if the index is None, so the more we can reuse the code the less error prone and the easier it will be to have testing code coverage. 

We can have two helper functions: 1) prepare_chat_request 2) save_ai_response. If we do this, then we can pull the saving of the AI response out of the provider which will be really nice -- we don't need to carve out a case of the inline completer in the provider.ts file (which you helped establish as a really sensible rule in our white boarding session). This also means we don't need to do any of the circlular import avoiding stuff.

aarondr77: .... I guess one other option is to just not support non-streaming responses for chat, agent, etc. Is there a world in which we are not going to want it to stream?

aarondr77: Actually this would def be the easiest ^. I'm down for that. There's no reason not to stream, right?

mito-ai/mito_ai/providers.py
```diff
@@ -354,3 +359,63 @@ async def stream_completions(
...
+            
+            # Yield the chunk to the caller
+            yield reply
+        
+        # Save the accumulated response to the message history if provided
+        if message_history is not None and message_type != MessageType.INLINE_COMPLETION:
+            ai_response_message: ChatCompletionMessageParam = {
+                "role": "assistant", 
+                "content": accumulated_response
+            }
```
aarondr77, line 412: I think we should do this in the completion_handler instead of here, so that we can continue to isolate all of the specific message type logic into the completion handlers and not bleed out in the provider which should just be the LLM functionality. 

I think we could either do that by:
1. Making this `stream_and_save_completions` return the accumulated response at the end 
2. Just building the accumulated response in the completion_handler.

mito-ai/mito_ai/handlers.py
```diff
@@ -181,13 +182,30 @@ async def on_message(self, message: str) -> None: # type: ignore
...
+                if stream and self._llm.can_stream:
+                    # Use stream_chat_completion to stream the response
+                    async for chunk in stream_chat_completion(
+                        chat_metadata, 
+                        self._llm, 
+                        message_history, 
+                        message_id,
+                    ):
+                        self.reply(chunk)                    
+                    return
```
aarondr77, line 205: We have this `async for chunk in ... ` logic in three places: 

1. Handler
2. Completion Handler 
3. Provider

We send the chunks all the way back up 3 -> 2 -> 1 and then just fire off the replies when we get them back. 

I wonder if we could simplify this by just passing the `self.reply` function down to the provider and then fire off replies from there instead. Then the provider can also just return the accumulated result to the Completion Handler and we can save it in the message history

aarondr77 (commented):
Two small refactor ideas that I think will streamline this even more. Let me know what you think

PR #1594, opened 2025-03-28
Fix yarn and python caching

fcollonval:
# Description

This fixes the workflow configuration to activate caching especially for yarn as it is currently almost not working as seen in the cache sizes for node-cache: https://github.com/mito-ds/mito/actions/caches

This implies ensuring `actions/setup-node` uses yarn 3 (by default it is using v1). To achieve that, the `packageManager` is set in the package.json files and `corepack` is enable prior to the setup node action.


fcollonval:
Closing as unfortunately corepack does not work as expected on windows.

PR #1599, opened 2025-04-01
Streaming user key frontend - take 2

ngafar:
# Description

Added streaming for chat messages. 

# Testing

Clone this PR, and send a chat message. 

You may also need to run:

```bash
git pull origin streaming-user-key
```

# Documentation

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
@@ -271,7 +271,10 @@ export class ChatHistoryManager {
         // Find the last AI message in the history
         const lastAIMessageIndex = this.getLastAIMessageIndex();
         
-        if (lastAIMessageIndex === undefined) {
+        if (
+            lastAIMessageIndex === undefined || 
+            this.displayOptimizedChatHistory.length !== lastAIMessageIndex + 1
+        ) {
             // If no AI message exists, create a new one
             this.addAIMessageFromResponse(messageContent, promptType);
         } else {
--- a/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
@@ -309,6 +309,10 @@ export class ChatHistoryManager {
     }
 
     getLastAIMessageIndex = (): number | undefined => {
+        // We assume that assistant messages are always separated by user messages.
+        // This allows us to simply find the last assistant message in the history.
+        // If this invariant changes (e.g., if we need to support consecutive assistant messages),
+        // we should modify this to use message IDs instead.
         const displayOptimizedChatHistory = this.getDisplayOptimizedHistory()
         const aiMessageIndexes = displayOptimizedChatHistory.map((chatEntry, index) => {
             if (chatEntry.message.role === 'assistant') {
--- a/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
@@ -264,6 +264,25 @@ export class ChatHistoryManager {
         );
     }
 
+    addStreamingAIMessage(
+        messageContent: string,
+        promptType: PromptType
+    ): void {
+        // Find the last AI message in the history
+        const lastAIMessageIndex = this.getLastAIMessageIndex();
+        
+        if (lastAIMessageIndex === undefined) {
+            // If no AI message exists, create a new one
+            this.addAIMessageFromResponse(messageContent, promptType);
+        } else {
+            // Update the last AI message with the new content
+            const lastMessage = this.displayOptimizedChatHistory[lastAIMessageIndex];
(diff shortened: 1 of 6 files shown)


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -473,59 +478,148 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+            // Create the stream handler function and store it in the ref
+            const streamHandler = (
+                _: CompletionWebsocketClient, chunk: ICompletionStreamChunk
+            ) => {                
+                // Use a ref to accumulate the content properly
+                streamingContentRef.current += chunk.chunk.content;
+                
+                // Force a re-render to update the UI
+                forceUpdate();
+
```
aarondr77, line 505: If we are forcing an update every time we update the streamingContentRef with new content, aren't we just duplicating the behavior that is already built into useState? Every time the state updates the component will rerender. 

Maybe there's a reason for doing it this way that I don't know though! If there is, let's document it!

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -473,59 +478,148 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
-            addAIMessageFromResponseAndUpdateState(
-                (error as any).hint ? (error as any).hint : `${error}`,
-                completionRequest.metadata.promptType,
-                newChatHistoryManager,
-                true
-            )
-        } finally {
-            // Reset states to allow future messages to show the "Apply" button
-            setCodeReviewStatus('chatPreview');
+        } else {
```
aarondr77, line 570: The giant try statement here and above are exactly the same (I just did a CMD+F). So can't we just change this up so that we have

```typescript
if (streaming) {
    // set up stream handling stuff
}

try {
    // send the message and do all of the handling
}
//... etc

```

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -473,59 +478,148 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+            
+            // Connect the handler
+            websocketClient.stream.connect(streamHandler, null);
 
-                if (completionRequest.metadata.promptType === 'agent:execution' || completionRequest.metadata.promptType === 'agent:autoErrorFixup') {
-                    // Agent:Execution prompts return a CellUpdate object that we need to parse
-                    const agentResponse: AgentResponse = JSON.parse(content)
-                    newChatHistoryManager.addAIMessageFromAgentResponse(agentResponse)
+            try {
+                const aiResponse = await websocketClient.sendMessage<ICompletionRequest, ICompletionReply>(completionRequest);
```
aarondr77, line 521: When we're streaming, this aiResponse is just the very first chunk of the response. So what we're doing here is:

1. Get the first response from the AI 
2. Save it in the ChatHistory Manager 
3. In the streamHandler when we call the `addStreamingAIMessage` each time we get a chunk back, we overwrite the last AI message with the new content. 

I think a simple change that we could make here is just:

Once we call `await websocketClient.sendMessage`, we do error handling incase the connection fails, but then we don't do any of the message chatHistory stuff and let the ChatHistory Manager  just create the message. 

Does that make sense?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
aarondr77: Actually, see other comment, I think we don't need all of this stuff for the streaming case

mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
```diff
@@ -264,6 +264,25 @@ export class ChatHistoryManager {
         );
     }
 
+    addStreamingAIMessage(
+        messageContent: string,
+        promptType: PromptType
+    ): void {
+        // Find the last AI message in the history
+        const lastAIMessageIndex = this.getLastAIMessageIndex();
```
aarondr77, line 272: So we are making an assumption here that there are never going to be an assistant message followed directly by another assistant message... that seems reasonable! Want to just document that. If that invariant changes, then we can just start associated ID's with the messages and checking it that way.

aarondr77 (approved):
Sweet! This PR is 10x better than the first approach. Really glad that you took a step back and re-evaluated how we could get rid of all that extra stuff. 

A little bit of cleanup I think we can do here to simplify things even more and then LGTM

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
ngafar: You're right, we can use `setChatHistoryManager` for this. I just needed to create a new instance of the chat history manager so that a re-render is triggered. 

I've opted to remove the `forceUpdate` code in favor of this solution, since we're more familiar with the chat hsitory manager.

PR #1601, opened 2025-04-02
Condensed chat history

aarondr77:
# Description

Drops the metadata from old chat messages in the ai chat history. We do not need to store all old states of the notebook, the files that were in the current working directory, etc. Those go stale quickly! 

# Testing

Try it out. Any user message older than 3 messages should not contain all of the old extra metadata. 

Also check out the new pytests. 

# Documentation

No.


mito-ai/mito_ai/tests/message_history_test.py
```diff
@@ -0,0 +1,137 @@
+# Copyright (c) Saga Inc.
```
aarondr77, line 1: delete

mito-ai/mito_ai/utils/message_history_utils.py
```diff
@@ -0,0 +1,62 @@
...
+    Removes specific metadata sections from message content to reduce token count so
+    that users don't exceed the token limit for the LLM.
+
+    These sections are replaced with a placeholder text.
+    """
+    if not isinstance(content, str):
+        return content
+
+    # Replace "Files in the current directory:" section
+    content = re.sub(
```
ngafar, line 22: Alternatively:

```suggestion
    # Replace metadata sections with placeholders
    section_headings = [
        FILES_SECTION_HEADING,
        VARIABLES_SECTION_HEADING,
        JUPYTER_NOTEBOOK_SECTION_HEADING
    ]
    
    for heading in section_headings:
        content = re.sub(
            f"{re.escape(heading)}\n(?:.+\n)+",
            f"{heading} {CONTENT_REMOVED_PLACEHOLDER}\n",
            content,
        )
```

mito-ai/mito_ai/utils/message_history_utils.py
```diff
@@ -0,0 +1,62 @@
...
+        f"{VARIABLES_SECTION_HEADING} {CONTENT_REMOVED_PLACEHOLDER}\n",
+        content,
+    )
+
+    # Replace "{JUPYTER_NOTEBOOK_SECTION_HEADING}" section
+    content = re.sub(
+        f"{re.escape(JUPYTER_NOTEBOOK_SECTION_HEADING)}\n(?:.+\n)+",
+        f"{JUPYTER_NOTEBOOK_SECTION_HEADING} {CONTENT_REMOVED_PLACEHOLDER}\n",
+        content,
+    )
```
ngafar, line 40:
```suggestion
    # Replace metadata sections with placeholders
    section_headings = [
        FILES_SECTION_HEADING,
        VARIABLES_SECTION_HEADING,
        JUPYTER_NOTEBOOK_SECTION_HEADING
    ]
    
    for heading in section_headings:
        content = re.sub(
            f"{re.escape(heading)}\n(?:.+\n)+",
            f"{heading} {CONTENT_REMOVED_PLACEHOLDER}\n",
            content,
        )
```

ngafar: I will concede that there is something nice about having everything spelt out, an almost quaint feel. And if we're sure that we wont have more sections like this, well then I'm fine with things as is.

mito-ai/mito_ai/prompt_builders/smart_debug_prompt.py
```diff
@@ -15,10 +20,10 @@ def create_error_prompt(
     return f"""You are debugging code in a JupyterLab 4 notebook. Analyze the error and provide a solution that maintains the original intent.
 
 <Example 1>
-Files in the current directory:
+{FILES_SECTION_HEADING}
```
ngafar, line 23: I wonder if this would be better as a function, that way we could group this with:

```python
files_str = '\n'.join([f"{file}" for file in files])
```

and have the function return both. 

It would get reused across a half dozen prompt files.

mito-ai/mito_ai/tests/utils/test_message_history_utils.py
```diff
@@ -0,0 +1,350 @@
...
+        {"role": "system", "content": system_message_with_sections},
+        {"role": "user", "content": user_message_with_sections},
+        {"role": "assistant", "content": assistant_message_with_sections},
+        {"role": "user", "content": "Recent user message"},
+    ]
+    
+    # Keep only the most recent message
+    result = trim_old_messages(messages, keep_recent=1)
+    
+    # System message should remain unchanged even though it's old
```
ngafar, line 245: Good catch, this completly slipped my mind while reviewing.

ngafar (commented):
Looks good, I'd say we should also remove the instructions from these every cell except the last. For example in the chat prompt we'd remove everything from, "You are an expert python programmer writing a script in a Jupyter notebook..." to "</Example>."

ngafar (approved):

PR #1602, opened 2025-04-02
mito-ai: do not add mito functions to context

aarondr77:
# Description

Removes the Mito exported functions from the variables we pass to the LLM. These are mostly irrelevant and the LLM can probably figure it out if it needs too. 

Making the Defined Variables in my testing notebook go from 5608 to 993 characters 

In retrospect, not sure this was really worth the 35 mins it took... I don't think this is where users are running into messages being too long

# Testing

Try it out! 

# Documentation

No


aarondr77 (commented):
LGTM

ngafar (approved):

PR #1604, opened 2025-04-03
Smart debug streaming - user key

ngafar:
# Description

Added streaming for smart debug and code explain.

Note that smart debug will display the inner thoughts. Any solution I could think of seemed too fragile, i.e. looking for specific words before streaming. I don't think it's a huge deal, but I am open to discussion.  

# Testing

Make sure you have a user key setup and test:

1. Smart debug
2.  Code explain

# Documentation

N/A


mito-ai/mito_ai/providers.py
```diff
@@ -368,3 +382,48 @@ async def stream_completions(
                     error=CompletionError.from_exception(e),
                 )
                 break
+
+    async def stream_and_save_completions(
```
aarondr77, line 386: We should update this to `stream_completions` since it's no longer doing the saving. nice!

mito-ai/mito_ai/completion_handlers/smart_debug_handler.py
```diff
@@ -1,66 +1,169 @@
...
+
+__all__ = ["get_smart_debug_completion", "stream_smart_debug_completion"]
+
 
 class SmartDebugHandler(CompletionHandler[SmartDebugMetadata]):
     """Handler for smart debug completions."""
-    
+
     @staticmethod
     async def get_completion(
```
aarondr77, line 31: will be nice when we can delete this whole function for the functions we want to have streaming!

mito-ai/mito_ai/providers.py
```diff
@@ -41,6 +51,10 @@
     log_ai_completion_success,
 )
 
+# Use TYPE_CHECKING to avoid circular imports
+if TYPE_CHECKING:
+    from mito_ai.message_history import GlobalMessageHistory
```
aarondr77, line 56: This can be deleted now, right?

aarondr77 (approved):

mito-ai/mito_ai/providers.py
ngafar: This file and the `stream_completions` function (which already exists in this file) are so heavily edited in https://github.com/mito-ds/mito/pull/1609#pullrequestreview-2747590361, that it might create too much merge headaches to make edits here. So I'll make the changes in #1609.

PR #1606, opened 2025-04-04
Show the Active Cell Output to the AI Chat

aarondr77:
# Description

Show the active cell output to the AI Chat prompt so it can answer questions, figure out how to improve graphs, etc. 

# Testing

Try it out. 

# Documentation

Yes, we should update our docs to let users know this functionality exists!


mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
```diff
@@ -312,10 +313,6 @@ export class ChatHistoryManager {
     }
 
     getLastAIMessageIndex = (): number | undefined => {
-        // We assume that assistant messages are always separated by user messages.
```
ngafar, line 315: We should probably keep this, right?

ngafar (approved):

PR #1607, opened 2025-04-04
Enabling Ollama Support via Environment Variables

hegdes-usc:
# Description

Modified the `mito-ai/mito_ai/providers.py` to make calls to local LLMs running on Ollama. Modified the `OpenAIProvider` class to include new ollama properties also modified the `request_completions` and `stream_completions` methods to account for Ollama changes.

https://github.com/user-attachments/assets/865e8814-f654-49f3-a92d-9557fef643ab

# Testing

- [ ] If I changed the interaction with JupyterLab, I tested that it does not break other programs (like VS Code), and tested that it works "multiple times" in the same notebook. (N/A)

# Note

This can be followed up by a UI update where we allow users to chose Ollama models from a dropdown. This still needs to be worked on. 

Three new environment variables are introduced as part of this PR:

- `USE_OLLAMA`: This is a boolean flag that allows the user to Enable/Disable Ollama (This would have to be done in the UI in the future)

- `OLLAMA_MODEL`: Specifies the LLM that is to be chosen.
  
-  `OLLAMA_HOST` (Optional) : Specifies the Ollama Host URL where Ollama is running. This is set to `http://localhost:11434` by default and this env variable need not be set if it is running on the default port.

Note that I have not added unit tests yet 🫣 and plan to add them if this approach is approved.

# Documentation

### DIFF
--- a/mito-ai/setup.py
+++ b/mito-ai/setup.py
@@ -95,7 +95,6 @@ setup(
         "traitlets",
         "pydantic",
         "ollama",
-        "aiohttp",
     ],
     extras_require = {
         'deploy': [
--- a/mito-ai/setup.py
+++ b/mito-ai/setup.py
@@ -94,7 +94,6 @@ setup(
         "tornado>=6.2.0",
         "traitlets",
         "pydantic",
-        "ollama",
     ],
     extras_require = {
         'deploy': [
--- a/mito-ai/setup.py
+++ b/mito-ai/setup.py
@@ -94,6 +94,8 @@ setup(
         "tornado>=6.2.0",
         "traitlets",
         "pydantic",
+        "ollama",
+        "aiohttp",
     ],
     extras_require = {
         'deploy': [
--- a/mito-ai/mito_ai/providers.py
+++ b/mito-ai/mito_ai/providers.py
@@ -3,13 +3,16 @@
 
 from __future__ import annotations
 
+import asyncio
 import os
-from typing import Any, AsyncGenerator, Dict, List, Literal, Optional, Union, Callable
+from typing import Any, AsyncGenerator, Dict, List, Optional, Union, Type
 
 import openai
+import ollama
+from aiohttp import ClientSession
 from openai._streaming import AsyncStream
 from openai.types.chat import ChatCompletionChunk, ChatCompletionMessageParam
-from traitlets import  Instance, Unicode, default, validate
+from traitlets import Instance, Unicode, default, validate
 from pydantic import BaseModel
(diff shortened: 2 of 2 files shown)


mito-ai/setup.py
```diff
@@ -94,6 +94,7 @@ def get_data_files_from_data_files_spec(
         "tornado>=6.2.0",
         "traitlets",
         "pydantic",
+        "ollama",
```
aarondr77, line 97: This new approach means we can remove ollama as a dependency!

aarondr77 (commented):
Hey @hegdes-usc thanks for opening this PR! It's a great first contribution! 

There's one architectural change that I think we should make: Instead of using the actual ollama sdk, I think it will be better for us long term to support Ollama through its OpenAI SDK compatibility https://ollama.com/blog/openai-compatibility

So what I'm thinking is: 
1. If the user has an `OPENAI_API_KEY` set, we use that. 
2. If the user has an `OLLAMA_MODEL` and `OLLAMA_BASE_URL` set, we use that
3. Otherwise, we use the Mito AI server 

So the checks for 1 and 2 should determine how we set up the `_openAI_async_client` and `_openAI_sync_client`.

The reason that I think this approach is better longterm is: This will let us support a bunch of different models without a bunch of code branching. We'll be able to take the same approach for Anthopic and Google models because you can also use them through the OpenAI SDK. And then, in a separate PR once we add support for other models, we can add a dropdown in the UI to let the user select which model they want to use instead of using this waterfall of environment variable checks.

aarondr77:
@hegdes-usc since this is your first contribution to this codebase, can you check out our CLA here: https://github.com/mito-ds/monorepo/blob/dev/CLA.md

If you agree, just hit me back with "I have read the CLA Document and I hereby sign the CLA" -- and we'll get this merged in as soon as this small review is addressed!

mito-ai/setup.py
hegdes-usc: I have addressed this comment in the subsequent commit: [654f35e](https://github.com/mito-ds/mito/pull/1607/commits/654f35e2d44fb1745d28e1166f6a39f781642c35)

hegdes-usc:
> @hegdes-usc since this is your first contribution to this codebase, can you check out our CLA here: https://github.com/mito-ds/monorepo/blob/dev/CLA.md
> 
> If you agree, just hit me back with "I have read the CLA Document and I hereby sign the CLA" -- and we'll get this merged in as soon as this small review is addressed!

I have read the CLA Document and I hereby sign the CLA

PR #1611, opened 2025-04-07
Quota accounting

aarondr77:
# Description

Don't count naming the chat against the user's quota. 

# Testing

Double check that when we automatically name the chat, we don't count that against the user's quota so they actually get a chance to use the tool :) 

# Documentation

Nope


ngafar (approved):
Good catch!

Issue #1613, opened 2025-04-08
Agent should wait for cell to finish executing before going to next step


aarondr77:
Agent should wait for cell to finish executing before going to next step so that it can get updated variables, etc.

PR #1615, opened 2025-04-08
mito-ai: wait for execution to finish

aarondr77:
# Description

Makes sure that we wait before constructing the next message to send to the agent. Without it, if we had a long running data import, the message we send back to the agent would not contain the successful dataframe creation. 

It used to say things like this: 

> Since the bike rides dataframe was not successfully displayed, we need to verify that the cell ran without any errors and that the dataframe bike_rides_df is defined. Please confirm that no errors occurred and that the dataframe is loaded properly, or let me know if there's a specific output or error message.

# Testing

Ask the agent to import a large excel file and tell you something about it. It should now not stop execution with a message like the above. 

# Documentation

No


aarondr77:
This resolves https://github.com/mito-ds/mito/issues/1613

mito-ai/src/Extensions/ContextManager/VariableInspector.tsx
```diff
@@ -115,6 +115,8 @@ print(structured_globals())
 // Function to fetch variables and sync with the frontend
 export function fetchVariablesAndUpdateState(notebookPanel: NotebookPanel, setVariables: (variables: Variable[]) => void): void {
     const kernel = notebookPanel.context.sessionContext.session?.kernel;
+
+    console.log("A1")
```
ngafar, line 119: We can probably drop these before merging. Same for "A2" on line 137.

mito-ai/src/utils/agentActions.tsx
```diff
@@ -46,7 +47,14 @@ export const acceptAndRunCode = async (
...
     previewAICodeToActiveCell()
     acceptAICode()
+
+    // This awaits until after the execution is finished
     await app.commands.execute("notebook:run-cell");
+
+    // By sleeping here, we make sure that this function returns after the variable manager
+    // has updated the state of the variables. This ensures that on the next Ai message
+    // gets the most up to date data.
+    await sleep(1000)
```
ngafar, line 57: Not opposed to sleep, but we should proably look for a more deterministic way of doing this.

ngafar (approved):

PR #1618, opened 2025-04-10
mito-ai: log thread_id

aarondr77:
# Description

Better thread id handling

### DIFF
--- a/mito-ai/mito_ai/providers.py
+++ b/mito-ai/mito_ai/providers.py
@@ -199,8 +199,9 @@ This attribute is observed by the websocket provider to push the error to the cl
         message_type: MessageType,
         messages: List[ChatCompletionMessageParam], 
         model: str,
+        thread_id: Optional[str],
         user_input: Optional[str] = None,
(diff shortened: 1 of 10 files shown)


mito-ai/mito_ai/providers.py
```diff
@@ -275,23 +277,14 @@ async def stream_completions(
...
         message_id: str,
+        thread_id: str,
         reply_fn: Callable[[Union[CompletionReply, CompletionStreamChunk]], None],
         user_input: Optional[str] = None,
         response_format_info: Optional[ResponseFormatInfo] = None
     ) -> str:
         """
         Stream completions from the OpenAI API and return the accumulated response.
-        
-        Args:
```
ngafar, line 285: Looks like we lost a bunch of docstring here.

ngafar (approved):

mito-ai/mito_ai/providers.py
aarondr77: ahh yeah I just deleted this cause I think docstrings like these should be intuitive enough from the parameter names and they get out of date pretty quickly.

PR #1622, opened 2025-04-11
Streaming Stop Button

ngafar:
# Description

Added a stop streaming button. It replaces the submit button when streaming. 

# Testing

Send a message then use the stop button to stop the stream. 

Be sure to test code explain and Smart Debug. 

# Documentation

N/A

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
@@ -1252,7 +1252,7 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
                         title="Agent can create plans and run code."
                     />
                     <button
-                        className="button-base submit-button"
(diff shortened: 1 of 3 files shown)


aarondr77:
hmm let's sit on this one until after the release. Hitting stop and then having it still get appended to the chat history (which means what the user is seeing and what the AI gets is different) gives me a little bit of pause. Thoughts?

aarondr77:
After hitting `stop` I'm getting stuck with this `thinking` message still 
Screenshot 2025-04-11 at 2 22 51 PM

aarondr77:
The pause icon should be vertically centered with the text

Screenshot 2025-04-11 at 2 22 17 PM

ngafar:
I'm no longer getting the "thinking" message after hitting stop. I have a strong feeling one of my pre-release fixes resolved this. 

As for vertical centering (my never ending nightmare, and fiercest enemy), I think it may be pretty close to centered. The stop button is a unicode character, not an SVG. Here are a few options (we're using the last one):

Screenshot 2025-04-15 at 1 09 49 PM

PR #1623, opened 2025-04-11
Fix issues with webscockets

ktaletsk:
# Description

## Replace HEAD request with query param on the main GET request
In situation where JupyterLab is running behind OAuth 2, like in JupyterHub with HTTPS enabled, the HEAD method is not supported. We use HEAD request to check the connection before opening a websocket. This PR replaces HEAD request with a query parameter in the main GET request thus enabling the extension to work with OAuth 2 flow.

## Fix websocket reconnection issues

1. Added error handling to the `initialize()` method to better report connection failures
2. Implemented a flag `_initializationInProgress` to prevent multiple concurrent initialization attempts
3. Added proper cleanup of existing sockets before creating new ones
4. Added a connection timeout to detect stalled connections
5. Reset reconnection counters after failures to ensure future attempts can succeed
6. Created a more robust error propagation

### DIFF
--- a/mito-ai/mito_ai/handlers.py
+++ b/mito-ai/mito_ai/handlers.py
@@ -73,11 +73,11 @@ class CompletionHandler(JupyterHandler, WebSocketHandler):
         return get_logger()
 
     @tornado.web.authenticated
-    def head(self) -> None:
-        """Handle a HEAD request for the websocket."""
+    def get_service_status(self) -> None:
+        """Handle a GET request for service availability check."""
         self.set_status(HTTPStatus.OK)
         self.finish()
-
+        
     async def pre_get(self) -> None:
         """Handles websocket authentication/authorization."""
         # authenticate the request before opening the websocket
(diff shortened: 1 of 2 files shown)


mito-ai/mito_ai/handlers.py
```diff
@@ -73,8 +73,8 @@ def log(self) -> logging.Logger:
         return get_logger()
 
     @tornado.web.authenticated
-    def head(self) -> None:
-        """Handle a HEAD request for the websocket."""
+    def get_service_status(self) -> None:
```
aarondr77, line 76: On dev this old `head` function was getting triggered, but in this new branch the `get_service_status` function doesn't seem to get called. Is that an intentional change? If so, can we document what's going on here.

aarondr77 (changes requested):

mito-ai/mito_ai/handlers.py
ktaletsk: Good catch! `get_service_status` was called initially and then through refactors I lost that call. Ultimately, I just moved what the function does inside of `get` function now

PR #1645, opened 2025-04-16
Agent can create markdown cells

ngafar:
# Description

The Agent should now be able to create markdown cells. 

# Testing

# Documentation

N/A

### DIFF
--- a/mito-ai/mito_ai/models.py
+++ b/mito-ai/mito_ai/models.py
@@ -27,6 +27,7 @@ class CellUpdate(BaseModel):
 # that we show it in the system prompt.
 class AgentResponse(BaseModel):
     type: Literal['cell_update', 'get_cell_output', 'finished_task']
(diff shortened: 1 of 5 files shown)


aarondr77 (approved):
Nice work! This looks pretty good and the mechanics are working pretty well. I think there is some additional prompt engineering we can do in a separate PR to make the markdown responses better, but we can save that for a different PR. 

For example, I asked it to document a notebook and it added these markdown cells at the bottom, which are not documenting any code sections. This is wrong. 
Screenshot 2025-04-16 at 3 00 18 PM


### Tests
Let's add an integration test with a simple prompt like 

"Write <span color: blue>Hello World</span> in a markdown cell"

And then we can test that `<span color: blue>` is not part of the result. Since we render this as markdown it should be hidden and we should just see "Hello World"

Issue #1627, opened 2025-04-14
Agent Stuck on Missing Packages


ngafar:
When the agent encounters a "package not found" error, it just wraps it in a try/except and stops progressing. 

We could add an instruction to the system prompt, telling the agent to add something like `!pip install scikit-learn --quiet` to the top of the cell, if it encounters a package not found.

PR #1649, opened 2025-04-17
Agent mode can install packages

ngafar:
# Description

Fixes https://github.com/mito-ds/mito/issues/1627. Adds instructions on how the Agent can install packages it needs. 

# Testing

Uninstall matplotlib:

```bash
!pip uninstall matplotlib
```

Then ask the Agent to "make a chart." 

The Agent should assume matplotlib is already installed, and start to create a chart. When it runs the code you should get a ModuleNotFoundError. 

Alternatively, you could ask the Agent to use a library you don't have installed (seaborn, pytorch, etc). However, it may install the package before getting started. To avoid this behavior you may need to prompt it to not install anything until it hits an error. 

# Documentation

N/A


aarondr77 (approved):
Let's add this same line to the non-agent smart debug and then LGTM

PR #1651, opened 2025-04-17
Azure for enterprise

aarondr77:
# Description

Setting Azure OpenAI endpoint is an enterprise feature. 

# Testing

See the tests + try setting the environment variables 

# Documentation

We will update on release

### DIFF
--- a/mito-ai/mito_ai/providers.py
+++ b/mito-ai/mito_ai/providers.py
@@ -85,7 +85,7 @@ This attribute is observed by the websocket provider to push the error to the cl
         client = openai.OpenAI(api_key=api_key)
         try:
             # Make an http request to OpenAI to make sure it works
-            self.models = client.models.list()
+            client.models.list()
         except openai.AuthenticationError as e:
             self.log.warning(
                 "Invalid OpenAI API key provided.",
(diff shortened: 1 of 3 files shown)


mito-ai/mito_ai/providers.py
```diff
@@ -254,7 +254,7 @@ def _build_openai_client(self) -> Optional[Union[openai.AsyncOpenAI, openai.Asyn
         return client
 
     def _resolve_model(self, model: Optional[str] = None) -> str:
-        if constants.AZURE_OPENAI_MODEL and constants.AZURE_OPENAI_API_KEY and constants.AZURE_OPENAI_ENDPOINT:
+        if is_azure_openai_configured():
             return constants.AZURE_OPENAI_MODEL
```
ngafar, line 258: This is throwing a linter error in Cursor:

> Expression of type "str | None" is incompatible with return type "str"

We can get rid of this by adding a fallback:

```python
return constants.AZURE_OPENAI_MODEL or "gpt-4o"
```

For the fallback we should probably use a global var, so we can change the defaults in one place. constants.py seems like a good spot.

ngafar: Same comment, but for line 223.

ngafar (approved):

ngafar:
Looks good, LGTM.

PR #1658, opened 2025-04-18
Adding UI dropdown for model selection

hegdes-usc:
# Description

UI dropdown prototype for model selection.

Note that as of this PR, the additional models are not supported without setting API Keys as env variables as in [https://github.com/mito-ds/mito/pull/1644](url)

https://github.com/user-attachments/assets/6d453890-92bd-4c85-ac06-4e43c4e94551

# Testing

### Note: This feature requires additional work in order to make it work without API keys as discussed with @aarondr77. Hence this will serve as the ground work for the same

### DIFF
--- a/mito-ai/mito_ai/constants.py
+++ b/mito-ai/mito_ai/constants.py
@@ -3,6 +3,12 @@
 
 import os
 
+# Active model selections (updated by UI)
+ACTIVE_CLAUDE_MODEL = None
+ACTIVE_GEMINI_MODEL = None
+ACTIVE_OLLAMA_MODEL = None
+ACTIVE_OPENAI_MODEL = None
+
 # Claude Base URL
 CLAUDE_MODEL = os.environ.get("CLAUDE_MODEL")
 CLAUDE_BASE_URL = "https://api.anthropic.com/v1"
@@ -20,8 +26,60 @@ OLLAMA_BASE_URL = os.environ.get("OLLAMA_BASE_URL", "http://localhost:11434/v1")
 # OpenAI API KEY
 OPENAI_API_KEY = os.environ.get("OPENAI_API_KEY")
 
-# Azure OpenAI Config 
+# Azure OpenAI Config
 AZURE_OPENAI_API_KEY = os.environ.get("AZURE_OPENAI_API_KEY")
 AZURE_OPENAI_API_VERSION = os.environ.get("AZURE_OPENAI_API_VERSION")
 AZURE_OPENAI_ENDPOINT = os.environ.get("AZURE_OPENAI_ENDPOINT")
 AZURE_OPENAI_MODEL = os.environ.get("AZURE_OPENAI_MODEL")
+
+
+def set_active_model(model):
+    """
+    Set the active model based on the model prefix.
+    This will override the environment variable settings.
+    """
+    global ACTIVE_CLAUDE_MODEL, ACTIVE_GEMINI_MODEL, ACTIVE_OLLAMA_MODEL, ACTIVE_OPENAI_MODEL
+
+    # Reset all active models
+    ACTIVE_CLAUDE_MODEL = None
+    ACTIVE_GEMINI_MODEL = None
+    ACTIVE_OLLAMA_MODEL = None
+    ACTIVE_OPENAI_MODEL = None
+
+    # Set the appropriate active model based on the prefix
+    if model.startswith('claude'):
+        ACTIVE_CLAUDE_MODEL = model
+    elif model.startswith('gemini'):
+        ACTIVE_GEMINI_MODEL = model
+    elif model.startswith('ollama'):
+        ACTIVE_OLLAMA_MODEL = model
+    elif model.startswith('gpt') or model.startswith('o3'):
+        ACTIVE_OPENAI_MODEL = model
+
+    return model
+
+
+def get_active_model(model_type=None):
+    """
+    Get the active model for a given type, with fallback to env variables
+    """
+    if model_type == 'claude':
+        return ACTIVE_CLAUDE_MODEL or CLAUDE_MODEL
+    elif model_type == 'gemini':
+        return ACTIVE_GEMINI_MODEL or GEMINI_MODEL
+    elif model_type == 'ollama':
+        return ACTIVE_OLLAMA_MODEL or OLLAMA_MODEL
+    elif model_type == 'openai':
+        return ACTIVE_OPENAI_MODEL or None
+
+    # If no type specified, return the first active model
+    for model in [ACTIVE_CLAUDE_MODEL, ACTIVE_GEMINI_MODEL, ACTIVE_OLLAMA_MODEL, ACTIVE_OPENAI_MODEL]:
+        if model:
+            return model
+
+    # Default to environment variables if no active models
+    for model in [CLAUDE_MODEL, GEMINI_MODEL, OLLAMA_MODEL]:
+        if model:
+            return model
+
+    return None
--- a/mito-ai/mito_ai/providers.py
+++ b/mito-ai/mito_ai/providers.py
@@ -70,6 +70,8 @@ This attribute is observed by the websocket provider to push the error to the cl
         super().__init__(log=get_logger(), **kwargs)
         self.last_error = None
         self._async_client: Optional[openai.AsyncOpenAI] = None
+        # Track the active model to know when to rebuild the client
+        self._active_model = None
 
     @default("api_key")
     def _api_key_default(self) -> Optional[str]:
@@ -129,7 +131,7 @@ This attribute is observed by the websocket provider to push the error to the cl
     @property
     def capabilities(self) -> AICapabilities:
         """Get the provider capabilities."""
-        
+
         if is_azure_openai_configured():
             return AICapabilities(
                 configuration={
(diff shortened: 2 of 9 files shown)


mito-ai/mito_ai/constants.py
```diff
@@ -3,6 +3,12 @@
 
 import os
 
+# Active model selections (updated by UI)
```
aarondr77, line 6: We'd be able to delete these

mito-ai/mito_ai/constants.py
```diff
@@ -20,8 +26,60 @@
...
+# Azure OpenAI Config
 AZURE_OPENAI_API_KEY = os.environ.get("AZURE_OPENAI_API_KEY")
 AZURE_OPENAI_API_VERSION = os.environ.get("AZURE_OPENAI_API_VERSION")
 AZURE_OPENAI_ENDPOINT = os.environ.get("AZURE_OPENAI_ENDPOINT")
-AZURE_OPENAI_MODEL = os.environ.get("AZURE_OPENAI_MODEL")
\ No newline at end of file
+AZURE_OPENAI_MODEL = os.environ.get("AZURE_OPENAI_MODEL")
+
+
+def set_active_model(model):
```
aarondr77, line 36: We could delete all of this too

mito-ai/mito_ai/providers.py
```diff
@@ -187,8 +219,14 @@ def capabilities(self) -> AICapabilities:
 
     @property
     def _active_async_client(self) -> Optional[openai.AsyncOpenAI]:
-        if not self._async_client or self._async_client.is_closed():
+        # Get the current active model
+        current_model = self._resolve_model()
+
+        # If model has changed or client is closed, rebuild the client
+        if self._active_model != current_model or not self._async_client or self._async_client.is_closed():
```
aarondr77, line 226: Nice good catch

mito-ai/mito_ai/providers.py
```diff
@@ -198,24 +236,27 @@ def key_type(self) -> str:
...
             return "claude"
 
-        if constants.GEMINI_MODEL and constants.GEMINI_API_KEY:
+        active_gemini = constants.get_active_model('gemini')
+        if (active_gemini or constants.GEMINI_MODEL) and constants.GEMINI_API_KEY:
             return "gemini"
 
         return MITO_SERVER_KEY
 
     def _build_openai_client(self) -> Optional[Union[openai.AsyncOpenAI, openai.AsyncAzureOpenAI]]:
```
aarondr77, line 253: I think we'd update this function to take in the `active_model` as a parameter and then we would do the following:

```python

if get_active_model_provider(active_model) == 'claude' and  constants.CLAUDE_MODEL and constants.CLAUDE_API_KEY:
             base_url = constants.CLAUDE_BASE_URL
             llm_api_key = constants.CLAUDE_API_KEY
             self.log.debug(f"Using Claude with model: {active_claude}")


# Then the same thing for the other providers ...
```

aarondr77 (changes requested):
Nice work on this. It's all coming together and I'm excited to get this released to folks. 

I think we can pretty dramatically simplify this PR if we make one design change: There is only ever one active model and that model is always whatever the user selects in the UI. We can then:

1. Just store the active model in the provider.py class 
2. When a completion message is sent, we can check if the user has environment variables set so we can build a provider for that model. If they do, we use their key to use that model. If they don't, we use the Mito Server instead. 


TLDR: Unless I'm thinking about it incorrectly, there's no need to have an active model per model provider. We can just have one active model and either use the user's key to execute it or use the Mito server.

aarondr77:
Feel free to ignore the Mito AI Frontend Playwright failing tests. Those are flakey and fail often, but let's fix up the rest of the tests before merging

hegdes-usc:
Hey @aarondr77 I have addressed all the review comments and have pushed a new commit with other changes as well. As discussed on slack

Different models return different types of outputs, and in order to handle those I have made a few changes in the way chunks are processed along with the changes you requested.

There are still a few issues in the way these models return outputs for example some responses from the claude models do not enclose the code snippets in the right format, which can block the user from overriding the current cell with the output. 

It is sure that there will be a lot of these edge cases which would require a much more thorough approach specific to each model provider.

mito-ai/mito_ai/providers.py
```diff
@@ -353,60 +399,89 @@ async def stream_completions(
...
+                            parent_id=message_id,
+                            chunk=CompletionItem(
+                                content=content,
+                                isIncomplete=True,
+                                token=message_id
+                            ),
+                            done=False,
+                        ))
+
+                    if is_finished:
```
aarondr77, line 466: What's the purpose of emitting the full final response at the end? I think the stream handler on the frontend doesn't need this information.

mito-ai/mito_ai/constants.py
```diff
@@ -20,8 +23,53 @@
...
+    Get the active model with fallback to env variables
+    """
+    # Return the active model if set
+    if ACTIVE_MODEL:
+        return ACTIVE_MODEL
+
+    # Default to environment variables if no active model
+    for model in [CLAUDE_MODEL, GEMINI_MODEL, OLLAMA_MODEL]:
+        if model:
+            return model
```
aarondr77, line 54: There should always be an `active model` I think since the UI is always going to display a model. So I think we can remove this

aarondr77 (changes requested):
Nice work. I like these simplifications that you made. 

Yeah, so the problem with Claude is [Claude's OpenAI SDK compatibility does not support the response_format parameter](https://docs.anthropic.com/en/api/openai-sdk#simple-fields). 

I think we should do the following since adding the model UI selector + adding support for all OpenAI models and Gemini models is going to be a really exciting and useful feature for people:
1. Let's keep everything as is, but comment out the anthropic models from the UI selector. That way, users can never end up with an anthropic model for now. 
2. Let's get this PR merged in. 
3. Let me invite you to our lambda function repo so you can get the code implemented on the mito server for users to be able to select their model even when they are using our server. This is going to be the most common user flow. I think we'll need to copy over the `get_model_type` function so we can choose which API key / endpoint to use. 

Then let's touch base on some next steps. A few things that we might want to do after that:
1. Move the AI model selector to the bottom like cursor. I showed the existing implementation to a few users and they shared this feedback that at the bottom is more expected since it's where the rest of their chat configuration options are. Before we do this, we can make a tiny mockup so we can align on the design.
2. Currently, we are setting the provider for all types of AI completions with this UI. However, the model that is best for users for chat is likely not the right model for inline code completions. For chat, thinking models might make sense. For inline code completion, speed is crucial. So one thing I could imagine doing is: when the user selects the model, we use that for the chat, and then the use the max (fast, capable) model from that same LLM provider for inline code completion. Let's create a little spec on how we would want to handle this before implementing it. 
3. Add support for Anthropic with the Anthropic SDK. This might be a bigger list if we have to abstract out all of the OpenAI types that we're currently relying on...

mito-ai/mito_ai/constants.py
hegdes-usc: This function is needed during initial startup before the UI fully loads else it throws an error since this appears to load first. 

This also helps if there's a bug in the UI model selection as well. This just makes it more robust to unexpected issues.

mito-ai/mito_ai/providers.py
hegdes-usc: This is to handle a case with the gemini flash model. It responds so quickly, that it returns the whole response in one chunk. 

Sidenote: We should use one of the gemini flash models to do inline completions.

mito-ai/mito_ai/providers.py
```diff
@@ -70,6 +70,7 @@ def __init__(self, **kwargs: Dict[str, Any]) -> None:
         super().__init__(log=get_logger(), **kwargs)
         self.last_error = None
         self._async_client: Optional[openai.AsyncOpenAI] = None
+        self._active_model = None
```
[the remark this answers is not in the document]
hegdes-usc: `constants.ACTIVE_MODEL` records the user's selection in the UI
`self._active_model` tracks what model the current client was built with

Removing `self._active_model` would eliminate the optimization that prevents rebuilding the client when unnecessary.

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,91 @@
...
+    // Save to localStorage
+    localStorage.setItem('llmModelConfig', JSON.stringify(newConfig));
+    
+    // Update environment variables based on model prefix
+    if (newModel.startsWith('claude')) {
+      localStorage.setItem('CLAUDE_MODEL', newModel);
+    } else if (newModel.startsWith('gemini')) {
+      localStorage.setItem('GEMINI_MODEL', newModel);
+    } else if (newModel === 'ollama') {
+      localStorage.setItem('OLLAMA_MODEL', newModel);
```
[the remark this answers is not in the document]
hegdes-usc: Fixed in: [87c6f6f](https://github.com/mito-ds/mito/pull/1658/commits/87c6f6f45d72fc18d92db22cb949c8167ca4f906)

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -140,6 +141,24 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
     const streamingContentRef = useRef<string>('');
     const streamHandlerRef = useRef<((sender: CompletionWebsocketClient, chunk: ICompletionStreamChunk) => void) | null>(null);
 
+    const updateModelOnBackend = async (model: string): Promise<void> => {
```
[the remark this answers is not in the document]
hegdes-usc: Fixed it via a change in the [mito-ai/src/components/ModelSelector.tsx](https://github.com/mito-ds/mito/pull/1658/files/a803de5c906607df4a5a5d2ab24ea9100f2ee90d#diff-f6fec345d523f09a1955ac6b5d58f9dd4b6e9d34b937e37b06c14b56a41d5e9d)

Commit id: [87c6f6f](https://github.com/mito-ds/mito/pull/1658/commits/87c6f6f45d72fc18d92db22cb949c8167ca4f906)

aarondr77 (commented):
Okay, very nice. We're so close to finishing this. I love the transition polishes you added when the dropdown resizes. Looks great! 

I found a few bugs that we need to address before we can merge it into dev. Let me know if you want to chat through any of them.

aarondr77:
Oh let's also make sure that the `mito-ai src lint`, `mito-ai mypy`, and `mito ai backend` tests are passing.

Issue #1629, opened 2025-04-14
Collapse Code Preview in Past Messages


ngafar:
Code previews show the full code (not truncated), which makes it hard to scan messages. 

We could collapse these previews by default—keeping the code accessible but hidden unless expanded. I suspect users primarily rely on the message text (what they typed) when scrolling through old conversations, not the full contents of the code cell. 

Note that this only for sent messages. This should not affect the input box.

PR #1659, opened 2025-04-18
Collapsable Cell Previews

ngafar:
# Description

For user input messages, the active cell preview will now only show the first 5 lines. You can click on the preview to expand. Additionally, the preview has a muted overlay to signify that it is context. 

Addresses https://github.com/mito-ds/mito/issues/1629

# Testing

Create a few cells with lots of content (>5 lines), and send a few messages with the cells selected. You should see the new style. 

# Documentation

N/A


aarondr77 (approved):
Awesome. I changed the look of this slightly. Instead of adding a comment, I now just add a button. I think this is clearer to users since they don't need to read the code to understand that they need to click on it to expand. 

Screenshot 2025-04-21 at 10 52 47 AM

If you like this look, let's go with it! 

Before merging can we throw a quick jest test together so we don't break this in the future?

PR #1679, opened 2025-04-25
Database Awareness - Chat

ngafar:
# Description

The Mito AI is now aware of database connections, and can write SQL queries.

For now this is limited to chat, and the connections must be hardcoded. 

# Testing

1. In the `.mito/db` dir add the new `connections.json` and `schemas.json` files. 
2. Then start a Jupyter sever, and ask a question that required a db connection.

Be sure to ask about data that is relevant to a table, but does not exist. 

# Documentation

N/A - We should add documentation, but after we add the taskpane to add new db connections.


aarondr77:
The schema has column headers in uppercase, so if I sk it to write code like this: How does the location of a customer effect their likelihood to churn?

It ends up referencing the columns in pandas with all uppercase, even though the resulting dataframe has lowercase. If snowflake sql-alchemy always ends up with lowercase dataframe headers, (idk if this is the case), then we should update the schema probably. 

```python

import pandas as pd
from sqlalchemy import create_engine

# Database connection setup
user = "XXX"
password = "XXX"
account = "XXX"
warehouse = "COMPUTE_WH"
database = "TELCO_CHRUN"
schema = "PUBLIC"

conn_str = (
    f"snowflake://{user}:{password}@{account}/"
    f"{database}/{schema}?warehouse={warehouse}"
)
engine = create_engine(conn_str)

# Query for churn status and location data
query = """
SELECT 
    l.COUNTRY,
    l.STATE,
    l.CITY,
    s.CHURN_LABEL
FROM LOCATION_DATA l
JOIN STATUS_ANALYSIS s
ON l.CUSTOMER_ID = s.CUSTOMER_ID
"""

df = pd.read_sql(query, engine)

# Analyze churn rate by state (you can modify to COUNTRY or CITY)
churn_by_state = (
    df.groupby('STATE')['CHURN_LABEL']
    .mean()
    .sort_values(ascending=False)
    .reset_index()
    .rename(columns={'CHURN_LABEL': 'churn_rate'})
)
churn_by_state
```

mito-ai/mito_ai/prompt_builders/prompt_constants.py
```diff
@@ -54,4 +59,41 @@ def cell_update_output_str(has_cell_update_output: bool) -> str:
...
+    """
+    Reads the user's database configurations,
+    and returns the rules for the AI to follow.
+    """
+
+    # Get the db configuration from the user's mito folder
+
+    APP_DIR_PATH: Final[str] = os.path.join(MITO_FOLDER)
+    
+    with open(os.path.join(APP_DIR_PATH, 'db', 'connections.json'), 'r') as f:
```
aarondr77, line 72: For this PoC I think this okay, but instead of hardcoding the username and password in the notebook, we should be importing from the config file. Hardcoding credentials in a notebook is obviously bad practice ...

mito-ai/mito_ai/prompt_builders/prompt_constants.py
```diff
@@ -54,4 +59,41 @@ def cell_update_output_str(has_cell_update_output: bool) -> str:
...
+
+    if connections is not None:
+        DATABASE_RULES = f"""DATABASE RULES:
+If the user has requested data that you belive is stored in the database:
+- Use the provided schema.
+- Only use SQLAlchemy to query the database.
+- Do not use a with statement when creating the SQLAlchemy engine. Instead, initialize it once so it can be reused for multiple queries.
+- Always return the results of the query in a pandas DataFrame, unless instructed otherwise.
+- Column names in query results may be returned in lowercase. Always refer to columns using their lowercase names in the resulting DataFrame (e.g., df['date'] instead of df['DATE']).
+- If you think the requested data is stored in the database, but you are unsure, then ask the user for clarification.
```
aarondr77, line 88: Nice! I've been having it respond with a question pretty often

mito-ai/mito_ai/prompt_builders/chat_system_message.py
```diff
@@ -37,6 +44,9 @@ def create_chat_system_message_prompt() -> str:
 
 Notice in the example above that the citation uses line number 2 because citation line numbers are 0-indexed.
 
+===
+{get_database_rules()}
```
aarondr77, line 48: Putting this in the system prompt so we don't send it over and over again makes sense. 

Eventually, it might even make sense to move this to a tool use that the agent and chat are allowed to use called `check_snowflake_schema` or something like that so it can decide when to use it...

aarondr77:
I made a bit bigger of a schema (still I think a lot smaller than real schemas our users will have) and the performance is definetly getting worse. I'll send you the schema so you can check it out. 

I asked Claude to generate 12 questions that I could ask about the data and here is how it did: 

### Prompts it responded with something like 'no data available'
1. Which emplpoyee has the highest commission percentage
2. How many people are we paying more than 100k per year ?
3. Compare the sales performance of our top 5 products across different regions for the past 12 months.
4. Which marketing campaigns delivered the highest ROI last year, and what channels were most effective for each?

### Prompts it asked a targetted question about: 
1. Which departments have exceeded their annual budget allocations, and by how much? => "Is your budget/spending data stored in the database? If so, should I query the table FINANCE_DB.PUBLIC.tbl_budget_2024? Or is this data in one of your files (e.g., Prospects_16042025_500linesdataset.xlsx/csv), and if so, which one?"
2. How has our S&P 500 stock portfolio performed compared to the sector averages over the past 3 years? => "What constitutes "your portfolio"? (e.g., do you have a list of stock symbols and weights, or is it an equally weighted portfolio of all S&P 500 stocks?) Should sector averages be equally weighted or weighted by market cap or sector weights?"

### Prompts it identified the correct tables for 
1. Find all of the emplpoyees who have worked at the company for more than 1 year 
2. Find the customer in each state who has made the single biggest purchase from our enterprise data
3. What's the average monthly revenue per customer in our telecom business, broken down by service type?
4. What's our customer churn rate for the telecom business, and how does it correlate with service plan types and monthly charges?
5. Show me a breakdown of IoT device performance by location, highlighting any devices with abnormal temperature readings.
6. Which employees have changed departments in the past year, and what was the impact on their salary?

mito-ai/setup.py
```diff
@@ -94,6 +94,7 @@ def get_data_files_from_data_files_spec(
         "tornado>=6.2.0",
         "traitlets",
         "pydantic",
+        "SQLAlchemy",
```
aarondr77, line 97: Is this enough or do we need https://pypi.org/project/snowflake-sqlalchemy/ also?

mito-ai/mito_ai/prompt_builders/prompt_constants.py
```diff
@@ -54,4 +59,61 @@ def cell_update_output_str(has_cell_update_output: bool) -> str:
...
+If the user has requested data that you belive is stored in the database:
+- Use the provided schema.
+- Only use SQLAlchemy to query the database.
+- Do not use a with statement when creating the SQLAlchemy engine. Instead, initialize it once so it can be reused for multiple queries.
+- Always return the results of the query in a pandas DataFrame, unless instructed otherwise.
+- Every schema includes a connection field that specifies which database connection to use.
+- Connection details are stored in a JSON file located at: `{connections_path}`
+- This connections.json file follows the structure below:
+
+{{
```
aarondr77, line 93: I think we need to give the AI more information about what the connection names are. It guesses that my connection was called 'SP_500'. Maybe we can just pass the json keys or something? 

```python
# Load the database connection information
with open('/Users/aarondiamond-reivich/.mito/db/connections.json', 'r') as f:
    connections = json.load(f)

conn_info = connections['SP_500']
engine = create_engine(
    f"snowflake://{conn_info['username']}:{conn_info['password']}@{conn_info['account']}/SP_500/PUBLIC?warehouse={conn_info['warehouse']}"
)
```

Even though my schema is: 

```
{
    "snowflake": {
        "username": "XXX",
        "password": "XXX",
        "account": "XXX",
        "warehouse": "XXX"
    }
}
```

mito-ai/mito_ai/prompt_builders/prompt_constants.py
```diff
@@ -54,4 +59,61 @@ def cell_update_output_str(has_cell_update_output: bool) -> str:
...
+}}
+
+- Do not hard-code connection credentials into your code. Instead, load the connections.json file and access connection fields dynamically like so:
+
+```
+connections[connection_name]["username"]
+```
+
+- The user may colloquially as for a "list of x", always assume they want a pandas DataFrame. 
+- When a SQL query is executed and the results are loaded into a pandas DataFrame, all column names are automatically converted to lowercase. Always reference columns using their lowercase names, regardless of how they appear in the database. For example, if the database column is named "DATE", you should reference it as "date" in your code.
```
aarondr77, line 109: This rule is pretty consistently not working for me: For example it keeps writing SYMBOL instead of symbol
```python
sp500_combined = pd.merge(sp500_companies, sp500_stocks, on='SYMBOL', how='inner')
```

mito-ai/mito_ai/prompt_builders/prompt_constants.py
```diff
@@ -54,4 +59,61 @@ def cell_update_output_str(has_cell_update_output: bool) -> str:
...
+        connections = json.load(f)
+
+    with open(schemas_path, 'r') as f:
+        schemas = json.load(f)
+
+    # If there is a db configuration, add return the rules
+
+    if connections is not None:
+        DATABASE_RULES = f"""DATABASE RULES:
+If the user has requested data that you belive is stored in the database:
```
aarondr77, line 84: "belive" -> "believe"

mito-ai/mito_ai/prompt_builders/prompt_constants.py
```diff
@@ -54,4 +59,61 @@ def cell_update_output_str(has_cell_update_output: bool) -> str:
...
+    }}
+}}
+
+- Do not hard-code connection credentials into your code. Instead, load the connections.json file and access connection fields dynamically like so:
+
+```
+connections[connection_name]["username"]
+```
+
+- The user may colloquially as for a "list of x", always assume they want a pandas DataFrame.
```
aarondr77, line 108: "as" -> "ask"

aarondr77: We should create some evals that are phrased this way so we can test how well we actually handle this!

mito-ai/mito_ai/prompt_builders/prompt_constants.py
aarondr77: We should create some evals so we can iterate this part of the prompt!

mito-ai/mito_ai/prompt_builders/prompt_constants.py
```diff
@@ -54,4 +59,61 @@ def cell_update_output_str(has_cell_update_output: bool) -> str:
...
+- When a SQL query is executed and the results are loaded into a pandas DataFrame, all column names are automatically converted to lowercase. Always reference columns using their lowercase names, regardless of how they appear in the database. For example, if the database column is named "DATE", you should reference it as "date" in your code.
+- If you think the requested data is stored in the database, but you are unsure, then ask the user for clarification.
+
+Here is the schema:
+{schemas}
+        """
+    else:
+        DATABASE_RULES = ""
 
+    print("DATABASE_RULES\n", DATABASE_RULES)
```
aarondr77, line 118: remove

aarondr77 (approved):
I'm noticing that the tool is struggling to know when to query the database more often then I anticipated based on the evals. For example:

Unless I am very specific, it often tries to import data from a file (even if that file doesn't exist). For example: Whenever I say "Import the SP500 dataset", it tries to "sp500_df = pd.read_csv('SP500.csv')". But if I say "Import the SP500 **stock** dataset", then it will choose to import. 

Can we turn this into an eval so we can iterate how we handle this? Maybe we should give it more clear instructions about how to check if it should query the database or not.

mito-ai/mito_ai/prompt_builders/prompt_constants.py
ngafar: This might be a bit hard to eval, since we are extracting the SQL from the response, and throwing away everything else.

mito-ai/mito_ai/prompt_builders/prompt_constants.py
ngafar: Mentioned this below, but this is probably not something we can do in the evals (without a major refactor). The evals extract and work exclusively with SQL, throwing away the rest of the response.

mito-ai/mito_ai/prompt_builders/prompt_constants.py
ngafar: Re-adding the schema to the system prompt, but sanitizing the fields to redact all of the users info.

mito-ai/setup.py
ngafar: Good catch

mito-ai/mito_ai/prompt_builders/prompt_constants.py
ngafar: Updated the rule to simply say:

> When working with dataframes created from an SQL query, ALWAYS use lowercase column names. 

From my (limited) testing, this seems to work. The AI now always seems to use `col.lower()`.

PR #1681, opened 2025-05-01
[WIP] SQL Eval Funnels

ngafar:
# Description

# Testing

# Documentation

### DIFF
--- a/evals/ai_api_calls/get_open_ai_completion.py
+++ b/evals/ai_api_calls/get_open_ai_completion.py
@@ -2,15 +2,22 @@
 # Distributed under the terms of the GNU Affero General Public License v3.0 License.
 
 import os
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 from evals.test_cases.agent_find_and_update_tests.simple import CellUpdate
 from openai import OpenAI
+from pydantic import BaseModel
 
-def get_open_ai_completion_function_params(prompt: str, model: str) -> Dict[str, Any]:
+class SQLDetails(BaseModel):
+    query: str
+    tables: List[str]
+    columns: List[str]
+
+def get_open_ai_completion_function_params(prompt: str, model: str, system_prompt: Optional[str] = None) -> Dict[str, Any]:
+    default_system_prompt = "You are an expert Python programmer."
     completion_function_params = {
         "model": model,
         "messages": [
-            {"role": "system", "content": "You are an expert Python programmer."},
+            {"role": "system", "content": system_prompt or default_system_prompt},
             {"role": "user", "content": prompt}
         ],
     }
@@ -21,8 +28,8 @@ def get_open_ai_completion_function_params(prompt: str, model: str) -> Dict[str,
         
     return completion_function_params
 
-def get_open_ai_completion_code_block(prompt: str, model: str) -> str:
-    completion_function_params = get_open_ai_completion_function_params(prompt, model)
+def get_open_ai_completion_code_block(prompt: str, model: str, system_prompt: Optional[str] = None) -> str:
+    completion_function_params = get_open_ai_completion_function_params(prompt, model, system_prompt)
 
     client = OpenAI(api_key=os.getenv("OPENAI_API_KEY"))
     response = client.chat.completions.create(**completion_function_params)
@@ -41,7 +48,8 @@ def get_open_ai_parsed_response(prompt: str, model: str, response_format: type[C
     response = client.beta.chat.completions.parse(**completion_function_params)
 
     response_content = response.choices[0].message.parsed
-
+    if response_content is None:
+        raise ValueError("No parsed response received")
     return response_content
        
 
@@ -57,3 +65,22 @@ def get_code_block_from_message(message: str) -> str:
         return message
     
     return message.split('```python\n')[1].split('\n```')[0]
+
+def get_sql_from_message(message: str, model: str) -> SQLDetails:
+    """
+    Extract the first SQL query + other metadata from a message.
+    """
+    client = OpenAI(api_key=os.getenv("OPENAI_API_KEY"))
+    response = client.beta.chat.completions.parse(
+        model=model,
+        messages=[
+            {"role": "system", "content": "You are a helpful assistant that extracts SQL queries from messages."},
+            {"role": "user", "content": message}
+        ],
+        response_format=SQLDetails,
+    )
+    parsed = response.choices[0].message.parsed
+    
+    if parsed is None:
+        raise ValueError("No SQL details parsed from response")
+    return parsed
--- a/evals/test_cases/sql_tests/simple_tests.py
+++ b/evals/test_cases/sql_tests/simple_tests.py
@@ -1,0 +1,3 @@
+SIMPLE_TESTS = [
+
+]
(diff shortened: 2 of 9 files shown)


evals/ai_api_calls/get_open_ai_completion.py
```diff
@@ -57,3 +60,22 @@ def get_code_block_from_message(message: str) -> str:
...
+    Extract the first SQL query + other metadata from a message.
+    """
+    client = OpenAI(api_key=os.getenv("OPENAI_API_KEY"))
+    response = client.beta.chat.completions.parse(
+        model=model,
+        messages=[
+            {"role": "system", "content": "You are a helpful assistant that extracts SQL queries from messages."},
+            {"role": "user", "content": message}
+        ],
+        response_format=SQLDetails,
```
aarondr77, line 75: Maybe just for a little bit more clarity we can call this `ParsedSQLDetails` or something like that.

evals/funnels/sql/steps.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+@dataclass
+class FunnelStepResult:
+    name: str
+    passed: bool
+    notes: Optional[str] = None
+
+
+def sql_generated_test(
```
aarondr77, line 15: Maybe we can call this `is_sql_generated_test`, I wasn't sure exactly what this was going to test until I looked at the docstring

evals/funnels/sql/steps.py
```diff
@@ -0,0 +1,222 @@
...
+    Args:
+        expected_output: The expected output string or None if no SQL query was expected
+        sql_query_recieved: The actual SQL query string generated or None if no query was generated
+    """
+    name = "sql_generated_test"
+
+    if (expected_output is None) and (sql_query_recieved is None):
+        # No SQL query was expected and none was generated
+        return FunnelStepResult(name=name, passed=True)
+    elif (expected_output is not None) and (sql_query_recieved is not None):
```
aarondr77, line 34: Maybe just so we are not so fragile to AI to return None when generating the SQLDetails, we could do:

```python
expected_sql_query_exists = expected_output is not None and expected_output != ''
actual_sql_query_exists = sql_query_recieved is not None and sql_query_recieved != ''

...
```

evals/funnels/sql/steps.py
```diff
@@ -0,0 +1,222 @@
...
+        return FunnelStepResult(
+            name=name,
+            passed=False,
+            notes="SQL query was not generated when it was expected, or generated when it was not expected",
+        )
+
+
+def correct_tables_test(
+    expected_tables: List[str],
+    tables_in_query: List[str],
```
aarondr77, line 47: Nitpick, but I like the convention `expected_tables` and `actual_tables`. The symmetry makes it very clear what these variables are

evals/funnels/sql/utils.py
```diff
@@ -0,0 +1,52 @@
+import os
+import pandas as pd
+from typing import Tuple, Optional
+from sqlalchemy import create_engine
+from dotenv import load_dotenv
+
+load_dotenv()
+
+
+def parse_table_path(table_path: str) -> Tuple[str, str, str]:
```
aarondr77, line 10: I could imagine naming this `parse_db_schema_table_names`

evals/eval_types.py
```diff
@@ -171,3 +183,23 @@ def get_prompt(self, error_message: str, notebook_state: NotebookState) -> str:
...
     
+
+class TableDetails(BaseModel):
+    name: str = Field(
+        description="The name of the table, should be in the format: database_name.schema_name.table_name"
+    )
+    columns: List[str] = Field(
+        description="List of column names used from this table"
+    )
+    alias: Optional[str] = Field(
```
aarondr77, line 194: What is a table alias? Is this something we want?

evals/funnels/sql/steps.py
```diff
@@ -0,0 +1,222 @@
...
+        for column in columns:
+            if column.startswith("*"):
+                # Skip if the column is a wildcard
+                continue
+            elif column not in columns_in_schema:
+                return FunnelStepResult(
+                    name=name,
+                    passed=False,
+                    notes=f"Column '{column}' not found in table '{table_name}'",
+                )
```
aarondr77, line 154: We could write it like this to get all of the columns 

```python
missing_columns = [col for col in columns if not col.startswith("*") and col not in columns_in_schema]
if missing_columns:
    return FunnelStepResult(name=name, passed=False, notes=f"Columns not found in table '{table_name}': {', '.join(missing_columns)}")
```

evals/funnels/sql/steps.py
```diff
@@ -0,0 +1,222 @@
...
+
+    return FunnelStepResult(name=name, passed=True)
+
+
+def syntax_check_test(
```
aarondr77, line 159: Let's call this `execute_without_errors_test` since it could fail for other reasons than the syntax not being correct

evals/funnels/sql/steps.py
```diff
@@ -0,0 +1,222 @@
...
+    # Return an empty dataframe which should cause future tests to fail.
+    if error is not None:
+        return pd.DataFrame(), FunnelStepResult(
+            name=name, passed=False, notes=str(error)
+        )
+    # All good, return the dataframe
+    return result, FunnelStepResult(name=name, passed=True)
+
+
+def correct_data_shape_test(
```
aarondr77, line 189: Sweet. I'm down for `correct_data_shape_test` and `correct_data_test` as is. My one thought is that maybe these are slightly unrealistic expectations since they require the actual query to return the exact same result as our expected query. Which just puts a lot of pressure on the AI to select the minimum required data, instead of overshooting and getting a table that it can then continue to filter or transform. Let's see how it goes with our actual evals!

evals/test_cases/sql_tests/simple_tests.py
```diff
@@ -0,0 +1,24 @@
...
+        expected_columns=[],
+    ),
+    SQLTestCase(
+        name="simple_test_1",
+        user_input="get the company with the largest market cap",
+        schema=SMALL_SCHEMA,
+        notebook_state=EMPTY_NOTEBOOK,
+        expected_output="SELECT * FROM SP_500.PUBLIC.SP500_COMPANIES ORDER BY MARKETCAP DESC LIMIT 1",
+        expected_tables=["SP_500.PUBLIC.SP500_COMPANIES"],
+        expected_columns=["*"],
```
aarondr77, line 22: Instead of us having to add these manually for each test, we could just use our same utils that we use for the `actual_query` to generate these. That's going to be easier to maintain and less error prone probably

aarondr77 (commented):
This is super awesome. Designing this as a funnel already seems like 5x more valuable than what we have currently. Excited to build more evals and see how we can use this new infra to optimize performance.

evals/eval_types.py
ngafar: Imagine the query `SELECT CustomerID AS ID` the alias would be `ID`. 

My hope was that the alias field would ensure that we parsed the cleanest possible version of the table name, and nothing else. But I think it's safe to remove this.

evals/funnels/sql/steps.py
ngafar: Will definitely keep an eye on this. Once we have some data, I'm sure we can adjust this test to make it more lenient.

PR #1684, opened 2025-05-08
SQL Evals - Test Cases

ngafar:
# Description

- Added a medium and large schema.
- Added new test cases for basic filtering, joins and grouping.

# Testing

Install new dependencies (tqdm):

```bash
cd evals
pip install -r requirements.txt
```

then move up to the root dir, and run the evals:

```bash
cd ..
python -m evals.main --test_type=sql
```

# Documentation

N/A


aarondr77 (approved):
Very nice work! 

One thing that might be useful is to add the actual user input into the dashboard also. It's hard to exactly interpret without going back to the code to see what input. 

Screenshot 2025-05-08 at 10 39 36 AM

PR #1685, opened 2025-05-08
SQL Evals - Test cases: Wave 2

ngafar:
# Description

- Added pattern matching and time-based tests.
- Update dashboard to show schema performance, test type performance.
- Can now execute sql from the dashboard. 
- Added two new steps to the funnel: `minimum_expected_cols_test` and `row_count_test`. 

# Testing

```bash
python -m evals.main --test_type=sql
```

# Documentation

N/A


evals/test_cases/sql_tests/time_tests.py
```diff
@@ -0,0 +1,118 @@
...
+        expected_output="""
+            SELECT COUNT(*) AS total_orders
+            FROM SALES_DB.PUBLIC.ORDERS
+            WHERE ORDER_DATE >= DATEADD(day, -7, (SELECT MAX(ORDER_DATE) FROM SALES_DB.PUBLIC.ORDERS))
+        """.strip(),
+        test_type=TEST_TYPE,
+    ),
+    SQLTestCase(
+        name="daily_order_totals_last_30_days",
+        user_input="Show the total order amount per day for the past 30 recorded days.",
```
aarondr77, line 106: These are sweet. This seems like a realistic question that a user might ask

evals/test_cases/sql_tests/time_tests.py
```diff
@@ -0,0 +1,118 @@
...
+        schema=MEDIUM_SCHEMA,
+        notebook_state=EMPTY_NOTEBOOK,
+        expected_output="""
+            SELECT ORDER_DATE, SUM(TOTAL_AMOUNT) AS daily_total
```
aarondr77, line 69: Interesting, I won't be surprised if the AI doesn't also name it `daily_total`. It would be interesting (not needed now) to develop a funnel step that checks if it gets the right data, regardless of what it names the column

evals/test_cases/sql_tests/grouping_tests.py
```diff
@@ -5,71 +5,117 @@
...
-        expected_output="SELECT AVG(NUMBER_OF_REFERRALS) FROM TELCO_CHRUN.PUBLIC.SERVICE_OPTIONS WHERE INTERNET_SERVICE = TRUE AND PHONE_SERVICE = TRUE",
+        expected_output="""
+            SELECT AVG(NUMBER_OF_REFERRALS) 
+            FROM TELCO_CHRUN.PUBLIC.SERVICE_OPTIONS 
+            WHERE INTERNET_SERVICE = TRUE AND PHONE_SERVICE = TRUE
+        """.strip(),
+        test_type=TEST_TYPE,
     ),
     SQLTestCase(
         name="avg_referrals_for_all_services",
```
aarondr77, line 73: I think this ie misnamed. It's not related to referrals. I think just copy and pasted from last one

aarondr77 (approved):
Great work on this! I'm really excited to be able to see what users actually query, then update this and make sure that we're actually supporting user user cases. I think all your work here (it's architected super well + the funnels are super informative + the streamlit dashboard is so digestible) is going to make our evals a useful tool. I already have a better understanding of how sql performs than python stuff.

evals/test_cases/sql_tests/time_tests.py
ngafar: Whops, this should not use an alias. I've added specific instructions for the eval prompt to not use aliases, so we can make better comparisons when checking the data.

PR #1690, opened 2025-05-13
Notify users about new version

ktaletsk:
# Summary

This PR implements a version check mechanism that notifies users when a new version of Mito AI is available. The system checks for updates once per day and displays a notification only when an update is available.

## Implementation details
- Created a new version check plugin that initializes when JupyterLab starts
- Added logic to check for updates once every 24 hours
- Implemented localStorage caching to avoid excessive version checks
- Added periodic background checks while the application is running
- Ensured graceful error handling to prevent disruption to user experience

## Benefits
- Users will be aware of new features and bug fixes
- Encourages users to stay on the latest version
- Non-intrusive notification system that only appears when updates are relevant
- Minimal performance impact with reasonable check interval

# Testing
- Verified version check works correctly on initial load
- Confirmed localStorage properly tracks last check time
- Tested error handling paths
- Ensured notification only appears when newer versions are available


mito-ai/mito_ai/version_check.py
```diff
@@ -0,0 +1,59 @@
...
+import time
+from functools import lru_cache
+import pkg_resources
+import requests
+import tornado.web
+from tornado.web import RequestHandler
+
+logger = logging.getLogger(__name__)
+
+class VersionCheckHandler(RequestHandler):
```
ngafar, line 18: We should create a pytest for this. It seems unlikely that the pypi endpoint will change, but it would be good monitor. Otherwise, I can imagine this being broken for a while before we realize it.

mito-ai/src/utils/version_check.ts
```diff
@@ -0,0 +1,102 @@
...
+    autoClose: false,
+    actions: [
+      {
+        label: 'Learn how to update',
+        callback: () => {
+          window.open('https://docs.trymito.io/getting-started/installing-mito', '_blank');
+        }
+      },
+      {
+        label: 'Release notes',
```
ngafar, line 24: Minor nitpick, but may we should change this to read, "What's new." Only because "release notes" feels like something you'd see after you finish the update.

mito-ai/src/utils/version_check.ts
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright (c) Saga Inc.
+ * Distributed under the terms of the GNU Affero General Public License v3.0 License.
+ */
+
+import { ServerConnection } from '@jupyterlab/services';
+import { Notification } from '@jupyterlab/apputils';
+import * as semver from 'semver';
```
ngafar, line 8: Just learning about semver, this looks way better than writing our own logic. I can just think of all the edgecases we don't have to worry about.

mito-ai/src/utils/version_check.ts
```diff
@@ -0,0 +1,102 @@
...
+ * Display a notification when Mito AI is outdated
+ */
+export function showVersionOutdatedNotification(currentVersion: string, latestVersion: string): void {
+  Notification.emit(`Your Mito AI version (${currentVersion}) is outdated. Latest version is ${latestVersion}.`, 'warning', {
+    autoClose: false,
+    actions: [
+      {
+        label: 'Learn how to update',
+        callback: () => {
+          window.open('https://docs.trymito.io/getting-started/installing-mito', '_blank');
```
ngafar, line 20: We should turn this url into a variable, and stick it at the top of the file. I could see these variables as constants:

```ts
const VERSION_CHECK_TIMEOUT = 5000;
const VERSION_CHECK_ENDPOINT = 'mito-ai/version-check';
const DOCS_UPDATE_URL = 'https://docs.trymito.io/getting-started/installing-mito';
const RELEASE_NOTES_URL = 'https://docs.trymito.io/misc/release-notes';
```

ngafar:
Let's also resolve the mypy error before merging.

mito-ai/src/utils/version_check.ts
```diff
@@ -0,0 +1,104 @@
...
+const VERSION_CHECK_TIMEOUT = 5000;
+const VERSION_CHECK_ENDPOINT = 'mito-ai/version-check';
+const DOCS_UPDATE_URL = 'https://docs.trymito.io/getting-started/installing-mito';
+const RELEASE_NOTES_URL = 'https://docs.trymito.io/misc/release-notes';
+
+/**
+ * Display a notification when Mito AI is outdated
+ */
+export function showVersionOutdatedNotification(currentVersion: string, latestVersion: string): void {
+  Notification.emit(`Your Mito AI version (${currentVersion}) is outdated. Latest version is ${latestVersion}.`, 'warning', {
```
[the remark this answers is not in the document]
ktaletsk: This is incorrect. Check the docs: https://jupyterlab.readthedocs.io/en/latest/api/functions/apputils.Notification.emit.html

There is no `.show` method mentioned in the docs at all

ngafar (approved):

PR #1691, opened 2025-05-14
Fix dev setup

aarondr77:
# Description
Two days ago there was a new release of jupyter-server 2.16.0. This PR https://github.com/jupyter-server/jupyter_server/pull/1509 introduced stricter validation for server extensions. Specifically: 

Before this release: The [toggle_server_extension](https://github.com/jupyter-server/jupyter_server/pull/1509/files) function would create an ExtensionPackage object with the module name and then attempt to validate it directly using extpkg.validate(). If the validation succeeded, great! If not, it would still attempt to continue.

After the release: The `toggle_server_extension` function checks if the import_name exists in the jpserver_extensions configuration and raises an error if it does not exist. 

Two issues appeared (potentially because of bad design decisions I made earlier on) as a result of this:

### Namespace Mismatch

Our extension was registered with an inconsistent naming convention:
    - Python module name: mito_ai (with underscore)
    - Entry points namespace: mito-ai (with hyphen)

I think as a result, the new check to see if the import_name exists did not pass. I fixed this by moving everything over to `mito_ai`. 

### Validation of Package
The patch in jupyter-server 2.16.0 added a new validation step that checks if the extension is properly registered in the configuration before attempting to validate it. In editable mode (`pip install -e .`), we don't pass this validation because the package is not actually installed in the site-packages, instead it is linked to the development directory. 

As a result, when the validation does not pass and the error is thrown, it prevents the extension from being enabled.

This was the result of us using a setuptools for our build process. I've updated our build process to hatchling which will copy the serverApp config into the correct location. 

This is the key line in the new pyproject.toml file
```
"jupyter-config/jupyter_server_config.d" = "etc/jupyter/jupyter_server_config.d"
```


mito-ai/test_mito_ai.py
```diff
@@ -0,0 +1,16 @@
+import sys
```
aarondr77, line 1: delete

mito-ai/mito_ai/__init__.py
```diff
@@ -13,13 +13,15 @@
     # in editable mode with pip. It is highly recommended to install
     # the package from a stable release or in editable mode: https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs
     import warnings
+    import os
+    import sys
 
-    warnings.warn("Importing 'mito_ai' outside a proper installation.")
+    if not __file__.startswith(os.path.join(sys.prefix, 'lib')):
```
aarondr77, line 19: Do I need this?

PR #1692, opened 2025-05-15
Deploy app UI

aarondr77:
# Description

Adds very basic UI for informing users when the app has finished deploying! 

# Testing

Try it out! 

# Documentation

Nope


ngafar (approved):
This looks good. I especially love how granular the updates are. imo, nothing worse that getting stuck on a generic "loading..." screen. 

Something to consider for the future, when I think of these types of notifications (the mini banner at the bottom), I tend to associate them with short running, or quick prompts. For example, a new update available notification. However, our use case seems  to take a bit of time. Perhaps a sidebar might make more sense?

PR #1694, opened 2025-05-16
fix buttons

aarondr77:
# Description

The build system refactor that I did broke all of the toolbar buttons we added through the `toolbar-buttons.json` file: Deploy App, Accept, Reject, explain code

# Testing

Make sure those buttons are there again! 

# Documentation

No


ngafar (approved):
Good catch!

PR #1695, opened 2025-05-16
Gemini Support for Mito

hegdes-usc:
# Description

This is intended to enable Gemini Models for the Mito AI chat and Agents.

# Testing

I have added unit tests and will add integration tests in the same PR

# Note

In order for this to work (since this is not the UI PR), one would need to set the `GEMINI_API_KEY` and `GEMINI_MODEL` environment variables.


mito-ai/mito_ai/gemini_client.py
```diff
@@ -0,0 +1,70 @@
...
+        self.client = genai.Client(api_key=api_key)
+        self.model = model
+
+    async def generate_content(self, contents: str) -> str:
+        try:
+            response = self.client.models.generate_content(
+                model=self.model,
+                contents=contents,
+                config={
+                "response_mime_type": "application/json",
```
aarondr77, line 20: The generate_content is used for inline code completion too. In that case we don't want to use AgentReponse format. Currently, we are getting this. 
 
Screenshot 2025-05-19 at 10 54 36 AM

You can see in the open ai that we are passing the response info from the specific prompt type completion_handler when we need it. We can use that again here. If the `ResponseFormatInfo` is provided, then we use that. Otherwise we don't set a response format.

mito-ai/mito_ai/gemini_client.py
```diff
@@ -0,0 +1,70 @@
...
+from google import genai
+from mito_ai.completions.models import AgentResponse
+
+
+class GeminiClient:
+    def __init__(self, api_key: str, model: str):
+        self.client = genai.Client(api_key=api_key)
+        self.model = model
+
+    async def generate_content(self, contents: str) -> str:
```
aarondr77, line 14: Maybe we can call this function `request_completion` to keep it in sync with the language in the provider.py file

mito-ai/mito_ai/gemini_client.py
```diff
@@ -0,0 +1,70 @@
...
+                        return " ".join(str(part) for part in content.parts)
+                    return str(content)
+                return str(candidate)
+
+            return str(response)
+
+        except Exception as e:
+            return f"Error generating content: {str(e)}"
+
+    async def stream_content(self, contents: str) -> AsyncGenerator[str, None]:
```
aarondr77, line 58: Maybe we can call this `stream_completion` also to keep it in sync with the provider.py file

mito-ai/mito_ai/tests/providers_test.py
```diff
@@ -183,6 +192,169 @@ def test_gemini_model_resolution(monkeypatch: pytest.MonkeyPatch) -> None:
...
+        assert llm._gemini_client is not None
+        # Verify the client was initialized with correct parameters
+        mock_gemini_class.assert_called_once_with(
+            api_key="gemini-key",
+            model="gemini-2-pro"
+        )
+
+
+@pytest.mark.asyncio
+async def test_gemini_completion_request(monkeypatch: pytest.MonkeyPatch) -> None:
```
aarondr77, line 246: It seems like we are mocking so much in this test, what are we actually testing?

mito-ai/mito_ai/gemini_client.py
```diff
@@ -0,0 +1,70 @@
...
+                }
+            )
+            
+            if not response:
+                return "No response received from Gemini API"
+                
+            if hasattr(response, 'text') and response.text:
+                return response.text
+
+            # Handle streaming response
```
aarondr77, line 31: When we talked through this code there was some good additional context you shared about why we needed this code. Can you add extra context like that as comments here. I already forget what you told me, and it would be required info if I ever needed to refactor this code.

aarondr77 (changes requested):
Nice work. 

### Small Refactor
Can you just extract the OpenAI provider code into a `OpenAI Client` as well so we can have symmetric handling for each provider. 

And then we can try to keep the provider.py file just handling all of the repeated stuff across every provider, like: logging, error handling, etc. 

### Get Cell Output Loops
I am testing with Gemini Model `gemini-2.0-flash` and it seems to be getting caught often when using the `get_cell_output` tool. 
Screenshot 2025-05-19 at 10 57 09 AM
Screenshot 2025-05-19 at 10 58 37 AM

aarondr77:
In a separate PR, we should fix this: https://github.com/mito-ds/mito/issues/1698

aarondr77 (commented):
Check out the PR I branched off of this. If that looks good, merge it in here and then merge this into dev! Nice work.

PR #1696, opened 2025-05-18
Eval framework for Agent

PoojasreeDwarakanath:
# Description

This pull request contains a set of deterministic evals for testing the functionality of the agent. The set of evals and what they need to test are specified in the test_cases.json file. Based on the expected responses, expected output notebook and comparison criteria, the evals are tested and we get a report of how many tests have passed.

# Testing

### How to run the evals:
1. Run the file handler.py. This will collect the evals from test_cases.json and evaluate them.

# Documentation

## How to add test cases:
1. Upload your existing notebook with the specific test case name to the test_case_inputs/input_notebooks folder. <br />
2. Upload your expected output notebook with the specific test case name to the test_case_inputs/expected_output_notebooks folder. <br />
3. Add an eval to test_cases.json. Provide the following details in the json file - <br />
    a. test_case_name - Name of the eval<br />
    b. user_task - The task you want the agent to execute<br />
    c. description - description of the eval<br />
    d. conversation_history_present - true/ false depending on whether a conversation history is present for the task<br />
    e. evals_to_test - List of all the tests you want to run. This is of the following format - <br />
        ```
        {
            "eval_name": "compare_next_response",
            "params": {
            "type": "cell_update", "cell_type": "code"
            }
        }
        ```


evals/agent_evals/default_system_prompt.py
```diff
@@ -0,0 +1,350 @@
+from mito_ai.prompt_builders.prompt_constants import (
```
aarondr77, line 1: When I try to run these evals, its not working because this import is broken.

```zsh
(venv) aarondiamond-reivich@aaron-2 agent_evals % python3 handler.py              
Traceback (most recent call last):
  File "/Users/aarondiamond-reivich/Mito/mito/evals/agent_evals/handler.py", line 6, in <module>
    from default_system_prompt import create_agent_system_message_prompt
  File "/Users/aarondiamond-reivich/Mito/mito/evals/agent_evals/default_system_prompt.py", line 1, in <module>
    from mito_ai.prompt_builders.prompt_constants import (
ModuleNotFoundError: No module named 'mito_ai'
```

evals/agent_evals/testing_constants.py
```diff
@@ -0,0 +1,4 @@
+FILES_SECTION_HEADING = "Files in the current directory:"
```
aarondr77, line 1: I think thisis duplicate with `evals.prompts.constants`. Let's use those so we minimize how many places we need to keep in sync.

aarondr77 (changes requested):

evals/agent_evals/test_cases.json
```diff
@@ -0,0 +1,115 @@
+[
+  {
+    "test_case_name": "test_case_1",
```
aarondr77, line 3: This is a pretty cool json format. It seems like this will be really easy to define new tests and also get a quick understanding of what the tests are supposed to do

evals/agent_evals/test_case_inputs/input_notebooks/test_case_2.ipynb
aarondr77, line 1: Nice! I like having the actual notebook as the input. This is much cleaner than what we did before where we made a fake notebook as a json object and then also build the global variables. 

It's also nice that there is one notebook per test. Then if we want to make changes for one test case, we can easily do that without being worried that we are breaking other test cases.

evals/agent_evals/run_test_case.py
```diff
@@ -0,0 +1,41 @@
...
+
+def get_code_block_from_message(message: str) -> str:
+    # If ```python is not part of the message, then we assume that the
+    # entire message is the code block
+    if "```python" not in message:
+        return message
+
+    return message.split('```python\n')[1].split('\n```')[0]
+
+def get_openai_code_and_conversation_history(user_task: str, model: str, system_prompt: Optional[str] = None, conversation_history: Optional[List] = []):
```
aarondr77, line 33: It looks like we are never returning the `conversation_history` from this function, so maybe this function should just be called `get_openai_code`

evals/agent_evals/run_test_case.py
```diff
@@ -0,0 +1,41 @@
+import os
+from typing import Any, Dict, List, Optional, Tuple
+from openai import OpenAI
+
+
+def get_open_ai_completion_function_params_for_agent(user_task: str, model: str, system_prompt: str, conversation_history: Optional[List] = []) -> Dict[str, Any]:
+    if not conversation_history:
```
aarondr77, line 7: I think the order should be the following because the converastion_history should come before the next user_task right?

```python
full_conversation_history = []
full_conversation_history.append({"role": "system", "content": system_prompt})
full_conversation_history.extend(conversation_history)
full_conversation_history.append({"role": "user", "content": user_task})
```

evals/agent_evals/agent_testing_utils.py
```diff
@@ -0,0 +1,103 @@
...
+            if cell.get("id") == target_cell_id:
+                cell["source"] = code
+                break
+    else:
+        raise ValueError(f"Unsupported update type: {update_type}")
+
+    all_code = "\n\n".join(
+        cell.source for cell in notebook.cells if cell.cell_type == "code"
+    )
+    # print(all_code)
```
aarondr77, line 33: Before we merge this in, let's just get rid of these commented out print statements across the PR

evals/agent_evals/handler.py
```diff
@@ -0,0 +1,140 @@
...
+def execute_test_case(input_nb_path, user_task, output_nb_path, output_response_path, output_conversation_history_path, input_conversation_history_path=""):
+    with open(input_nb_path, "r", encoding="utf-8") as f:
+        input_nb = nbformat.read(f, as_version=4)
+    existing_response_json = {"response": []}
+
+    if input_conversation_history_path:
+        with open(input_conversation_history_path, 'r', encoding='utf-8') as f:
+            conversation_history = json.load(f)
+    else:
+        conversation_history = start_new_conversation_history(PROMPT)
```
aarondr77, line 31: I don't think we are going to want to store the system prompt in the conversation_history json file because we are going to want to easily experiment with different system prompts. 

So in the json file we should just keep the raw user input (ie: what the user types into the chat taskpane) and the raw Ai outputs (just what the AI responds in the taskpane). 

Then, when we build the conversation history we can add the system prompt by calling the `create_agent_system_message_prompt` function and turn the raw user inputs into the actual messages that we send to the AI by calling the `create_prompt_from_code_and_user_task` function. 


Sooo... 

I think what this means is that: We can get rid of the `start_new_conversation_history(PROMPT)` code here. And just let the `get_openai_code_and_conversation_history` function build the system prompt and put it in the correct location.

aarondr77 (approved):
Very cool stuff! This is a great start to building a ton of useful evals infrastructure for the agent. I'm happy with this as a starting point. Great job. 

I think we can get this merged into dev after just a little bit of cleanup. I left a few todo's in the code that we don't need to do right now. But let's just address the code review comments. It shouldn't be too much work. 

Let me know if you have any questions about it of if you want to push back on any of the comments!

evals/agent_evals/run_test_case.py
PoojasreeDwarakanath: I currently assume that system prompt is part of the conversation history, but I'll go ahead and make this change

PR #1700, opened 2025-05-20
Beta settings

aarondr77:
# Description

1. Adds a new tab to the settings page that lets the user turn on/off beta features. 
2. Uses new beta toggle to display/hide the deploy app button

# Testing

1. Make sure on first load the Deploy App button is not visible
2. Go to the new General tab in the settings page and toggle beta mode on
3. Refresh the browser
4. Check that the Deploy App button is visible. 
5. Shut down Jupyter, relaunch it, make sure the Deploy App button is still visible. Your settings should be persistent! 

# Documentation

Nope


mito-ai/src/Extensions/SettingsManager/SettingsWidget.tsx
```diff
@@ -2,17 +2,23 @@ import React, { useState } from 'react';
...
 import { SupportPage } from './support/SupportPage';
+import { GeneralPage } from './general/GeneralPage';
 import '../../../style/SettingsWidget.css';
 
 const TABS_CONFIG = {
     database: {
         label: 'Database',
         component: DatabasePage
     },
+    general: {
```
ngafar, line 13: We can probably move this to the top so users start on the "General" page. 

You'll also need to change the default tab which can be found on the same line as the inital `setActiveTab` state.

mito-ai/src/RestAPI.tsx
```diff
@@ -0,0 +1,31 @@
+import { PageConfig } from "@jupyterlab/coreutils";
+
+const baseUrl = PageConfig.getBaseUrl();
+
+/************************************
+
+SETTINGS ENDPOINTS
```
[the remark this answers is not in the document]
ngafar: nvm, I see now that this is used in ToolbarButtons

mito-ai/src/Extensions/SettingsManager/general/GeneralPage.tsx
```diff
@@ -0,0 +1,39 @@
...
+        await updateSettings('beta_mode', newBetaMode.toString());
+        setBetaMode(newBetaMode);
+    };
+    
+    return (
+        <div>
+            <div className="settings-header">
+                <h2>General</h2>
+            </div>
+            <div className="settings-option">
```
ngafar, line 28: Looks good. In the future I can see this easily becoming an object we add stuff into. But might be overkill for today.

mito-ai/mito_ai/settings/handlers.py
```diff
@@ -0,0 +1,62 @@
...
+        if 'value' not in data:
+            self.set_status(400)
+            self.finish(json.dumps({"error": "Value is required"}))
+            return
+            
+        set_settings_field(key, data['value'])
+        self.finish(json.dumps({"status": "updated", "key": key, "value": data['value']}))
+
+
+def set_settings_field(field: str, value: Any) -> None:
```
ngafar, line 46: This might make more sense in a utils.py file. Same for get_settings_field.

mito-ai/src/Extensions/SettingsManager/SettingsWidget.tsx
aarondr77: Maybe we leave it for now just cause we want people going to the database page. But I agree we'll want to move this around once there are more things in the general tab!

PR #1703, opened 2025-05-21
DB logging

ngafar:
# Description

Added basic logging when users connect to db:

1. Log attempted database connections
2. Log successful database connections
3. Log unsuccessful database connections

# Testing

Need to test:

- correct db creds:  should log `mito_ai_db_connection_attempt` and `mito_ai_db_connection_success`.
- incorrect db cred: should log `mito_ai_db_connection_attempt` and `mito_ai_db_connection_error`.

# Documentation

N/A

### DIFF
--- a/mito-ai/mito_ai/utils/telemetry_utils.py
+++ b/mito-ai/mito_ai/utils/telemetry_utils.py
@@ -278,3 +278,5 @@ def log_ai_completion_success(
 
         log(f"mito_ai_{message_type.value}_success", params=final_params, key_type=key_type, thread_id=thread_id)
         
+def log_db_connection_attempt() -> None:
+    log("mito_ai_db_connection_attempt")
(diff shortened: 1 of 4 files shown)


mito-ai/mito_ai/utils/telemetry_utils.py
```diff
@@ -278,3 +278,22 @@ def log_ai_completion_success(
...
+
+def log_db_connection_success(connection_type: str, schema: dict[str, Any]) -> None:
+    log(
+        "mito_ai_db_connection_success", 
+        params={
+            "connection_type": connection_type, 
+            }
+        )
+
+def log_db_connection_error(connection_type: str, error_message: str) -> None:
```
aarondr77, line 292: It would be nice for these to still log when the user is on the pro + mito_server because for initial rollout, I think that most users will be pro people we do testing with. However, right now because we are not passing the key_type to the log function, we don't turn on telemetry for pro users. 

I don't know the best way to fix this though because of how we've set up our backend. The providers.py file is where we know about the key_type and the db connection doesn't go through that at all. 

Take a look at how `log_ai_completion_success` for more context on how this works. 

Any quick solutions? Lmk if you need more context.

aarondr77 (commented):
Just reminder to make sure that the mito src listing tests pass before merging

PR #1705, opened 2025-05-21
initial commit for anthropic client

sid-sagacollab:
# Description

This is to enable support for anthropic models/

# Testing

### DIFF
--- a/mito-ai/mito_ai/anthropic_client.py
+++ b/mito-ai/mito_ai/anthropic_client.py
@@ -3,41 +3,13 @@
 
 import json
 import anthropic
-from typing import Dict, Any, Optional, Union, Callable, List
+from typing import Dict, Any, Optional, Tuple, Union, Callable, List
 
-from anthropic.types import Message
+from anthropic.types import Message, MessageParam, ToolUnionParam
 from mito_ai.completions.models import ResponseFormatInfo, CompletionReply, CompletionStreamChunk, CompletionItem, AgentResponse
 from openai.types.chat import ChatCompletionMessageParam
 
 MAX_TOKENS = 2_000
-
-AGENT_RESPONSE_EXAMPLES = {
-    "cell_update_example": {
-        "type": "cell_update",
-        "message": "I've created a new cell with a function to calculate fibonacci numbers.",
-        "cell_update": {
-            "type": "new",
-            "index": 2,
-            "id": None,
-            "code": "def fibonacci(n):\n    if n <= 1:\n        return n\n    return fibonacci(n-1) + fibonacci(n-2)",
-            "cell_type": "code"
-        },
-        "get_cell_output_cell_id": None
-    },
-    "get_output_example": {
-        "type": "get_cell_output",
-        "message": "Let me check the output of the previous cell.",
-        "cell_update": None,
-        "get_cell_output_cell_id": "cell-123"
-    },
-    "finished_task_example": {
-        "type": "finished_task",
-        "message": "I've completed the requested analysis.",
-        "cell_update": None,
-        "get_cell_output_cell_id": None
-    }
-}
-
 
 def _extract_and_parse_json_response(response: Message) -> Union[object, Any]:
     """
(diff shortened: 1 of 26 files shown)


mito-ai/mito_ai/completions/providers.py
```diff
@@ -199,14 +230,28 @@ async def stream_completions(
...
+                    self._anthropic_client = AnthropicClient(
+                        api_key=constants.CLAUDE_API_KEY,
+                        model=model
+                    )
+
+                accumulated_response = await self._anthropic_client.stream_response(
+                    messages=messages,
+                    message_id=message_id,
+                    reply_fn=reply_fn
+                )
```
aarondr77, line 245: Sweet! It's awesome how this code is so simple for new clients. Nice!

mito-ai/mito_ai/anthropic_client.py
```diff
@@ -0,0 +1,261 @@
...
+
+import json
+import anthropic
+from typing import Dict, Any, Optional, Union, Callable, List
+
+from anthropic.types import Message
+from mito_ai.completions.models import ResponseFormatInfo, CompletionReply, CompletionStreamChunk, CompletionItem
+from openai.types.chat import ChatCompletionMessageParam
+
+AGENT_RESPONSE_SCHEMA = {
```
aarondr77, line 12: Is it possible to auto generate this from the Pydantic base class? That way we don't need to make sure we're keeping both in sync?

mito-ai/mito_ai/anthropic_client.py
```diff
@@ -0,0 +1,261 @@
...
+        messages: List of chat messages in OpenAI format
+
+    Returns:
+        A single string prompt formatted for Claude
+    """
+    # Filter out messages without content and convert to string format
+    formatted_messages = []
+    for msg in messages:
+        if isinstance(msg, dict) and 'content' in msg and msg['content']:
+            role = msg.get('role', 'user')
```
aarondr77, line 155: In what cases would the role not be set?

mito-ai/mito_ai/anthropic_client.py
```diff
@@ -0,0 +1,261 @@
...
+                # Make the API request with tool use
+                response = self.client.messages.create(
+                    model=self.model,
+                    max_tokens=2000,
+                    temperature=0,
+                    messages=[
+                        {"role": "user", "content": _formatted_prompt(prompt)}
+                    ],
+                    tools=tools,
+                    tool_choice={"type": "tool", "name": "agent_response"}
```
aarondr77, line 197: My understanding is that if we set the `tool_choice` field, we are forcing the Claude to always respond with this tool. If that is the case, do we need all of the `_formatted_prompt` stuff? It would obviously be ideal if we didn't need to maintain this additional set of prompt instructions, but if we need it, then we need it.

mito-ai/mito_ai/anthropic_client.py
```diff
@@ -0,0 +1,261 @@
...
+                )
+
+                result = _extract_and_parse_json_response(response)
+                return json.dumps(result) if not isinstance(result, str) else result
+
+            else:
+                response = self.client.messages.create(
+                    model=self.model,
+                    max_tokens=2000,
```
aarondr77, line 206: Can we create some variables for max_tokens and temperature at the top of this file so we can easily keep them consistent. 

I would also be down to remove max_tokens, I don' think we're setting that for other providers, right?

mito-ai/mito_ai/anthropic_client.py
```diff
@@ -0,0 +1,261 @@
...
+                    temperature=0,
+                    messages=[
+                        {"role": "user", "content": prompt}
```
aarondr77, line 209: Is this right? Why are we collapsing the entire conversation into one user prompt?. I think this is supposed to be a list of messages from the conversation. And it looks like the system prompt has its own field

From the docs here: https://docs.anthropic.com/en/docs/get-started#call-the-api

Screenshot 2025-05-22 at 11 38 23 AM

mito-ai/mito_ai/anthropic_client.py
```diff
@@ -0,0 +1,261 @@
...
+
+                Important requirements:
+                - If type is "cell_update", you MUST include a valid cell_update object with "type" and "code" fields.
+                - If type is "get_cell_output", you MUST include a get_cell_output_cell_id.
+                - For "finished_task", no additional fields are required.
+                - The message field is always required.
+                """
+
+
+def _convert_messages_to_prompt(messages: List[ChatCompletionMessageParam]) -> str:
```
aarondr77, line 141: I think this function is not properly handling the get_cell_output tool. Even with `claude-3-7-sonnet-20250219`, I'm getting stuck in loops of using the `get_cell_output` tool. 

I think it might be the case that we're not actually sending the cell output to the AI! When we send the cell output, the message is constructed in `create_ai_optimized_message`. Notice that it has `image_url` part to the message as well. 

This might also be the case for Gemini Client! Can we add pytest tests for this `_convert_messages_to_prompt` function and also the gemini one?


Screenshot 2025-05-22 at 11 47 22 AM

mito-ai/mito_ai/anthropic_client.py
aarondr77: I tested this a little bit with the `_formatted_prompt` removed and it is adhering to the json output. But maybe you've tested it more than me!

aarondr77 (commented):
Nice work. This is working pretty well (except I keep getting rate limited). I think we need to use our Claude keys more so they bump our Anthropic tier! I have a few questions here about the anthropic speficic stuff.

Let me know what you think.

mito-ai/mito_ai/anthropic_client.py
sid-sagacollab: This piece of code gets triggered in two scenarios,

1. For inline completions

2. For setting the title of the chat.

Both of these are concerned with the code in the current cell and the first response from the chat/agent respectively and not the whole conversation history afaik. So, I think this can be as is. Lmk if there is a hole in my understanding and if this needs additional work.

mito-ai/mito_ai/anthropic_client.py
sid-sagacollab: will be dealt with here: https://github.com/mito-ds/mito/issues/1707

mito-ai/mito_ai/anthropic_client.py
sid-sagacollab: The `max_tokens` is a mandatory field. I have moved it to the top of the file in my subsequent commit.

mito-ai/mito_ai/anthropic_client.py
sid-sagacollab: This is just a guard rail to prevent the code from breaking when the role is set to `system`.

mito-ai/mito_ai/anthropic_client.py
sid-sagacollab: Done

mito-ai/mito_ai/completions/providers.py
sid-sagacollab: :)

aarondr77:
Can you make sure that the `mito-ai src` and `mito-ai mypy` tests are passing before merging. I'll take care of the otheres

PR #1706, opened 2025-05-22
Mito rules

aarondr77:
# Description

Adds rules so you can customize how the Ai responds, provide it additional context, etc! Good communication makes your colleagues better at collaborating with you! 

There's some polish that would be nice, but not needed to show to client https://github.com/mito-ds/mito/issues/1713

# Testing

Add a new rule! 

# Documentation

Yes sir!


mito-ai/mito_ai/completions/prompt_builders/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    if selected_rules is None:
+        return ''
+    
+    rules_str = ''
+    for rule in selected_rules:
+        rule_content = get_rule(rule)
+        if rule_content is None or rule_content == '':
+            continue
+        
+        rules_str += f"===========\n\nCustom Instructions Provided by User: {rule}\n\n{rule_content}\n\n==========="
```
sid-sagacollab, line 17: Would it be more useful to add stronger wording here such that it does not get shrouded amidst all the other context ? 


```suggestion
        rules_str += f"===========\n\ESSENTIAL BEHAVIORAL MODIFICATION:
Integrate and enforce this user-defined rule as a core operational parameter: {rule}
\n\n{rule_content}\n\n==========="
```

mito-ai/mito_ai/rules/handlers.py
```diff
@@ -0,0 +1,41 @@
...
+        """Update or create a specific setting"""
+        data = json.loads(self.request.body)
+        if 'content' not in data:
+            self.set_status(400)
+            self.finish(json.dumps({"error": "Content is required"}))
+            return
+            
+        set_rules_file(key, data['content'])
+        self.finish(json.dumps({"status": "updated", "rules file ": key}))
+
```
sid-sagacollab, line 40: Wouldn't it make sense to also add delete method ? something like this?


```suggestion
@tornado.web.authenticated
def delete(self, key):
    """Delete a specific rule"""
    try:
        if not key:
            return self._handle_error(400, "Rule key is required")
        
        if delete_rule(key):  # Assuming this function exists
            self._handle_success({"status": "deleted", "key": key})
        else:
            self._handle_error(404, f"Rule '{key}' not found")
    except Exception as e:
        self._handle_error(500, str(e))
```


Or is the expectation that the rules page dictates what rules to be added and the deletion of rules would be happening only in that page by the user

mito-ai/src/restAPI/RestAPI.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export const getRules = async(): Promise<string[]> => {
+    const resp = await requestAPI<string[]>(`rules`)
+    if (resp.error) {
+        throw new Error(resp.error.message);
+    }
+    return resp.data || [];
+}
+
```
sid-sagacollab, line 63: Same comment about implementing deletion of rules

mito-ai/mito_ai/rules/handlers.py
aarondr77: I'm going to leave off DELETE for now because deleting files on the user's file system is always a little bit more dangerous incase the path gets fucked up and we accidentally delete something we don't want to be deleting. I've seen this happen in other Open Source Jupyter Extension.

mito-ai/mito_ai/completions/prompt_builders/utils.py
ngafar: I'd be curious to see how well this works (evals anyone?). fwiw, just based on past prompting experience, the simplest explanations usually work best.

mito-ai/mito_ai/tests/rules/rules_test.py
```diff
@@ -0,0 +1,117 @@
...
+# Distributed under the terms of the GNU Affero General Public License v3.0 License.
+
+import requests
+
+RULE_NAME = "test_rule"
+RULE_CONTENT = "This is a test rule for data analysis."
+
+from mito_ai.tests.conftest import TOKEN
+
+# --- PUT RULES ---
```
ngafar, line 11: These tests are looking awfully familiar!

ngafar:
This looks good. I would suggest an alternate, and perhaps, simpler architecture for this feature. One where all of the rules are contained in a single JSON file, and where we simply add them all to the prompt. 

I only suggest this because our "at" feature severely underused, based on the call and videos. Second, I can't imagine a time where I'd want to specifically reference a rule. 

For example, I'd use this feature to specify that package management should be done by uv. However, I'm not going to "at" this rule every time, but I'd still like it to be included with every prompt. Just incase the code call for an installation. 

Along these lines, I would consider creating a single JSON file. It would be formatted like so:

```json
{
    "uuid-for-rule1": {
        "name": "only use uv",
        "content": "Always use uv for package management"
    },
    "uuid-for-rule2": {
        "name": "...",
        "content": "..."
    }
}
```

Then we could give users a limit on the number of characters allocatable to rules, and create an endpoint that keeps track of this. 

By switching to a single JSON file we'd:

1. Only have to worry about a single file.
2. Make deleting easier, using UUIDs.
3. Not have to worry about invalid names with spaces.

mito-ai/src/restAPI/RestAPI.tsx
```diff
@@ -0,0 +1,63 @@
...
```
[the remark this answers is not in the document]
aarondr77: This file is using the requestAPI function!

aarondr77:
These are all cool ideas. Especially being able to set an option to always include the rules in every new chat makes a lot of sense for the uv use case. However, there's another set of use cases (the team that this was originally built for) where for specific projects they want to give it additional context about specific python packages. 

In this case, what they are looking to do is copy and paste a huge markdown file of documentation into the context. Just since that's the main use case right now, I'm going to leave it as is. And then if we decide that improving rules is worthwhile, let's make some of these changes for sure!

PR #1715, opened 2025-05-27
UI Component for Model Selection

sid-sagacollab:
# Description

https://github.com/user-attachments/assets/6d453890-92bd-4c85-ac06-4e43c4e94551

# Testing

# Documentation

Note that currently for a model apart from the default OpenAI model to work, the user needs to set the `GEMINI_API_KEY` or `CLAUDE_API_KEY` or `OPENAI_API_KEY` respectively depending upon the provider they would like to deal with. In a future PR, the mito server will support all the models.


mito-ai/style/ModelSelector.css
```diff
@@ -0,0 +1,183 @@
+/*
+ * Copyright (c) Saga Inc.
+ * Distributed under the terms of the GNU Affero General Public License v3.0 License.
+ */
+
+/* ModelSelector.css - responsive design with compact mode */
+
+.model-selector {
```
aarondr77, line 8: This button is very similar in looks to a button with the class `button-base` and `button-purple` defined in buttons.css.  Can we use those classes to handle the bulk of the formatting and then just add the extra classes that are not covered by those. That way we can use a consistent formatting. For example, I think the border radius on this button is larger than the border radius of our buttons. 


Screenshot 2025-05-27 at 3 02 44 PM

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+interface ModelConfig {
+  model: string;
+}
+
+const ALL_MODELS = [
+  'gpt-4.1',
+  'o3-mini',
+  'claude-opus-4-20250514',
+  'claude-sonnet-4-20250514	',
```
aarondr77, line 20: We don't need the big space before ending `'`, right?

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,221 @@
...
+import '../../style/ModelSelector.css';
+// Import the AI icon image
+// Note: You'll need to adjust this import path to match your project structure
+import AIIcon from "../icons/AiIcon";
+
+interface ModelConfig {
+  model: string;
+}
+
+const ALL_MODELS = [
```
aarondr77, line 16: I think we're giving users too much model choices here. Its choice overload. 

Can we instead just give them the best options. 

I think we could go with: 
1. Open AI: gpt-4.1
2. Gemini 2.5 pro
3. Claude-4-Sonnet 
4. Claude-4-opus

For all of them, we can default to `latest`

Which other models should we include?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1274,6 +1293,10 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
                         }}
                         title="Agent can create plans and run code."
                     />
+                    <ModelSelector onConfigChange={(config) => {
```
aarondr77, line 1296: Instead of evenly spacing this, I think we can just always keep it on the left next to the Chat/Agent selection code
Screenshot 2025-05-27 at 3 11 52 PM

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,221 @@
...
+interface ModelSelectorProps {
+  onConfigChange: (config: ModelConfig) => void;
+}
+
+const ModelSelector: React.FC<ModelSelectorProps> = ({ onConfigChange }) => {
+  const [selectedModel, setSelectedModel] = useState<string>('gpt-4o-mini');
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const [dropdownWidth, setDropdownWidth] = useState<number>(0);
+  const [isCompact, setIsCompact] = useState<boolean>(false);
+  const selectedModelRef = useRef<HTMLSpanElement>(null);
```
aarondr77, line 45: I don't think we need this `selectedModelRef`, can't we just use the `selectedModel` instead?

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+// Maximum length for displayed model name before truncating
+const MAX_MODEL_NAME_LENGTH = 18;
+
+interface ModelSelectorProps {
+  onConfigChange: (config: ModelConfig) => void;
+}
+
+const ModelSelector: React.FC<ModelSelectorProps> = ({ onConfigChange }) => {
+  const [selectedModel, setSelectedModel] = useState<string>('gpt-4o-mini');
```
aarondr77, line 41: let's default to `gpt-4.1`

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,221 @@
+/*
+ * Copyright (c) Saga Inc.
+ * Distributed under the terms of the GNU Affero General Public License v3.0 License.
+ */
+
+import React, { useState, useEffect, useRef } from 'react';
+import '../../style/ModelSelector.css';
+// Import the AI icon image
+// Note: You'll need to adjust this import path to match your project structure
```
aarondr77, line 9: We can delete this

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+    document.addEventListener('mousedown', handleClickOutside);
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside);
+    };
+  }, []);
+
+  const dropdownRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
```
aarondr77, line 128: It looks like all of this code is not really helpful, if I'm understanding it correctly. 

1. The first section of code is checking if the dropdown should be above or below? As far as I can tell there's no case where the dropdown should be below. 
2. The second chunk of code figures out where the the dropdown should be located horizontally? But even when I delete all of this code it seems positioned perfectly fine. 

I think if we delete this code this entire useEffect.

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,221 @@
...
+      }
+    }
+  }, [isOpen, isCompact]);
+
+  return (
+    <div className="model-selector" ref={containerRef}>
+      <div
+        className={`model-selector-dropdown ${isCompact ? 'compact-mode' : ''}`}
+        onClick={() => setIsOpen(!isOpen)}
+        style={{ width: `${dropdownWidth}px` }}
```
aarondr77, line 177: Instead of hardcoding the `width` here, I think it's better practice to just let the css take care of it. You can set something like `width: min-content` and I think that will have the same impact.

mito-ai/src/components/ModelSelector.tsx
aarondr77: Let's also make this a variable at the top of the file under `MAX_MODEL_NAME_LENGTH` and use the variable here as well to keep it in sync `const model = parsedConfig.model || 'gpt-4o-mini';`

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,221 @@
...
+  useEffect(() => {
+    if (selectedModelRef.current && !isCompact) {
+      const textWidth = selectedModelRef.current.scrollWidth;
+      setDropdownWidth(Math.min(textWidth + 30, 150)); // Set a maximum width of 150px
+    } else if (isCompact) {setDropdownWidth(28);
+    }
+  }, [selectedModel, isCompact]);
+
+  // Shorten model name if it's too long
+  const truncateModelName = (name: string): string => {
```
aarondr77, line 91: I think we can just handle this in the css by saying `overflow-x: ellipsis`

mito-ai/style/ModelSelector.css
```diff
@@ -0,0 +1,183 @@
...
+    display: flex;
+    align-items: center; /* Center vertically */
+  }
+  
+  .model-selector-dropdown {
+    position: relative;
+    min-width: 80px; /* Minimum width for very short model names */
+    height: 20px; /* Fixed height to match other elements */
+    border-radius: 4px;
+    border: 1px solid #c3a4fe;
```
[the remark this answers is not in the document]
aarondr77: This will, ie, make it look better in dark mode. Let's try using similar colors to `chatDropdown.css` for consistency
Screenshot 2025-05-27 at 3 34 26 PM

mito-ai/mito_ai/completions/handlers.py
```diff
@@ -187,12 +185,40 @@ async def on_message(self, message: str) -> None: # type: ignore
...
+            return
+        
         try:
             # Get completion based on message type
             completion = None
             message_id = parsed_message.get('message_id')
             stream = parsed_message.get('stream')
 
+            # When handling completions, always use the selected model
+            model = self._selected_model or MESSAGE_TYPE_TO_MODEL.get(type, "gpt-4.1") or ""
```
aarondr77, line 221: Would the `self._selected_model` ever be None?

aarondr77: This case will definetly never get to the `or ""` right? because at minimum it will get `gpt-4.1`

mito-ai/mito_ai/completions/completion_handlers/agent_auto_error_fixup_handler.py
```diff
@@ -20,7 +20,8 @@ class AgentAutoErrorFixupHandler(CompletionHandler[AgentSmartDebugMetadata]):
     async def get_completion(
         metadata: AgentSmartDebugMetadata,
         provider: OpenAIProvider,
-        message_history: GlobalMessageHistory
+        message_history: GlobalMessageHistory,
+        model: str = None
```
aarondr77, line 24: Based on the handlers.py code, I think this will never be None, so can we get rid of this default and have it be a required field?

mito-ai/mito_ai/completions/handlers.py
```diff
@@ -67,9 +67,7 @@ def initialize(self, llm: OpenAIProvider) -> None:
         self.log.debug("Initializing websocket connection %s", self.request.path)
         self._llm = llm
         self.is_pro = is_pro()
-        
-        # Re-identify the user the user once we have the key type so that 
-        # we can check if the user is on the mito server or their own key
+        self._selected_model = None
```
aarondr77, line 70: Should we default to `gpt4.1` to start? There is never really such a thing as no model selected

mito-ai/mito_ai/completions/providers.py
```diff
@@ -55,135 +51,83 @@ class OpenAIProvider(LoggingConfigurable):
...
     @property
     def capabilities(self) -> AICapabilities:
-        """Get the provider capabilities."""
-        if constants.CLAUDE_MODEL and constants.CLAUDE_API_KEY and not self.api_key:
+        if constants.CLAUDE_API_KEY and not self.api_key:
             return AICapabilities(
-                configuration={
-                    "model": constants.CLAUDE_MODEL
-                },
+                configuration={"model": "<dynamic>"},
```
aarondr77, line 69: why `dynamic` instead of returning the actual model?

mito-ai/mito_ai/completions/providers.py
```diff
@@ -55,135 +51,83 @@ class OpenAIProvider(LoggingConfigurable):
...
-            # Handle OpenAI and other providers
-            elif self._openai_client:
+            if model_type == "claude":
+                api_key = constants.CLAUDE_API_KEY or ""
+                anthropic_client = AnthropicClient(api_key=api_key, model=model)
+                completion = await anthropic_client.request_completions(messages, response_format_info)
+            elif model_type == "gemini":
+                api_key = constants.GEMINI_API_KEY or ""
+                gemini_client = GeminiClient(api_key=api_key, model=model)
+                from typing import cast, Dict, Any
```
aarondr77, line 118: Let's move these imports to the top of the file

mito-ai/mito_ai/completions/providers.py
```diff
@@ -55,135 +51,83 @@ class OpenAIProvider(LoggingConfigurable):
...
-                        model=model
-                    )
-
-                completion = await self._gemini_client.request_completions(messages, response_format_info)
-
-            # Handle OpenAI and other providers
-            elif self._openai_client:
+            if model_type == "claude":
+                api_key = constants.CLAUDE_API_KEY or ""
+                anthropic_client = AnthropicClient(api_key=api_key, model=model)
```
aarondr77, line 113: Do we want to save this client in the class? Currently it looks like `self._anthropic_client` is always just going to be None. So either let's save it there or I guess delete `self._anthropic_client`

aarondr77: Same with gemini client

mito-ai/mito_ai/completions/providers.py
```diff
@@ -55,135 +51,83 @@ class OpenAIProvider(LoggingConfigurable):
...
+                completion = await anthropic_client.request_completions(messages, response_format_info)
+            elif model_type == "gemini":
+                api_key = constants.GEMINI_API_KEY or ""
+                gemini_client = GeminiClient(api_key=api_key, model=model)
+                from typing import cast, Dict, Any
+                completion = await gemini_client.request_completions(cast(List[Dict[str, Any]], messages), response_format_info)
+            elif model_type == "openai":
+                if not self._openai_client:
```
aarondr77, line 121: Why is this different for open ai? I think we want to be treating all clients the same just so our mental model of how this code works is easy

mito-ai/mito_ai/completions/providers.py
```diff
@@ -55,135 +51,83 @@ class OpenAIProvider(LoggingConfigurable):
...
-            model: The model to request completions for.
-        Returns:
-            The completion from the AI provider.
         """
-        # Reset the last error
         self.last_error = None
         completion = None
         last_message_content = str(messages[-1].get('content', '')) if messages else ""
-
+        model_type = get_model_type(model)
```
aarondr77, line 109: Maybe a better name is `model_provider` and `get_model_provider`?

mito-ai/mito_ai/completions/providers.py
```diff
@@ -192,74 +136,59 @@ async def request_completions(
...
-                        api_key=constants.GEMINI_API_KEY,
-                        model=model
-                    )
-
-                accumulated_response = await self._gemini_client.stream_completions(
-                    messages=messages,
+            elif model_type == "gemini":
+                api_key = constants.GEMINI_API_KEY or ""
+                gemini_client = GeminiClient(api_key=api_key, model=model)
+                from typing import cast, Dict, Any
```
aarondr77, line 183: Let's make these global imports also

mito-ai/mito_ai/completions/providers.py
```diff
@@ -55,135 +51,83 @@ class OpenAIProvider(LoggingConfigurable):
...
-            elif self._openai_client:
+            if model_type == "claude":
+                api_key = constants.CLAUDE_API_KEY or ""
+                anthropic_client = AnthropicClient(api_key=api_key, model=model)
+                completion = await anthropic_client.request_completions(messages, response_format_info)
+            elif model_type == "gemini":
+                api_key = constants.GEMINI_API_KEY or ""
+                gemini_client = GeminiClient(api_key=api_key, model=model)
+                from typing import cast, Dict, Any
+                completion = await gemini_client.request_completions(cast(List[Dict[str, Any]], messages), response_format_info)
```
aarondr77, line 119: I don't think this is the best way to handle converting from OpenAIMessageParam to the Gemini client message type. Can we do the same thing that we're doing with the `anthropic client` where we have a function that does this for us. That we are actually being type safe.

mito-ai/mito_ai/completions/providers.py
```diff
@@ -192,74 +136,59 @@ async def request_completions(
...
-                    )
-
-                accumulated_response = await self._gemini_client.stream_completions(
-                    messages=messages,
+            elif model_type == "gemini":
+                api_key = constants.GEMINI_API_KEY or ""
+                gemini_client = GeminiClient(api_key=api_key, model=model)
+                from typing import cast, Dict, Any
+                accumulated_response = await gemini_client.stream_completions(
+                    messages=cast(List[Dict[str, Any]], messages),
```
aarondr77, line 185: Same as above for this casting

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -141,6 +142,24 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
     const streamingContentRef = useRef<string>('');
     const streamHandlerRef = useRef<((sender: CompletionWebsocketClient, chunk: ICompletionStreamChunk) => void) | null>(null);
 
+    const updateModelOnBackend = async (model: string): Promise<void> => {
```
aarondr77, line 145: Maybe in the useEffect where we call the `initializeChatHistory` function we should also call `updateModelOnBackend` to the websocket model is updated when the taskpane is opened

mito-ai/mito_ai/completions/handlers.py
```diff
@@ -220,12 +247,13 @@ async def on_message(self, message: str) -> None: # type: ignore
...
                         message_id,
-                        self.reply
+                        self.reply,
+                        model=model or ""
                     )
                     return
                 else:
                     # Regular non-streaming completion
-                    completion = await get_smart_debug_completion(smart_debug_metadata, self._llm, message_history)
+                    completion = await get_smart_debug_completion(smart_debug_metadata, self._llm, message_history, model=model or "")
```
aarondr77, line 256: If we just handle setting the model right at the top of this function then we don't need any of these `model=model or ""` either. That's one of the benefits of pushing up our type handling

mito-ai/mito_ai/completions/providers.py
```diff
@@ -55,135 +51,83 @@ class OpenAIProvider(LoggingConfigurable):
...
-            },
+            configuration={"model": "gpt-4.1"},
             provider="Mito server",
         )
 
     @property
     def key_type(self) -> str:
-        """Returns the authentication key type being used."""
-        if constants.CLAUDE_MODEL and constants.CLAUDE_API_KEY and not self.api_key:
+        if constants.CLAUDE_API_KEY and not self.api_key:
```
aarondr77, line 86: These checks also depends on what the selected model is. If the user has:
1. A Claude API key set
2. No OpenAI key set
3. No Gemini iKey set, 
4. but selected a Gemini model

The key_type should be `mito_server`, but I think right now it would say User Claude Key

aarondr77: Can we just make sure we have a few simple pytests for this?

sid-sagacollab: This will also be taken care of in the Mito-server PR and basically the flow to providers is going to change a bit, so I feel it's a good idea to let it be as is for now and that I handle it there.

mito-ai/src/components/ModelSelector.tsx
sid-sagacollab: Some of them do not have latest attributes 

https://ai.google.dev/gemini-api/docs/models

https://docs.anthropic.com/en/docs/about-claude/models/overview

mito-ai/src/components/ModelSelector.tsx
sid-sagacollab: `selectedModelRef` is used to measure the rendered width of the selected model's name (via `selectedModelRef.current.scrollWidth`) in order to set the dropdown width dynamically. However, we cannot directly use the `selectedModel` string to measure the width, because the width depends on how the text is rendered in the DOM (font, size, etc.), not just the string length.

mito-ai/mito_ai/completions/providers.py
sid-sagacollab: This will be fixed in the Mito server work where this section is bound to change.

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,200 @@
...
+import AIIcon from "../icons/AiIcon";
+
+interface ModelConfig {
+  model: string;
+}
+
+const ALL_MODELS = [
+  'gpt-4.1',
+  'claude-opus-4-20250514',
+  'claude-sonnet-4-20250514',
```
aarondr77: Can we update this to that what we show to the user is `claude-opus-4`, `gemini-2.5-pro`, etc. instead of the the full thing. And then we can send these full names to the backend

Screenshot 2025-06-06 at 11 13 22 AM

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,200 @@
...
+    return () => observer.disconnect();
+  }, []);
+
+  useEffect(() => {
+    if (selectedModelRef.current && !isCompact) {
+      // Removed dropdownWidth logic
+    } else if (isCompact) {
+      // Removed dropdownWidth logic
+    }
+  }, [selectedModel, isCompact]);
```
aarondr77: Looks like we can remove this...

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+    document.addEventListener('mousedown', handleClickOutside);
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside);
+    };
+  }, []);
+
+  const dropdownRef = useRef<HTMLDivElement>(null);
+
+  // useEffect(() => {
```
aarondr77: We can remove all of this obviously..

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+// Maximum length for displayed model name before truncating
+export const DEFAULT_MODEL = 'gpt-4.1';
+
+interface ModelSelectorProps {
+  onConfigChange: (config: ModelConfig) => void;
+}
+
+const ModelSelector: React.FC<ModelSelectorProps> = ({ onConfigChange }) => {
+  const [selectedModel, setSelectedModel] = useState<string>('gpt-4o-mini');
```
aarondr77: Don't forget to update this to `DEFAULT_MODEL` its already out of sync :)

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -270,28 +319,31 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+                    (error as { hint?: string }).hint ? (error as { hint?: string }).hint! : `${error}`,
                     'chat',
                     newChatHistoryManager,
                     true
                 );
             }
         };
 
+        // Set the model on backend when the taskpane is opened
+        void updateModelOnBackend(DEFAULT_MODEL); // Placeholder/default model
```
aarondr77: Does it make sense to always be setting this as default model? Shouldn't we first check if there is a model set in the local storage and use that if there is?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1109,7 +1131,7 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
             accelYDisposable.dispose();
             accelDDisposable.dispose();
         };
-    }, [codeReviewStatus]);
+    }, [codeReviewStatus, app.commands]);
```
aarondr77: why are we adding app.comands here? is this related to this pr?

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,200 @@
...
+        const model = parsedConfig.model || DEFAULT_MODEL;
+        setSelectedModel(model);
+
+        onConfigChange({ model });
+
+      } catch (e) {
+        console.error('Failed to parse stored LLM config', e);
+      }
+    }
+  }, [onConfigChange]);
```
aarondr77: Yeah this logic should get moved to the to the ChatTaskpane, so we are checking if the local storage right away. 

Currently, there's some infinite loop of setting the models going on. 

https://github.com/user-attachments/assets/5c5996bc-9eb8-49e9-98e3-e8f3fa9deaf0

mito-ai/mito_ai/completions/completion_handlers/inline_completer_handler.py
```diff
@@ -19,7 +19,8 @@ class InlineCompleterHandler(CompletionHandler[InlineCompleterMetadata]):
     async def get_completion(
         metadata: InlineCompleterMetadata,
         provider: OpenAIProvider,
-        message_history: GlobalMessageHistory
+        message_history: GlobalMessageHistory,
+        model: Optional[str] = None
```
aarondr77: I think by the time we call each of these handlers, the model should defiantly be set, right? Since we are defaulting to the Fallback model instead of None. 

So I think we can:
1. Update these types to be str instead of Optional[str]
2. Delete the `MESSAGE_TYPE_TO_MODEL` 

We can make this update in all of the handlers.

mito-ai/src/components/ModelSelector.tsx
sid-sagacollab: This section of code is useful in this scenario: 

When the agent starts running, once it finishes running, it re-renders the `chat-controls` (i.e the bottom section of agent/chat, model selector and submit button). If we do not fetch the model from localstorage, the dropdown selection will be the default model (so always get-4.1).

aarondr77:
I got this error after the agent used the `get_cell_output` tool with Claude 4 Sonnet

Screenshot 2025-06-10 at 10 37 44 AM

mito-ai/mito_ai/completions/providers.py
```diff
@@ -58,145 +55,81 @@ class OpenAIProvider(LoggingConfigurable):
...
-                        self._gemini_client = GeminiClient(
-                            api_key=None,
-                            model=constants.GEMINI_MODEL
-                        )
-                completion = await self._gemini_client.request_completions(messages, response_format_info, message_type)
-
-            # Handle OpenAI and other providers
-            elif self._openai_client:
+            if model_type == "claude":
+                api_key = constants.CLAUDE_API_KEY or None
```
aarondr77: I think these contents are already either the string or None because `os.environ.get("CLAUDE_API_KEY")` will return None if they don't exist.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1301,21 +1335,27 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+                                const chatInput = document.querySelector('.chat-input') as HTMLTextAreaElement;
+                                if (chatInput) {
+                                    chatInput.focus();
+                                }
+                            }}
+                            title="Agent can create plans and run code."
+                        />
+                        <ModelSelector onConfigChange={(config) => {
+                        // Just update the backend
+                        void updateModelOnBackend(config.model);
```
aarondr77: Indent

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,156 @@
...
+interface ModelMapping {
+  displayName: string;
+  fullName: string;
+}
+
+const MODEL_MAPPINGS: ModelMapping[] = [
+  { displayName: 'GPT 4.1', fullName: 'gpt-4.1' },
+  { displayName: 'Claude 4 Opus', fullName: 'claude-opus-4-20250514' },
+  { displayName: 'Claude 4 Sonnet', fullName: 'claude-sonnet-4-20250514' },
+  { displayName: 'Gemini 2.5 Pro', fullName: 'gemini-2.5-pro-preview-03-25' }
```
aarondr77, line 23: Heck yeah. This is great

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -0,0 +1,156 @@
...
+];
+
+const ALL_MODELS = MODEL_MAPPINGS.map(mapping => mapping.displayName);
```
aarondr77: Can we call this variable `ALL_MODEL_DISPLAY_NAMES`?

PR #1720, opened 2025-05-27
Filtering databases based kind

ngafar:
# Description

We are only importing standard and imported databases. Snowflake applications are not supported.  

# Testing

Connect to Snowflake, make sure the SNOWFLAKE apple is not in the schema.json.

# Documentation

Yes, we should update the docs to let users know what types of databases are supported.


aarondr77 (approved):
Nice work. LGTM. Do you think there is a simple test we should add for this? ie: maybe that error thing I called out?

Issue #1698, opened 2025-05-19
Selecting Model should only effect Chat/Agent completions.


aarondr77:
When the user sets a model, like `Gemini 2.5 Pro Preview` we cannot also use this for the inline completer it is way too slow. 

We are sending a lot of inline completion requests and as a result, we dramatically hurt Jupyter performance. For example, while waiting for the inline completion to come back, you cannot kill the Jupyter server. 

I think since some users want to set an API key instead of using our key we should do the following: 

1. When the user sets a model in the chat/agent taskpane, we use that model for the chat/agent completions
2. For all other completions, we use a fast model from the same provider.
 ie: If the user selects  `Gemini 2.5 Pro Preview`, then we chose a fast Gemini model, ie: `Gemini 2.0 Flash-Lite`

Or if the user selects gpt-4.1 for

PR #1729, opened 2025-05-31
Mito server support

sid-sagacollab:
# Description

This is a PR to support mito server completions when the user does not bring their own key.

# Testing

# Documentation

None


aarondr77 (changes requested):
A few changes here to simplify things across all providers. I think we want to make shared utilities whenever possible and make sure the approaches across all three providers are as similar as possible (they are pretty similar right now already, nice!) so its easier to understand how they all work without having to relearn the mental model for each one them. 

We also need some pytests for these things! Let's make sure to do that before merging because it will help us test these things faster!

mito-ai/mito_ai/constants.py
```diff
@@ -22,4 +22,21 @@
 AZURE_OPENAI_API_KEY = os.environ.get("AZURE_OPENAI_API_KEY")
 AZURE_OPENAI_API_VERSION = os.environ.get("AZURE_OPENAI_API_VERSION")
 AZURE_OPENAI_ENDPOINT = os.environ.get("AZURE_OPENAI_ENDPOINT")
-AZURE_OPENAI_MODEL = os.environ.get("AZURE_OPENAI_MODEL")
\ No newline at end of file
+AZURE_OPENAI_MODEL = os.environ.get("AZURE_OPENAI_MODEL")
+
+# Mito AI Base URLs and Endpoint Paths
+MITO_BASE_URL = "https://x3rafympznv4abp7phos44gzgu0clbui.lambda-url.us-east-1.on.aws"
+MITO_DEV_BASE_URL = "https://x3rafympznv4abp7phos44gzgu0clbui.lambda-url.us-east-1.on.aws"
```
aarondr77: Hmmm these shouldn't be the same urls

mito-ai/mito_ai/constants.py
```diff
@@ -22,4 +22,21 @@
...
+
+# Endpoint paths
+ANTHROPIC_PATH = "anthropic/completions"
+GEMINI_PATH = "gemini/completions"
+OPENAI_PATH = "openai/completions"
+
+# Full URLs (for reference, but use base + path in code)
```
aarondr77: Instead of building the PROD_URL and the DEV_URL can we do it this way: 

1. Have the PROD Base URL and the DEV BASE URL above
2. Have another variable called ACTIVE_BASE_URL that we either set to the prod or dev one
3. Then here we can just build one url for each provider using the active url

This way, when we want to test the dev end point we can just change the ACTIVE_BASE_URL instead of having to change each provider. We don't want to accidentally ship code that has the dev url set for one of the providers!

mito-ai/mito_ai/gemini_client.py
```diff
@@ -7,6 +7,7 @@
     ResponseFormatInfo, MessageType
 from mito_ai.utils.gemini_utils import get_gemini_completion_from_mito_server, stream_gemini_completion_from_mito_server
 
+INLINE_COMPLETION_MODEL = "gemini-2.0-flash-lite"
```
aarondr77, line 10: Nice! This should work well! We need to do the same thing for anthropic still, right?

mito-ai/mito_ai/utils/anthropic_utils.py
```diff
@@ -0,0 +1,221 @@
...
+        if fetch_future:
+            try:
+                await fetch_future
+            except Exception:
+                pass
+        raise
+    finally:
+        http_client.close()
+
+def get_anthropic_completion_function_params(
```
aarondr77, line 188: Looks like we're still missing this function for gemini

aarondr77: Let's definetly try to keep the structure of all the providers the same

aarondr77 (approved):
Just a few tiny changes and then LGTM!

mito-ai/mito_ai/gemini_client.py
sid-sagacollab: The issue was describing only gemini and openai, so I just stuck to those https://github.com/mito-ds/mito/issues/1698.

I'll add the anthropic Claude-3.5-haiku for that, it is their fastest model.

mito-ai/mito_ai/utils/anthropic_utils.py
sid-sagacollab: Whoops. Missed that one

mito-ai/mito_ai/constants.py
sid-sagacollab: This is actually smart.

mito-ai/mito_ai/constants.py
sid-sagacollab: I just kept them as the same since in PROD we do not have them deployed and it would break the code if someone just tried cloning it in dev once its merged since the utils point to the PROD URLs

sid-sagacollab: I think it would be a good idea to just replace it with the new prod url once we have it.

Issue #1718, opened 2025-05-27
Clear Variable Manager on restart kernel


aarondr77:
Make sure that the variable manager is cleared when the kernel restarts, currently it is only reset (I think) after the user executes an another code cell. This leads to the variable manager being out of date directly after reseting the kernel when all of the variables are deleted!

ngafar:
Possible duplicate of https://github.com/mito-ds/mito/issues/1640

PR #1730, opened 2025-06-02
Clearing Variable Manager on Kernel Refresh 

ngafar:
# Description

Fixes https://github.com/mito-ds/mito/issues/1718

When the kernel is refreshed, the variable manager is set to an empty array. 

# Testing

1. Create a notebook with some variables. 
2. Take note of what is in the Context Manager. 
3. Then refresh the kernel.
4. The Context Manager's `_variable` field should be empty. 

# Documentation

N/A


aarondr77 (approved):
Love it and love the tests!

PR #1734, opened 2025-06-03
Metapackage

ktaletsk:
# Description

This PR introduces the **Mito metapackage** - a convenience package that automatically installs both `mitosheet` and `mito-ai` with a single `pip install mito` command. This addresses user requests for simplified installation and provides a unified entry point for the Mito ecosystem.

### 🎯 **Pure Metapackage Implementation**
- Created `mito/` directory with metapackage configuration
- Uses **Hatchling** as the modern build backend
- Pure metapackage approach (no Python source code) - users import `mitosheet` and `mito_ai` directly
- **Calendar Versioning (CalVer)** with `YYYY.MM[.PATCH]` scheme (e.g., `2025.06`, `2025.06.1`)

### 🤖 **Automated CI/CD Pipeline**
- **Daily automated checks** for new mito-ai/mitosheet releases via GitHub Actions
- **Smart version detection** - only publishes when dependencies are updated
- **CalVer generation** - automatically increments patch versions within the same month
- **Manual trigger** option with force-deploy capability

### 📦 **Package Configuration**
- **Dependencies**: `mito-ai>=0.1.20`, `mitosheet>=0.2.20` (automatically updated)
- **Python Support**: >=3.9 (aligned with both packages)
- **License**: AGPL-3.0-or-later
- **Metadata**: Comprehensive project URLs, keywords, and classifiers

## Motivation

**Before**: Users needed to install both packages separately:
```bash
pip install mitosheet mito-ai
```

**After**: Single command installs everything:
```bash
pip install mito
```

This simplifies:
- 📖 **Documentation** - Single installation instruction
- 🎯 **User Experience** - One package name to remember
- 🔄 **Version Management** - Coordinated releases
- 📊 **Analytics** - Unified download metrics

# Testing

## ✅ Core Functionality Testing

- [ ] **Basic Installation**: `pip install mito` successfully installs both dependencies
- [ ] **Dependency Resolution**: Verify correct versions of mito-ai and mitosheet are installed
- [ ] **Import Testing**: Confirm `import mitosheet` and `import mito_ai` work after metapackage installation
- [ ] **Jupyter Integration**: Test both packages work correctly in JupyterLab and Jupyter Notebook
- [ ] **VS Code Integration**: Verify functionality in VS Code with Jupyter extension

## 🏗️ **Build & Deployment Testing**

- [ ] **Hatchling Build**: `python -m build` creates correct wheel and sdist
- [ ] **Package Contents**: Wheel contains only metadata (no Python source code)
- [ ] **PyPI Upload**: Test deployment to TestPyPI first
- [ ] **Version Constraints**: Verify dependency ranges work correctly

## 🤖 **CI/CD Workflow Testing**

- [ ] **Manual Trigger**: Test `workflow_dispatch` with and without force-deploy
- [ ] **Version Detection**: Verify script correctly identifies version differences  
- [ ] **CalVer Generation**: Test version generation logic for current and future months
- [ ] **PyPI API Integration**: Confirm API calls work for version checking
- [ ] **Git Commits**: Verify automated commits work correctly

## 🔄 **Real-World Scenarios**

- [ ] **Fresh Environment**: Test installation in clean Python environment
- [ ] **Existing Dependencies**: Test with pre-existing mito-ai or mitosheet installations
- [ ] **Version Conflicts**: Test behavior with conflicting version requirements
- [ ] **Large Datasets**: Verify both packages work with substantial data after metapackage install
- [ ] **Multiple Sessions**: Test repeated use in same notebook/environment

# Documentation

### **Installation Documentation**
- [ ] Update main README.md with simplified installation instructions
- [ ] Add metapackage section to documentation website
- [ ] Update getting-started tutorials to use `pip install mito`

### **New Documentation**
- [ ] Add metapackage README.md explaining purpose and versioning
- [ ] Document CalVer versioning scheme for users
- [ ] Create migration guide for existing separate installations

### **Developer Documentation** 
- [ ] Document CI/CD workflow for maintainers
- [ ] Add troubleshooting guide for metapackage issues
- [ ] Update release process documentation

## 🔗 **Documentation Links to Review**
- Installation pages across docs.trymito.io
- PyPI package descriptions
- GitHub repository README files
- Tutorial notebooks and examples

---

## 🚀 **Next Steps After Merge**

1. **Coordination**: Ensure PyPI tokens are configured for automated deployment
2. **Communication**: Announce metapackage availability to users
3. **Monitoring**: Watch for any installation issues or user feedback
4. **Documentation**: Complete all documentation updates listed above

This metapackage represents a significant improvement in user experience while maintaining full backward compatibility. Users can continue using individual packages or adopt the convenient metapackage approach.

### DIFF
--- a/mito/README.md
+++ b/mito/README.md
@@ -1,0 +1,25 @@
+# Mito
+
+A convenience metapackage that installs both `mitosheet` and `mito-ai` packages.
+
+## Installation
+
+```bash
+pip install mito
+```
+
+This automatically installs:
+
+- **mitosheet** - Interactive spreadsheet for pandas DataFrames in Jupyter notebooks
+- **mito-ai** - AI-powered data analysis and code generation tools
+
+## Versioning
+
+This metapackage uses [Calendar Versioning (CalVer)](https://calver.org/) with the `YYYY.MM[.PATCH]` scheme:
+
+- **YYYY.MM** (e.g., `2025.06`) - First bundle in a month
+- **YYYY.MM.PATCH** (e.g., `2025.06.1`) - Additional bundles or hot-fixes in the same month
+- **Pre-releases** (e.g., `2025.06b1`) - Optional beta/RC versions
+- **Post-releases** (e.g., `2025.06.post1`) - Re-uploads with packaging-only fixes
+
+The version indicates when the package was bundled and what dependency versions were current at that time.
--- a/.github/workflows/deploy-mito-metapackage.yml
+++ b/.github/workflows/deploy-mito-metapackage.yml
@@ -1,0 +1,205 @@
+name: Deploy - Mito Metapackage
+on:
+  schedule:
+    # Run daily at 10:00 UTC to check for new versions
+    - cron: '0 10 * * *'
+  workflow_dispatch:
+    # Allow manual triggering
+    inputs:
+      force_deploy:
+        description: 'Force deploy even if no version changes detected'
+        required: false
+        default: false
+        type: boolean
+
+jobs:
+  check-and-deploy:
+    name: Check versions and deploy metapackage
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        python-version: [3.11]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install requests toml build twine hatchling
+          
+      - name: Setup Auth for PyPi
+        run: |
+          echo -e "[distutils]" >> ~/.pypirc
+          echo -e "index-servers =" >> ~/.pypirc
+          echo -e "    pypi" >> ~/.pypirc
+          echo -e "    testpypi" >> ~/.pypirc
+          echo -e "[pypi]" >> ~/.pypirc
+          echo -e "repository = https://upload.pypi.org/legacy/" >> ~/.pypirc
+          echo -e "username = __token__" >> ~/.pypirc
+          echo -e "password = ${{ secrets.PYPI_API_TOKEN }}" >> ~/.pypirc
+          echo -e "" >> ~/.pypirc
+          echo -e "[testpypi]" >> ~/.pypirc
+          echo -e "repository = https://test.pypi.org/legacy/" >> ~/.pypirc
+          echo -e "username = __token__" >> ~/.pypirc
+          echo -e "password = ${{ secrets.TEST_PYPI_API_TOKEN }}" >> ~/.pypirc
+          
+      - name: Check for version updates and deploy
+        run: |
+          cd mito
+          python << 'EOF'
+          import requests
+          import toml
+          import subprocess
+          import sys
+          from datetime import datetime
+          import os
+          
+          def get_latest_pypi_version(package_name):
+              """Get the latest version of a package from PyPI."""
+              try:
+                  response = requests.get(f"https://pypi.org/pypi/{package_name}/json")
+                  response.raise_for_status()
+                  return response.json()["info"]["version"]
+              except Exception as e:
+                  print(f"Error fetching version for {package_name}: {e}")
+                  return None
+          
+          def parse_version_constraint(constraint):
+              """Parse version constraint like '>=0.1.20' to get the minimum version."""
+              if ">=" in constraint:
+                  return constraint.split(">=")[1].strip()
+              elif ">" in constraint:
+                  return constraint.split(">")[1].strip()
+              elif "==" in constraint:
+                  return constraint.split("==")[1].strip()
+              return constraint.strip()
+          
+          def generate_calver():
+              """Generate CalVer version in YYYY.MM format."""
+              now = datetime.now()
+              base_version = f"{now.year}.{now.month:02d}"
+              
+              # Check if this version already exists on PyPI
+              try:
+                  response = requests.get("https://pypi.org/pypi/mito/json")
+                  if response.status_code == 200:
+                      existing_versions = list(response.json()["releases"].keys())
+                      # Find if base version exists and get next patch number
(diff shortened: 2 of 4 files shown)


mito/README.md
```diff
@@ -0,0 +1,25 @@
...
+```bash
+pip install mito
+```
+
+This automatically installs:
+
+- **mitosheet** - Interactive spreadsheet for pandas DataFrames in Jupyter notebooks
+- **mito-ai** - AI-powered data analysis and code generation tools
+
+## Versioning
```
aarondr77, line 16: We can get rid of this since we moved away from calver

deployment/check_versions.py
```diff
@@ -0,0 +1,114 @@
...
+        with open(pyproject_path, 'r') as f:
+            content = f.read()
+            import re
+            match = re.search(r'version = "([^"]*)"', content)
+            return match.group(1) if match else "NOT FOUND"
+    else:
+        return "FILE NOT FOUND"
+
+
+def get_mitosheet_version() -> str:
```
aarondr77, line 38: In our old deployment system, instead of checking the version from the pyproject.toml files, we were instead reading it directly from testpypi or pypi. 

I'm not particularly wed to either approach. What do you think is best?

It was code like this. 

```python
if on_dev:
        url = f"https://test.pypi.org/pypi/{package_name}/json"
    else:
        url = f"https://pypi.org/pypi/{package_name}/json"

    try:
        response = urllib.request.urlopen(url).read().decode()
        data = json.loads(response)
        return data['info']['version']
```

aarondr77: Oh I guess one benefit of the old approach is that we can deploy often to test_pypi (every time we merge anything into dev) without skipping a bunch of versions on the prod pypi version ...

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,167 @@
...
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git add version.json
+          git add mito/pyproject.toml
+          git add mitosheet/package.json
+          git add mito-ai/package.json
+          git commit -m "Bump version to $NEW_VERSION [skip ci]" || exit 0
+          git push
+          
+      - name: Create Git Tag
```
aarondr77, line 201: This is sweet! I've always been curious how we get those release tags! 

But I think we only want to do this on a release to pypi, not dev pypi, right?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,167 @@
...
+          source venv/bin/activate
+          python -m build --sdist --wheel
+          twine upload dist/*
+          
+      - name: Deploy mito metapackage
+        run: |
+          cd mito
+          rm -rf dist/
+          python -m build --sdist --wheel
+          twine upload dist/*
```
aarondr77: It looks like all of these are missing the logic to decide if we want to deploy to testpypi (when we merge into dev) or pypi (when we merge into main)

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,167 @@
...
+          source venv/bin/activate
+          python setup.py sdist bdist_wheel
+          twine upload dist/*
+          
+      - name: Deploy mito-ai
+        run: |
+          cd mito-ai
+          source venv/bin/activate
+          python -m build --sdist --wheel
+          twine upload dist/*
```
aarondr77: I could be totally wrong on this, but I think this is not the correct build commands for a hatchling project? In the deploy_hatch.py file you can see our current process uses the command `python3 -m hatchling build`

aarondr77: the mito-ai and mito packages are hatchling build and the mitosheet package is still setuptools

.github/workflows/deploy.yml
aarondr77: To deploy to test pypi, we need this `twine upload -repository testpypi dist/*`.

aarondr77:
Oh and also, if we end up not using the `deploy_hatch.py` and `deploy.py` files anymore, let's get rid of them

.github/workflows/deploy.yml
ktaletsk: Yes, I am addressing this now

.github/workflows/deploy.yml
ktaletsk: Yes, I made sure to only create tags on main branch merge / pypi release

.github/workflows/deploy.yml
ktaletsk: Thanks for taking a look!  
`python -m build` is the **PyPA-recommended PEP 517 frontend**. It still invokes **Hatchling** for `mito-ai`/`mito`, and **setuptools** for `mitosheet`; the only thing that changes is *how* the backend is called.

| Aspect | `python -m hatchling build` | `python -m build --sdist --wheel` |
|--------|----------------------------|-----------------------------------|
| Who drives the build? | Hatchling CLI only | PyPA **build** (backend-agnostic) |
| Environment | Uses the current Python env | Creates an isolated temp venv, installs `[build-system].requires`, then cleans up |
| Artifacts | Wheel + sdist (since Hatchling ≥ 1.6) | Same artefacts, produced directly from the source tree |
| Cross-backend consistency | Works for Hatchling projects only | Works for **Hatchling *and* setuptools** out of the box – ideal for our mixed repo |
| Reproducibility | Depends on whatever is already installed | Deterministic: every run starts from a clean slate |

### Why switch?

1. **One command for every package** – Hatchling and setuptools alike. No special-casing `mitosheet` down the road.  
2. **Clean, repeatable CI builds** – the temp venv guarantees we don’t leak dev-env cruft into published wheels.  
3. **Future-proof & community standard** – `build` is the tool used in PyPI’s own docs and by most modern projects (NumPy, pandas, JupyterLab, etc.).  
4. **Zero change to the resulting artifacts** – you’ll get the exact same `.whl` and `.tar.gz` that `deploy_hatch.py` produces today.

If there’s no strong need to pin ourselves to the backend-specific CLI, I’d love to standardise on:

```bash
python -m build --sdist --wheel

deployment/check_versions.py
ktaletsk: This should be fixed now:
- deploy to test PyPI from dev branch
- build adds `.dev<timestamp>` to the version to create unique developer versions

PR #1735, opened 2025-06-04
More DB Support 

ngafar:
# Description

Added support for PostgreSQL and SQLite.

# Testing

Connect to a Postgres and SQLite db.

A sample Postgres db can be launched with:

```bash
docker-compose -f mito_ai/tests/docker/postgres.yml up
```

When you're done, stop and remove the container and its volumes with:

```bash
docker-compose -f mito_ai/tests/docker/postgres.yml down -v
```

An SQLite db is also included in`/mito_ai/tests/data`

# Documentation


mito-ai/README.md
```diff
@@ -151,7 +151,16 @@ To run the performance tests, just run `python -m pytest mito_ai/tests/performan
 
 Note that you'll have to edit `open_ai_utils.py`, specifically the `is_running_test` condition.
 
+#### Running Databases
 
+To ensure reproducibility databases, like Postgres, are created using Docker. To run:
```
aarondr77:
```suggestion
To ensure reproducibility, databases, like Postgres, are created using Docker. To run:
```

mito-ai/mito_ai/db/crawlers/snowflake.py
```diff
@@ -2,84 +2,83 @@
...
-            metadata[db][schema][table].append(column_name)
-
-    return metadata
-
-
-def crawl_snowflake(username: str, password: str, account: str, warehouse: str):
+
+class ColumnInfo(TypedDict):
+    name: str
+    type: str
```
aarondr77: This is duplicated with Postgress, can we make it a shared model + any other shared models

mito-ai/mito_ai/tests/db/test_db_constants.py
```diff
@@ -9,10 +9,28 @@
 CONNECTIONS_PATH: Final[str] = os.path.join(DB_DIR_PATH, "connections.json")
 SCHEMAS_PATH: Final[str] = os.path.join(DB_DIR_PATH, "schemas.json")
 BACKUP_DB_DIR_PATH: Final[str] = os.path.join(MITO_FOLDER, "db_backup")
-CONNECTION_JSON = {
+POSTGRES = {
```
aarondr77: Maybe these could be called `POSTGRES_CONNECTION_DETAILS` or something like that?

mito-ai/mito_ai/tests/settings/settings_test.py
```diff
@@ -55,14 +55,14 @@ def test_get_settings_with_auth(jp_base_url):
...
-def test_get_settings_with_no_auth(jp_base_url):
+def test_get_settings_with_no_auth(jp_base_url: str) -> None:
     response = requests.get(
         jp_base_url + f"/mito-ai/settings/{SETTINGS_KEY}",
     )
     assert response.status_code == 403  # Forbidden
 
 
-def test_get_settings_with_incorrect_auth(jp_base_url):
+def test_get_settings_with_incorrect_auth(jp_base_url: str) -> None:
```
aarondr77, line 65: thanks! :)

mito-ai/mito_ai/utils/utils.py
```diff
@@ -27,6 +30,38 @@ def is_running_test() -> bool:
...
+    
+    for package in packages:
+        try:
+            subprocess.check_call([sys.executable, "-m", "pip", "install", package])
+        except subprocess.CalledProcessError as e:
+            result['success'] = False
+            result['error'] = str(e)
+    
+    return result
+
```
aarondr77, line 64: One question about this is if this python code always runs in the same environment as the kernel. 

@ktaletsk do you have context on this? The goal is to detect if the user has the required Python driver for the database that they are connecting to. If they do, then we don't need to do anything. If they do not, then we try to install it for them.

mito-ai/src/Extensions/SettingsManager/database/ConnectionForm.tsx
```diff
@@ -35,6 +36,46 @@ export const ConnectionForm: React.FC<ConnectionFormProps> = ({
         }
     };
 
+    const renderField = (field: DatabaseField) => {
```
aarondr77, line 39: This is sweet! Nice work 👍

mito-ai/src/Extensions/SettingsManager/database/ConnectionList.tsx
```diff
@@ -41,9 +41,20 @@ export const ConnectionList: React.FC<ConnectionListProps> = ({
                         <h3>{connection.type}</h3>
                     </div>
                     <div className="connection-details">
-                        <p><strong>Username:</strong> {connection.username}</p>
-                        <p><strong>Account:</strong> {connection.account}</p>
-                        <p><strong>Warehouse:</strong> {connection.warehouse}</p>
+                        {databaseConfigs[connection.type]?.fields.map(field => {
+                            // Skip password fields for security
+                            if (field.type === 'password') return null;
```
aarondr77, line 46: Maybe this would be a good jest test to make sure we don't break it in the future? I agree we don't want to be showing the password field!

mito-ai/src/Extensions/SettingsManager/database/model.ts
```diff
@@ -5,12 +5,115 @@
...
+export const databaseConfigs: Record<string, DatabaseConfig> = {
+    postgres: {
+        type: 'postgres',
+        displayName: 'PostgreSQL',
+        fields: [
+            {
+                name: 'username',
+                type: 'text',
+                label: 'Username',
+                placeholder: 'postgres',
```
aarondr77: This isn't a good placeholder for the username field. Maybe `'john.doe'` again. Or you could make some little easter egg lol

mito-ai/src/Extensions/SettingsManager/database/model.ts
```diff
@@ -5,12 +5,115 @@
...
+                type: 'text',
+                label: 'Host',
+                placeholder: 'localhost',
+                required: true
+            },
+            {
+                name: 'port',
+                type: 'number',
+                label: 'Port',
+                placeholder: '5432',
```
aarondr77, line 60: Nice. 5432 is apparently the default port. 

This made me realize that it could be nice to have a `subtext` field that we could optionally add. You can see an example of this on the `beta_mode` field. For fields like `warehouse` and `port` it would be nice to provide some extra help to the user since we're seeing them struggle with things like this. 

It could even link out to our documentation for extra help.

mito-ai/src/Extensions/SettingsManager/database/model.ts
ngafar: The default username for PostgreSQL is typically "postgres." But I can change for clarity.

mito-ai/src/Extensions/SettingsManager/database/model.ts
ngafar: This is a great idea. I'm going to add this as a ticket. Most of the infra is there, but it does require a bit of css adjustment.

ngafar: https://github.com/mito-ds/mito/issues/1747

mito-ai/mito_ai/utils/utils.py
ngafar: I tried this logic in a few virtual envs to test it out, and the packages seem to be contained to the active env. However, also curious to get your take Konstantin

PR #1740, opened 2025-06-04
Apply when code is generated

aarondr77:
# Description

Updates the taskpane so that it does not show the `Overwrite` or `Copy` button until the entire code cell is generated by the AI. This is useful for streaming when where we don't want the user to apply the code to the cell before it's done generating. If they do, it leads to pretty confusing behavior. 

I saw a user do this yesterday where they:
1. Pressed apply before the code was done generating
2. Ran the code and it errored
3. Pressed the `fix in ai chat` button
4. Got the AI response and the the original code was still rendering at the same time, leading to a bunch of unstyled code that was just very hard for them to understand what was going on. 

Note: Sometimes the Ai generated code is displayed in markdown instead of the code cell until the entire code cell is done rendering. We should fix that in a future PR too. 

# Testing

Try it out + see the new jest tests. 

# Documentation

No

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
@@ -127,6 +127,11 @@ const ChatMessage: React.FC<IChatMessageProps> = ({
         return <></>
     }
 
+    // While the code is streaming back we don't want to show the overwrite button. 
+    // Users end up applying the code in the middle of streaming and it gets very confusing
+    // very quickly for users. 
+    let isCodeComplete = false;
+
     return (
         <div className={classNames(
             "message",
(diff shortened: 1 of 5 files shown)


mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
sid-sagacollab, line 1: If there are multiple code blocks, only the last one determines the final value. If this is the approach we are going for, then no worries at all. But I just want to highlight this.

sid-sagacollab:
```
// Alternate approach to calculate completion per code block in a response
{messageContentParts.map((messagePart, index) => {
    if (messagePart.startsWith(PYTHON_CODE_BLOCK_START_WITHOUT_NEW_LINE)) {
        const isThisCodeComplete = messagePart.endsWith('```');
        
        return (
            <>
                <CodeBlock
                    key={index + messagePart}
                    code={messagePart}
                    isCodeComplete={isThisCodeComplete}
                    // ... other props
                />
                
                {/* Only show buttons after the LAST code block that is complete */}
                {isLastAiMessage && isThisCodeComplete && 
                 index === getLastCodeBlockIndex(messageContentParts) &&
                 codeReviewStatus === 'chatPreview' && 
                    <div className='chat-message-buttons'>
                        {/* buttons */}
                    </div>
                }
            </>
        );
    }
    // ... handle non-code blocks
})}

```

ngafar (approved):

mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
aarondr77: Good callout. I think there's a bug right now where the if the chat response has multiple components, the overwrite button always applies the final code cell...

PR #1741, opened 2025-06-05
Gemini bug fix: Flattening string and view cell output issues

sid-sagacollab:
# Description

This addresses the issue of gemini models not being able to see the cell output and also the flat string that was passed into the gemini context.

Uploading Screen Recording 2025-06-04 at 5.02.23 PM.mov…

# Testing

# Documentation

N/A

### DIFF
--- a/mito-ai/mito_ai/gemini_client.py
+++ b/mito-ai/mito_ai/gemini_client.py
@@ -3,11 +3,14 @@
 
 from typing import Any, AsyncGenerator, Callable, Dict, List, Optional, Union
 from google import genai
+from google.genai import types
 from mito_ai.completions.models import AgentResponse, CompletionItem, CompletionReply, CompletionStreamChunk, \
     ResponseFormatInfo, MessageType
-from mito_ai.utils.gemini_utils import get_gemini_completion_from_mito_server, stream_gemini_completion_from_mito_server, get_gemini_completion_function_params
+from mito_ai.utils.gemini_utils import get_gemini_completion_from_mito_server, \
+    stream_gemini_completion_from_mito_server, get_gemini_completion_function_params
 
 INLINE_COMPLETION_MODEL = "gemini-2.0-flash-lite"
+
 
 class GeminiClient:
     def __init__(self, api_key: Optional[str], model: str):
@@ -17,40 +20,49 @@ class GeminiClient:
             self.client = genai.Client(api_key=api_key)
 
     async def request_completions(
-        self, 
-        messages: List[Dict[str, Any]], 
-        response_format_info: Optional[ResponseFormatInfo] = None,
-        message_type: MessageType = MessageType.CHAT
+            self,
+            messages: List[Dict[str, Any]],
+            response_format_info: Optional[ResponseFormatInfo] = None,
+            message_type: MessageType = MessageType.CHAT
     ) -> str:
         try:
             contents = self.convert_messages_for_gemini(messages)
-            config = None
-            if response_format_info:
-                config = {
-                    "response_mime_type": "application/json",
-                    "response_schema": AgentResponse.model_json_schema()
-                }
+
             # Build provider_data once
             provider_data = get_gemini_completion_function_params(
                 model=self.model if not response_format_info else INLINE_COMPLETION_MODEL,
                 contents=contents,
                 message_type=message_type,
-                config=config,
                 response_format_info=response_format_info
             )
+
             if self.api_key:
                 # Unpack provider_data for direct API call
-                response = self.client.models.generate_content(
-                    model=provider_data["model"],
-                    contents=provider_data["contents"]
-                )
+                # Extract the generation config if it exists
+                generation_config = provider_data.get("generation_config")
+
+                if generation_config:
+                    # Call with generation config for structured output
+                    response = self.client.models.generate_content(
+                        model=provider_data["model"],
+                        contents=provider_data["contents"],
+                        config=types.GenerateContentConfig(
+                            response_mime_type=generation_config.get("response_mime_type"),
+                            response_schema=generation_config.get("response_schema")
+                        )
+                    )
+                else:
+                    # Call without generation config for regular responses
+                    response = self.client.models.generate_content(
+                        model=provider_data["model"],
+                        contents=provider_data["contents"]
+                    )
             else:
                 # Only pass provider_data to the server
                 return await get_gemini_completion_from_mito_server(
                     model=provider_data["model"],
                     contents=provider_data["contents"],
                     message_type=message_type,
-                    config=config,
                     response_format_info=response_format_info
                 )
 
@@ -75,7 +87,7 @@ class GeminiClient:
                 for chunk in response:
                     if isinstance(chunk, str):
                         collected_response += chunk or ''
-                    elif hasattr(chunk, 'text') and chunk.text: # type: ignore[attr-defined]
+                    elif hasattr(chunk, 'text') and chunk.text:  # type: ignore[attr-defined]
                         collected_response += chunk.text or ''  # type: ignore[attr-defined]
                     else:
                         collected_response += str(chunk)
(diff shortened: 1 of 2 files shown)


mito-ai/mito_ai/gemini_client.py
```diff
@@ -146,26 +158,25 @@ async def stream_completions(
...
                     accumulated_response += chunk_text or ''
 
             return accumulated_response
 
         except Exception as e:
             return f"Error streaming content: {str(e)}"
-            
-            
+
     def convert_messages_for_gemini(self, messages: List[Dict[str, Any]]) -> str:
```
aarondr77: Sorry if I was unclear, but updating this function was what I meant by "Add Gemini system prompt (instead of flattening entire conversation into one string, send the list of messages)"

Right now, we're flattenting the entire list of messages down to one string. Instead, Gemini expects that we pass a list of messages!

So we should build a function [like this](https://github.com/mito-ds/mito/blob/8f44d5336bd8b91b7da6d94a8e38ff21089dbcd4/mito-ai/mito_ai/anthropic_client.py#L50) that converts the OpenAI messages into the [Gemini format](https://stackoverflow.com/a/78545506). There's another example of someone building a function like this here: https://www.googlecloudcommunity.com/gc/AI-ML/Implementing-System-Prompts-in-Gemini-Pro-for-Chatbot-Creation/m-p/712132

It looks like there are some important differences, like (I think this is not an exhaustive list)
1. Instead of the AI's role being called 'assistant', it's called 'model'
2. Instead of the field being called 'contents', it is nested in the 'parts' list
3. Instead of the system message being in the chat, its elevated to a [specific param](https://cloud.google.com/vertex-ai/generative-ai/docs/learn/prompts/system-instructions) (like anthropic)
4. We need to make sure that we're passing the base64encodedImaged in the right way


Definetly confirm this is all right, I just did a little bit of googling. But we should definetly share the model history in the way that model has been trained to receive it!

mito-ai/mito_ai/utils/gemini_utils.py
```diff
@@ -34,9 +35,15 @@ def _prepare_gemini_request_data_and_headers(
...
-        # Ensure config is serializable
-        inner_data["config"] = json.loads(json.dumps(config))
+
+        # Add generation config for structured output
+        # Convert AgentResponse schema to Gemini's expected format
+        schema = _convert_pydantic_to_gemini_schema(AgentResponse)
+        inner_data["generation_config"] = {
+            "response_mime_type": "application/json",
+            "response_schema": schema
+        }
```
aarondr77: How come in the `get_gemini_completion_function_params` we are able to just use the AgentResposne schema directly, but here we need to convert the pydantic schema into a gemini specific schema? Two points here: 

1. If I understand [this documentation correctly](https://ai.google.dev/gemini-api/docs/structured-output), Gemini supports pydantic models directly. If we can do that, let's definetly do that instead of having to do this complex conversion code. 
2. It looks like this function is duplicating a lot of the work that is done by the `get_gemini_completion_function_params` function. Can we just use that function for things like building the `inner data`?

sid-sagacollab: 1. Yes we can use the pydantic model directly, but if we do that, we will not be able to pass it to the mito server in an API request, so its much easier to convert it into a json and pass it like that.
2. Will do

mito-ai/mito_ai/gemini_client.py
```diff
@@ -1,56 +1,91 @@
...
+        Returns a sequence of properly formatted messages for Gemini, including the system prompt as the first message if present.
+        """
+        gemini_messages: List[Content] = []
+        for message in messages:
+            if not isinstance(message, dict) or 'content' not in message:
+                continue
+            role = message.get('role', 'user')
+            # Convert system messages to user role as Gemini doesn't support system role
+            if role == 'system':
+                role = 'user'
```
aarondr77, line 35: Gemini does support system messages, just not in the `messages` field. Take a look at this https://cloud.google.com/vertex-ai/generative-ai/docs/learn/prompts/system-instructions. 

This is similar to how we handle it for the anthropic client, like our code here:

```python
def _get_system_prompt_and_messages(messages: List[ChatCompletionMessageParam]) -> Tuple[Union[str, anthropic.NotGiven], List[MessageParam]]:
    """
    Convert a list of OpenAI messages to a list of Anthropic messages.
    """
    anthropic_messages: List[MessageParam] = []
    system_prompt: Union[str, anthropic.NotGiven] = anthropic.NotGiven()
    
    for message in messages:
        if 'content' not in message:
            continue
        
        # We assume that the converastion only has one system message.
        # Or if there are multiple, we take the last one.
        if message['role'] == 'system':
            system_prompt = str(message['content'])

        # Construct the messages for the user and assistant in Anthropic format.
        if message['role'] == 'user':
            anthropic_messages.append(MessageParam(role='user', content=str(message['content'])))
        elif message['role'] == 'assistant':
            anthropic_messages.append(MessageParam(role='assistant', content=str(message['content'])))

    return system_prompt, anthropic_messages

```

mito-ai/mito_ai/gemini_client.py
```diff
@@ -19,19 +20,41 @@ def __init__(self, api_key: Optional[str], model: str):
         if api_key:
             self.client = genai.Client(api_key=api_key)
 
+    def convert_messages_for_gemini(self, messages: List[Dict[str, Any]]) -> Sequence[Content]:
```
aarondr77, line 23: Instead of returning a Sequence, can we just return a List so we don't need to cast it later?

mito-ai/mito_ai/utils/gemini_utils.py
```diff
@@ -17,15 +18,18 @@
...
-        contents: str,
+        contents: Sequence[Content],
         message_type: MessageType,
         response_format_info: Optional[Any] = None,
+        system_prompt: Optional[str] = None,
 ) -> Tuple[Dict[str, Any], Dict[str, str]]:
     inner_data = {
         "model": model,
-        "contents": contents,
+        "contents": [{"role": c.role, "parts": [{"text": p.text} for p in c.parts if p.text is not None]} for c in contents if c.parts],
```
aarondr77, line 28: Why do we need to do this? The `convert_messages_for_gemini` function should be getting the messages into the right format so we don't have to do any more preprocessing of it, right?

mito-ai/mito_ai/utils/gemini_utils.py
```diff
@@ -227,19 +235,22 @@ async def wait_for_fetch() -> None:
 
 def get_gemini_completion_function_params(
         model: str,
-        contents: str,
+        contents: Sequence[Content],
```
aarondr77, line 238: These should all be changed to List[Content] I think

mito-ai/mito_ai/utils/gemini_utils.py
```diff
@@ -45,14 +54,77 @@ def _prepare_gemini_request_data_and_headers(
     headers = {"Content-Type": "application/json"}
     return data, headers
 
+
+def _convert_pydantic_to_gemini_schema(pydantic_model) -> Dict[str, Any]:
```
aarondr77, line 58: We're no longer using this function, nice! So we can delete it.

mito-ai/mito_ai/utils/gemini_utils.py
```diff
@@ -227,19 +235,22 @@ async def wait_for_fetch() -> None:
 
 def get_gemini_completion_function_params(
         model: str,
-        contents: str,
+        contents: Sequence[Content],
         message_type: MessageType,
         response_format_info: Optional[Any] = None,
+        system_prompt: Optional[str] = None,
```
aarondr77, line 241: It doesn't look like we are ever providing the `system_prompt` param right now... but once we update the `convert_messages_for_gemini` function to handle the system messages properly, then we will probably want to provide the system message here.

mito-ai/mito_ai/utils/gemini_utils.py
```diff
@@ -227,19 +235,22 @@ async def wait_for_fetch() -> None:
...
 ) -> Dict[str, Any]:
     """
     Build the provider_data dict for Gemini completions, mirroring the OpenAI/Anthropic approach.
     Only includes fields needed for the Gemini API.
     """
     provider_data = {
         "model": model,
-        "contents": contents,
+        "contents": [{"role": c.role, "parts": [{"text": p.text} for p in c.parts if p.text is not None]} for c in contents if c.parts],
```
aarondr77, line 249: same thing here as above

aarondr77 (changes requested):

PR #1751, opened 2025-06-09
Gemini String flattening issue fix

sid-sagacollab:
# Description
This is a PR to fix the flat string being passed to the gemini client without specifying roles.

# Testing

# Documentation

N/A

### DIFF
--- a/mito-ai/mito_ai/gemini_client.py
+++ b/mito-ai/mito_ai/gemini_client.py
@@ -1,10 +1,10 @@
 # Copyright (c) Saga Inc.
 # Distributed under the terms of the GNU Affero General Public License v3.0 License.
 import base64
-from typing import Any, AsyncGenerator, Callable, Dict, List, Optional, Union, Tuple, Sequence
+from typing import Any, Callable, Dict, List, Optional, Union, Tuple
 from google import genai
 from google.genai import types
-from google.genai.types import GenerateContentConfig, Content, Part
+from google.genai.types import GenerateContentConfig
 from mito_ai.completions.models import AgentResponse, CompletionItem, CompletionReply, CompletionStreamChunk, \
     ResponseFormatInfo, MessageType
 from mito_ai.utils.gemini_utils import get_gemini_completion_from_mito_server, stream_gemini_completion_from_mito_server, get_gemini_completion_function_params
@@ -96,15 +96,17 @@ class GeminiClient:
         try:
             config = None
 
-            # Extract system instructions and Gemini-compatible contents
+            # Extract system instructions and contents
             system_instructions, contents = extract_system_instruction_and_contents(messages)
 
+            # Configure response format if provided
             if response_format_info:
                 config = {
                     "response_mime_type": "application/json",
                     "response_schema": AgentResponse.model_json_schema()
                 }
 
+            # Get provider data for Gemini completion
             provider_data = get_gemini_completion_function_params(
                 model=self.model if not response_format_info else INLINE_COMPLETION_MODEL,
                 contents=contents,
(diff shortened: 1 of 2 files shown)


mito-ai/mito_ai/gemini_client.py
```diff
@@ -1,14 +1,86 @@
...
+                        text_content = item.get("text", "")
+                        if text_content:
+                            parts.append({"text": text_content})
+                    elif item.get("type") == "image_url":
+                        image_url_data = item.get("image_url", {})
+                        image_url = image_url_data.get("url", "")
+
+                        if image_url:
+                            # Handle base64 data URLs
+                            if image_url.startswith("data:"):
```
aarondr77:
```suggestion
                        if image_ur and  image_url.startswith("data:"):
```

aarondr77 (changes requested):
I'm getting this error every time I try to use Gemini. I know that you're doing a lot of manual testing, but I think your efforts would be better spent on creating pytests. I think a good method for a PR like this would be to: figure out what the correct response type should be, create pytests for it, then implement the function. 

Doing this test-driven development strategy will probably: 1) Save you a ton of time in development because you can just run your pytests and test all of the edge cases after each change without having to spend time manually testing 2) Make it more easier for Cursor agent to write some of the implementation code for you 3) Save us a lot of time in the review cycles because the first round of reviews won't be about bug catching and we can skip straight to the code review part. 

Ideally, we are only doing 1 review per PR. 

Screenshot 2025-06-10 at 11 49 22 AM

mito-ai/mito_ai/gemini_client.py
```diff
@@ -17,71 +88,80 @@ def __init__(self, api_key: Optional[str], model: str):
...
+                        config=GenerateContentConfig(
+                            system_instruction=""   # No system instruction needed for inline completions
+                        )
+                    )
             else:
-                # Only pass provider_data to the server
                 return await get_gemini_completion_from_mito_server(
                     model=provider_data["model"],
-                    contents=provider_data["contents"],
+                    contents=str(contents),  # Use the processed contents instead of str(messages)
```
aarondr77: In the openai_client, we just leave this as the list of contents and then then the `get_ai_completion_from_mito_server` function just converts the entire request body like this `body=json.dumps(data), `

Do we need to convert it to a string here because `types.Part.from_bytes` is not JSON serializable or something? If so, can you document that here.

mito-ai/mito_ai/gemini_client.py
```diff
@@ -17,71 +88,80 @@ def __init__(self, api_key: Optional[str], model: str):
...
+                        )
+                    )
+                else:
+                    response = self.client.models.generate_content(
+                        model=provider_data["model"],
+                        contents=contents,
+                        config=GenerateContentConfig(
+                            system_instruction=""   # No system instruction needed for inline completions
+                        )
+                    )
```
aarondr77: Would this simplify things?

Screenshot 2025-06-10 at 4 08 00 PM

mito-ai/mito_ai/tests/test_gemini_client.py
```diff
@@ -0,0 +1,53 @@
...
+
+import pytest
+from mito_ai.gemini_client import extract_system_instruction_and_contents
+
+# Dummy base64 image (1x1 PNG)
+DUMMY_IMAGE_DATA_URL = (
+    "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mP8/wcAAgMBAp9l9AAAAABJRU5ErkJggg=="
+)
+
+def test_mixed_text_and_image():
```
aarondr77, line 12: yay! Thanks for creating these!

aarondr77 (approved):
Feel free to implement these small changes if they seem reasonable and then merge it in.

PR #1752, opened 2025-06-09
MySQL Support in Mito AI

ngafar:
# Description

Added support for MySQL connections in Mito AI. 

# Testing

To create a mysql database, run the following command:

```bash
docker-compose -f mito_ai/docker/mysql/compose.yml up
```

You can then use the credentials from the compose file. 

# Documentation

### DIFF
--- a/mito-ai/mito_ai/db/utils.py
+++ b/mito-ai/mito_ai/db/utils.py
@@ -120,7 +120,7 @@ def crawl_and_store_schema(
         conn_str = f"sqlite:///{connection_details['database']}"
         schema = base_crawler.crawl_db(conn_str, "sqlite")
     elif connection_details["type"] == "mysql":
-        conn_str = f"mysql+mysqlconnector://{connection_details['username']}:{connection_details['password']}@{connection_details['host']}:{connection_details['port']}/{connection_details['database']}"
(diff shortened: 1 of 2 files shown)


mito-ai/mito_ai/db/utils.py
```diff
@@ -119,6 +119,9 @@ def crawl_and_store_schema(
     elif connection_details["type"] == "sqlite":
         conn_str = f"sqlite:///{connection_details['database']}"
         schema = base_crawler.crawl_db(conn_str, "sqlite")
+    elif connection_details["type"] == "mysql":
+        conn_str = f"mysql+mysqlconnector://{connection_details['username']}:{connection_details['password']}@{connection_details['host']}:{connection_details['port']}/{connection_details['database']}"
```
aarondr77: When I had the AI generate code, I ended up with this as the start of my connection string `f"mysql+pymysql://`. I then got the error `pymysql` is not installed. 

I'm pretty happy to use either. But maybe we should add a line to the system prompt that tells the AI which one to use?

aarondr77 (approved):
Awesome work on this!

mito-ai/mito_ai/db/utils.py
ngafar: Whops, this should be resolved now. Alternatively, instead of specifying a mapping for each db driver, we could include a list of install packages in the system prompt. This should be enough for the AI to deduce which driver to use.

PR #1754, opened 2025-06-09
Modifying commands to generate the correct requirements.txt file

PoojasreeDwarakanath:
# Description

This PR includes the change to generate the correct requirements.txt file from the code present in the notebook. 

# Testing

- [ ] I have tested this on real data that is reasonable

# Documentation
The file is generated in 3 steps:
1. Create a temporary directory with the notebook code saved as a .py file
2. Run pipreqs to get the list of imports needed and save it in a requirements.in file
3. Since some package names dont match their canonical names which need to be used for import, match the packages in the requirements.in file with the packages and versions present in the user's terminal

### DIFF
--- a/mito-ai/src/Extensions/AppBuilder/requirementsUtils.tsx
+++ b/mito-ai/src/Extensions/AppBuilder/requirementsUtils.tsx
@@ -46,15 +46,36 @@ with tempfile.TemporaryDirectory() as temp_dir:
     with open(temp_file, "w") as f:
         f.write("""${codeContent}""")
     
-    # Make sure pipreqs is installed
+    # Make sure pipreqs is installed. Then
+    # 1. Create a requirements.in file
+    # 2. From the requirements.in file, generate the requirements.txt file with the canonical PyPI name of the packages
+    # and the versions as they exist on the user's terminal
     try:
         # Run pipreqs on the temporary directory
-        result = subprocess.run(
-            ['pipreqs', '--force', '--savepath', '-', temp_dir],
+        generate_req_in_file = subprocess.run(
+            ['pipreqs', '--savepath', 'requirements.in', '--force', temp_dir],
             capture_output=True, 
             text=True
         )
-        print(result.stdout)
+
+        command_for_generating_req_txt_file = r"""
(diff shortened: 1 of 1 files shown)


mito-ai/src/Extensions/AppBuilder/requirementsUtils.tsx
```diff
@@ -46,15 +46,36 @@ with tempfile.TemporaryDirectory() as temp_dir:
     with open(temp_file, "w") as f:
         f.write("""${codeContent}""")
     
-    # Make sure pipreqs is installed
+    # Make sure pipreqs is installed. Then
```
aarondr77, line 49: Hmm for some reason this doesn't seem to be working for me. Am I doing something wrong?

Here's a zip file with my notebook:
[Untitled (1).ipynb.zip](https://github.com/user-attachments/files/20674430/Untitled.1.ipynb.zip)

https://github.com/user-attachments/assets/75e7f6cd-3ed9-47ab-ac9d-0cb6d0de0750

aarondr77: Can we write some pytests for this function? I guess we would need to have a python file that contains this code and then we can pass in various app.py files and see if it generates the correct requirements.

mito-ai/src/tests/AppBuilder/NotebookToStreamlit.test.tsx
```diff
@@ -0,0 +1,401 @@
...
+    });
+
+    test('should convert a markdown cells correctly in the streamlit app', async () => {
+        // Create a mock MarkdownCell instance with a model getter
+        const mockMarkdownCell = Object.create(MarkdownCell.prototype);
+        Object.defineProperty(mockMarkdownCell, 'model', {
+            get: () => ({
+                type: 'markdown',
+                sharedModel: {
+                    source: '# This is a heading\nSome *italic* text.'
```
aarondr77: It might just be nice to turn this text into a variable and then use it in the source here and in the expect statement so its easy to keep them in sync

PR #1762, opened 2025-06-11
Microsoft SQL Server Support

ngafar:
# Description

Added support for Microsoft SQL Server.

# Testing

To run the server:

```bash
docker-compose -f mito_ai/docker/mssql/compose.yml up
```

You can then setup a database as you normally would. 

The credentials:

```python
username=SA
password=YourStrong!Passw0rd
database=test_db
```

Note that you'll also need to install an additional driver manually. As part of the testing process, please let me know if the directions are too difficult to follow. 

# Documentation

Updated https://docs.trymito.io/mito-ai/database-connectors/database-drivers
Added https://docs.trymito.io/mito-ai/database-connectors/microsoft-sql-server


mito-ai/src/Extensions/SettingsManager/database/model.ts
```diff
@@ -24,10 +24,60 @@ export interface DatabaseField {
...
     displayName: string;
+    alertText?: string;  // Optional HTML content for alerts/notifications
     fields: DatabaseField[];
 }
 
 export const databaseConfigs: Record<string, DatabaseConfig> = {
+    mssql: {
+        type: 'mssql',
+        displayName: 'Microsoft SQL Server',
+        alertText: 'Microsoft SQL Server requires an additional driver to be installed. For more info, consult the <a href="https://docs.trymito.io/mito-ai/database-connectors/microsoft-sql-server" target="_blank">Mito docs</a>.',
```
aarondr77: Once I installed the driver, I was still surprised to see this alert. I don't think we need to add logic now to detect if its there or not, but maybe we can just update the messaging slightly to let users know that they might already have it???

aarondr77 (approved):
Very nice very nice

PR #1768, opened 2025-06-12
Unit tests and Jest tests for the multi modal providers work.

sid-sagacollab:
# Description

- [x]  A basic jest test to make sure the model selector calls the update model on backend function
- [x]  Pytests for the `_extract_and_parse_json_response` for gemini, openai anthropic
- [x]  Pytests for `_prepare_gemini_request_data_and_headers` for gemini, openai, anthropic
- [x]  Basic pytests to make sure that the `request_completions` and `stream_completions` use the correct provider
- [x]  Extract_system_instruction_and_contents for gemini and anthropic

# Testing

# Documentation

N.A

### DIFF
--- a/mito-ai/mito_ai/tests/open_ai_utils_test.py
+++ b/mito-ai/mito_ai/tests/open_ai_utils_test.py
@@ -9,8 +9,8 @@ from mito_ai.utils.server_limits import (
     OS_MONTHLY_AI_COMPLETIONS_LIMIT,
     OS_MONTHLY_AUTOCOMPLETE_LIMIT,
 )
-from mito_ai.utils.telemetry_utils import MITO_SERVER_FREE_TIER_LIMIT_REACHED
 from mito_ai.completions.models import MessageType
+from mito_ai.utils.open_ai_utils import _prepare_request_data_and_headers
 
 REALLY_OLD_DATE = "2020-01-01"
 TODAY = datetime.now().strftime("%Y-%m-%d")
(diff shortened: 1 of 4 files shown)


mito-ai/mito_ai/tests/open_ai_utils_test.py
```diff
@@ -71,3 +71,92 @@ def test_check_mito_server_quota_pro_user() -> None:
...
+            assert data["data"] == {"key": "value"}
+            assert data["user_input"] == "test message"
+
+            # Verify headers
+            assert headers == {"Content-Type": "application/json"}
+
+def test_prepare_request_data_and_headers_null_message() -> None:
+    """Test handling of null message content"""
+    from mito_ai.utils.open_ai_utils import _prepare_request_data_and_headers
+    from mito_ai.completions.models import MessageType
```
aarondr77: We could move these all to the top of the file I think

mito-ai/mito_ai/tests/utils/test_anthropic_utils.py
```diff
@@ -0,0 +1,154 @@
...
+    monkeypatch.setattr("mito_ai.utils.anthropic_utils.get_user_field", mock_get_field)
+
+
+def test_basic_request_preparation():
+    """Test basic request preparation with minimal parameters"""
+    model = "claude-3-sonnet"
+    max_tokens = 100
+    temperature = 0.7
+    # Use NotGiven to ensure system is not included in inner_data
+    system = cast(Tuple[Union[str, anthropic.NotGiven]], (anthropic.NotGiven(),))
```
aarondr77: Can we just set `system = anthropic.NotGiven()`?

aarondr77: Speaking of, why are we setting the type of the system message to `Tuple[Union[str, anthropic.NotGiven]`

Whenever we use the system message, it looks like we are just providing it like `(system_message, )`. The return type of `_get_system_prompt_and_messages` already says that the system_message is `Union[str, anthropic.NotGiven]` (not a tuple). 

So can we update the system_message type to that everywhere for the anthropic client? ie: stream_anthropic_completion_from_mito_server, get_anthropic_completion_from_mito_server, etc.

mito-ai/src/tests/AiChat/ModelSelector.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const mockOnConfigChange = jest.fn();
+
+  beforeEach(() => {
+    // Clear mock calls before each test
+    mockOnConfigChange.mockClear();
+    // Clear localStorage before each test
+    localStorage.clear();
+  });
+
+  it('calls onConfigChange when a model is selected', () => {
```
aarondr77, line 15: Nice!

mito-ai/mito_ai/tests/utils/test_anthropic_utils.py
sid-sagacollab: Fixed this implementation.

aarondr77 (commented):
It looks like the new anthropic unit tests are not passing. Let's fix that and then LGTM.

PR #1769, opened 2025-06-13
Oracle Support

ngafar:
# Description

Added support for Oracle databases.

# Testing

Run

```bash
docker-compose -f mito_ai/docker/oracle/compose.yml up
```

Then try to connect using the following:

```python
ORACLE_CONNECTION_DETAILS = {
    "type": "oracle",
    "username": "test_user",
    "password": "test_pass",
    "host": "localhost",
    "port": "1521",
    "service_name": "xepdb1",
}
```

# Documentation

### DIFF
--- a/mito-ai/src/Extensions/SettingsManager/database/model.ts
+++ b/mito-ai/src/Extensions/SettingsManager/database/model.ts
@@ -119,6 +119,47 @@ export const databaseConfigs: Record<string, DatabaseConfig> = {
             }
         ]
     },
+    oracle: {
+        type: 'oracle',
+        displayName: 'Oracle',
+        fields: [
+            {
+                name: 'username',
+                type: 'text',
+                label: 'Username',
+                placeholder: 'john.doe',
+                required: true
+            },
+            {
+                name: 'password',
+                type: 'password',
+                label: 'Password',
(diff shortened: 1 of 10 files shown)


mito-ai/src/Extensions/SettingsManager/database/model.ts
```diff
@@ -119,6 +119,47 @@ export const databaseConfigs: Record<string, DatabaseConfig> = {
...
+            },
+            {
+                name: 'port',
+                type: 'number',
+                label: 'Port',
+                placeholder: '1521',
+                required: true
+            },
+            {
+                name: 'service_name',
```
aarondr77, line 155: When we have the ability to add subtext to a field, I think this would be a good one to add some subtext for?

aarondr77 (commented):
I'm getting this error when trying to connect to Oracle after starting up the docker container. Any thoughts?

Screenshot 2025-06-14 at 10 25 50 PM

ngafar:
hmm, my best best for this error is that the database was not properly setup in Docker. I can only recreate this error if I change the host name, which probably points to there being an issues during setup, any other discrepancy (service name, port, etc) triggers a more specific error.

PR #1771, opened 2025-06-13
Next steps

aarondr77:
# Description

# Testing

# Documentation


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -416,6 +427,9 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
     const sendSmartDebugMessage = async (errorMessage: string): Promise<void> => {
         // Step 0: Reject the previous Ai generated code if they did not accept it
         rejectAICode()
+        
+        // Clear next steps when sending a new message
+        setNextSteps([])
```
ngafar: Steps 0 and 1 repeat quite a bit, maybe we should break them out into a utility function

mito-ai/src/components/NextStepsPills.tsx
```diff
@@ -0,0 +1,83 @@
...
+    displayedNextStepsIfAvailable: boolean;
+    setDisplayedNextStepsIfAvailable: (displayedNextStepsIfAvailable: boolean) => void;
+}
+
+const NextStepsPills: React.FC<NextStepsPillsProps> = ({ nextSteps, onSelectNextStep, displayedNextStepsIfAvailable, setDisplayedNextStepsIfAvailable }) => {
+    const [isVisible, setIsVisible] = useState(displayedNextStepsIfAvailable);
+    const [isAnimatingOut, setIsAnimatingOut] = useState(false);
+
+    /*
+    Triggers entrance animation for next steps pills.
```
ngafar, line 21: This is a nice touch!

mito-ai/style/NextStepsPills.css
```diff
@@ -0,0 +1,189 @@
...
+.next-steps-pills-container {
+    display: flex;
+    flex-direction: column;
+    gap: 8px;
+    padding: 8px 20px;
+    margin-bottom: 0;
+    border-top-left-radius: 12px;
+    border-top-right-radius: 12px;
+    margin-left: 10px;
+    margin-right: 10px;
```
ngafar, line 22: Screenshot 2025-06-16 at 11 45 28 AM


What do you think about making the next steps div the same size as the input box?

ngafar: Alternatively, we could keep it this smaller size, but add a drop shadow to the front to make it seem like it is sitting behind the input box. This way we can visually seperate the two.

ngafar (approved):

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -113,6 +115,10 @@ const ChatInput: React.FC<ChatInputProps> = ({
             setDropdownVisible(false);
             setDropdownFilter('');
         }
+
+        if (onInputChange) {
```
aarondr77: I don't think we're using this anymore. Delete if so

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
aarondr77: Good call!

PR #1776, opened 2025-06-13
Fixed the flow of data to the mito server for gemini

sid-sagacollab:
# Description

Based on this comment, we are changing the flow of data to the mito server for gemini: https://github.com/mito-ds/mito-private-monorepo/pull/896#discussion_r2145225112
# Testing

# Documentation

N/A


aarondr77 (changes requested):
When testing this with the new backend by setting up the localhost server, I ran into some issues. 

### Error (only one time)

Screenshot 2025-06-15 at 8 03 15 PM

### Parsing issues on refresh every time 

After having a chat with Gemini (this only happens on Gemini) and then refreshing the page, the Json parsing doesn't work the same as it did the first time you receive the message. Notice that all of the new line characters are there, the quotes at the start and end of the interaction, etc. 


https://github.com/user-attachments/assets/0fd21480-0904-411c-a740-b3c224ab9704

mito-ai/mito_ai/gemini_client.py
```diff
@@ -19,6 +19,7 @@ def extract_system_instruction_and_contents(messages: List[Dict[str, Any]]) -> T
     - system_instructions: list of strings (for system_instruction param)
     - contents: list of dicts for Gemini (excluding system role)
     """
+    print(messages)
```
aarondr77: delete

aarondr77 (changes requested):
Nice unicode issues are gone and code looks good, but on Gemini I'm not getting streamed responses. It's all loading in go. 

https://github.com/user-attachments/assets/2aa0d273-f6f1-4bc8-a29b-206ed7dc1844

mito-ai/mito_ai/gemini_client.py
sid-sagacollab: This was removed in [33e010a](https://github.com/mito-ds/mito/pull/1776/commits/33e010a7c5d9f6499fa49adf135f7195c473700d)

PR #1777, opened 2025-06-16
Adding support for sympy

PoojasreeDwarakanath:
# Description

This PR adds support for sympy plots in addition to matplotlib and poorly.

# Testing

### DIFF
--- a/mito-ai/src/Extensions/AppBuilder/visualizationConversionUtils.tsx
+++ b/mito-ai/src/Extensions/AppBuilder/visualizationConversionUtils.tsx
@@ -8,23 +8,31 @@ export const generateDisplayVizFunction = (): string => {
     return `
 def display_viz(fig):
     """Display a visualization in Streamlit based on its type."""
-    
+    import streamlit as st
+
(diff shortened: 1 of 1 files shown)


mito-ai/src/Extensions/AppBuilder/visualizationConversionUtils.tsx
```diff
@@ -8,23 +8,31 @@ export const generateDisplayVizFunction = (): string => {
     return `
 def display_viz(fig):
     """Display a visualization in Streamlit based on its type."""
-    
+    import streamlit as st
```
aarondr77: Streamlit is already imported at the top of the file we are generating, so I don't think we need it here, right?

aarondr77 (approved):
Looks good. I didn't realize how popular sympy is, but it has about 1/2 as many daily installs as matplotlib! So pretty relevant! Glad you caught that. 

This looks good, but since we're making changes to the `transformVisualizationCell` can we just add some new test cases to the jest tests? They should go in `transformVisualizationCell.test.tsx`

Issue #1749, opened 2025-06-05
Stop Agent Immediately


aarondr77:
Currently, when you press `Stop Agent` it doesn't stop until the next Ai message is finished. Effectively, it's just not sending the next message back to the Ai. 

This means that it can sometimes take 10+ seconds from when you press stop for the agent to actually stop. 

To actually stop it, we need to:
1. Turn off agent mode
2. Make sure that we ignore the response from that previous message. We can't stop the frontend from receiving that response, so we need to make sure that when it comes back it's ignored instead of being added to the chat history manager.

PR #1779, opened 2025-06-17
Stop Agent Immediately

sid-sagacollab:
# Description

Intends to solve: https://github.com/mito-ds/mito/issues/1749

Currently, when you press Stop Agent it doesn't stop until the next Ai message is finished. Effectively, it's just not sending the next message back to the Ai.

This means that it can sometimes take 10+ seconds from when you press stop for the agent to actually stop.

To address this issue, I have done the following:

Fix agent stopping functionality to properly handle new agent tasks

- Removed unused shouldIgnoreNextResponse ref that was causing TypeScript errors
- Ensured shouldContinueAgentExecution flag is properly reset at the start of each new agent task
- Simplified the agent stopping logic to use a single check at the start of each execution loop
- Fixed issue where starting a new agent task immediately after stopping would cause it to stop prematurely

This change ensures that the agent can be properly stopped and new agent tasks can be started without interference from previous stop states. 

This should reduce the stop time of the agent significantly.

# Testing

# Documentation


aarondr77 (changes requested):
I don't think this is exactly what we're looking for. In this implementation, I still have to wait for the most recent message to come back before the agent stops. 

In this video you notice two things:
1. After I click `Stop Agent`, I still have to wait for the agent to finish before it returns me to the original state of the taskpane. 
2. The message still gets processed when it returns. 

https://github.com/user-attachments/assets/938618d4-e9bb-476a-adc5-6856c485ed1f

I don't think we can stop the frontend end websocket from receiving the response from the backend, so I think instead we need to figure out how to ignore the messages associated with the stopped request. I'm not exactly sure the best approach to do this, but this isn't a new thing, so I'm sure we can find some best practices online. Can you investigate this?

Let me know if it's unclear what we want the UX to be. To see how it should work, you can try Stopping the Cursor agent. We want the same immediate stopping !

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -795,13 +823,27 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
         // Update UI to show stopping state
         setAgentExecutionStatus('stopping');
     }
 
     const finalizeAgentStop = (): void => {
         // Notify user that agent has been stopped
         shouldContinueAgentExecution.current = false;
+        
+        // Ensure any remaining pending requests are cancelled
+        websocketClient.cancelPendingRequests();
```
aarondr77: I think the only way we can get here is if we already called markAgendForStopping, which already calls `websocketClient.cancelPendingRequests();`. So I think this is overkill

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -795,13 +823,27 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+        
+        // Disconnect any active stream handlers
+        if (streamHandlerRef.current) {
+            websocketClient.stream.disconnect(streamHandlerRef.current, null);
+            streamHandlerRef.current = null;
+        }
```
aarondr77: We currently don't do any streaming in agent mode, so we don't need this for now.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -462,7 +462,12 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
             metadata: agentSmartDebugMessage,
             stream: false
         }
-        await _sendMessageAndSaveResponse(smartDebugCompletionRequest, newChatHistoryManager)
+        const success = await _sendMessageAndSaveResponse(smartDebugCompletionRequest, newChatHistoryManager)
+        
+        // If the request was cancelled, throw an error to break out of the agent execution loop
+        if (!success) {
+            throw new Error('Agent execution cancelled');
+        }
```
aarondr77: Can we move this into the `_sendMessageAndSaveResponse` function so we don't have to handle it everywhere? 

Also, instead of checking just !success are we able to do a more specific check, because I think it could fail for a number of reasons.

sid-sagacollab: Done !

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
sid-sagacollab: Done sir !

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
sid-sagacollab: You're right. I had added this in another flow I was testing initially, getting rid of this

mito-ai/src/websockets/BaseWebsocketClient.ts
```diff
@@ -132,6 +132,20 @@ export abstract class BaseWebsocketClient<RequestType, ResponseType, StreamType
...
   }
 
+  /**
+   * Cancel all pending requests by rejecting them with a cancellation error.
+   * This is useful for immediately stopping agent execution.
+   */
+  public cancelPendingRequests(): void {
+    // Reject all pending requests with a cancellation error
+    for (const [messageId, resolver] of this._pendingRepliesMap.entries()) {
+      resolver.reject(new Error('Request cancelled by user'));
```
aarondr77, line 142: Can we generate a specific Error Type just to make type checking easier? Something like "AgentInteruptedError" or something like that? If not, let's at least turn the error message into a constant so we don't have to keep it in sync manually.

sid-sagacollab: Yes this can be done!

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -675,6 +675,12 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
 
             // Create the stream handler function and store it in the ref
             const streamHandler = (_: CompletionWebsocketClient, chunk: ICompletionStreamChunk): void => {
+                // Safety check: Don't process responses if agent execution has been stopped
+                if (!shouldContinueAgentExecution.current) {
+                    console.log('Ignoring response chunk because agent execution was stopped');
+                    return;
+                }
+
                 if (chunk.error) {
```
sid-sagacollab, line 684: If this streamHandler is registered and then the user clicks on stop and if we do not make a check again for `shouldContinueAgentExecution `, the UI would update even after we stop.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -873,127 +964,147 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
-            // Only the first message sent to the Agent should contain the user's input.
-            // All other messages only contain updated information about the state of the notebook.
-            if (agentExecutionDepth === 1) {
-                await sendAgentExecutionMessage(input, messageIndex, false, selectedRules)
-            } else {
-                await sendAgentExecutionMessage('', undefined, sendActiveCellOutput)
+                // Only the first message sent to the Agent should contain the user's input.
+                // All other messages only contain updated information about the state of the notebook.
+                if (agentExecutionDepth === 1) {
+                    await sendAgentExecutionMessage(input, messageIndex, false, selectedRules)
```
sid-sagacollab, line 979: Also related to my comment above, if we do not add a stopping check in `sendAgentExecutionMessage `, the control would not come back to this until all the chunks from that request are sent and we'd have to wait for the agent to stop again for a long time.

sid-sagacollab (commented):
> I've dropped the ball on getting the Stop Agent immedietly PR reviewed. Sorry!! My one thought on this is that there is a lot of places we are interjecting the logic for checking if the stop has occured and then handling it. And I wonder if there is a better way to do that where we can handle it all in one place and interupt the flow somehow? Relying on those stops in a bunch of different places feels really fragile.
>
> Maybe we could do something like. When the user clicks the stop button: 0. Set shouldContinueAgentExecution.current = false
>
> 1. we drop the message from pending messages in the websocket so we don't process any more requests
> 2. We immediately respond with a UserInteruptMessage right away so that the agent stops waiting and we can hit the shouldContinueAgentExecution.current = false check right away
>
> Would that work?

This in theory should work, but the issue is we need to be at a stable state in order to even set the `shouldContinueAgentExecution.current  = false`, let me attempt to explain what I mean by that in the following.

Right now, with the way our web socket works, if we have sent a request from our front end through the web socket, we await for its response. We do not get the control back into our code until at-least one chunk of response is received. If we were to set this flag `shouldContinueAgentExecution.current  = false ` when this response is being processed from the web socket, it would not actually come into effect as our code is stuck up with the async call. Let me know if this does not make sense, we can touch base on this.

ngafar:
Closing in favor of https://github.com/mito-ds/mito/pull/1971

PR #1780, opened 2025-06-17
Checkpoints in Agent Mode

ngafar:
# Description

Before the Agent starts to work, a checkpoint is created. This checkpoint is a snapshot of of the user's notebook, it's cells and content. Upon the Agent's completion the user is presented with the option to undo these changes, and restore the checkpoint. 

# Testing

Create a notebook with some content. Then send a message in agent mode. Upon completion you should see the option to restore the checkpoint.

Use this button to ensure that the notebook returns to its original state. 

# Documentation

We should call it out in our next release note, otherwise this seems self-explanatory.


ngafar:
There are 6 Jest tests that are failing. However, these are all from `src/tests/NotebookFooter`.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -181,6 +185,17 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
           }
     };
 
+    const handleRestoreCheckpoint = async (): Promise<void> => {
```
aarondr77: Maybe since this file is already so large, we can keep this function in the checkpoint.ts file?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1325,6 +1349,29 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+                    agentExecutionStatus === 'idle' &&
+                    displayOptimizedChatHistory.length > 0 && (
+                        <div className='message message-assistant-chat'>
+                            <button
+                                className="button-base button-gray button-small"
+                                title="Restore from Checkpoint"
+                                onClick={() => void handleRestoreCheckpoint()}
+                                disabled={!hasCheckpoint}
+                            >
+                                ⏪ Restore checkpoint
```
aarondr77: I think `restore to checkpoint` language makes more sense when it's located in the place you are restoring the state to. But it being at the end of the chat is a bit confusing because it doesn't signal what state you are restoring back to. 

So what if we changed this language to: `Revert changes` and then the subtext stays as `Undo the most recent changes made by the agent`

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1325,6 +1349,29 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+                            </button>
+                            <div style={{
+                                marginTop: '5px',
+                                fontSize: '12px',
+                                color: 'var(--jp-ui-font-color3)'
+                            }}>
```
aarondr77: Let's move all the styles to a css class

mito-ai/src/utils/checkpoint.ts
```diff
@@ -0,0 +1,188 @@
...
+
+// Simple in-memory storage for the current session
+let currentCheckpoint: INotebookCheckpoint | null = null;
+
+/**
+ * Saves a checkpoint to storage (currently in-memory for simplicity)
+ */
+export const saveCheckpoint = (checkpoint: INotebookCheckpoint): void => {
+    currentCheckpoint = checkpoint;
+    console.log(`💾 Checkpoint saved: ${checkpoint.checkpointId}`);
```
aarondr77: We can probably delete this, it doesn't seem particularly useful for debugging

mito-ai/src/utils/checkpoint.ts
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Generates a unique checkpoint ID
+ */
+const generateCheckpointId = (): string => {
+    return `checkpoint_${Date.now()}_${Math.random().toString(36).substr(2, 9)}`;
+};
+
+// Simple in-memory storage for the current session
+let currentCheckpoint: INotebookCheckpoint | null = null;
```
aarondr77: Oh interesting. I didn't think of this simplification. 

Okay, so I guess the limitations here are that if the user refreshes their browser at any point after the checkpoint is created, they will lose the checkpoint. 

Right now if you refresh the page, the `restore to checkpoint` button doesn't even remain in the chat.

aarondr77 (commented):
### Stream of Consciousness Thoughts that I think we can ignore , but want to share anyways. 

One other implementation strategy is to use the actual Jupyter checkpointing system. What that would look like is: 

1. Use the create checkpoint function to create the checkpoint https://jupyterlab.readthedocs.io/en/stable/api/classes/services.ContentsManager-1.html#createcheckpoint
2. I think we could still store the checkpoint in memory for now
3. use the restoreCheckpoints function. But it looks like the restore checkpoint function takes a checkpoint ID and then in the Python code it goes and finds that checkpoint id and restores to it. So I guess what we have to do is: when the user presses the restore checkpoint button, we make sure to save the checkpoint as the real checkpoint, then call the restore button.. 

I don't think this is necessarily better... it sounds much more complex, but there is an appeal to using the Jupyter checkpoint system ..

aarondr77:
> There are 6 Jest tests that are failing. However, these are all from src/tests/NotebookFooter.

These tests should pass if you merge dev into this branch

mito-ai/src/utils/checkpoint.ts
```diff
@@ -0,0 +1,188 @@
...
+                source: string;
+                metadata: Partial<ICellMetadata>;
+                outputs?: PartialJSONValue[];
+                execution_count?: number | null;
+            };
+
+            // Add code cell specific data
+            if (cellCheckpoint.cellType === 'code') {
+                cellData.outputs = cellCheckpoint.outputs || [];
+                cellData.execution_count = cellCheckpoint.executionCount;
```
aarondr77: I don't think we're doing anything with these execution counts. We set them, but then they all get updated when we rerun th enotebook from top to bottom anyways, so I think we can delete?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -181,6 +185,17 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
           }
     };
 
+    const handleRestoreCheckpoint = async (): Promise<void> => {
+        const success = await restoreFromCurrentCheckpoint(notebookTracker);
+        if (success) {
+            console.log('🔄 Re-executing code cells to restore kernel state...');
+            await app.commands.execute("notebook:run-all-cells");
```
aarondr77: To restore the state properly I think we first need to restart the kernel, otherwise any variables that were created by the agent will still exists in the global state. So I think this operation should be restart and run all cells (which I think is an available command). 

Note that this doesn't perfectly restore the state because if the user had ran cells in non-top to bottom order, there could be a disrepancy in the output, but I think this is good enough.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -181,6 +185,17 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+            setHasCheckpoint(false);
```
aarondr77: A few other thoughts on what could happen when you restore a checkpoint. Just ideations, what do you think?

1. The Next Steps should be cleared because the next steps are built off code that no longer exists anymore
2. All of the messages in the chat that were undone, should be deleted (or marked as reverted) because the agent might get confused with the discrepancy. 
3. An AI message is appended to the chat that says something to the effect of "I've reverted all previous changes". This replaces the button that was there previously.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1325,6 +1349,29 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
                     </div>
                 }
+                {/* Agent restore button - shows after agent completes and when agent checkpoint exists */}
+                {hasCheckpoint &&
+                    agentModeEnabled &&
+                    agentExecutionStatus === 'idle' &&
+                    displayOptimizedChatHistory.length > 0 && (
+                        <div className='message message-assistant-chat'>
+                            <button
+                                className="button-base button-gray button-small"
```
aarondr77: I think we could maybe make this button look a little bit better? A few ideas: 

1. I like having an icon, but so far emojis are not in our ~design language~. What about something like the vertically flipped version of our submit button? (Similar to Cursor's icon)
2. maybe the padding shout be a little bit more? Or none? We don't want it to be too obtrusive, but also it looks a little bit squished right now.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
ngafar: > if the user had ran cells in non-top to bottom order, there could be a disrepancy in the output

ooh, this is a good catch. I don't have a good solution for this right now, but it is something we should consider for a v2.

mito-ai/src/utils/checkpoint.ts
ngafar: Yeah, definitely a limitation, but I think its ok for v1.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
ngafar: Implemented # 3, though the message is temporary. 

For next steps: what if we had the restore button, but it also asks "what went wrong?" we could then take this feedback, and send another agent message to get it fixed.

This is for sure a v2 feature, but it could be a good way to keep users engaged with the agent.

PR #1781, opened 2025-06-17
Fix non-gemini mypy, jest, and pytests

aarondr77:
# Description

- Fixes a bunch of failing mypy, jest, and pytests. 
- Updates the CI testing workflow so that we should get the same results locally as we do on CI


mito-ai/mito_ai/tests/utils/test_gemini_utils.py
```diff
@@ -9,7 +9,7 @@
 def test_basic_request_preparation():
```
sid-sagacollab, line 16: Let's get rid of this file in favour of my PR that was merged ?

sid-sagacollab (approved):

PR #1782, opened 2025-06-17
Fixed the failing gemini pytest

sid-sagacollab:
# Description

Fixed the failing gemini pytest
# Testing

# Documentation

N/A

### DIFF
--- a/mito-ai/mito_ai/tests/utils/test_gemini_utils.py
+++ b/mito-ai/mito_ai/tests/utils/test_gemini_utils.py
@@ -5,16 +5,22 @@ import pytest
 from mito_ai.utils.gemini_utils import _prepare_gemini_request_data_and_headers
 from mito_ai.completions.models import MessageType
 
+TEST_CONTENTS = [
+    {'role': 'system', 'content': 'You are Mito Data Copilot, an AI assistant for Jupyter.'},
+    {'role': 'user', 'content': 'Help me complete the following task. print 10'},
+    {'role': 'assistant', 'content': 'python\nprint(10)\n\nPrinted the number 10 '},
+    {'role': 'user', 'content': 'Update to print 11'}
+]
+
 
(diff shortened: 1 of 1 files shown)


mito-ai/mito_ai/tests/utils/test_gemini_utils.py
```diff
@@ -24,30 +24,12 @@ def test_basic_request_preparation():
...
-    system_instructions = "You are a helpful assistant"
-
-    data, headers = _prepare_gemini_request_data_and_headers(
-        model=model,
-        contents=contents,
-        message_type=message_type,
-        system_instructions=system_instructions
-    )
-
-    assert data["data"]["system_instructions"] == system_instructions
```
sid-sagacollab, line 44: This test is no longer needed as we are not passing the system instructions separately, instead we are passing the entire `message` object to the mito server and it reconstructs the system instructions in the server.

mito-ai/mito_ai/tests/utils/test_gemini_utils.py
```diff
@@ -9,7 +9,7 @@
 def test_basic_request_preparation():
     """Test basic request preparation with minimal parameters."""
     model = "gemini-pro"
-    contents = "test content"
+    contents = [{"text": "test content"}]
```
aarondr77: Can we make the contents have a realistic type. This is what we're passing in right now: 

[
{'role': 'system', 'content': 'You are Mito Data Copilot, an AI assistant for Jupyter.'},
{'role': 'user', 'content': 'Help me complete the following task.  print 10'}]
{'role': 'assistant', 'content': '```python\nprint(10)\n```\nPrinted the number 10 '}, 
{'role': 'user', content: 'Update to print 11'}
]

aarondr77 (approved):
Nice, just one tiny change then LGTM

PR #1784, opened 2025-06-18
Gemini mypy fixes

sid-sagacollab:
# Description

This is a PR to fix the failing mypy checks

### DIFF
--- a/mito-ai/mito_ai/anthropic_client.py
+++ b/mito-ai/mito_ai/anthropic_client.py
@@ -3,7 +3,7 @@
 
 import json
 import anthropic
-from typing import Dict, Any, Optional, Tuple, Union, Callable, List
+from typing import Dict, Any, Optional, Tuple, Union, Callable, List, cast
 
 from anthropic.types import Message, MessageParam, ToolUnionParam
 from mito_ai.completions.models import ResponseFormatInfo, CompletionReply, CompletionStreamChunk, CompletionItem, AgentResponse, MessageType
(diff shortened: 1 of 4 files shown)


mito-ai/mito_ai/anthropic_client.py
```diff
@@ -103,7 +109,7 @@ def _get_system_prompt_and_messages(messages: List[ChatCompletionMessageParam])
                                 }
                             })
 
-                anthropic_messages.append(MessageParam(role='user', content=anthropic_content))
+                anthropic_messages.append(MessageParam(role='user', content=cast(Any, anthropic_content)))
```
aarondr77, line 112: Instead of casting this to Any should we be using the actual Anthropic types? 
Screenshot 2025-06-18 at 4 54 17 PM

mito-ai/mito_ai/gemini_client.py
```diff
@@ -32,7 +32,7 @@ def extract_system_instruction_and_contents(messages: List[Dict[str, Any]]) -> T
                 # We assume that that there is only one system message 
                 system_instructions = content
         elif role in ("user", "assistant"):
-            parts = []
+            parts: List[Any] = []
```
aarondr77: Same for this. Is there a Gemini type we should be using? If we use their types then we can make sure the code we're writing is type safe

mito-ai/mito_ai/gemini_client.py
```diff
@@ -143,23 +143,9 @@ async def request_completions(
...
-                collected_response = ""
-                for chunk in response:
-                    if isinstance(chunk, str):
-                        collected_response += chunk or ''
-                    elif hasattr(chunk, 'text') and chunk.text:
-                        collected_response += chunk.text or ''
-                    else:
-                        collected_response += str(chunk)
-                return collected_response
-
```
aarondr77, line 162: Ohh wow that's awesome we can get rid of this code

mito-ai/mito_ai/gemini_client.py
```diff
@@ -197,8 +183,6 @@ async def stream_completions(
                     next_chunk = ""
                     if hasattr(chunk, 'text'):
                         next_chunk = chunk.text or ''
-                    elif isinstance(chunk, str):
```
aarondr77, line 200: Nice!

aarondr77 (changes requested):
Nice nice! Love that this let us delete some of those extra if statements. A few small comments about not using Any and instead using the Anthropic and Gemini types, lmk if you think it's worth it. Otherwise LGTM

mito-ai/mito_ai/anthropic_client.py
sid-sagacollab: The issue is, when we try to force a type into it, we'd have to change the image content to be of `ImageBlock` type and this is requires that we provide a media type. This is causing issues when I try to set an image type since we are converting our base64 data into the image, it's causing some issues. It is not able to recognize the image_type that we are providing. Hence I let it be `any`

mito-ai/mito_ai/gemini_client.py
sid-sagacollab: Yessir !

mito-ai/mito_ai/gemini_client.py
sid-sagacollab: Addressed in [e13ffef](https://github.com/mito-ds/mito/pull/1784/commits/e13ffef7f7a45df017b48101f80fd9c95b338c25)

PR #1791, opened 2025-06-19
[WIP] Cell Numbering Spike

ngafar:
# Description

Added cell numbers above every cell. 

# Testing

Open a notebook, and you should see numbered cells. Move the cells around, delete them, etc.
 
# Documentation

N/A - self explanatory.

### DIFF
--- a/mito-ai/src/Extensions/CellHeaders/CellHeader.tsx
+++ b/mito-ai/src/Extensions/CellHeaders/CellHeader.tsx
@@ -7,7 +7,7 @@ interface CellHeaderProps {
 const CellHeader: React.FC<CellHeaderProps> = ({ number }) => {
   return (
     <div
-      className="cell-number"
+      className="mito-cell-number"
       style={{
         fontSize: '12px',
         color: '#666',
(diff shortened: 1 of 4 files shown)


aarondr77:
I'm seeing some issues with this implementation that I think is a result of Jupyter's notebook virtualization. 

The virtualization system uses the WindowedList class to manage viewport-based cell rendering. [github](https://github.com/jupyterlab/jupyterlab/pull/10131) Only cells within the current viewport plus a small buffer are attached to the DOM, while off-screen cells remain detached to improve performance. 

So I think what's happening in the video here is that the we're adding the cell header to cells that don't exist in the DOM! And as a result, in some cases, the Cell Header is below the cell when the cell is far away from what was rendered on the screen when the last modification occurred. 

https://github.com/user-attachments/assets/cfea60b2-7c65-4948-a17e-9f4223ef0069

I guess there are a few different approaches we can take to fix this: 
1. We could try to adjust the CellHeaders whenever the rendered cells change and whenever there is a modification. But that sounds very fragile to me. 
2. Maybe we can edit the Cell Object itself and therefore Jupyter will handle figuring out when the Cell Headers should be rendered.

ngafar:
Closing in favor of https://github.com/mito-ds/mito/pull/1797

Issue #1717, opened 2025-05-27
Don't auto scroll to bottom when user is scrolled up in chat to read


aarondr77:
When user is scrolled up in taskpane, we should not automatically scroll down to the next message when it arrives. It makes it very hard to read the conversation and try to understand what is happening.

aarondr77:
# Spec

### Problem 

Often, the user wants to scroll up in the chat to read what the agent is doing while it is working. (I do this often). However, every time the agent sends back a new message, we automatically scroll to the bottom. See the video below. 

https://github.com/user-attachments/assets/f70bb2ac-c755-470a-af74-f4426387cef0

### Implementation

What we want: 
1. If the user has scrolled up in the chat to view previous messages, their scroll position should take priority. It **should not** automatically scroll them back down to the bottom of the chat every time a new message is sent. 
2. If the user did not scroll in the chat, then every time a new message is sent back, it **should** scroll to the bottom. This makes is easy to follow along. 

I think what we can do is: 
1. Whenever a user sends a new agent message, we put them back into `follow the most recent message` mode
2. If the user then scrolls, we break them out of `follow the most recent message` mode and keep them at their current scroll position.

PR #1795, opened 2025-06-20
Disabled autoscroll when the user has scrolled up

sid-sagacollab:
# Description

Fix for: [Don't auto scroll to bottom when user is scrolled up in chat to read](https://github.com/mito-ds/mito/issues/1717)

# Testing

# Documentation

N/A

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
@@ -420,7 +420,7 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
         if (autoScrollFollowMode) {
             scrollToDiv(chatMessagesRef);
         }
-    }, [chatHistoryManager.getDisplayOptimizedHistory().length, autoScrollFollowMode]);
+    }, [chatHistoryManager.getDisplayOptimizedHistory().length, chatHistoryManager, autoScrollFollowMode]);
 
     // Add scroll event handler to detect manual scrolling
     useEffect(() => {
@@ -488,6 +488,9 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
     const sendAgentSmartDebugMessage = async (errorMessage: string): Promise<void> => {
         // Step 0: reset the state for a new message
         resetForNewMessage()
+
+        // Enable follow mode when sending agent debug message (same behavior as other modes)
+        setAutoScrollFollowMode(true);
 
         // Step 1: Create message metadata
         const newChatHistoryManager = getDuplicateChatHistoryManager()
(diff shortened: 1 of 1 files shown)


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -136,6 +136,16 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
         agentModeEnabledRef.current = agentModeEnabled;
     }, [agentModeEnabled]);
 
+    /* 
+        Auto-scroll follow mode: tracks whether we should automatically scroll to bottom
+        when new messages arrive. Set to false when user manually scrolls up.
+    */
+    const [autoScrollFollowMode, setAutoScrollFollowMode] = useState<boolean>(true);
```
aarondr77, line 143: My understanding is that we should only use useState when we want to trigger a rerender of the component when the state changes. In this case, I don't think we want to do that, so instead we can just use the ref to keep track of how we should scroll. 

Does that sound right to you?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -402,11 +415,35 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+        const chatContainer = chatMessagesRef.current;
+        if (!chatContainer) return;
+
+        const handleScroll = (): void => {
+            const { scrollTop, scrollHeight, clientHeight } = chatContainer;
+            const isAtBottom = scrollTop + clientHeight >= scrollHeight - 10; // 10px threshold
+            
+            // If user is not at bottom and we're in follow mode, break out of follow mode
+            if (!isAtBottom && autoScrollFollowModeRef.current) {
+                setAutoScrollFollowMode(false);
```
aarondr77, line 436: ie; this would just become: 
```suggestion
            // If user is not at bottom and we're in follow mode, break out of follow mode
            if (!isAtBottom && autoScrollFollowModeRef.current) {
                autoScrollFollowModeRef.current = false;
```

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -449,6 +489,9 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
         // Step 0: reset the state for a new message
         resetForNewMessage()
 
+        // Enable follow mode when sending agent debug message (same behavior as other modes)
+        setAutoScrollFollowMode(true);
```
aarondr77, line 493: Let's add this into the `resetForNewMessage()` function

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
sid-sagacollab: When we use only useRef, the scroll position check happens in the same useEffect that handles auto-scrolling:

The problem is that when new content is added, the DOM hasn't been updated yet when this useEffect runs. So scrollHeight might still be the old value, making the user appear to be "at bottom" when they're actually not.

This leads to the scrolling functionality to glitch.

Issue #1697, opened 2025-05-19
Gemini models cite multiple lines


aarondr77:
Gemini models are more likely to generate citations that include multiple lines ie: `[MITO_CITATION:84146866-5426-417b-bc58-716509924f29:18-41]` which is citing lines 18-41. We don't currently have the ability to handle this. 

We should either make sure it only picks on line or add UI support for multiple lines.

PR #1796, opened 2025-06-20
Multiline citations

sid-sagacollab:
# Description

Enabled multiline citations as per: [Gemini and Anthropic models cite multiple lines often](https://github.com/mito-ds/mito/issues/1697)

https://github.com/user-attachments/assets/047c6549-cd7b-4282-858a-531545630f05


evals/agent_evals/default_system_prompt.py
```diff
@@ -317,7 +317,7 @@ def create_agent_system_message_prompt(isChromeBrowser: bool) -> str:
 Output:
 {{
     type: 'finished_task', 
-    message: "The all time high tesla stock closing price was $265.91 [MITO_CITATION:9c0d5fda-2b16-4f52-a1c5-a48892f3e2e8:1] on 2025-03-16 [MITO_CITATION:9c0d5fda-2b16-4f52-a1c5-a48892f3e2e8:2]",
+    message: "The all time high tesla stock closing price was $265.91 [MITO_CITATION:9c0d5fda-2b16-4f52-a1c5-a48892f3e2e8:1] on 2025-03-16 [MITO_CITATION:9c0d5fda-2b16-4f52-a1c5-a48892f3e2e8:2]. This analysis[MITO_CITATION:9c0d5fda-2b16-4f52-a1c5-a48892f3e2e8:0-2] shows Tesla's stock performance.",
```
aarondr77: I think we actually wouldn't want this new citation to be part of the response since the agent just gave those previous two citations. This one doesn't add anything. It's just citing the same thing again. 

So can we just make this a better example?

evals/prompts/system/production_prompt.py
```diff
@@ -30,16 +30,20 @@ def get_system_prompt(database_rules: str) -> str:
...
-The average value is 53[MITO_CITATION:7b3a9e2c-5d14-4c83-b2f9-d67891e4a5f2:2]
+The average value is 53[MITO_CITATION:7b3a9e2c-5d14-4c83-b2f9-d67891e4a5f2:2] and the standard deviation is 15.2[MITO_CITATION:7b3a9e2c-5d14-4c83-b2f9-d67891e4a5f2:3]. These statistical measures[MITO_CITATION:7b3a9e2c-5d14-4c83-b2f9-d67891e4a5f2:2-3] provide insights into the data distribution.
 
 </Example>
 
-Notice in the example above that the citation uses line number 2 because citation line numbers are 0-indexed.
+Notice in the example above:
+- Single line citations use line number (e.g., :2 for line 2)
+- Multiline citations use range format (e.g., :2-3 for lines 2-3)
+- All line numbers are 0-indexed
```
aarondr77: Nice! Thanks for updating the evals one too!

mito-ai/mito_ai/completions/prompt_builders/prompt_constants.py
```diff
@@ -25,19 +25,24 @@
...
 2. When choosing the citation, select the code that will most help the user validate the fact or statement that you shared with them.
 3. Place the citation immediately after the statement it supports. Do not explain the citation with phrases like "See", "Derived from", etc. Just provide the citation object.
 4. For the "line_number" field, use the line number within the cell that is most relevant to the citation. Important: The cell line number should be 0-indexed and should not skip comments.
-5. If you cannot find relevant information in the notebook to answer a question, clearly state this and do not provide a citation.
-6. You ONLY need to provide a citation when sharing an insight from the data in the message part of the response. If all you are doing is writing/updating code, then there is no need to provide a citation.
-7. Do not include the citation in the code block as a comment. ONLY include the citation in the message field of your response.
+5. For multiline citations, use the "start_line-end_line" format when the insight spans multiple lines of code. Both line numbers should be 0-indexed.
```
aarondr77, line 42: Do you think it would be better to always have it respond with the start and end line. 

If they only want to reference one line then its in the format 1-1, if its multiple lines then its 1-3 for example.

Not sure which is better, but maybe worth a short experiment?

aarondr77: And it might make the citation parsing code simpler?

sid-sagacollab: Let's hold off on that thought for now, because doing so would mess up the older existing chats that have only 1 line not work anymore. So I wouldn't recommend doing that, at-least for parsing the citation code.

evals/prompts/system/production_prompt.py
sid-sagacollab: 👍

evals/agent_evals/default_system_prompt.py
sid-sagacollab: Done sir!

PR #1797, opened 2025-06-20
Cell header extension

aarondr77:
# Description

# Testing

# Documentation

### DIFF
--- a/mito-ai/package-lock.json
+++ b/mito-ai/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "mito_ai",
-  "version": "0.1.5",
+  "version": "0.1.228",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "mito_ai",
-      "version": "0.1.5",
+      "version": "0.1.228",
       "license": "AGPL-3.0-only",
       "dependencies": {
         "@codemirror/state": "^6.2.0",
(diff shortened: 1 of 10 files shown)


aarondr77:
TODO: Fill in learnings

aarondr77:
# Header-Attribute Approach for Cell Numbering

### Idea
Leverage JupyterLab's empty `.jp-Cell-header` that already sits above every cell (even virtualized).  Copy the cell's global index (`data-windowed-list-index`) onto that header so CSS can display **"Cell N"**.

### Code Sketch
```ts
function addHeaderFunctionality(cell: Cell) {
  const update = () => {
    const header = cell.node.querySelector('.jp-Cell-header');
    const idx     = cell.node.getAttribute('data-windowed-list-index');
    if (header && idx) header.setAttribute('data-cell-number', idx);
  };
  update();                          // initial
  const orig = cell.onUpdateRequest?.bind(cell);
  cell.onUpdateRequest = (msg) => {  // each re-attach (virtualization)
    orig?.(msg);
    update();
  };
}
```
```css
.jp-Cell-header{display:block;height:32px;line-height:32px;}
.jp-Cell-header::before{
  content:"Cell " attr(data-cell-number);
  padding-left:8px;
  font:14px/32px var(--jp-ui-font-family);
  color:var(--jp-ui-font-color2);
  border-bottom:1px solid var(--jp-border-color2);
  user-select:none;
}
```

### Why This Works
* Virtualization-safe — header appears each time a cell attaches; JS copies index instantly.
* No layout hacks — keeps JupyterLab's own header placement.
* Tiny runtime cost — one attribute copy per attach, done via lightweight factory wrappers.

PR #1798, opened 2025-06-21
Fix/model selector polish

aarondr77:
# Description

> If everything is highlighted in purple, then nothing is highlighted in purple" - @ngafar

- Updates the chat controls to make it fit a bit better on narrow screens. I was able to simplify the compact logic a lot by moving to media queries. I discovered that you can build a media query off of a parent's width. Now, we are casing on the width of the taskpane. This means we don't need all of those resize observer logic and additional js. 
- Uses purple more sparingly.
- Fixes the tooltips for Chat/Agent mode 

### Before: 
When the taskpane is wide: 
Screenshot 2025-06-21 at 1 47 35 PM

When the taskpane is compact: 
Screenshot 2025-06-21 at 1 47 43 PM

### After
When the taskpane is wide: 
Screenshot 2025-06-21 at 1 44 47 PM

When the taskpane is compact: 
Screenshot 2025-06-21 at 1 44 40 PM

# Testing

Try it out


mito-ai/style/ModelSelector.css
```diff
@@ -6,155 +6,152 @@
...
+/* Ensure dropdown appears above other elements */
+.model-options {
+  z-index: 1010;
+  /* Higher z-index to ensure visibility */
+}
+
+/* Make sure dropdown shows up correctly on the page boundaries */
+.model-selector {
+  position: relative;
+}
```
sid-sagacollab: This is a duplicate

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -56,23 +53,6 @@ const ModelSelector: React.FC<ModelSelectorProps> = ({ onConfigChange }) => {
...
-    });
-
-    const chatControls = containerRef.current?.closest('.chat-controls');
-    if (chatControls) {
-      observer.observe(chatControls);
-    }
-
-    return () => observer.disconnect();
-  }, []);
-
```
sid-sagacollab, line 75: I like this !!

It makes it much simpler

mito-ai/style/ModelSelector.css
aarondr77: good catch!

Issue #1788, opened 2025-06-19
[Agent] agent gets stuck in loop on simple requests


aarondr77:
When I just sent the message `hi`, the agent can get stuck in very long loops of responding with intro material. 

The issue here is that it decided to use the `CELL_UPDATE` instead of a `IS_FINISHED`. Then, once it made that wrong choice once, it just kept making the choice over and over. 

So maybe the solution is to clarify that if the response does not require updated code, ie: its not part of the analysis and just something it wants to tell to the user, then it should use the is_finished tool. Or something like that. 

Here's an example of the chat:

[sample-chat.txt](https://github.com/user-attachments/files/20821802/sample-chat.txt)

https://github.com/user-attachments/assets/4bc8a0b6-d9ba-4b36-96fa-5c7f2ca1e6f9

PR #1799, opened 2025-06-23
Decision tree scheme for decision making in the agent mode

sid-sagacollab:
# Description

This is to fix https://github.com/mito-ds/mito/issues/1788

Added a new "DECISION MAKING LOGIC" section that clearly categorizes message types
Provided specific examples for each category (greetings, questions, tasks)
Explicitly shows "hi" should result in FINISHED_TASK, not CELL_UPDATE

# Testing

This sometimes can fail, especially with gemini model. But this is not bulletproof as this is prompt engineering

# Documentation

Alternate approaches

- Strengthen FINISHED_TASK Guidance

```
WHEN TO USE FINISHED_TASK:

Use FINISHED_TASK in these situations:
- Responding to greetings or casual conversation
- Answering questions that don't require code execution
- Providing explanations or information
- When the user's message doesn't request any notebook changes
- When you're unsure if a task requires code - err on the side of conversation

Remember: Not every message needs code. Many interactions are purely conversational.

```
- Add Pre-Flight Check

```
Always ask yourself these questions:
1. Does this message require writing or modifying code? (Yes/No)
2. Does this message request data analysis or notebook changes? (Yes/No)
3. Is this a greeting, question, or casual conversation? (Yes/No)

If you answered "No" to questions 1 and 2, OR "Yes" to question 3, use FINISHED_TASK.
Only use CELL_UPDATE when you clearly need to write or modify code.
```

### DIFF
--- a/mito-ai/mito_ai/completions/prompt_builders/agent_system_message.py
+++ b/mito-ai/mito_ai/completions/prompt_builders/agent_system_message.py
@@ -215,14 +215,26 @@ When you have completed the user's task, respond with a message in this format:
 }}
 
 Important information:
-1. The message is a short summary of the ALL the work that you've completed on this task. It should not just refer to the final message. It could be something like "I've completed the sales strategy analysis by exploring key relationships in the data and summarizing creating a report with three recommendations to boost sales.""
-2. The message should include citations for any insights that you shared with the user.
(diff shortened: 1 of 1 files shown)


aarondr77 (changes requested):
I was still getting quite a few `cell update` responses to message like `hi` with Gemini in particular, so I tried another route here which seems to be working pretty well for me. Let me know what you think: https://github.com/mito-ds/mito/pull/1802

aarondr77:
The decision tree is a cool idea. it also adds a bit more complexity for us to manage. For example, currently, we never direct it to use the GET_CELL_OUTPUT tool. I feel like before we do this, we should create some more evals so we can evaluate the performance impact vs added complexity.

PR #1806, opened 2025-06-24
Autoscroll Fix

sid-sagacollab:
# Description

This is the PR for the changes that disappeared from https://github.com/mito-ds/mito/pull/1795


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -571,6 +617,9 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
         // Step 0: reset the state for a new message
         resetForNewMessage()
 
+        // Enable follow mode when user sends a new message
+        setAutoScrollFollowMode(true);
```
aarondr77, line 621: Let's add this into the resetForNewMessage() function, so we don't have to remember to call this everywhere?

sid-sagacollab: Sadly somehow this one change breaks it 😢

PR #1808, opened 2025-06-24
Cancel Checkpoint Restore (Fix)

ngafar:
# Description

Previously, if you were to select "restore checkpoint," but then hit cancel on the modal, the chat interface would still show the restore successful message. This PR fixes that.

Users should now be able to cancel a checkpoint restore, and still see the restore button. 

# Testing

Create a notebook, use the agent, attempt to restore a checkpoint. 

Make sure:

1. Restoring checkpoint still work.
2. You can cancel a restore. 

# Documentation

N/A


aarondr77 (approved):
LGTM after we add a jest test. I think we can just do two tests. They can: 

- mock `getNotebookStateHash`. Return the the same values in one test and different values in the other test to simulate the user pressing cancel or submit. 
- check if `setHasCheckpoint` was called or not.

aarondr77:
@claude can you write these jest tests for me?

PR #1809, opened 2025-06-24
Modifying the API_BASE_URL to point to the test environment 

PoojasreeDwarakanath:
# Description
Modifying the API_BASE_URL to point to the test environment url (instead of the development environment url).

# Testing


mito-ai/mito_ai/app_builder/handlers.py
```diff
@@ -19,7 +19,9 @@
 import requests
 
 # API endpoint for getting pre-signed URL
-API_BASE_URL = "https://fr12uvtfy5.execute-api.us-east-1.amazonaws.com"
+# For DEV testing use - "https://fr12uvtfy5.execute-api.us-east-1.amazonaws.com"
+# TODO: Change the API before production deployment
+API_BASE_URL = "https://iyual08t6d.execute-api.us-east-1.amazonaws.com"
```
aarondr77: Two quick changes here: 

1. Can we do something similar here to what we do in the mito_ai/completions/constants.py file where we have a PROD and DEV URL set and then we can switch which one is the active one
2. Can we create pytests like in mito_ai/tests/tests-constants.py to ensure that we have the correct ones set!

PR #1812, opened 2025-06-26
Fix/Gemini System Prompt and Model

aarondr77:
# Description

- Fixes bug where we were always using the fast GEMINI model for agent because of incorrect `if` statement logic
- Fixes bug here https://github.com/mito-ds/mito-private-monorepo/pull/899 where the Gemini system prompt was never set! 
- Adds fast models for OpenAI and Anthropic
- Does a little bit of cleaning up of the gemini and anthropic clients with the purpose of making sure the clients are structured / names as similarly as possible to make it easier to compare their logic. 

# Testing

- Checkout the new pytests to confirm that we are using the correct model! 

# Documentation

### DIFF
--- a/mito-ai/mito_ai/gemini_client.py
+++ b/mito-ai/mito_ai/gemini_client.py
@@ -10,19 +10,16 @@ from mito_ai.utils.gemini_utils import get_gemini_completion_from_mito_server, s
 
 GEMINI_FAST_MODEL = "gemini-2.0-flash-lite"
 
-def extract_system_instruction_and_contents(messages: List[Dict[str, Any]]) -> Tuple[str, List[Dict[str, Any]]]:
+def get_gemini_system_prompt_and_messages(messages: List[Dict[str, Any]]) -> Tuple[str, List[Dict[str, Any]]]:
     """
-    Separates system instructions from user/assistant messages and handles image content.
-    Returns:
-    - system_instructions: list of strings (for system_instruction param)
-    - contents: list of dicts for Gemini (excluding system role)
+    Converts a list of OpenAI messages to a list of Gemini messages.
     
     IMPORTANT: THIS FUNCTION IS ALSO USED IN THE LAMDBA FUNCTION. IF YOU UPDATE IT HERE,
     YOU PROBABLY NEED TO UPDATE THE LAMBDA FUNCTION AS WELL.
     """
 
-    system_instructions = ""
-    contents = []
+    system_prompt = ""
+    gemini_messages: List[Dict[str, Any]] = []
 
     for msg in messages:
         role = msg.get("role")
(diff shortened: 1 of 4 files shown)


mito-ai/mito_ai/gemini_client.py
```diff
@@ -97,24 +95,14 @@ async def request_completions(
...
-                config = {
-                    "response_mime_type": "application/json",
-                    "response_schema": AgentResponse.model_json_schema()
-                }
+            system_instructions, contents = get_gemini_system_prompt_and_messages(messages)
 
             # Get provider data for Gemini completion
             provider_data = get_gemini_completion_function_params(
-                model=self.model if not response_format_info else INLINE_COMPLETION_MODEL,
+                model=self.model if response_format_info else GEMINI_FAST_MODEL,
```
aarondr77, line 103: This was the key bug that was causing us to use the wrong model!! We were using the Fast Model when the response_format_info was set instead of when it was NOT set. The rest of the code in this PR is just cleaning up code and adding tests.

mito-ai/mito_ai/utils/anthropic_utils.py
```diff
@@ -208,13 +208,21 @@ def get_anthropic_completion_function_params(
         "max_tokens": max_tokens,
         "temperature": temperature,
         "messages": messages,
+        "system": system,
     }
     if response_format_info is not None:
         provider_data["model"] = INLINE_COMPLETION_MODEL
-    if system is not None and system is not anthropic.NotGiven:
-        provider_data["system"] = system
```
sid-sagacollab, line 215: We'll need this check for when the system is `anthropic.NotGiven` in inline completions I believe

sid-sagacollab: I know we've turned the inline completions off, but if we want the chat title completion to be from another these models, we'd want this logic to stay.

mito-ai/mito_ai/tests/test_anthropic_client.py
```diff
@@ -226,4 +229,42 @@ def test_tool_use_without_agent_response():
...
+
+CUSTOM_MODEL = "claude-3-5-sonnet-latest"
+@pytest.mark.parametrize("response_format_info, expected_model", [
+    (ResponseFormatInfo(name="agent_response", format=AgentResponse), CUSTOM_MODEL),  # With response_format_info - should use self.model
+    (None, ANTHROPIC_FAST_MODEL),  # Without response_format_info - should use ANTHROPIC_FAST_MODEL
+])
+@pytest.mark.asyncio
+async def test_model_selection_based_on_response_format_info(response_format_info, expected_model):
+    """
+    Tests that the correct model is selected based on whether response_format_info is provided.
```
sid-sagacollab, line 242: This is a great test to add!

mito-ai/mito_ai/gemini_client.py
sid-sagacollab: Oopsie !

sid-sagacollab (approved):
Small change in the anthropic util, the rest LGTM!

mito-ai/mito_ai/utils/anthropic_utils.py
aarondr77: I think this is okay because using a `anthropic.NotGiven` is the same as not having a system message, right? It seems to be working for the chat title completion anyways..

PR #1813, opened 2025-06-26
[WIP] Collapsed Code Blocks - Agent Mode

ngafar:
# Description

# Testing

# Documentation


mito-ai/src/Extensions/AiChat/ChatMessage/AssistantCodeBlock.tsx
```diff
@@ -0,0 +1,187 @@
...
+            <span className="agent-mode-toggle-content">
+                <CodeIcon />
+                Generated {lineCount} lines of code
+            </span>
+            <ExpandIcon isExpanded={isCodeExpanded} />
+        </div>
+    );
+};
+
+const renderLastAiMessageToolbar = (
```
aarondr77: Instead of making these functions, they should just be react components I think! Just like the `AssistantCodeBlock`. Its more react-y

mito-ai/src/Extensions/AiChat/ChatMessage/AssistantCodeBlock.tsx
```diff
@@ -0,0 +1,187 @@
...
+        acceptAICode();
+    }, [acceptAICode]);
+
+    const handleRejectCode = useCallback(() => {
+        rejectAICode();
+    }, [rejectAICode]);
+
+    // Memoize conditional logic
+    const shouldShowToolbar = isLastAiMessage || isCodeComplete;
+    const shouldShowFullToolbar = isLastAiMessage && isCodeComplete;
```
aarondr77: I don't think this is memoized, which is totally fine. I don't think we need to memoize it, but let's delete that comment

mito-ai/src/Extensions/AiChat/ChatMessage/AssistantCodeBlock.tsx
```diff
@@ -0,0 +1,187 @@
...
+        previewAICode();
+    }, [previewAICode]);
+
+    const handleAcceptCode = useCallback(() => {
+        acceptAICode();
+    }, [acceptAICode]);
+
+    const handleRejectCode = useCallback(() => {
+        rejectAICode();
+    }, [rejectAICode]);
```
aarondr77: Can't we just call these functions, like `setIsCodeExpanded` directly? Why are we doing this useCallback stuff?

aarondr77: Instead of creating these `handle.. ` function, we can just call the passed in function directly wherever we call the `handle..` funcition.

mito-ai/src/Extensions/AiChat/ChatMessage/AssistantCodeBlock.tsx
```diff
@@ -0,0 +1,187 @@
...
+        return renderOtherAiMessageToolbar(code);
+    }, [
+        shouldShowToolbar,
+        shouldShowFullToolbar,
+        codeReviewStatus,
+        code,
+        handlePreviewCode,
+        handleAcceptCode,
+        handleRejectCode
+    ]);
```
aarondr77: Hmm I think this is all pretty overkill too. I think instead of this whole function we can just write our render function like 

```typescript
{shouldShowFullToolbar && 

<LastAiMessageToolbar
codeReviewStatus = codeReviewStatus,
code = code
...

/>
}
```

mito-ai/src/Extensions/AiChat/ChatMessage/AssistantCodeBlock.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+    return (
+        <div
+            className={`code-block-container ${agentModeEnabled ? 'agent-mode' : ''} ${agentModeEnabled && !isCodeExpanded ? 'agent-mode-collapsed' : ''}`}
```
aarondr77: Let's use our classNames utility function here. It makes this code a bit less fragile.

mito-ai/src/Extensions/AiChat/ChatMessage/AssistantCodeBlock.tsx
```diff
@@ -0,0 +1,187 @@
...
+                    {toolbarElement}
+                    <PythonCode
+                        code={code}
+                        renderMimeRegistry={renderMimeRegistry}
+                    />
+                </>
+            )}
+        </div>
+    );
+};
```
aarondr77: It also might just be easier to understand what we have going on here if we express this return like this:

```typescript

if (agentModeEnabled) {

    return (
        <div className={classNames('code-block-container', 'agent-mode', {!isCodeExpanded: 'agent-mode-collapsed'}}>
            <AgentModeToggle
                isCodeExpanded=isCodeExpanded,
                lineCount=lineCount,
                handleToggleExpanded=handleToggleExpanded
            />
            {isToggleExpanded && 
                <toolbarElement />
                <PythonCode
                    code={code}
                    renderMimeRegistry={renderMimeRegistry}
                />
            } 
    )
} else {
    return (
        <div className={classNames('code-block-container')}>
            <ToolbarElement />
            <PythonCode
                code={code}
                renderMimeRegistry={renderMimeRegistry}
            />
        </div>
    )
}

Comment (aarondr77) [reply mito-ai/src/Extensions/AiChat/ChatMessage/AssistantCodeBlock.tsx]: That's not exactly correct, but you get the point..

File: mito-ai/src/Extensions/AiChat/ChatMessage/UserCodeBlock.tsx [hunk 7]
```diff
@@ -0,0 +1,52 @@
...
+    code, renderMimeRegistry, agentModeEnabled
+}) => {
+    const [isExpanded, setIsExpanded] = useState(false);
+    const numCodePreviewLines = 5;
+    const isCodeExpandable = code.split('\n').length > numCodePreviewLines;
+    const previewCode = code.split('\n').slice(0, numCodePreviewLines).join('\n');
+
+    return (
+        <div
+            className={`code-block-container active-cell-code-block ${agentModeEnabled ? 'agent-mode' : ''}`}
```
aarondr77, line 28: Let's move this to the classNmaes utitls also

mito-ai/src/tests/AiChat/AssistantCodeBlock.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    previewAICode: jest.fn(),
+    acceptAICode: jest.fn(),
+    rejectAICode: jest.fn(),
+    isLastAiMessage: true,
+    codeReviewStatus: 'chatPreview' as CodeReviewStatus,
+    agentModeEnabled: false,
+    ...overrides
+});
+
+describe('AssistantCodeBlock Component', () => {
```
aarondr77, line 48: LFG. Thanks for creating these. They look great and like they are testing the right things :) No mocking eveyrhtin!

aarondr77 (changes requested):
Nice. This is a step in the right direction I think. 

I think there are two main things we should do: 

1. The AssistantCodeBlock code can use a bit of cleaning up I think, just to make it more react-y and similar to the rest of our components. 
2. I think we should get rid of the toolbar for the agent code cells. I don't think it's important for users to be able to copy the code right now. If they want to, they can always just select it and copy it. I originally thought that maybe we should add the copy button next to the expand button, but I think that would actually make it less clear that the code was already applied for them. 

I think this looks better
Screenshot 2025-06-26 at 5 03 54 PM

Than this
Screenshot 2025-06-26 at 5 08 31 PM

mito-ai/src/Extensions/AiChat/ChatMessage/AssistantCodeBlock.tsx
```diff
@@ -0,0 +1,187 @@
...
+    renderMimeRegistry: IRenderMimeRegistry;
+    previewAICode: () => void;
+    acceptAICode: () => void;
+    rejectAICode: () => void;
+    isLastAiMessage: boolean;
+    codeReviewStatus: CodeReviewStatus;
+    agentModeEnabled: boolean;
+}
+
+const renderAgentModeToggle = (
```
aarondr77: Maybe we can call this one CollapsedCellUpdateBlock

PR #1814, opened 2025-06-26
Use the correct provider to name the chat 

aarondr77:
# Description

Use the selected provider to name the chat. Now, for example, if Open AI is experiencing outages, the user can switch the provider to Anthropic and have full functionality. 

# Testing

See the new tests. Also:
1. Run the mito server locally
2. Select the Anthropic model
3. Update the server so that the API key returned for OpenAI and Gemini are None (therefore we can't use them!)
4. Make sure you can name a chat 
5. Then, do the same for OpenAI and Gemini

# Documentation

No

### DIFF
--- a/mito-ai/mito_ai/gemini_client.py
+++ b/mito-ai/mito_ai/gemini_client.py
@@ -25,7 +25,8 @@ def extract_and_parse_gemini_json_response(response: GenerateContentResponse) ->
                 return " ".join(str(part) for part in content.parts)
             return str(content)
         return str(candidate)
-
+    
+    return None
 
 def get_gemini_system_prompt_and_messages(messages: List[Dict[str, Any]]) -> Tuple[str, List[Dict[str, Any]]]:
     """


mito-ai/mito_ai/completions/message_history.py
```diff
@@ -20,15 +19,34 @@
...
-async def generate_short_chat_name(user_message: str, assistant_message: str, llm_provider: OpenAIProvider) -> str:
+async def generate_short_chat_name(user_message: str, assistant_message: str, model: str, llm_provider: OpenAIProvider) -> str:
     prompt = create_chat_name_prompt(user_message, assistant_message)
 
     completion = await llm_provider.request_completions(
         messages=[{"role": "user", "content": prompt}], 
-        model=MESSAGE_TYPE_TO_MODEL[MessageType.CHAT_NAME_GENERATION],
+        # We set the model so we can use the correct model provider, but request_completions will decide to 
+        # use the fast model from that provider to make the request.
+        model=model,
```
aarondr77, line 29: This is the key change. By passing the selected model all the way through, the request_completions function will now use that client and choose the Fast model from that provider!

mito-ai/mito_ai/tests/message_history/test_generate_short_chat_name.py
```diff
@@ -0,0 +1,125 @@
...
+        # Test the function
+        result = await generate_short_chat_name(
+            user_message="What is the capital of France?",
+            assistant_message="The capital of France is Paris.",
+            model=selected_model,
+            llm_provider=llm_provider
+        )
+    
+    # Verify the result
+    assert result == "Test Chat Name"
```
sid-sagacollab: Shouldn't we assert that the fast model is being used ? instead of asserting the mocked response ?

sid-sagacollab (approved):
Small comment on the pytest, rest LGTM

mito-ai/mito_ai/tests/message_history/test_generate_short_chat_name.py
aarondr77: Yeah good callout

PR #1817, opened 2025-06-27
Cleanup 6 26 25

aarondr77:
# Description

# Testing

# Documentation


mito-ai/mito_ai/constants.py
```diff
@@ -43,7 +43,7 @@ def get_model_provider(model: str) -> Union[str, None]:
 
 
 # Mito AI Base URLs and Endpoint Paths
-MITO_PROD_BASE_URL = "https://x3rafympznv4abp7phos44gzgu0clbui.lambda-url.us-east-1.on.aws"
+MITO_PROD_BASE_URL = "http://0.0.0.0:8001"
```
aarondr77: No!

PR #1818, opened 2025-06-27
Fixed the OpenAI bug and also changed the default model

sid-sagacollab:
# Description

This resolves the issue of default model selection on a private window which simulates a new user. 

# Testing

### DIFF
--- a/mito-ai/mito_ai/constants.py
+++ b/mito-ai/mito_ai/constants.py
@@ -30,13 +30,15 @@ def get_model_provider(model: str) -> Union[str, None]:
     if not model:
         return None
 
-    if model.startswith('claude'):
+    model_lower = model.lower()
+
+    if model_lower.startswith('claude'):
         return 'claude'
-    elif model.startswith('gemini'):
+    elif model_lower.startswith('gemini'):
         return 'gemini'
-    elif model.startswith('ollama'):
+    elif model_lower.startswith('ollama'):
         return 'ollama'
-    elif model.startswith('gpt') or model.startswith('o3'):
+    elif model_lower.startswith('gpt'):
         return 'openai'
 
     return None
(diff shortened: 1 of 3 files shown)


mito-ai/mito_ai/constants.py
```diff
@@ -30,13 +30,15 @@ def get_model_provider(model: str) -> Union[str, None]:
...
+    if model_lower.startswith('claude'):
         return 'claude'
-    elif model.startswith('gemini'):
+    elif model_lower.startswith('gemini'):
         return 'gemini'
-    elif model.startswith('ollama'):
+    elif model_lower.startswith('ollama'):
         return 'ollama'
-    elif model.startswith('gpt') or model.startswith('o3'):
+    elif model_lower.startswith('gpt'):
```
sid-sagacollab, line 41: Updated this logic since it was outdated

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -39,18 +39,31 @@ const ModelSelector: React.FC<ModelSelectorProps> = ({ onConfigChange }) => {
...
+    let displayName: string | undefined;
+
     if (storedConfig) {
       try {
         const parsedConfig = JSON.parse(storedConfig);
-        const fullModelName = parsedConfig.model || MODEL_MAPPINGS.find(m => m.displayName === DEFAULT_MODEL)?.fullName;
-        const displayName = MODEL_MAPPINGS.find(m => m.fullName === fullModelName)?.displayName || DEFAULT_MODEL;
-        setSelectedModel(displayName);
-
-        onConfigChange({ model: fullModelName });
```
sid-sagacollab, line 49: When the model defaulted, it was not getting resolved to the right model `fullname`, instead, it was just falling back into the `displayname`

aarondr77 (approved):
Looks great! Thanks for all the callouts too!

PR #1819, opened 2025-06-30
Collapse Error Fix-up 

ngafar:
# Description

When the Agent enters an error fix-up loop, the initial message is no longer displayed as a user chat message. Instead, the chat log show a collapsed "fixing error" message that can be expanded. 

# Testing

Prompt the Agent to complete a task you know will error, look to make sure the new UI element is displayed. Refresh to make sure it works.

If you are unsure of how to trigger an error try this:

1. Run `!pip uninstall numpy -y`
2. Then refresh the kernel. 
3. In Agent mode, ask "please import numpy."

# Documentation

N/A - This should be self-explanatory.


mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
```diff
@@ -87,7 +88,19 @@ const ChatMessage: React.FC<IChatMessageProps> = ({
...
+        // Initially, messages are labeled with the prompt type 'agent:autoErrorFixup'
+        promptType === 'agent:autoErrorFixup' ||
+        // However, when the chat history is saved, this field is stripped, and every message is labeled as 'chat'.
+        // In this case we have to manually determine if the message is an error fixup message.
+        (message.role === 'user' && 
+            messageContent && 
+            messageContent.includes('---->') && 
+            /\w+Error:/.test(messageContent)
+        )
+    );
```
aarondr77: Dang that's annoying. Good fix for now. Eventually, we should do a refactor where we store the metadata in the chat history. That would include everything!

mito-ai/src/components/AgentToolComponents/ErrorFixupToolUI.tsx
```diff
@@ -0,0 +1,48 @@
...
+            >
+                <span className="error-fixup-toggle-content">
+                    <WrenchAndScrewdriverIcon />
+                    Fixing {parsePythonErrorType(messageContent)}
+                </span>
+                <ExpandIcon isExpanded={isExpanded} />
+            </div>
+            {isExpanded && (
+                <div className="error-fixup-expanded">
+                    {messageContent}
```
aarondr77: The message content contains python code. Currently, the python code not formatted and is displayed in ```

Maybe it would be better to just delete the python code section (we have utility functions, like `splitStringWithCodeBlocks` that should make this pretty easy) and just display the error? 

Or if you think the python code should remain, then let's render it like python code! 

Screenshot 2025-07-01 at 1 09 47 PM

aarondr77 (changes requested):
### Even more collapsed!!! 

> "Perfectly balanced, as all things should be - Thanos

We are still cluttering up the screen with a lot of information about things erroring. I think ideally we would collapse this even more.
Screenshot 2025-07-01 at 1 06 12 PM

What if we collapsed the entire error fix up process into one element like this? 

Inside of this toggle is: 
- The first error
- The first attempted fix
- The second error
- The second attempted fix
- The third error
- The third, correct fix. 

To implement this: 
- Implementing this would be slightly more tricky because we'd need to group ChatMessages a bit before we render each component. But I think that won't be too hard! 
- Remove the hardcoded "Hmm this didn't work" text.

Screenshot 2025-07-01 at 1 23 52 PM

Comparing these two screenshots (I generated the second one just by editing the rendered html btw), I think the second one is a lot more readable and worth an afternoon of implementation. Thoughts? 

### Testing

Let's make sure to add our Jest tests!

mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
aarondr77: Ohh actually, I'm still seeing this when I refresh the page
Screenshot 2025-07-01 at 1 35 17 PM

ngafar: Yeah, this is not the best solution. It looks like there are a few edge cases which I've updated the `isErrorFixupMessage` logic to handle:

1. Something the error message has an arrow, but the number of dashes in the arrow changes. I've updated it to look for at least one dash, followed by a `>`. 
2. Certain errors, like syntax errors don't create an arrow, but instead have a `Did you mean`.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1286,6 +1289,18 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+    
+    if (agentModeEnabled) {
+        processedDisplayOptimizedChatHistory = processChatHistoryForErrorGrouping(
+            chatHistoryManager.getDisplayOptimizedHistory()
+        );
+    } else {
+        processedDisplayOptimizedChatHistory = chatHistoryManager.getDisplayOptimizedHistory()
+    }
+
+    console.log('processedDisplayOptimizedChatHistory', processedDisplayOptimizedChatHistory)
```
aarondr77: We can delete this

mito-ai/src/components/AgentToolComponents/ErrorFixupToolUI.tsx
```diff
@@ -0,0 +1,109 @@
...
+                    <AlertIcon />
+                    {errorType} Detected
+                </span>
+                <ExpandIcon isExpanded={expandedError} />
+            </div>
+            {expandedError && errorContent && (
+                <div className="error-fixup-expanded">
+                    <PythonCode
+                        code={errorContent}
+                        renderMimeRegistry={renderMimeRegistry}
```
aarondr77: This isn't Python code, and it's also not markdown. So maybe we should just display it as raw text??

aarondr77: Hmmmm it does contain python most of the time though... maybe having it rendered as Python is good!

mito-ai/src/components/AgentToolComponents/ErrorFixupToolUI.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    const errorContent = getContentStringFromMessage(errorMessage);
+    const errorType = parsePythonErrorType(errorContent);
+
+    const toggleError = (): void => {
+        setExpandedError(!expandedError);
+    };
+
+    return (
+        <div className={classNames('error-fixup-container', { expanded: expandedError })}>
```
aarondr77: Can we give these a little effect on hover to make it more clear that they are clickable?

mito-ai/src/Extensions/AiChat/ChatMessage/GroupedErrorMessagesBlock.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react"
+import { getContentStringFromMessage } from "../../../utils/strings"
+import { IDisplayOptimizedChatItem } from "../ChatHistoryManager"
+
+interface IGroupedErrorMessagesProps {
+    displayOptimizedChat: IDisplayOptimizedChatItem[]
+}
+
+export default function GroupedErrorMessagesBlock({ displayOptimizedChat }: IGroupedErrorMessagesProps) {
```
aarondr77: Doesn't look like we are using this component anywhere. We can delete!

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1286,6 +1289,18 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
 
     const lastAIMessagesIndex = chatHistoryManager.getLastAIMessageIndex()
 
+    let processedDisplayOptimizedChatHistory: (IDisplayOptimizedChatItem | GroupedErrorMessages)[] = []
+    
+    if (agentModeEnabled) {
```
aarondr77, line 1320: Want to add a small comment that just explains how we handle things differently in agent vs chat mode?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1343,34 +1358,44 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
-                            rejectAICode={rejectAICode}
-                            onUpdateMessage={handleUpdateMessage}
-                            contextManager={contextManager}
-                            codeReviewStatus={codeReviewStatus}
-                            setNextSteps={setNextSteps}
-                            agentModeEnabled={agentModeEnabled}
-                        />
-                    )
+                {processedDisplayOptimizedChatHistory.map((displayOptimizedChat, index) => {
+                    if (Array.isArray(displayOptimizedChat)) {
```
aarondr77: Can we turn this into a better check? Instead of checking if it's an array, ideally we'd check if the element is of type `GroupedErrorMessages`.

mito-ai/src/components/AgentToolComponents/ErrorFixupToolUI.tsx
```diff
@@ -0,0 +1,109 @@
+import React, { useState } from 'react';
```
aarondr77: Nice new folder of components! Want to move over the other AgentToolComponents like the Get Cell Output, etc? 

.. also, nitpick, but maybe we should just call this AgentComponents because Error fixing up is not a tool, its just something the agent does.

mito-ai/src/components/AgentToolComponents/ErrorFixupToolUI.tsx
```diff
@@ -0,0 +1,109 @@
...
+                className={classNames('error-fixup-toggle', {
+                    expanded: expandedError
+                })}
+            >
+                <span className="error-fixup-toggle-content">
+                    <AlertIcon />
+                    {errorType} Detected
+                </span>
+                <ExpandIcon isExpanded={expandedError} />
+            </div>
```
aarondr77: It would be nice to make this div into a component, called something like `AgentComponentHeader` or something like that. Because we have this same thing duplicated across a bunch of agent components now and we want them to be styled the same every time. 

I think it would need to be a component with properties: 
1. Icon
2. Text
3. OnClick
4. DisplayBorder

If making this a component isn't easy, then we can at least re-use the same classes for it.

mito-ai/src/Extensions/AiChat/ChatMessage/AssistantCodeBlock.tsx
```diff
@@ -47,12 +49,13 @@ const AssistantCodeBlock: React.FC<IAssistantCodeBlockProps> = ({
         return (
             <div className={classNames('code-block-container', {
                 'agent-mode': true,
```
aarondr77: If this is always true, then we can just write it as classNames('code-block-container', 'agent-mode', { 'agent-mode-collapsed': !isCodeExpanded}

mito-ai/src/components/AgentToolComponents/ErrorFixupToolUI.tsx
```diff
@@ -0,0 +1,109 @@
...
+                    }
+                    return (
+                        <AssistantCodeBlock
+                            key={`assistant-${index}`}
+                            code={messageItem.message.content as string}
+                            isCodeComplete={true}
+                            renderMimeRegistry={renderMimeRegistry}
+                            previewAICode={() => { }}
+                            acceptAICode={() => { }}
+                            rejectAICode={() => { }}
```
aarondr77: Hmm this makes my hacky code senses go off. 

If we want to reuse the AssistantCodeBlock, then we should type it more strongly at least. Let's make each one of those fields potentially undefined. Note, let's make it undefined instead of optional so we at least have to consciously set the field still.

This is maybe a little fragile because we don't ever want to set previewAICode to a function and acceptAICode to undefined. So I think this probably means it's not the best factoring of our components, but it's okay for now! 

.

mito-ai/src/utils/chatHistory.tsx
```diff
@@ -0,0 +1,72 @@
...
+                displayOptimizedChatHistory[next_index]?.type === 'openai message' &&
+                displayOptimizedChatHistory[next_index]?.message.role === 'assistant'
+            ) {
+                groupedErrorMessages.push(displayOptimizedChatHistory[next_index]!);
+                i = next_index;
+            } 
+        } else {
+            if (groupedErrorMessages.length > 0) {
+                processedDisplayOptimizedChatHistory.push(groupedErrorMessages);
+                groupedErrorMessages = new Array<IDisplayOptimizedChatItem>();
```
aarondr77, line 61: Let's add a small comment here that explains we need to make a new array object because lists are immutable. right?

mito-ai/src/utils/chatHistory.tsx
```diff
@@ -0,0 +1,72 @@
...
+            messageContent
+        );
+
+        if (_isErrorFixupMessage) {
+            // If the current message is an error fixup message, we need to group it with the next message
+            // which is the ai response to the error fixup message.
+            groupedErrorMessages.push(displayOptimizedChatItem);
+            
+            // TODO: Make sure the next message is an ai response to the error fixup message. 
+            // If the user has stopped the agent, it might not be an ai response to the error fixup message.
```
aarondr77: We fixes this, so we can remove the todo, but let's just turn this into a comment, since it's an edge case we will probably forget when we come back to it. And let's make sure we have a test for it! I think we do :)

mito-ai/src/components/AgentToolComponents/ErrorFixupToolUI.tsx
```diff
@@ -0,0 +1,109 @@
...
+    renderMimeRegistry: IRenderMimeRegistry;
+}
+
+const parsePythonErrorType = (content: string | undefined): string => {
+    if (!content) return 'Error';
+    const errorMatch = content.match(/(\w+Error):/);
+    return errorMatch?.[1] || 'Error';
+};
+
+const ErrorBlock = ({
```
aarondr77: Maybe we can call this component `ErrorDetectedBlock`

mito-ai/src/components/AgentToolComponents/ErrorFixupToolUI.tsx
```diff
@@ -0,0 +1,109 @@
...
+                        code={errorContent}
+                        renderMimeRegistry={renderMimeRegistry}
+                    />
+                </div>
+            )}
+        </div>
+    );
+}
+
+const ErrorFixupToolUI: React.FC<IErrorFixupToolUIProps> = ({
```
aarondr77: Maybe we can call this one `GroupedErrorsAndFixes`?

mito-ai/src/components/AgentToolComponents/ErrorFixupToolUI.tsx
aarondr77: Okay, well I did see this at some point. This is what I expect -- the error message is displayed. 
Screenshot 2025-07-05 at 9 27 34 PM

However, now every time I am using it, I am just seeing the code cell here! 

https://github.com/user-attachments/assets/3a32633f-26c8-4328-a770-dc529cad9296

This might be a weird react issue where react is treating this as the same component as the previous Python Code block we render??

aarondr77: Hmm yeah if I add this index code, then it seems to fix the issue.... I'm always a bit confused on how react handles these components and why errors like this occur? Do you know? 
Screenshot 2025-07-05 at 10 39 59 PM

aarondr77 (approved):
Sweet!! this rocks!! I think this is a great UI refactor that makes the agent much, much easier to follow. Great work! 

A few small changes and then LGTM! I'm glad we have a lot of tests for this. I feel confident that its working well :) 

I also made a few small styling changes, including that border looking a little funky. The fix for that is doing `overflow: hidden` btw. It makes sure that the python code container doesn't overflow the rounded borders.

mito-ai/src/components/AgentToolComponents/ErrorFixupToolUI.tsx
ngafar: Yeah, this is a weird one. Going to look into this a little more. 

Looks like if you refresh, the component displays the right error message.

ngafar: Fixed. It was the PythonCode component, specifically the `addMarkdownCodeFormatting` function which would strip the message so that only the code block is shown. 

Removing the code block means there is only an error message, and it gets shown.

PR #1831, opened 2025-07-04
Retry logic and refactor

aarondr77:
# Description

The primary goal of this PR was to adds centralized retry logic into the provider.py file. This means that we automatically add retry logic to all providers regardless of if the user is using the mito server or their own key. 

**This is intended to fix the slice error that we've seen with Anthropic which happens when Anthropic returns a 529 overload error** 

In the process of implementing this, I made a few refactors that made it much easier to add this logic. In particular:

1. Standardizes the mito server response type. If the LLM provider returns a response we pass it back to the client via the `{'completion': <response>}` format. If it fails, we return the error as `{'error': <error message>}`. We now use a helper function for extracting these completions and errors. This made it easy to: 
- Fix bug where we were not counting usage of anthropic or gemini against the free tier limit. 
- Create a more useful error message for the user that tells them there was a problem with the provider and they should try switching to a different provider. 

2. This PR also adds a whole bunch of tests that should actually test the right behavior. We had some issues in existing tests, for example the AI capabilities test, where instead of actually testing the logic of the app, we were just testing that the mocked objects were returning the correct thing. 

### Accompanying PR

This PR requires changes to the mito-server. Those changes are here: https://github.com/mito-ds/mito-private-monorepo/pull/901

The problem however, is that if we change the return type on the mito server, it will break all users who have existing mito installations that don't handle the new return type for anthropic and gemini.

**We need to figure out how we want to do this release.**

# Testing

To test this: 
1. Set up the dev mito server 
2. Success case: Use the dev mito server and make sure responses from all models are handled correctly
3. Error case: Update the mito server so that it returns an error by raising an exception in the Anthropic `generate_response` function. (ie: like this) and then make sure that it retries to get the correct response. 
Screenshot 2025-07-06 at 5 54 03 PM

# Documentation

No.


mito-ai/mito_ai/anthropic_client.py
```diff
@@ -148,51 +149,46 @@ async def request_completions(self, messages: List[ChatCompletionMessageParam],
         """
         Get a response from Claude or the Mito server that adheres to the AgentResponse format.
         """
-        try:
```
aarondr77, line 151: Instead of catching the exceptions in the clients, we let the exceptions bubble up to the provider.py so we can manage all of the exceptions in one centralized place.

mito-ai/mito_ai/constants.py
```diff
@@ -23,29 +23,8 @@
...
-        return 'ollama'
-    elif model_lower.startswith('gpt'):
-        return 'openai'
-
-    return None
-
-
 # Mito AI Base URLs and Endpoint Paths
-MITO_PROD_BASE_URL = "https://yxwyadgaznhavqvgnbfuo2k6ca0jboku.lambda-url.us-east-1.on.aws"
+MITO_PROD_BASE_URL = "http://0.0.0.0:8001"
```
aarondr77: move this back

mito-ai/mito_ai/gemini_client.py
```diff
@@ -112,50 +113,46 @@ async def request_completions(
...
     ) -> str:
-        try:
-            # Extract system instructions and contents
-            system_instructions, contents = get_gemini_system_prompt_and_messages(messages)
-
-            # Get provider data for Gemini completion
-            provider_data = get_gemini_completion_function_params(
-                model=self.model if response_format_info else GEMINI_FAST_MODEL,
-                contents=contents,
+        # Extract system instructions and contents
```
aarondr77, line 116: The only change here is that I removed the try block, again so we can handle the exceptions in the providers.py file

mito-ai/mito_ai/utils/mito_server_utils.py
```diff
@@ -0,0 +1,85 @@
...
+        self.user_friendly_hint = f"There was a problem with {provider_name}. Try switching to a different model and trying again."
+
+        # Set args[0] for fallback compatibility
+        super().__init__(self.user_friendly_title)
+
+    def __str__(self):
+        return f"{self.provider_name} Error: {self.error_message}"
+
+
+async def get_response_from_mito_server(
```
aarondr77, line 30: Creating this function was the main refactor done in this PR. We now extract the completions + handle any errors here. Using this function for all of our clients makes sure that we have the same return type for all clients, making it much easier to understand that data types we're working with!

mito-ai/mito_ai/utils/mito_server_utils.py
```diff
@@ -0,0 +1,85 @@
...
+            content = json.loads(res.body.decode("utf-8"))
+                        
+            if "completion" in content:
+                return content["completion"]
+            elif "error" in content:
+                # Server returned an error
+                raise ProviderCompletionException(content['error'], provider_name=provider_name)
+            else:
+                # Invalid response format
+                raise ProviderCompletionException(f"No completion found in response: {content}", provider_name=provider_name)
```
aarondr77, line 73: This is what lets us display better error messages for the user. When the server responds with an error object, we use that to build the error message we display to the user!

mito-ai/mito_ai/completions/providers.py
```diff
@@ -176,19 +212,23 @@ async def stream_completions(
...
+                gemini_client = GeminiClient(api_key=api_key)
+                # TODO: We shouldn't need to do this because the messages should already be dictionaries... 
+                # but if we do have to do some pre-processing, we should do it in the gemini_client instead.
                 messages_for_gemini = [dict(m) for m in messages]
                 accumulated_response = await gemini_client.stream_completions(
                     messages=messages_for_gemini,
+                    model=model,
                     message_id=message_id,
                     reply_fn=reply_fn,
                     message_type=message_type
```
sid-sagacollab, line 234: We would want retry logic here as well ?

mito-ai/mito_ai/tests/test_anthropic_client.py
```diff
@@ -231,38 +232,59 @@ def test_tool_use_without_agent_response():
...
+        with patch.object(client.client, 'messages') as mock_messages:
+            if response_format_info is not None:
+                # When response_format_info is provided, mock a tool use response
+                tool_use_block = ToolUseBlock(
+                    type="tool_use",
+                    id="test_id", 
+                    name="agent_response",
+                    input={"test": "value"}
+                )
+                mock_messages.create.return_value = Message(
```
sid-sagacollab, line 258: I like this test!

mito-ai/mito_ai/tests/providers/test_providers.py
```diff
@@ -0,0 +1,382 @@
...
+                )
+
+                assert completion == "Mito server response"
+                mock_mito_function.assert_called_once()
+
+
+@pytest.mark.parametrize("mito_server_config", [
+    {
+        "name": "openai_fallback",
+        "model": "gpt-4o-mini",
```
sid-sagacollab, line 312: Isn't this supposed to be gpt4.1 ?

mito-ai/mito_ai/completions/providers.py
aarondr77: Yeah we do! I want to do that in a different PR though cause this one is getting pretty big!

mito-ai/mito_ai/tests/providers/test_retry_logic.py
```diff
@@ -0,0 +1,386 @@
...
+    ):
+        """Test comprehensive logging behavior for all retry scenarios."""
+        monkeypatch.setenv("OPENAI_API_KEY", FAKE_API_KEY)
+        monkeypatch.setattr("mito_ai.constants.OPENAI_API_KEY", FAKE_API_KEY)
+        
+        # Clear other API keys to ensure OpenAI path is used
+        monkeypatch.delenv("CLAUDE_API_KEY", raising=False)
+        monkeypatch.delenv("GEMINI_API_KEY", raising=False)
+        monkeypatch.setattr("mito_ai.constants.CLAUDE_API_KEY", None)
+        monkeypatch.setattr("mito_ai.constants.GEMINI_API_KEY", None)
```
sid-sagacollab, line 329: Can you please tell me what this is trying to ?

mito-ai/mito_ai/tests/providers/test_retry_logic.py
```diff
@@ -0,0 +1,386 @@
...
+                model="gpt-4o-mini",
+                max_retries=max_retries
+            )
+            
+            assert result == "Success!"
+            assert mock_client.return_value.request_completions.call_count == expected_call_count
+            assert provider.last_error is None  # Should be reset on success
+            
+            # Verify sleep was called for retries
+            assert mock_sleep.call_count == attempts_before_success
```
sid-sagacollab, line 79: Not important: The current tests only verify that asyncio.sleep is called the correct number of times, but they don't validate that the exponential backoff timing is implemented correctly.

sid-sagacollab: We can try something like this 

```suggestion
            expected_calls = [call(2**i) for i in range(attempts_before_success)]
            mock_sleep.assert_has_calls(expected_calls)

```

sid-sagacollab (approved):
Overall this PR looks solid !

I've left minor comments and questions. Thanks!

mito-ai/mito_ai/tests/providers/test_providers.py
aarondr77: This test is just making sure whatever model we pass get used by the function

mito-ai/mito_ai/tests/providers/test_retry_logic.py
aarondr77: We're just making sure that we don't have any environment variables set so this can test the mito-server route. Raising = False makes sure that if the environment variable wasn't set already, then it doesn't throw an error

PR #1832, opened 2025-07-06
Fix/azure open ai

aarondr77:
# Description

# Testing

# Documentation

### DIFF
--- a/mito-ai/mito_ai/openai_client.py
+++ b/mito-ai/mito_ai/openai_client.py
@@ -194,10 +194,11 @@ This attribute is observed by the websocket provider to push the error to the cl
             # The format for using Azure OpenAI is different than using
             # other providers, so we have a special case for it here.
             # Create Azure OpenAI client with explicit arguments
+            print("HERE!!: BUILDING AZURE OPENAI CLIENT")
             return openai.AsyncAzureOpenAI(
                 api_key=constants.AZURE_OPENAI_API_KEY,
                 api_version=constants.AZURE_OPENAI_API_VERSION,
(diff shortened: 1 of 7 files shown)


mito-ai/mito_ai/openai_client.py
```diff
@@ -253,14 +266,18 @@ async def request_completions(
...
+            model = self._resolve_model(model, response_format_info)
 
             # Handle other providers as before
             completion_function_params = get_open_ai_completion_function_params(
                 model, messages, False, response_format_info
             )
 
             if self._active_async_client is not None:
+                print("HERE")
+                print("sending request with messages", messages)
```
aarondr77: remove

mito-ai/mito_ai/openai_client.py
```diff
@@ -253,14 +266,18 @@ async def request_completions(
         completion = None
 
         try:
-            model = self._resolve_model(model) if response_format_info else OPENAI_FAST_MODEL
+            
+            # Make sure we are using the correct model
+            model = self._resolve_model(model, response_format_info)
```
aarondr77, line 272: This was the key change. Before this change, we were only using the set OpenAI Azure model for agent mode requests and were defaulting to the FAST model for things like naming the chat. Because that was not configured with the user's Azure client, it would fail!

mito-ai/mito_ai/openai_client.py
```diff
@@ -253,14 +266,18 @@ async def request_completions(
         completion = None
 
         try:
-            model = self._resolve_model(model) if response_format_info else OPENAI_FAST_MODEL
+
```
aarondr77, line 268: Add a comment here that if we ever add back inline completions, then we need to force the user to configure a fast model and a slow model.

PR #1836, opened 2025-07-06
mito-ai: fix weakmap error with try block

aarondr77:
# Description

# Testing

# Documentation


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1249,52 +1249,49 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
-                        compartment = new Compartment();
-                        codeDiffStripesCompartments.current.set(codeCell, compartment);
-
-                        // Apply the initial configuration
-                        editorView.dispatch({
-                            effects: StateEffect.appendConfig.of(
-                                compartment.of(unifiedDiffLines !== undefined && isActiveCodeCell ? codeDiffStripesExtension({ unifiedDiffLines: unifiedDiffLines }) : [])
-                            ),
-                        });
+        try {
```
aarondr77, line 1252: I added this try block since this is non-critical behavior. That is the only change.

PR #1843, opened 2025-07-08
Set model in helper

aarondr77:
# Description

Creates a helper function for deciding if we need to use a fast model or the selected model. In doing so, we make the decision criteria based on the message type instead of if there is response_format_info provided or not. And moves all of the decision logic into the client utils to keep the clients clean. 

# Testing

See the new tests + try it out. 

# Documentation

Nope.

### DIFF
--- a/mito-ai/mito_ai/constants.py
+++ b/mito-ai/mito_ai/constants.py
@@ -24,7 +24,7 @@ AZURE_OPENAI_ENDPOINT = os.environ.get("AZURE_OPENAI_ENDPOINT")
 AZURE_OPENAI_MODEL = os.environ.get("AZURE_OPENAI_MODEL")
 
 # Mito AI Base URLs and Endpoint Paths
-MITO_PROD_BASE_URL = "http://0.0.0.0:8001"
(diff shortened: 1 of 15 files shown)


mito-ai/mito_ai/constants.py
```diff
@@ -24,7 +24,7 @@
 AZURE_OPENAI_MODEL = os.environ.get("AZURE_OPENAI_MODEL")
 
 # Mito AI Base URLs and Endpoint Paths
-MITO_PROD_BASE_URL = "https://yxwyadgaznhavqvgnbfuo2k6ca0jboku.lambda-url.us-east-1.on.aws"
+MITO_PROD_BASE_URL = "http://0.0.0.0:8001"
```
aarondr77: move this back

mito-ai/mito_ai/tests/providers/test_model_resolution.py
```diff
@@ -0,0 +1,125 @@
+
```
sid-sagacollab, line 1: I think we should add a pytest case to assert that passing an unsupported MessageType raises ValueError in `does_message_require_fast_model()`. Otherwise the else branch of that logic isn’t covered.

aarondr77: good call!

PR #1852, opened 2025-07-11
mito-ai: fix bug with get cell output tool

aarondr77:
# Description

Makes sure that we return the cell output of the cell requested by the AI, instead of whatever the active cell is. Often those are the same, but not always!

# Testing

See the new tests

# Documentation

No

### DIFF
--- a/mito-ai/src/Extensions/AiChat/utils.tsx
+++ b/mito-ai/src/Extensions/AiChat/utils.tsx
@@ -1,0 +1,28 @@
+import { INotebookTracker } from "@jupyterlab/notebook"
+import { getActiveCellOutput, getCellIndexByID, scrollToCell } from "../../utils/notebook"
+import { logEvent } from "../../restAPI/RestAPI"
+
+
+export const getBase64EncodedCellOutput = async (notebookTracker: INotebookTracker, cellID: string | undefined): Promise<string | undefined> => {
+    if (cellID === undefined) {
+        return undefined
+    }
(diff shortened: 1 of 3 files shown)


mito-ai/src/Extensions/AiChat/utils.tsx
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    // Check if the cell id is part of the notebook
+    const cellIndex = getCellIndexByID(notebookTracker, cellID)
+    
+    if (cellIndex === undefined) {
+        // Log that the cell id is not part of the notebook
+        console.log(`Cell with id ${cellID} not found in notebook`)
+        logEvent('get_cell_output_requested_non_existent_cell')
+        return undefined
```
sid-sagacollab, line 18: Design question here:

If the LLM hallucinates and messes up the cell ID by a bit, i.e some characters are wrong in the ID (since it often does that with line numbers and mito citations), would it not be sensible to find an ID that is the closest match to this ID and return that if the exact ID does not exist ? I know this is probably over-engineering it, but this issue is prevalent with mito citations atleast, and I was considering this solution for that issue.

aarondr77: Ohh interesting. This could be a reasonable idea to use some heuristic. I haven't paid attention to what types of hallucinations it makes -- like if its just completely off or still somewhat close.

PR #1856, opened 2025-07-14
Streamlit dashboard backend code

PoojasreeDwarakanath:
# Description

This PR contains the backend code for agent based Streamlit code generation and validation.

# Testing


mito-ai/src/utils/auth.ts
```diff
@@ -0,0 +1,210 @@
+/*
+ * Copyright (c) Saga Inc.
+ * Distributed under the terms of the GNU Affero General Public License v3.0 License.
+ */
+
+import { requestAPI } from '../restAPI/utils';
+
+// AWS Cognito configuration
+const COGNITO_CONFIG = {
+    SIGNUP_URL: 'https://mito-app-auth.auth.us-east-1.amazoncognito.com/signup?client_id=6ara3u3l8sss738hrhbq1qtiqf&response_type=code&scope=email+openid+profile&redirect_uri=http://localhost:8888/lab',
```
aarondr77: We don't always want to redirect users back to `http://localhost:8888/lab`. Here's a few examples: 

1. Sometimes the user is not running Jupyter on that port. ie: I just had it running on port 8889 a second ago and got very confused when it redirected me back to 8888. I was also running Jupyter on that port, but it was a different Jupyter environment. 
2. Other times, users are going to be logged into a JupyterHub (like launch.trymito.io) 

Can we get the current url and redirect back to that?

mito-ai/src/utils/auth.ts
```diff
@@ -0,0 +1,210 @@
...
+    SIGNIN_URL: 'https://mito-app-auth.auth.us-east-1.amazoncognito.com/login?client_id=6ara3u3l8sss738hbq1qtiqf&response_type=code&scope=email+openid+profile&redirect_uri=http://localhost:8888/lab',
```
aarondr77: Same url thing here. We can abstract that into a constant variable

mito-ai/mito_ai/app_builder/handlers.py
```diff
@@ -115,12 +124,40 @@ async def _handle_build_app(self, message: dict) -> None:
...
+                return
+            else:
+                self.log.info("JWT token validation successful")
+        else:
+            self.log.warning("No JWT token provided or using placeholder token")
+
         try:
             # This is a placeholder for the actual app building logic
             # In a real implementation, this would deploy the app to a hosting service
             # and return the URL
```
aarondr77: I guess this is out of date now!

mito-ai/mito_ai/streamlit_conversion/streamlit_agent_handler.py
```diff
@@ -0,0 +1,110 @@
...
+from typing import List
+
+from mito_ai.logger import get_logger
+from mito_ai.streamlit_conversion.streamlit_system_prompt import streamlit_system_prompt
+from mito_ai.streamlit_conversion.validate_and_run_streamlit_code import streamlit_code_validator
+from mito_ai.streamlit_conversion.streamlit_utils import extract_code_blocks, create_app_file, parse_jupyter_notebook_to_extract_required_content
+from mito_ai.utils.anthropic_utils import stream_anthropic_completion_from_mito_server
+from mito_ai.completions.models import MessageType
+
+STREAMLIT_AI_MODEL = "claude-sonnet-4-20250514"
```
aarondr77: It took 45 seconds to build my streamlit app and I only had about 100 lines of code in my noteook. I'm concerend about how slow this is going to get. I think the biggest win is just using a faster model. Let's try using `ANTHROPIC_FAST_MODEL = "claude-3-5-haiku-latest"`

mito-ai/mito_ai/streamlit_conversion/streamlit_agent_handler.py
```diff
@@ -0,0 +1,110 @@
...
+
+    async def get_response_from_agent(self, message_to_agent: List[MessageParam]) -> str:
+        """Gets the streaming response from the agent using the mito server"""
+        model = STREAMLIT_AI_MODEL
+        max_tokens = 64_000
+        temperature = 0.2
+
+        self.log.info("Getting response from agent...")
+        accumulated_response = ""
+        async for stream_chunk in stream_anthropic_completion_from_mito_server(
```
aarondr77, line 43: I think this is okay for now. However, I do think that this breaks the mental model of how our completions work right now. The reason I say this is because we want the entire extension to be able to run from just one llm provider. That way, if someone wants to use it by providing their own API keys for OpenAI, for example, they can. With this implementation they would need to always provide an athropic key as well. 

This isnt so bad because they are using a feature that is going to deploy their app on the mito server anyways, so if they ended up defaulting to mito server, that would probably be fine!

aarondr77: Btw, make sure you merge dev into this branch because I think this function signature has changed a bit. My recconmendation is to merge dev into your branch at least once a day since we're adding new code so frequently !

mito-ai/mito_ai/streamlit_conversion/streamlit_agent_handler.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+    async def correct_error_in_generation(self, error: str) -> str:
+        """If errors are present, send it back to the agent to get corrections in code"""
+        self.messages.append(
+            {
+                "role": "user",
+                "content": [{
+                    "type": "text",
+                    "text": f"I am facing the following error. Correct it for me - {error}"
```
aarondr77: Maybe the prompt can be a little bit more descriptive. Something like: 

"""
When I run the streamlit app, I get the following error: 

{error}

Please return the FULL STREAMLIT app with the error corrected

{streamlit_app_code}
"""

mito-ai/src/Extensions/AppBuilder/NotebookToStreamlit.tsx
```diff
@@ -30,6 +31,13 @@ export const convertNotebookToStreamlit = async (
   notebookTracker: INotebookTracker,
   appBuilderService?: IAppBuilderService,
 ): Promise<void> => {
+  // Check authentication before proceeding with deployment
```
aarondr77, line 23: Note for the future when we implement the UI, we should be starting to notify the user that their app is getting processed as soon as this function is called. Currently, waiting until the streamlit app.py file is created is pretty confusing. Especially since it takes a long time!

mito-ai/mito_ai/auth/handlers.py
```diff
@@ -0,0 +1,196 @@
...
+        try:
+            # Basic JWT format validation (header.payload.signature)
+            if not token or '.' not in token:
+                return False
+            
+            parts = token.split('.')
+            if len(parts) != 3:
+                return False
+            
+            # TODO: Add proper JWT validation using AWS Cognito public keys
```
aarondr77: For TODO's like this, let's make sure to add them to the Notion doc so we don't foget to come back to them. I think there's probably going to be lots and lots of small details we need to keep track of

mito-ai/mito_ai/streamlit_conversion/streamlit_system_prompt.py
```diff
@@ -0,0 +1,40 @@
...
+3. Maintain the logical flow and structure of the original analysis
+4. Create an executive-friendly dashboard suitable for company leadership
+
+STREAMLIT IMPLEMENTATION GUIDELINES:
+- Use appropriate Streamlit components (st.title, st.header, st.subheader, st.markdown, etc.)
+- Display all visualizations using st.pyplot(), st.plotly_chart(), or st.altair_chart() as appropriate
+- Show dataframes and tables using st.dataframe() or st.table()
+- Include all text explanations and insights from markdown cells
+- Add interactive elements where beneficial (filters, selectors, etc.)
+- Ensure professional styling and layout suitable for executives
```
aarondr77, line 23: Let's do a little bit of prompt engineering (or add it as a todo in notion) to make these apps more professional looking. 

Some of the apps I've seen so far have:
1. A ton of emojis across the app
2. Overuse of the sidebase with a bunch of configuration options in it


I think while all of this could be intersting to explore more, the more complex the apps are, the more users are going to want to customize the apps. Since we don't have capacity to offer that to users right now, I think we want to go with a minimal app (probably always a single page app that is more linearly layed out, with professional colors, no emojis, etc.) Its a bit more boring, but at least it will get the information across and since we're using AI it should actually succeed in creating the app more often then our manual conversion script. 

TLDR: The more creative we let the AI be, the more users are going to want to control that creativity I think.

mito-ai/mito_ai/streamlit_conversion/streamlit_system_prompt.py
```diff
@@ -0,0 +1,40 @@
...
+- Organize content with clear sections and headers
+- Add error handling where appropriate
+- Include comments explaining key sections
+
+OUTPUT FORMAT:
+- Provide the complete app.py file code
+- Ensure all notebook outputs are faithfully reproduced
+- Make the dashboard professional and presentation-ready
+- Focus on clarity and executive-level communication
+- Don't give extra explanations, just give the python code
```
aarondr77, line 37: Yeah, I think this part of the prompt is very important to get across clearly: Its job is to take the existing notebook and make it into a streamlit app. IT SHOULD NOT:
1. Edit the analysis / graphs / etc.
2. Create its own explanations or interpretations of the data 
3. etc. 

Let's make sure that this is very clear to the LLM.

This streamlit conversion agent just doesn't enough context to know what it should be doing.

mito-ai/mito_ai/streamlit_conversion/streamlit_utils.py
```diff
@@ -0,0 +1,88 @@
...
+        with open(file_path+"/app.py", 'w') as f:
+            f.write(code)
+        return True, f"Successfully created {file_path}"
+    except IOError as e:
+        return False, f"Error creating file: {str(e)}"
+    except Exception as e:
+        return False, f"Unexpected error: {str(e)}"
+
+
+def parse_jupyter_notebook_to_extract_required_content(notebook_path: str) -> Dict:
```
aarondr77: Another way that we could've done this is by using the frontend code we already have for building a summary of the notebook, but this works too! 

Can we leave as a TODO in notion to not include cells that have the metadata tag to exclude it from the streamlit app. Currently, if the user presses that little include/exclude button in the toolbar of a cell it toggles the metadata. We should either support that here or remove those buttons for the MLV

mito-ai/mito_ai/streamlit_conversion/validate_and_run_streamlit_code.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+# Usage example
+def streamlit_code_validator(app_code):
+    """Convenience function to validate Streamlit code"""
+    has_validation_error = False
+    error_message = "Errors found: "
+
+    if "error" in app_code:
+        return has_validation_error, error_message
```
aarondr77: This code seems wrong to me. 

If "error" is part of the app_code, then we return has_validation_error = False. Is that what we want?

mito-ai/mito_ai/streamlit_conversion/validate_and_run_streamlit_code.py
```diff
@@ -0,0 +1,239 @@
...
+            'errors': []
+        }
+
+        try:
+            # Step 1: Check syntax
+            syntax_valid, syntax_msg = self.validate_syntax(app_code)
+            results['syntax_valid'] = syntax_valid
+            if not syntax_valid:
+                results['errors'].append(syntax_msg)
+                return results
```
aarondr77, line 151: Do we want to be returning here or continuing to build the full list of errors? 

Ie: We could return a list of syntax errors and import errors at the same tie and hopefully get them fixed in one go. 

App starts errors don't make sense to share until we fix all the import and syntax errors.

mito-ai/mito_ai/auth/handlers.py
```diff
@@ -0,0 +1,196 @@
...
+                headers={'Content-Type': 'application/x-www-form-urlencoded'}
+            )
+            
+            if response.status_code == 200:
+                token_response = response.json()
+                
+                # Add detailed logging for debugging timezone issues
+                from datetime import datetime, timezone
+                import base64
+                import json
```
aarondr77: Let's move to the imports to the top of the file

mito-ai/mito_ai/auth/handlers.py
```diff
@@ -0,0 +1,196 @@
...
+            data = json.loads(self.request.body)
+            code = data.get('code')
+            
+            if not code:
+                self.set_status(400)
+                self.finish(json.dumps({"error": "Authorization code is required"}))
+                return
+            
+            # Exchange authorization code for tokens
+            token_response = self._exchange_code_for_tokens(code)
```
aarondr77, line 36: Can you add a bit more documentation about how this authentication flow works? In particular, I'm not super clear on the purpose of the code and token. 

Is the code returned as the url parameter when you sign in and then we convert that into a token which we use to 1) authenticate to aws cognito when we go to deploy the app and 2) store it in the browser so we have the token the next time the user wants to deploy the app?

mito-ai/src/utils/auth.ts
aarondr77, line 1: Maybe we can move this into the Extensions/AppBuilder folder so its more clear where this code is being used?

mito-ai/src/utils/auth.test.ts
aarondr77, line 1: Our convention right now is that the tests live in the serc/tests/ folder. Let's move this there for now so we have all our tests in one spot.

mito-ai/mito_ai/app_builder/handlers.py
```diff
@@ -137,12 +174,47 @@ async def _handle_build_app(self, message: dict) -> None:
                 error=error
             ))
     
-    async def _deploy_app(self, app_path: str) -> str:
+    def _validate_jwt_token(self, token: str) -> bool:
```
aarondr77, line 173: It looks like this function exists here and also in `mito-ai/mito_ai/auth`. Let's make a helper function and use it in both places!

mito-ai/src/Extensions/AppBuilder/NotebookToStreamlit.tsx
```diff
@@ -115,29 +127,27 @@ export const convertNotebookToStreamlit = async (
     }
   });
 
-  // Create the streamlit app.py file
-  const streamlitSourceCode = streamlitCode.join('\n');
```
aarondr77, line 119: We still have all this code to actually build the app deterministically. Let's delete it all if we're not using it.

mito-ai/mito_ai/auth/handlers.py
```diff
@@ -0,0 +1,196 @@
...
+                                self.log.debug(f"Issued at: {current_time.isoformat()}")
+                                self.log.debug(f"Expires at: {exp_time.isoformat()}")
+                                self.log.debug(f"Expires in: {time_until_expiry / 60:.1f} minutes")
+                                self.log.debug(f"Token lifetime: {token_response.get('expires_in', 'unknown')} seconds")
+                    except Exception as e:
+                        self.log.warning(f"Could not decode JWT token for logging: {e}")
+                
+                return token_response
+            else:
+                self.log.error(f"Token exchange failed: {response.status_code} - {response.text}")
```
aarondr77, line 88: This sort of logging doesn't get sent to our analytics -- it just shows up in the jupyter server. 

We don't need to do it right now, but we should make sure that we have logs through the telemetry_utils file as well. Let's add this to our TODOs in Notion.

mito-ai/mito_ai/streamlit_conversion/streamlit_system_prompt.py
```diff
@@ -0,0 +1,40 @@
...
+- Include all text explanations and insights from markdown cells
+- Add interactive elements where beneficial (filters, selectors, etc.)
+- Ensure professional styling and layout suitable for executives
+
+CODE STRUCTURE:
+- Generate a complete, runnable app.py file
+- Include all necessary imports
+- Handle data loading and processing
+- Organize content with clear sections and headers
+- Add error handling where appropriate
```
aarondr77: I think this makes it sound like it should be editing the analysis to handle errors. Is that what we want? I think probably no for now until we have a better understanding of what these errors might be.

mito-ai/mito_ai/streamlit_conversion/validate_and_run_streamlit_code.py
```diff
@@ -0,0 +1,239 @@
...
+    def validate_syntax(self, app_code):
+        """Check if the Python code has valid syntax"""
+        try:
+            ast.parse(app_code)
+            return True, "Syntax is valid"
+        except SyntaxError as e:
+            error_msg = ''.join(traceback.format_exception(type(e), e, e.__traceback__))
+            return False, f"Syntax error: {error_msg}"
+
+    def check_imports(self, app_code):
```
aarondr77: A few thoughts on this check: 

1. Since we're not allowing the AI to edit the requirement.txt file right now, what do we expect to happen if this check fails?
2. Based on my understanding of what is going on here, it seems like we are not checking the right thing because this is the not mirroring the environment that the streamlit app is going to be hosted in aka: we are not installing the requirements.txt before doing trying to import the modules?

aarondr77 (commented):
Sweet! The authentication stuff seems to be working well, and I got a streamlit app up and running! In aws, I got an error that said matplotlib is not defined. But fixing that seems like it is out of scope for this PR, right? 

There's a few changes that I wrote about here. We don't need to make all of them right now -- use your judgment on what we should fix in this PR vs what we should add to our Notion doc as a later TODO. I just want to make sure we are tracking them!

mito-ai/mito_ai/streamlit_conversion/streamlit_agent_handler.py
PoojasreeDwarakanath: Added a TODO for supporting other models and API keys
Also synced changes with the latest dev branch

mito-ai/mito_ai/app_builder/handlers.py
PoojasreeDwarakanath: Removed this function being called in 2 places. Now the token only gets validated in 1 place

mito-ai/mito_ai/auth/handlers.py
PoojasreeDwarakanath: Added documentation for this at mito-ai/mito_ai/auth/README.md

mito-ai/mito_ai/auth/handlers.py
PoojasreeDwarakanath: This logging was to help me debug the cookie issue I was facing. I thought I had removed it before committing but looks like I pushed it by mistake. I have removed this logic as it isnt needed.

I have also added a TODO for telemetry

mito-ai/mito_ai/streamlit_conversion/streamlit_system_prompt.py
PoojasreeDwarakanath: Modified the system prompt in a couple of places.
Also added a TODO to make the prompt better as we go forward

mito-ai/mito_ai/streamlit_conversion/streamlit_utils.py
PoojasreeDwarakanath: Added this TODO in the notion doc

mito-ai/mito_ai/streamlit_conversion/validate_and_run_streamlit_code.py
PoojasreeDwarakanath: Removed import check as discussed on call. We assume that the user has the list of required imports to support running the dashboard locally

mito-ai/src/Extensions/AppBuilder/NotebookToStreamlit.tsx
PoojasreeDwarakanath: Added a TODO

mito-ai/mito_ai/streamlit_conversion/streamlit_system_prompt.py
PoojasreeDwarakanath: Modified the prompt to be more clear

mito-ai/mito_ai/streamlit_conversion/validate_and_run_streamlit_code.py
PoojasreeDwarakanath: Removing this check for now

mito-ai/mito_ai/streamlit_conversion/streamlit_agent_handler.py
PoojasreeDwarakanath: Added this change to the prompt

mito-ai/mito_ai/streamlit_conversion/streamlit_agent_handler.py
PoojasreeDwarakanath: Modified the model

mito-ai/mito_ai/auth/handlers.py
PoojasreeDwarakanath: Done!

mito-ai/src/utils/auth.ts
PoojasreeDwarakanath: Done!

mito-ai/src/utils/auth.ts
PoojasreeDwarakanath: Modified the code to get current url and redirect back to that url

mito-ai/mito_ai/app_builder/handlers.py
PoojasreeDwarakanath: Removed the comments

mito-ai/src/Extensions/AppBuilder/NotebookToStreamlit.tsx
PoojasreeDwarakanath: Removed this code

PoojasreeDwarakanath: Deleted this code

mito-ai/mito_ai/auth/handlers.py
PoojasreeDwarakanath: I had used this for debugging and this function is no more required. I thought I had removed it before committing but apparently it got pushed. Removing this function

mito-ai/mito_ai/constants.py
```diff
@@ -46,4 +46,20 @@
 
 # Set ACTIVE_BASE_URL manually
 # TODO: Modify to PROD url before release
-ACTIVE_STREAMLIT_BASE_URL = MITO_STREAMLIT_TEST_BASE_URL  # Change to MITO_STREAMLIT_DEV_BASE_URL for dev
+ACTIVE_STREAMLIT_BASE_URL = MITO_STREAMLIT_DEV_BASE_URL  # Change to MITO_STREAMLIT_DEV_BASE_URL for dev
+
+SIGNUP_DEV_URL = "https://mito-app-auth.auth.us-east-1.amazoncognito.com/signup?client_id=6ara3u3l8sss738hrhbq1qtiqf&response_type=code&scope=email+openid+profile&redirect_uri=http://localhost:8888/lab"
+SIGNIN_DEV_URL = "https://mito-app-auth.auth.us-east-1.amazoncognito.com/login?client_id=6ara3u3l8sss738hrhbq1qtiqf&response_type=code&scope=email+openid+profile&redirect_uri=http://localhost:8888/lab"
```
aarondr77: Do we need to do something with the redirect urls here? They are both still `redirect_uri=http://localhost:8888/lab`?

mito-ai/mito_ai/utils/anthropic_utils.py
```diff
@@ -104,23 +104,25 @@ async def stream_anthropic_completion_from_mito_server(
...
-    message_id: str,
+    reply_fn: Optional[Callable[[Union[CompletionReply, CompletionStreamChunk]], None]] = None,
+    message_id: Optional[str] = None,
 ) -> AsyncGenerator[str, None]:
     data, headers = _prepare_anthropic_request_data_and_headers(
         model, max_tokens, temperature, system, messages, message_type, None, None, stream
     )
-    
     # Use the unified streaming function
+    # Provide default no-op for reply_fn and empty string for message_id if not provided
```
aarondr77: Can you just update this comment to explain why we're doing this so we don't forget that this is handling the special case of the streamlit converstion

aarondr77 (commented):
A few tiny comments and then LGTM! I added `streamlit` to the pyrpoject.toml since it is required to run the app locally. Without it, my app was not ever starting correctly

Issue #1847, opened 2025-07-09
Claude Not Showing Preview Code Button


ngafar:
When using Claude models in chat mode, only the first message has the preview button. Subsequent messages only have a copy button. This issue seems to be limited to Claude models. 

To recreate:

1. Start a new chat, and select a Claude model. 
2. Send a message
3. Preview and accept code.
4. Send a second message.

You should notice that the second message does not have a preview button.

aarondr77:
I got this bug again. Here is the conversation that produced the error: 
- Chat mode
- Claude 4 Sonnet

Notice, there is no overwrite button here

<img width="692" height="201" alt="Image" src="https://github.com/user-attachments/assets/bbd0b494-f8fe-48c1-bda5-202c781fd756" />


```
    {
      "role": "user",
[...]

PR #1858, opened 2025-07-14
[WIP] Bug Fix: No Preview Code Button w. Claude

ngafar:
# Description

Fixes https://github.com/mito-ds/mito/issues/1847

This PR creates a new, reusable `stream_response_from_mito_server` function. 

# Testing

Send messages using each of the supported models. Be sure to accept the suggest code, then send a follow up message. The followup message should have the preview/copy buttons that were previously missing.

# Documentation

N/A - Changes only made to the backend, nothing user-facing was introduced.

### DIFF
--- a/mito-ai/mito_ai/utils/mito_server_utils.py
+++ b/mito-ai/mito_ai/utils/mito_server_utils.py
@@ -1,12 +1,13 @@
 # Copyright (c) Saga Inc.
 # Distributed under the terms of the GNU Affero General Public License v3.0 License.
 
-from mito_ai.completions.models import MessageType
+import asyncio
+import json
+import time
+from typing import Any, Dict, Optional, Callable, Union, AsyncGenerator
+from mito_ai.completions.models import MessageType, CompletionReply, CompletionStreamChunk, CompletionItem
 from mito_ai.utils.server_limits import check_mito_server_quota, update_mito_server_quota
 from tornado.httpclient import HTTPResponse
-import time
-import json
-from typing import Any, Dict, Optional
 from mito_ai.constants import MITO_GEMINI_URL
 from mito_ai.utils.utils import _create_http_client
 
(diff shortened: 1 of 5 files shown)


mito-ai/mito_ai/utils/mito_server_utils.py
```diff
@@ -88,5 +89,147 @@ async def get_response_from_mito_server(
...
+                done=True,
+            ))
+    except Exception as e:
+        print(f"\n{provider_name} stream failed after {time.time() - start_time:.2f} seconds with error: {str(e)}")
+        # If an exception occurred, ensure the fetch future is awaited to properly clean up
+        if fetch_future:
+            try:
+                await fetch_future
+            except Exception:
+                pass
```
aarondr77, line 221: Do we think this is the right thing to be doing? I've sometimes seen issues where the AI just hangs for a long time on the `thinking` message without ever sending anything back. I wonder if this is the culprit?

mito-ai/mito_ai/utils/mito_server_utils.py
```diff
@@ -88,5 +89,147 @@ async def get_response_from_mito_server(
...
+                        parent_id=message_id,
+                        chunk=CompletionItem(
+                            content=processed_chunk,
+                            isIncomplete=True,
+                            token=message_id,
+                        ),
+                        done=False,
+                    ))
+                
+                yield chunk  # Yield original chunk for backward compatibility
```
aarondr77: I think this comment is wrong. The reason we are yielding the chunks is so that the backend can collect the full Ai response and save it the chat-history.json file.

mito-ai/mito_ai/utils/mito_server_utils.py
```diff
@@ -88,5 +89,147 @@ async def get_response_from_mito_server(
...
+async def stream_response_from_mito_server(
+    url: str,
+    headers: Dict[str, str],
+    data: Dict[str, Any],
+    timeout: int,
+    max_retries: int,
+    message_type: MessageType,
+    provider_name: str = "Mito Server",
+    reply_fn: Optional[Callable[[Union[CompletionReply, CompletionStreamChunk]], None]] = None,
+    message_id: Optional[str] = None,
```
aarondr77: Is there any reason that these should be Optional? seems like they are pretty required! That's where the bug came from.

aarondr77 (commented):
Seems to be working well, once we add some basic tests, then LGTM

mito-ai/mito_ai/utils/mito_server_utils.py
ngafar: Maybe we save this for a future investigation, just to keep a tight scope on this PR.

PR #1859, opened 2025-07-14
Mitosheet helper 4c2a

aarondr77:
# Description

# Testing

# Documentation


mitosheet/mitosheet/__init__.py
```diff
@@ -40,6 +40,8 @@
 from mitosheet.user import initialize_user
 initialize_user()
 
+print("HERE!")
```
aarondr77: remove

PR #1862, opened 2025-07-15
[WIP] Add Context Button

ngafar:
# Description

# Testing

# Documentation

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatMessage/ChatDropdown.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatMessage/ChatDropdown.tsx
@@ -3,7 +3,7 @@
  * Distributed under the terms of the GNU Affero General Public License v3.0 License.
  */
 
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import { classNames } from '../../../utils/classNames';
 import { ExpandedVariable } from './ChatInput';
 import { getRules } from '../../../restAPI/RestAPI';
@@ -14,6 +14,8 @@ interface ChatDropdownProps {
     filterText: string;
     maxDropdownItems?: number;
     position?: 'above' | 'below';
+    showSearchInput?: boolean;
+    onFilterChange?: (filterText: string) => void;
 }
 
 interface ChatDropdownVariableOption {
@@ -33,8 +35,12 @@ const ChatDropdown: React.FC<ChatDropdownProps> = ({
     onSelect,
     filterText,
     maxDropdownItems = 10,
+    showSearchInput = false,
+    onFilterChange,
 }) => {
     const [selectedIndex, setSelectedIndex] = useState(0);
+    const [localFilterText, setLocalFilterText] = useState(filterText);
+    const searchInputRef = useRef<HTMLInputElement>(null);
 
     const [rules, setRules] = useState<string[]>([]);
 
@@ -45,6 +51,16 @@ const ChatDropdown: React.FC<ChatDropdownProps> = ({
         };
         void fetchRules();
     }, []);
+
+    // Focus search input when dropdown opens with search input
+    useEffect(() => {
+        if (showSearchInput && searchInputRef.current) {
+            searchInputRef.current.focus();
+        }
+    }, [showSearchInput]);
+
+    // Use local filter text when search input is shown, otherwise use prop
+    const effectiveFilterText = showSearchInput ? localFilterText : filterText;
 
     // Create a list of all options with the format 
     // ['type': 'variable', "expandedVariable": variable]
@@ -62,16 +78,16 @@ const ChatDropdown: React.FC<ChatDropdownProps> = ({
 
     const filteredOptions = allOptions.filter((option) =>
         option.type === 'variable' ?
-            option.variable.variable_name.toLowerCase().includes(filterText.toLowerCase()) &&
+            option.variable.variable_name.toLowerCase().includes(effectiveFilterText.toLowerCase()) &&
             option.variable.type !== "<class 'module'>" &&
             option.variable.variable_name !== "FUNCTIONS" // This is default exported from mitosheet when you run from mitosheet import * as FUNCTIONS
         :
-            option.rule.toLowerCase().includes(filterText.toLowerCase())
+            option.rule.toLowerCase().includes(effectiveFilterText.toLowerCase())
     ).slice(0, maxDropdownItems);
 
     useEffect(() => {
         setSelectedIndex(0);
-    }, [options, rules, filterText]);
+    }, [options, rules, effectiveFilterText]);
 
     const handleKeyDown = (event: KeyboardEvent): void => {
         switch (event.key) {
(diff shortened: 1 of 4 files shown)


mito-ai/mito_ai/completions/prompt_builders/utils.py
```diff
@@ -4,19 +4,63 @@
...
-    rules_str = ''
+    selected_rules = [
+        context.split("Rule: ")[1] # Remove the "Rule: " prefix
+        for context in additional_context
+        if context.startswith("Rule:")
+    ]
+    if len(selected_rules) == 0:
+        return ""
+
+    print("selected_rules", selected_rules)
```
aarondr77: delete

mito-ai/src/Extensions/AiChat/ChatMessage/ChatDropdown.tsx
```diff
@@ -46,32 +59,54 @@ const ChatDropdown: React.FC<ChatDropdownProps> = ({
...
 
+    // Focus search input when dropdown opens with search input
+    useEffect(() => {
+        if (showSearchInput && searchInputRef.current) {
+            searchInputRef.current.focus();
+        }
+    }, [showSearchInput]);
+
+    // Use local filter text when search input is shown, otherwise use prop
+    const effectiveFilterText = showSearchInput ? localFilterText : filterText;
```
aarondr77: Instead of having this conditional on `showSearchInput` can we have it conditional on how the chatDrodown was opened? Just so if we change the showSearchInput logic at some point we don't break this. 

I think to do this, we can calculate showSearchInput inside this component instead since its not something that the ChatInput needs to worry about.

aarondr77: Ohh it actually looks like showSearchInput is just the same as `isDropdownFromButton`, can we just call this prop `isDropdownFromButton` instead?

mito-ai/src/Extensions/AiChat/ChatMessage/ChatDropdown.tsx
```diff
@@ -216,6 +311,43 @@ const VariableDropdownItem: React.FC<VariableDropdownItemProps> = ({ variable, i
...
 }
 
+interface FileDropdownItemProps {
+    file: ExpandedVariable;
+    index: number;
+    selectedIndex: number;
+    onSelect: (file: ExpandedVariable) => void;
+}
+
+const FileDropdownItem: React.FC<FileDropdownItemProps> = ({ file, index, selectedIndex, onSelect }) => {
```
aarondr77: Can we create a folder called `ChatDropdownItems` and have this in it? This file is just getting a bit big.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatDropdown.tsx
```diff
@@ -216,6 +311,43 @@ const VariableDropdownItem: React.FC<VariableDropdownItemProps> = ({ variable, i
...
+                    }
+                }}
+            >
+                {file.variable_name}
+            </span>
+        </li>
+    )
+}
+
 interface RuleDropdownItemProps {
```
aarondr77: And we can move this and the rest to the ChatDropdownItems folder too

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -134,13 +158,17 @@ const ChatInput: React.FC<ChatInputProps> = ({
             } else {
                 contextChatRepresentation = `\`${option.variable.variable_name}\``
             }
+        } else if (option.type === 'file') {
+            // For files, add them as both back-ticked elements and the additional context container
+            contextChatRepresentation = `\`${option.file.variable_name}\``
+            setAdditionalContext([...additionalContext, `File: ${option.file.variable_name}`]);
```
aarondr77: If we want to type these different types of context (which sounds like a great call to me), can we just update additionalContext from a list of strings to a list of dict objects that have 

```
{
type: string, 
value: string
}
```

It will make it more robust instead of having to parse these values and make sure we get the string formatting correctly.

mito-ai/style/SelectedContextContainer.css
```diff
@@ -3,6 +3,32 @@
...
 
+.context-container {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  gap: 4px;
+  padding: 8px 10px;
+}
+
+.add-context-button {
```
aarondr77: I think this class maybe belongs in the ChatInput stuff instead of here? Its not used the SelectedContextContainer

mito-ai/style/SelectedContextContainer.css
```diff
@@ -3,6 +3,32 @@
  * Distributed under the terms of the GNU Affero General Public License v3.0 License.
  */
 
+.context-container {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
```
aarondr77, line 9: Are things vertically centered? The X looks low to me? sorry! Don't spend more than 5 mins on it

Image

mito-ai/src/Extensions/AiChat/ChatMessage/ChatMessage.tsx
```diff
@@ -225,6 +228,17 @@ const ChatMessage: React.FC<IChatMessageProps> = ({
                                     </div>
 
                                 }
+                                {message.role === 'user' && additionalContext && additionalContext.length > 0 && 
+                                    <>
+                                        {additionalContext.map((context) => (
```
aarondr77: Can we update the spacing between these?

I think: 

1. The spacing between the button and the context containers should be the same as the spacing between the context containers themselves
2. Reduce the spacing between the rows 

I think we can fix this just by specifying the margin rules on all children of the `context-container` using `context-container > *` or something like that. 

Image

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -109,14 +109,38 @@ const ChatInput: React.FC<ChatInputProps> = ({
...
+                // For variables, we'll add them as a special context type
+                const contextName = option.variable.parent_df 
+                    ? `${option.variable.parent_df}.${option.variable.variable_name}`
+                    : option.variable.variable_name;
+                setAdditionalContext(prev => [...prev, `Variable: ${contextName}`]);
+            } else if (option.type === 'file') {
+                setAdditionalContext(prev => [...prev, `File: ${option.file.variable_name}`]);
+            } else if (option.type === 'rule') {
+                setAdditionalContext(prev => [...prev, `Rule: ${option.rule}`]);
+            }
```
aarondr77: I think we could make this more information dense. `Variable`, etc. can we just use a different icon for each? One for variable, one for file, and one for rule?

mito-ai/src/Extensions/AiChat/ChatMessage/ChatDropdown.tsx
```diff
@@ -46,32 +59,54 @@ const ChatDropdown: React.FC<ChatDropdownProps> = ({
...
+            .filter(variable => variable.file_name) // Only files
+            .map((file): ChatDropdownFileOption => ({ 
+                type: 'file', 
+                file: file 
+            })),
         ...rules.map((rule): ChatDropdownRuleOption => ({ 
             type: 'rule', 
             rule: rule 
         })),
     ];
```
aarondr77: Can we do a really quick pass on sorting these in a more useful way. 

Maybe we can do Rules, Files, Dataframes, Columns, Variables?

mito-ai/style/ChatDropdown.css
```diff
@@ -91,3 +93,32 @@
 .dropdown-item-disabled:hover {
   background-color: inherit;
 }
+
+.chat-dropdown-search {
```
aarondr77, line 87: I think we should update the background color of the chat dropdown? It doesn't look like the search bar is part of it. I've missed the search bar a few times.

Image
Image

mito-ai/mito_ai/completions/prompt_builders/utils.py
ngafar: dang, and I did a ctrl+f for print statements!

PR #1863, opened 2025-07-15
[Spike] Assumptions

aarondr77:
# Description

# Testing

# Documentation


ngafar:
Here's an idea, for instances where there are multiple assumptions, we can have them displayed in one tool call div, and display them as bullet points.

image

ngafar:
The bullets look great. 

One other way I think about assumptions is that they are almost always variables that the user can edit. For example, if you wanted to "filter by cheap stocks," you might set a variable called `price_to_earnings` that you filter against. Maybe a helpful idea for the system prompt?

PR #1867, opened 2025-07-22
Databases in Dropdown Menu

ngafar:
# Description

- Added `alias` field to database connections.
- Added databases to dropdown.
- Updated chat and agent prompts to make reference of selected databases (connection id and type). 

# Testing

Reference a database using the dropdown menu.

# Documentation

Yes, the [Using Your Database in Chat](https://docs.trymito.io/mito-ai/database-connectors#using-your-database-in-chat) section should be updated to let users know they can use the dropdown to reference a specific db.


mito-ai/src/Extensions/SettingsManager/database/model.ts
```diff
@@ -28,12 +28,21 @@ export interface DatabaseConfig {
...
+const ALIAS_PLACEHOLDER = 'Enter a nickname for this database';
+
 export const databaseConfigs: Record<string, DatabaseConfig> = {
     mssql: {
         type: 'mssql',
         displayName: 'Microsoft SQL Server',
         alertText: 'Microsoft SQL Server requires an additional driver. If you\'ve already installed it, you can safely ignore this message. For more info, consult the <a href="https://docs.trymito.io/mito-ai/database-connectors/microsoft-sql-server" target="_blank">Mito docs</a>.',
         fields: [
+            {
+                name: 'alias',
```
aarondr77, line 40: I think alias is maybe too technical? What if we just call this `nickname` to make it clear that this is just any name they want to give it as opposed to some name they need to go find.

mito-ai/style/SelectedContextContainer.css
```diff
@@ -12,41 +12,42 @@
...
+  text-overflow: ellipsis;
+  max-width: 200px;
 }
 
 .selected-context-container .icon {
-    width: 12px;
-    height: 12px;
+  width: 12px;
+  max-height: 14px;
+  align-self: center;
```
aarondr77, line 52: I just centered these things a little bit more vertically

aarondr77 (approved):

mito-ai/src/Extensions/SettingsManager/database/model.ts
ngafar: I agree; the name `alias` did seem too technical at first blush. But it seems like the accepted nomenclature across database apps. FWIW, the placeholder text is "enter a nickname," which should give users an idea.

PR #1868, opened 2025-07-22
Add Database Button

ngafar:
# Description

Added a new "add database" button in chat, with a color status notification.

# Testing

- Remove all databases, should show a yellow dot.
- Add a database, should show green dot.
- Use button, make sure it takes you to db settings page.

# Documentation

Probably not necessary, users can just click and see what happens.


mito-ai/style/IconButton.css
```diff
@@ -10,4 +10,27 @@
...
+
+.notification-dot-success {
+  background-color: var(--green-600);
+}
+
+.notification-dot-warning {
+  background-color: var(--yellow-600);
+}
+
+.notification-dot-danger {
```
aarondr77: I don't think we're using this

aarondr77 (approved):
I did a bit of polish. Checking that tests pass and then LGTM

Issue #1869, opened 2025-07-24
Assumptions Repeated


ngafar:
The AI will sometime make an assumption, which is correct, but continuously make note of it. Below is a screenshot where the assumption is mentioned six times. 

We could keep track of the assumptions in an array on the frontend, and not display duplicates. 

<img width="1445" height="993" alt="Image" src="https://github.com/user-attachments/assets/4569a987-0b21-442d-b924-da1a42818b57" />

PR #1871, opened 2025-07-25
Remove Duplicate Agent Assumptions 

ngafar:
# Description

Fixes https://github.com/mito-ds/mito/issues/1869

# Testing

Start a new Agent chat, work with it, and keep a note of the assumptions the Agent makes. You should not see duplicate assumptions.

You can switch branches, and refresh the chat to see the number of assumptions that were removed as part of the deduplication process. 

# Documentation

No, all changes are behind the scenes.

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
@@ -56,15 +56,7 @@ export class ChatHistoryManager {
     private displayOptimizedChatHistory: IDisplayOptimizedChatItem[];
     private contextManager: IContextManager;
     private notebookTracker: INotebookTracker;
-    private _allAssumptions: Set<string> = new Set();
-
-    get allAssumptions(): Set<string> {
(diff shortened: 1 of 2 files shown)


mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
```diff
@@ -299,13 +339,14 @@ export class ChatHistoryManager {
     }
 
     addAIMessageFromAgentResponse(agentResponse: AgentResponse): void {
+        const deduplicatedAgentResponse = this.deduplicateAssumptions(agentResponse);
```
aarondr77: Small nitpick but maybe we can just called this `agentResponseWithDeduplicatedAssumptions` just so it's more clear what is actually deduplicated.

aarondr77 (commented):
LGTM can we just add a few jest tests to make sure: 

1. When we call the `createDuplicateChatHistoryManager` function, we preserve the assumptions. 
2. The deduplication works correctly.

PR #1875, opened 2025-07-29
selectedRules Bug Fix

ngafar:
# Description

Trying to fix an error where a user is seeing:

```python
builtins.TypeError: AgentExecutionMetadata.__init__() got an unexpected keyword argument 'selectedRules'"
```

I can not recreate this issue without manually editing `agentExecutionMetadata` in the ChatHistoryManager, and adding a `selectedRules` field. 

However, I did notice that there are a few areas where `selectedRules` was being used as a variable name, when it should have been called `additionalContext`. The changes in this PR should not affect functionality, I've only renamed the variables for the sake of clarity. 

Also fixed a bug that was removing any context added on edit. 

# Testing

- Send an agent message with a rule.
- Edit an agent message with a rule. 

# Documentation

No, behind the scenes change.

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
@@ -24,7 +24,7 @@ interface ChatInputProps {
     app: JupyterFrontEnd;
     initialContent: string;
     placeholder: string;
-    onSave: (content: string, index?: number, selectedRules?: Array<{type: string, value: string}>) => void;
+    onSave: (content: string, index?: number, additionalContext?: Array<{type: string, value: string}>) => void;
     onCancel?: () => void;
     isEditing: boolean;
     contextManager?: IContextManager;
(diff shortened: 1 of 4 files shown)


mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -24,7 +24,7 @@ interface ChatInputProps {
     app: JupyterFrontEnd;
     initialContent: string;
     placeholder: string;
-    onSave: (content: string, index?: number, selectedRules?: Array<{type: string, value: string}>) => void;
+    onSave: (content: string, index?: number, additionalContext?: Array<{type: string, value: string}>) => void;
```
aarondr77, line 27: Hmm the rules don't seem to get applied on save for me.. Notice that the rule does not get applied when I edit the message, but it does get applied in a new message. 

Possible to add a test for this since we caught a bug?

https://github.com/user-attachments/assets/4d93f0cb-7d6d-4953-88c5-4c1b97d09719

PR #1877, opened 2025-07-30
Revert Logging + Questionnaire 

ngafar:
# Description

Logging when a user clicks revert, then showing a questionnaire.

# Testing

Send a question in Agent mode, and hit revert. 

# Documentation

N/A

### DIFF
--- a/mito-ai/mito_ai/__init__.py
+++ b/mito-ai/mito_ai/__init__.py
@@ -76,7 +76,7 @@ def _load_jupyter_server_extension(server_app) -> None: # type: ignore
     handlers.extend(get_db_urls(base_url))  # type: ignore
     handlers.extend(get_settings_urls(base_url))  # type: ignore
     handlers.extend(get_rules_urls(base_url))  # type: ignore
-    handlers.extend(get_log_urls(base_url))  # type: ignore
+    handlers.extend(get_log_urls(base_url, open_ai_provider.key_type))  # type: ignore
     handlers.extend(get_auth_urls(base_url))  # type: ignore
     handlers.extend(get_streamlit_preview_urls(base_url))  # type: ignore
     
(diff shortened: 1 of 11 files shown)


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1445,7 +1450,12 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
                                 text="Revert changes"
                                 icon={UndoIcon}
                                 title="Revert changes"
-                                onClick={() => restoreCheckpoint(app, notebookTracker, setHasCheckpoint, getDuplicateChatHistoryManager, setChatHistoryManager)}
+                                onClick={() => {
+                                    void restoreCheckpoint(app, notebookTracker, setHasCheckpoint)
+                                    setHasCheckpoint(false)
```
ngafar, line 1456: Do you know how I can clear the `nextSteps` array here? I've tried (with no luck):

```ts
setNextSteps([])
```

It would be nice to not overwhelm the user with choices at this stage.

aarondr77: This was happening because when you update a state variable, the app rerenders and it was just recalculating the nextSteps. So to make sure that we don't display the next steps we use the `setDisplayedNextStepsIfAvailable`

aarondr77: This was happening because when you update a state variable, the app rerenders and it was just recalculating the nextSteps. So to make sure that we don't display the next steps we use the `setDisplayedNextStepsIfAvailable`

mito-ai/mito_ai/__init__.py
```diff
@@ -76,7 +76,7 @@ def _load_jupyter_server_extension(server_app) -> None: # type: ignore
     handlers.extend(get_db_urls(base_url))  # type: ignore
     handlers.extend(get_settings_urls(base_url))  # type: ignore
     handlers.extend(get_rules_urls(base_url))  # type: ignore
-    handlers.extend(get_log_urls(base_url))  # type: ignore
+    handlers.extend(get_log_urls(base_url, open_ai_provider.key_type))  # type: ignore
```
aarondr77, line 79: Pro users weren't getting these logs (aka the confusing postgress log issue you saw today) because the frontend-triggered logging didn't know the key type that was being used. Now, we are passing the key type to the logging endpoint so we can use the same "If they are using the mito server, then we can log" logic.

PR #1884, opened 2025-08-01
Streamlit dashboard setup

PoojasreeDwarakanath:
# Description

This PR includes the following:
- If an app is already present from the preview, we deploy the same app instead of re-doing the agent flow to get the streamlit code
- If there are more than 10 files in a directory, we ask the user to create a new directory and process
- Fixing redirect_mismatch messages because of extra characters in the redirect url
- Fixing error where if the notebook was in the root, the notebook dir was resolved to '', and then it just errors saying that the file/ dir is not present. Fixed it by resolving to absolute path in such cases

### DIFF
--- a/mito-ai/mito_ai/streamlit_conversion/streamlit_utils.py
+++ b/mito-ai/mito_ai/streamlit_conversion/streamlit_utils.py
@@ -114,12 +114,15 @@ def parse_jupyter_notebook_to_extract_required_content(notebook_path: str) -> Di
         raise Exception(f"Error processing notebook: {str(e)}")
 
 
-def clean_directory_check(notebook_path: str) -> None:
+def resolve_notebook_path(notebook_path:str) -> str:
     # Convert to absolute path if it's not already absolute
     # Handle both Unix-style absolute paths (starting with /) and Windows-style absolute paths
     if not (notebook_path.startswith('/') or (len(notebook_path) > 1 and notebook_path[1] == ':')):
         notebook_path = os.path.join(os.getcwd(), notebook_path)
-    
+    return notebook_path
+
+def clean_directory_check(notebook_path: str) -> None:
+    notebook_path = resolve_notebook_path(notebook_path)
     # pathlib handles the cross OS path conversion automatically
     path = Path(notebook_path).resolve()
     dir_path = path.parent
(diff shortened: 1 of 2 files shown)


mito-ai/src/Extensions/AppBuilder/auth.ts
```diff
@@ -13,12 +13,21 @@ const client_id = '6ara3u3l8sss738hrhbq1qtiqf'
 const active_domain = domain_dev
 
 const currentUrl: string = window.location.href;
+let redirectUrl: string;
+
+if (currentUrl.includes('localhost')) {
+    redirectUrl = 'http://localhost:8888/lab';
```
aarondr77, line 21: Is it possible to redirect back to the original url instead? Sometimes users have Jupyter running on a different localhost port than 8888

mito-ai/mito_ai/app_builder/handlers.py
```diff
@@ -147,9 +147,16 @@ async def _handle_build_app(self, message: BuildAppRequest) -> None:
...
         try:
-
             notebook_path = str(notebook_path) if notebook_path else ""
-            success_flag, app_path, result_message = await streamlit_handler(notebook_path)
+
+            app_directory = os.path.dirname(notebook_path)
+            app_path = os.path.join(app_directory, "app.py")
+            if os.path.exists(app_path):
+                success_flag = True
+                result_message = "Could not find app.py to deploy"
```
aarondr77: In this case where the app was found, why are we using the message "Could not fina app.py to deploy"? I might be misunderstanding this code?

mito-ai/mito_ai/streamlit_conversion/streamlit_agent_handler.py
```diff
@@ -94,9 +94,20 @@ async def correct_error_in_generation(self, error: str) -> str:
...
+                      if os.path.isfile(os.path.join(dir_path, f))])
+    if file_count > 10:
+        raise ValueError(f"Too many files in directory: 10 allowed but {file_count} present. Create a new directory and retry")
 
 async def streamlit_handler(notebook_path: str) -> Tuple[bool, Optional[str], str]:
     """Handler function for streamlit code generation and validation"""
+
+    if not os.path.isabs(notebook_path):
+        notebook_path = os.path.join(os.getcwd(), notebook_path)
+    clean_directory_check(notebook_path)
```
aarondr77: I think this should be something that happens in the `deploy` steps instead of the streamlit conversion steps since that is where the constraints come in

aarondr77 (commented):
A few tiny comments and then LGTM once the tests pass!

mito-ai/mito_ai/app_builder/handlers.py
PoojasreeDwarakanath: ooh this message was part of another flow I had, wherein the message wouldn't be present in the else part. Thanks for catching that, I've removed it

PR #1889, opened 2025-08-04
Active Cell Context Card

ngafar:
# Description

Removed the active cell preview (in Chat mode), and replaced it with an _Active Cell_ card in the Selected Context Container. 

# Testing

- In chat mode you should be able to see the Active Cell card.
- In chat mode you should be able to click on the card, and go to the active cell.
- In Agent mode the card should not be visible. 

# Documentation

No, this should be self-explanatory.

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
@@ -8,10 +8,8 @@ import { classNames } from '../../../utils/classNames';
 import { IContextManager } from '../../ContextManager/ContextManagerPlugin';
 import ChatDropdown from './ChatDropdown';
 import { Variable } from '../../ContextManager/VariableInspector';
-import { getActiveCellID, getCellCodeByID } from '../../../utils/notebook';
+import { getActiveCellID } from '../../../utils/notebook';
 import { INotebookTracker } from '@jupyterlab/notebook';
-import { IRenderMimeRegistry } from '@jupyterlab/rendermime';
-import PythonCode from './PythonCode';
 import '../../../../style/ChatInput.css';
 import '../../../../style/ChatDropdown.css';
 import { useDebouncedFunction } from '../../../hooks/useDebouncedFunction';
@@ -30,8 +28,6 @@ interface ChatInputProps {
     isEditing: boolean;
     contextManager?: IContextManager;
     notebookTracker: INotebookTracker;
-    renderMimeRegistry: IRenderMimeRegistry;
-    displayActiveCellCode?: boolean;
     agentModeEnabled: boolean;
     agentExecutionStatus?: AgentExecutionStatus;
 }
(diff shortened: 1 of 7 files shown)


mito-ai/src/components/SelectedContextContainer.tsx
```diff
@@ -29,21 +40,46 @@ const SelectedContextContainer: React.FC<SelectedContextContainerProps> = ({ tit
...
+            const cell = getCellByID(notebookTracker, activeCellID);
+            if (cell) {
+                // Scroll to the cell
+                void notebookTracker.currentWidget?.content.scrollToCell(cell, 'center');
+                // Highlight the cell
+                setTimeout(() => {
+                    highlightCodeCell(notebookTracker, activeCellID);
+                }, 500);
+            }
+        } else if (onClick) {
```
aarondr77, line 64: If for some reason the type was active cell, but the notebookTracker was not yet defined then we would try calling this. Should we just make the above check first checking if it is the active_cell and then move the other checks inside of that case to avoid that potentially incorrect code flow?

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -266,31 +260,32 @@ const ChatInput: React.FC<ChatInputProps> = ({
         setExpandedVariables(expandedVariables);
     }, [contextManager?.variables, contextManager?.files]);
 
-    // If there are more than 8 lines, show the first 8 lines and add a "..."
-    const activeCellCode = getCellCodeByID(notebookTracker, activeCellID) || ''
-    const activeCellCodePreview = activeCellCode.split('\n').slice(0, 8).join('\n') + (
-        activeCellCode.split('\n').length > 8 ? '\n\n# Rest of active cell code...' : '')
+    // Automatically add active cell context when in Chat mode and there's active cell code
+    useEffect(() => {
+        if (!agentModeEnabled && activeCellCode && activeCellCode.trim().length > 0) {
```
aarondr77: There's a small bug here where: 
1. Open Chat mode
2. Click on a cell that has code in it
3. Click on an empty code cell

Notice that the active cell context button is still there. 

I think it's okay, and would just be down to do: whenever in chat mode, it has that icon.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -310,6 +305,8 @@ const ChatInput: React.FC<ChatInputProps> = ({
                             title={context.type === 'db' && context.display ? context.display : context.value}
                             type={context.type}
                             onRemove={() => setAdditionalContext(additionalContext.filter((_, i) => i !== index))}
+                            notebookTracker={context.type === 'active_cell' ? notebookTracker : undefined}
+                            activeCellID={context.type === 'active_cell' ? activeCellID : undefined}
```
aarondr77: I think for both of these it makes more sense to pass them through to the `SelectedContextContainer` regardless of if its of type active_cell or not. 

Then, we can let the `SelectedContextContainer` be in charge of the logic

aarondr77 (commented):
Nice! Seems to be working well. And looks pretty good! 

We're still showing the active cell preview in the user message once it is sent. Do you think we should remove that too? Can be in different PR if so. 
Image

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
ngafar: I definitely think we want to keep the icon. Even if the cell is empty, you'd still be referencing it.

PR #1902, opened 2025-08-08
Truncating long samples in VariableInspector

ngafar:
# Description

Fix for shapefiles blowing up the context window. When adding samples to the context manager, the VariableInspector will only add the first 50 characters. 

Screenshot 2025-08-08 at 11 15 41 AM

# Testing

Add a shapefile, and have the Agent open it. You can then print the `contextManager` and look at the samples. 

# Documentation

No

### DIFF
--- a/mito-ai/src/Extensions/ContextManager/VariableInspector.tsx
+++ b/mito-ai/src/Extensions/ContextManager/VariableInspector.tsx
@@ -34,7 +34,10 @@ def get_dataframe_structure(df, sample_size=5):
             # Handle None and NaN (convert to None, which maps to null in JSON)
             return None
         elif not isinstance(value, (str, int, float, bool, type(None))):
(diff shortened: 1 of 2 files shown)


aarondr77:
cursor review

mito-ai/src/Extensions/ContextManager/VariableInspector.tsx
```diff
@@ -34,7 +34,10 @@ def get_dataframe_structure(df, sample_size=5):
             # Handle None and NaN (convert to None, which maps to null in JSON)
             return None
         elif not isinstance(value, (str, int, float, bool, type(None))):
-            return str(value)            
+            return str(value)[:50] + "..."
```
aarondr77: This is going to add the `...` to every element regardless of if it's more than 50 characters or not.

PR #1906, opened 2025-08-11
Signin/ signup flow UI and cognito changes

PoojasreeDwarakanath:
# Description

This PR includes the changes in our signin flow. Instead of redirecting users to a sign-in page hosted by AWS Cognito, we are providing a custom sign-in popup, which removes redirects to other urls.

# Testing

# Documentation
None


mito-ai/src/Extensions/AppBuilder/auth-popup-deploy.tsx
```diff
@@ -0,0 +1,102 @@
...
+import '../../../style/AuthPopup.css';
+import TextButton from '../../components/TextButton';
+
+interface AuthPopupProps {
+  isOpen: boolean;
+  onClose: () => void;
+  onSuccess: (user: any) => void;
+}
+
+// Using Mito's modal styling classes
```
aarondr77: Wooo it looks pretty awesome!

mito-ai/src/Extensions/AppBuilder/auth-popup-deploy.tsx
```diff
@@ -0,0 +1,102 @@
...
+                placeholder: 'Enter your full name',
+                label: 'Full Name *',
+                required: true
+              },
+              email: {
+                order: 2,
+                placeholder: 'Enter your email address',
+                label: 'Email *'
+              },
+              password: {
```
aarondr77, line 64: Are we able to show the password requirements here? It was a bit annoying that I had to keep submitting in order to see what was missing. Ie: uppercase, lowercase, special characters etc. 

Image

mito-ai/src/Extensions/AppBuilder/aws-config.ts
```diff
@@ -0,0 +1,34 @@
...
+const userPoolClientId = userPoolClientIdDev
+
+export const configureAmplify = () => {
+  Amplify.configure({
+    Auth: {
+      Cognito: {
+        userPoolId: userPoolId,
+        userPoolClientId: userPoolClientId,
+
+        loginWith: {
```
aarondr77, line 22: Are we able to let users sign in with Google SSO here too? We can do in separate PR.

mito-ai/src/Extensions/AppBuilder/auth-popup-deploy.tsx
```diff
@@ -0,0 +1,102 @@
...
+                }
+              }, 0);
+            }
+
+            return (
+              <div style={{ padding: '20px', textAlign: 'center' }}>
+                <p style={{ color: 'var(--jp-ui-font-color1)', marginBottom: '16px' }}>
+                  Welcome! You're now signed in and ready to deploy your app.
+                </p>
+                <TextButton
```
aarondr77: having this button in the signed up modal is a bit confusing because you can't get back to it after you've signedup/ signed in.

mito-ai/src/Extensions/AppBuilder/auth-popup-deploy.tsx
```diff
@@ -0,0 +1,102 @@
...
+          <button
+            onClick={onClose}
+            className="modal-close-button"
+            title="Close"
+          >
+            ×
+          </button>
+        </div>
+
+        <Authenticator
```
aarondr77, line 48: I'm noticing that after I sign in or Sign up, I'm always redirected to this screen. It doesn't make a ton of sense to be on this screen because I didn't just reset my password. Is that something we have control over? I don't see this in our code.. 

Image

aarondr77 (approved):
This is awesome. I'm glad we were able to avoid the issues we were running into previously + keeping it all in the app seems like a great UX as well. 

Left a few comments about issues I'm running into

I'm also noticing that the apps aren't loading and the APP url that I get taken to looks like it might be using the notebook id? It doesn't look like the code was changed in this PR to me so not sure where that might have happened. Let me know what you think. Okay to ignore for now and make a plan to fix later if need be! 

Image

mito-ai/src/Extensions/AppBuilder/auth-popup-deploy.tsx
PoojasreeDwarakanath: okay, I'll check this out and add it

mito-ai/src/Extensions/AppBuilder/aws-config.ts
PoojasreeDwarakanath: Ya we can do that in a separate PR. I just wanted to get the basic auth working correctly with this one.

mito-ai/src/Extensions/AppBuilder/auth-popup-deploy.tsx
PoojasreeDwarakanath: ooh yeah this isnt getting displayed. Let me modify this to display it and then start the app deploy

mito-ai/src/Extensions/AppBuilder/auth-popup-deploy.tsx
PoojasreeDwarakanath: yeah this is happening because it is set like that on the Cognito side. Even I found it a bit annoying. I'll change those settings and then we shouldn't be seeing this verification as part of the flow.

PR #1907, opened 2025-08-15
Adding the run_all_cells tool to the agent

sid-sagacollab:
# Description

This feature enables the agent to use the run_all_cells feature. This helps the agent to bring back variables to scope.

There are a lot of scenarios where a variable is out of scope and the agent either rewrites code in old cells (such as redundant `pd.read_csv`) or just fails after 3 attempts. An elegant way to fix this is to actually just run all the cells once and refreshing the context.

Here's a usage example of this tool where agent uses this tool to refresh the context

https://github.com/user-attachments/assets/eb55babf-2ace-4580-a887-9598edfc45d3

Here's the tool in action when it encounters a name error.

https://github.com/user-attachments/assets/e7e174b3-1dc3-4600-b587-da00b8c97fcf

# Testing

# Documentation


mito-ai/style/RunAllCellsToolUI.css
```diff
@@ -0,0 +1,23 @@
+.run-all-cells-container {
```
aarondr77: Let's reuse the same component styling that we use for the get_cell_output tool for consistency

Image
Image

mito-ai/src/utils/agentActions.tsx
```diff
@@ -143,3 +150,14 @@ export const retryIfExecutionError = async (
     return 'success'
 }
 
+export const runAllCells = async (
+    app: JupyterFrontEnd,
+    notebookTracker: INotebookTracker
+): Promise<void> => {
+    await app.commands.execute("notebook:run-all-cells");
```
aarondr77, line 171: If there is an error while refreshing the notebook, we should either report that to the user and stop, or we should go through the error fixup process. I'll defer to you on which one to go with. But right now it's a bit confusing that it doesn't call any attention to the error and thinks it did good work. 

https://github.com/user-attachments/assets/44d121e8-d665-4ba0-91ec-7e99030f4d96

mito-ai/src/components/AgentComponents/ErrorFixupToolUI.tsx
```diff
@@ -100,7 +100,8 @@ const GroupedErrorsAndFixes: React.FC<IErrorFixupToolUIProps> = ({
                         <AssistantCodeBlock
                             key={`assistant-${index}`}
                             code={messageItem.message.content as string}
-                            codeSummary={messageItem.agentResponse?.cell_update?.code_summary ?? undefined}
+                            codeSummary={messageItem.agentResponse?.cell_update?.code_summary ??
```
aarondr77: Maybe we can move this logic into the AssistantCodeBlock to keep things organized into the component?

sid-sagacollab: Addressed!

mito-ai/style/RunAllCellsToolUI.css
sid-sagacollab: Addressed!

mito-ai/src/utils/agentActions.tsx
sid-sagacollab: Works now 

https://github.com/user-attachments/assets/0d68da9e-671a-4f0f-afc4-fc14f498f1aa

PR #1908, opened 2025-08-18
File Uploads

ngafar:
# Description

This PR adds a new _Attach File_ button to the chat input. This lets the user select a file from anywhere on their computer, the file is then opened on the frontend, and encoded. This encoding is then sent to the backend (via a POST request) where it is decoded, and saved to current working directory. 

# Testing

Attach a file, and reference it in chat. Then make sure the AI uses the attached file. 

Also, test different file types, and sizes. 

# Documentation

No, should be self-explanatory, like attaching files in an email.


aarondr77:
@cursoragent review this please

mito-ai/mito_ai/file_uploads/handlers.py
```diff
@@ -0,0 +1,120 @@
...
+        self.write(
+            {
+                "success": True,
+                "chunk_received": True,
+                "chunk_number": chunk_number,
+                "total_chunks": total_chunks,
+            }
+        )
+
+    def _handle_error(self, error_message: str, status_code: int = 500) -> None:
```
aarondr77, line 206: When this raises an error, we don't share any information with the user. Can we do the following: 

1. Use the Juptyer Notification system to display the error message
2. Log the error so we can figure out how to improve the feature!

mito-ai/src/components/AttachFileButton.tsx
```diff
@@ -0,0 +1,198 @@
...
+            console.log(`File ${file.name} is larger than 100MB (${(file.size / (1024 * 1024)).toFixed(2)}MB). Splitting into chunks...`);
+            void handleLargeFile(file);
+        } else {
+            // Upload file directly for files <= 100MB
+            void uploadFile(file);
+        }
+    };
+
+    const handleLargeFile = async (file: File): Promise<void> => {
+        setIsUploading(true);
```
aarondr77, line 54: Maybe we should move the loading handling into the `handleFileSelect` function so it is used for both small and large files.

mito-ai/src/components/AttachFileButton.tsx
```diff
@@ -0,0 +1,198 @@
...
+    onFileUploaded: (fileName: string) => void;
+}
+
+const AttachFileButton: React.FC<AttachFileButtonProps> = ({ onFileUploaded }) => {
+    const fileInputRef = useRef<HTMLInputElement>(null);
+    const [isUploading, setIsUploading] = useState(false);
+
+    // Constants for file handling
+    const MAX_FILE_SIZE = 100 * 1024 * 1024; // 100MB in bytes
+    const CHUNK_SIZE = 50 * 1024 * 1024; // 50MB chunks
```
aarondr77: A few small things here: 

0. I think Jupyter uses much smaller max_file_size and chunk sizes. If I were to upload a file that is 25MB I think I would get the popup that says "This is a large file are you sure you want to upload". Want to look at the Jupyter code and just use the same cutoffs + chink sizes since they are presumably very well tested.
1. It looks like this isn't a max file size it's just the cutoff for if we treat it as chunks or not. Max file size made me think we reject files larger than this. Can we just update the name?
2. Can we just move these contents to the top of the file outside of the component.

aarondr77 (approved):
Sweet. This is pretty awesome and seems to be working pretty well. I think I caught one high priority bug that we need to fix re: using the correct file path, and then just some small cleanup and polish stuff. Then LGTM!

mito-ai/mito_ai/file_uploads/handlers.py
```diff
@@ -0,0 +1,120 @@
...
+        with open(filename, "wb") as f:
+            f.write(file_data)
+
+        self.write({"success": True, "filename": filename, "path": filename})
+
+    def _save_chunk(self, filename: str, file_data: bytes, chunk_number: int) -> None:
+        """Save a chunk to a temporary file."""
+        chunk_filename = f"{filename}.part{chunk_number}"
+        with open(chunk_filename, "wb") as f:
+            f.write(file_data)
```
aarondr77: Instead of storing the files in the root directory and then having to clean them up, can we just use a python temporary directory to store the files. I think we can:

1. Create a temp directory
2. Store all the chunked files there
3. Once we get all the chunks, combine them into the real file and write it to the correct location
4. Let the temp directory auto clean itself up so we don't have to do a os.remove stuff

aarondr77: Here's a little explainer of temp directories

```python
import tempfile
import os

# Create a temporary directory using a context manager
with tempfile.TemporaryDirectory() as temp_dir_path:
    print(f"Temporary directory created at: {temp_dir_path}")

    # You can perform operations within this temporary directory
    temp_file_path = os.path.join(temp_dir_path, "my_temp_file.txt")
    with open(temp_file_path, "w") as f:
        f.write("This is temporary content.")

    print(f"File created inside temporary directory: {temp_file_path}")

# Once the 'with' block is exited, the temporary directory and its contents are automatically deleted.
print("Temporary directory and its contents have been cleaned up.")
```

PR #1911, opened 2025-08-19
Manage apps taskpane 

PoojasreeDwarakanath:
# Description

This PR adds a manage apps feature which allows the user to keep track of the Apps deployed by them using the streamlit feature.

# Testing

# Documentation
\-


mito-ai/src/Extensions/AppManager/apps-list.tsx
```diff
@@ -0,0 +1,192 @@
...
+          </button>
+          <button
+            onClick={() => {
+              console.log('Logout clicked');
+              void logoutAndClearJWTTokens();
+            }}
+            className="apps-list-button"
+            title="Logout"
+          >
+            Logout
```
aarondr77: Can we make the logout and refresh button background color turn slightly darker (just like the copy button) when you hover over it. It's good polish that makes it clear that things are clickable!

mito-ai/src/Extensions/AppManager/apps-list.tsx
```diff
@@ -0,0 +1,192 @@
...
+                    <span
+                      className="app-item-status-indicator"
+                      style={{ backgroundColor: getStatusColor(app.status) }}
+                    />
+                    <span className="app-item-status-text">
+                      {getStatusText(app.status)}
+                    </span>
+                  </div>
+                  <div className="app-item-created">
+                    Created: {app.createdAt}
```
aarondr77: Having the created at is awesome, but maybe its a little bit too exact of a time: 

What about if we do `Created: 2025-08-19 22:20` instead of `Created: 2025-08-19T22:20:00.564381Z`

aarondr77: Also, should this be `Created` or `Last deployed`? Created makes it sound to me like its the first time I deployed the app, not the most recently updated one.

mito-ai/src/Extensions/AppManager/list-apps-api.ts
```diff
@@ -0,0 +1,80 @@
+// list-apps-api.ts - Backend API function
+
+import { IManageAppReply, IManageAppRequest } from '../../websockets/appManager/appManagerModels';
+import { getJWTToken } from '../AppBuilder/auth';
+import { UUID } from '@lumino/coreutils';
+
+// Import the actual service interface from the plugin
+import { IAppManagerService } from './ManageAppsPlugin';
+
+export interface App {
```
aarondr77: Maybe we can call this `AppMetadata`

mito-ai/src/Extensions/AppManager/list-apps-api.ts
```diff
@@ -0,0 +1,80 @@
...
+          apps: [],
+          success: false,
+          message: 'User not authenticated'
+        };
+      }
+
+      // Create the request message with proper typing
+      const request: IManageAppRequest = {
+        type: 'manage-app',
+        jwt_token: jwtToken || appManagerService.client.serverSettings?.token
```
aarondr77: Can you leave a comment here that explains why we either use the jwtToken or the token from the client appManagerService. In particular, what's the difference between these and when do we expect to use each?

mito-ai/src/Extensions/AppManager/list-apps-api.ts
```diff
@@ -0,0 +1,80 @@
...
+      const response: IManageAppReply = await appManagerService.client.sendMessage<IManageAppRequest, IManageAppReply>(request);
+
+      // Check if the response indicates an error
+      if (response.error) {
+        throw new Error(`Error: ${response.error.title || 'Failed to fetch apps'}`);
+      }
+
+      // Transform the response to match expected format
+      const apps: App[] = (response.apps || []).map(app => ({
+        id: UUID.uuid4(), // Generate ID since not provided in response
```
aarondr77: If we're just creating a new UUID for the app each time we pull new info about it, do we really need the id at all? Can we just delete it?

mito-ai/src/Extensions/AppManager/list-apps-api.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export interface App {
+  id: string;
+  name: string;
+  url: string;
+  status: 'running' | 'stopped' | 'deploying';
+  createdAt: string;
+}
+
+export interface GetAppsResponse {
```
aarondr77: I think we can type this even stronger.. As far as I can tell it looks like we have two distinct response types: 

```typescript
export interface GetAppsSuccess {
     success: true  
     apps: App[]
}

export interface GetAppsFailure {
     success: false  
     errorMessage: string
}

export interface GetAppsResponse = GetAppsSuccess | GetAppsFailure
```

This might be nice so that we don't need to worry about setting the message to undefined in the case of success, for example.

aarondr77: But not the biggest deal if you don't think it's a good refactor now. Maybe we can at least change `message` to `errorMessage` if that's the only time it's being used.

mito-ai/src/Extensions/AppManager/list-apps-api.ts
```diff
@@ -0,0 +1,80 @@
...
+        apps,
+        success: !response.error,
+        message: undefined // No message field available in response
+      };
+
+      return data;
+    } catch (error) {
+      console.error('Error fetching apps via websocket:', error);
+      throw error;
+    }
```
aarondr77: We're using nested try, catch statements this function. But it doesn't look like this internal one is really needed since it's just raising a generic error that is going to get handled by the outer catch statement.  Can we get rid of the internal try, catch statement?

mito-ai/src/Extensions/AppManager/ManageAppsPlugin.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  render(): React.ReactElement {
+    // Pass the appManagerService as a prop to the AppsList component
+    return React.createElement(AppsList, {
+      appManagerService: this._appManagerService
+    });
+  }
+}
+
+const ManageAppsPlugin: JupyterFrontEndPlugin<IAppManagerService> = {
```
aarondr77, line 77: Nice work on building this Jupyter extension! These are pretty confusing and you did a great job on your first one! Woohoo.

mito-ai/src/websockets/appManager/appManagerWebsocketClient.ts
```diff
@@ -0,0 +1,101 @@
...
+            if (message.error) {
+              pendingReply.reject(new Error(message.error.title || 'Server error'));
+            } else {
+              pendingReply.resolve(message as any);
+            }
+          }
+        }
+      }
+    }
+  }
```
aarondr77: You can ignore this if you want for now, but just a few callouts about how I would structure this code since there's a lot going on here! 

### Change 0

This code is super nested. In those cases, you can use right guard style programming where instead of checking a condition like: 

```typescript 
if (pendingReply) {
     // A bunch of nested code here

```

You can write it like this: 

```typescript 
if (!pendingReply) {
     return

// The previously nested code goes here, not nested! 
```

It just makes the code a bit more organized! 

### Change 1

It looks like we have the same final section of code in the top level if and else statements. Specifically this at the end of both: 

```typescript 
this._pendingRepliesMap.delete(pendingId);

if (message.error) {
          pendingReply.reject(new Error(message.error.title || 'Server error'));
        } else {
          pendingReply.resolve(message as any);
        }
      }
```

So maybe we could update the code so that the if and else statements are just responsible for their unique code paths to get to the id and then we use the same lines of code to actually process the deleting. 


Just thoughts! No need to change now.

mito-ai/style/apps-list.css
aarondr77, line 1: Just small nitpick, but I think convention is for this file to be called AppsList.css instead of apps-list.css

mito-ai/style/ManageAppsWidget.css
```diff
@@ -0,0 +1,301 @@
...
+.manage-apps-header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 16px 20px;
+  border-bottom: 1px solid var(--jp-border-color1);
+  background: var(--jp-layout-color1);
+}
+
+.manage-apps-header h2 {
```
aarondr77: It doesn't look like we are using an h2 tag in our code, so I don't think this is doing anything.

mito-ai/style/ManageAppsWidget.css
```diff
@@ -0,0 +1,301 @@
...
+
+.visibility-badge.public {
+  background: var(--jp-success-color3);
+  color: var(--jp-success-color0);
+}
+
+.visibility-badge.private {
+  background: var(--jp-ui-font-color3);
+  color: var(--jp-ui-font-color1);
+}
```
aarondr77: I don't think we're using the visibility badge classes anywhere either

mito-ai/style/ManageAppsWidget.css
```diff
@@ -0,0 +1,301 @@
...
+  background: var(--jp-layout-color2);
+  border-color: var(--jp-accent-color1);
+}
+
+.copy-icon {
+  font-size: 14px;
+  transition: all 0.2s ease;
+}
+
+.copy-icon.copied {
```
aarondr77: Don't think we're using this either

mito-ai/style/ManageAppsWidget.css
```diff
@@ -0,0 +1,301 @@
...
+}
+
+.app-name {
+  margin: 0 0 8px 0;
+  font-size: 16px;
+  font-weight: 600;
+  color: var(--jp-ui-font-color0);
+}
+
+.app-description {
```
aarondr77: Also not using this

mito-ai/style/ManageAppsWidget.css
```diff
@@ -0,0 +1,301 @@
...
+  font-size: 14px;
+  color: var(--jp-ui-font-color2);
+  line-height: 1.4;
+}
+
+.app-url {
+  margin-bottom: 16px;
+}
+
+.url-label {
```
aarondr77: not using this either

mito-ai/style/ManageAppsWidget.css
```diff
@@ -0,0 +1,301 @@
...
+  margin-bottom: 6px;
+  text-transform: uppercase;
+  letter-spacing: 0.5px;
+}
+
+.url-container {
+  position: relative;
+}
+
+.url-input {
```
aarondr77: I don't think we're using any of the .url-input classes

aarondr77: Can you just go through the rest and make sure we're deleting classes that we aren't using. I think Cursor could be useful :)

mito-ai/style/ManageAppsWidget.css
```diff
@@ -0,0 +1,301 @@
...
+/* Dark theme adjustments */
+[data-jp-theme-light="false"] .app-card {
+  background: var(--jp-layout-color1);
+  border-color: var(--jp-border-color1);
+}
+
+[data-jp-theme-light="false"] .url-input {
+  background: var(--jp-layout-color0);
+  color: var(--jp-ui-font-color0);
+}
```
aarondr77: The way that themes work in Jupyter is that when you go into dark theme, the `--jp-...` variables update, so we don't need to do any special handling. We should be good to delete this.

mito-ai/style/ManageAppsWidget.css
```diff
@@ -0,0 +1,301 @@
...
+.open-app-button:hover {
+  background: var(--jp-accent-color2);
+}
+
+.button-icon {
+  font-size: 14px;
+}
+
+/* Responsive Design */
+@media (max-width: 768px) {
```
aarondr77: I don't think we need these responsive changes. It seems pretty good without it to me. What do you think?

mito-ai/src/Extensions/AppBuilder/auth.ts
```diff
@@ -36,4 +36,77 @@ export const getAuthHeaders = async (): Promise<Record<string, string>> => {
...
+        const keysToRemove: string[] = [];
+        for (let i = 0; i < localStorage.length; i++) {
+            const key = localStorage.key(i);
+            if (key && (
+                key.startsWith('aws-amplify-') ||
+                key.startsWith('amplify-') ||
+                key.startsWith('CognitoIdentityServiceProvider.') ||
+                key.includes('LastAuthUser') ||
+                key.includes('clockDrift') ||
+                key.includes('userData')
```
aarondr77: Since we're the ones setting these, shouldn't we know exactly what keys to delete instead of having to guess using these heuristics?

mito-ai/src/Extensions/AppBuilder/auth.ts
```diff
@@ -36,4 +36,77 @@ export const getAuthHeaders = async (): Promise<Record<string, string>> => {
...
+                const key = localStorage.key(i);
+                if (key && (
+                    key.startsWith('aws-amplify-') ||
+                    key.startsWith('amplify-') ||
+                    key.startsWith('CognitoIdentityServiceProvider.')
+                )) {
+                    allKeys.push(key);
+                }
+            }
+            allKeys.forEach(key => localStorage.removeItem(key));
```
aarondr77: If we want this to happen both in the try and catch phase, we can move it to a finally block, so it happens regardless of which case we are in and we don't need to duplicate the code.

mito-ai/src/Extensions/AppManager/apps-list.tsx
aarondr77, line 1: Can we rename these to camel Case? So AppList.tsx and ListAppsAPIS.ts

mito-ai/src/Extensions/AppManager/apps-list.tsx
```diff
@@ -0,0 +1,192 @@
...
+        setError(err instanceof Error ? err.message : 'An unexpected error occurred');
+        setApps([]);
+      } finally {
+        setLoading(false);
+      }
+    };
+
+    console.log('[AppsList] Component mounted, calling loadApps...');
+    void loadApps();
+  }, [appManagerService]); // Add appManagerService to dependency array
```
aarondr77: we can delete this comment

mito-ai/src/Extensions/AppManager/apps-list.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  // Refresh function that can be called manually
```
aarondr77: same with this comment, not really adding much

mito-ai/src/Extensions/AppManager/apps-list.tsx
```diff
@@ -0,0 +1,192 @@
...
+      const textArea = document.createElement('textarea');
+      textArea.value = url;
+      document.body.appendChild(textArea);
+      textArea.select();
+      document.execCommand('copy');
+      document.body.removeChild(textArea);
+    }
+  };
+
+  const getStatusColor = (status: string): string => {
```
aarondr77: Hmm I think there is a bit of type mismatch going on with the status. A few comments. 

### Type Mismatch

This is the status type that we are defining:

```typescript
export interface App {
  id: string;
  name: string;
  url: string;
  status: 'running' | 'stopped' | 'deploying';
  createdAt: string;
}
```

But here, we are handling cases like `processing`. That makes me think that our status type is not correct.

We can make stronger types to help with this by updating the interface to this: 

```typescript
export type AppStatus = 'running' | 'stopped' | 'deploying';

export interface App {
  id: string;
  name: string;
  url: string;
  status: AppStatus
  createdAt: string;
}
```

And then in this function instead of typing the input like this 

```typescript
const getStatusColor = (status: string): string => {
```

We can do this: 

```typescript
const getStatusColor = (status: AppStatus): string => {
```

That makes sure we handle all of the possible statuses! 

### Status Options 

Other thing, is that I'm not super clear on what each status means. In particular, what is the difference between deploying and processing? 

Maybe we should have: 

1. error
2. building  -> while we are creating the app.py file (we could remove this for now. not a big deal at all)
3. deploying -> while aws is doing its thing 
4. active -> app is running
5. shut down -> app is deployed, but shut down.

aarondr77: Currently, I have this app deployed at: demo.bc187455-abab-4c7b-b4bb-66695047a20b.app.dev.trymito.io

But the status is deploying

Image

mito-ai/src/Extensions/AppManager/apps-list.tsx
PoojasreeDwarakanath: resolving on the AWS side

mito-ai/src/Extensions/AppManager/list-apps-api.ts
PoojasreeDwarakanath: resolved

mito-ai/src/Extensions/AppManager/list-apps-api.ts
PoojasreeDwarakanath: resolved

mito-ai/style/apps-list.css
PoojasreeDwarakanath: resolved

mito-ai/src/Extensions/AppManager/apps-list.tsx
PoojasreeDwarakanath: resolved

mito-ai/src/Extensions/AppManager/apps-list.tsx
PoojasreeDwarakanath: resolved

mito-ai/style/ManageAppsWidget.css
PoojasreeDwarakanath: resolved

mito-ai/style/ManageAppsWidget.css
PoojasreeDwarakanath: resolved

mito-ai/src/Extensions/AppBuilder/auth.ts
PoojasreeDwarakanath: The keys usually have an ID after them, and hence we cannot do an exact match. Thats why we check what it starts with

PoojasreeDwarakanath: I've modified this to delete only 'CognitoIdentityServiceProvider' since thats all amplify is setting as far as I can see.

mito-ai/src/Extensions/AppBuilder/auth.ts
PoojasreeDwarakanath: resolved

mito-ai/src/Extensions/AppManager/list-apps-api.ts
PoojasreeDwarakanath: This isnt being used anymore. It was used in our appBuilder code as well. Thanks for this, good catch! I've removed it

mito-ai/src/Extensions/AppManager/list-apps-api.ts
PoojasreeDwarakanath: resolved

mito-ai/src/Extensions/AppManager/apps-list.tsx
PoojasreeDwarakanath: I have modified the status types. Now we have the following types - active, deploying, shut down, error.
I need to update these status types on the AWS side as well

PoojasreeDwarakanath: I'll investigate the error on the AWS status change lambdas and check

mito-ai/src/websockets/appManager/appManagerWebsocketClient.ts
PoojasreeDwarakanath: refactored

mito-ai/src/Extensions/AppManager/list-apps-api.ts
PoojasreeDwarakanath: I have made this modification, resolved!

mito-ai/src/Extensions/AppManager/apps-list.tsx
PoojasreeDwarakanath: resolved

PR #1934, opened 2025-08-28
Database Search by Alias

ngafar:
# Description

This fixes a bug where the ChatDropdown was using the underlying database value for filtering. However, the value is made up of `connectionId + " - " + connection.type` where the `connectionId` is a UUID. 

Now filtering is done against the alias or database type. 

# Testing

Add a database with an alias that is unique. It should not contain the database type (sqlite, psotgres, etc). Then search for the alias and the type. 

# Documentation

No, this fix is the expected behavior.


aarondr77:
LGTM once tests finish running!

PR #1935, opened 2025-08-29
Profiler MVP

ngafar:
# Description

Sometimes we forget how much goes into constructing prompts, the profiler provides a snapshot of all the things.

The profiler has two main components:

1. Context Manager - This is a view of everything in the context manager. 
2. Captured Completion Requests - The object sent to the backend, including the prompt, variables, etc. As long as you have the profiler open, it will automatically capture these requests. 

# Testing

The profiler should be open alongside your notebook when doing any testing. Switching between tabs in Jupyter will cause certain values to disappear. 

1. Send a few requests with the profiler open, make sure they are being captured.
2. Add variables to a notebook, make sure they show up in the profiler. 
4. Do a small workflow without the profiler open, to make sure nothing was inadvertently broken. 

# Documentation

No, for internal use only.

### DIFF
--- a/mito-ai/src/Extensions/SettingsManager/SettingsWidget.tsx
+++ b/mito-ai/src/Extensions/SettingsManager/SettingsWidget.tsx
@@ -9,9 +9,11 @@ import { DatabasePage } from './database/DatabasePage';
 import { SupportPage } from './support/SupportPage';
 import { GeneralPage } from './general/GeneralPage';
 import { RulesPage } from './rules/RulesPage';
+import { ProfilerPage } from './profiler/ProfilerPage';
+import { IContextManager } from '../ContextManager/ContextManagerPlugin';
 import '../../../style/SettingsWidget.css';
 
-const TABS_CONFIG = {
+const TABS_CONFIG = (contextManager: IContextManager) => ({
     database: {
         label: 'Database',
         component: DatabasePage
@@ -19,6 +21,10 @@ const TABS_CONFIG = {
     general: {
         label: 'General',
         component: GeneralPage
+    },
+    profiler: {
+        label: 'Profiler',
+        component: () => <ProfilerPage contextManager={contextManager} />
     },
     rules: {
         label: 'Rules',
@@ -29,13 +35,18 @@ const TABS_CONFIG = {
         component: SupportPage
     },
 
-} as const;
+}) as const;
 
-const App = (): JSX.Element => {
-    const [activeTab, setActiveTab] = useState<keyof typeof TABS_CONFIG>('database');
+interface AppProps {
+    contextManager: IContextManager;
+}
+
+const App = ({ contextManager }: AppProps): JSX.Element => {
+    const [activeTab, setActiveTab] = useState<keyof ReturnType<typeof TABS_CONFIG>>('database');
+    const tabsConfig = TABS_CONFIG(contextManager);
 
     const renderContent = (): JSX.Element => {
-        const TabComponent = TABS_CONFIG[activeTab].component;
+        const TabComponent = tabsConfig[activeTab].component;
         return <TabComponent />;
     };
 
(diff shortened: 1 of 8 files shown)


mito-ai/src/Extensions/SettingsManager/profiler/ProfilerPage.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const captureCompletionRequest = (request: any): void => {
+    const event = new CustomEvent('mito-ai-capture-request', {
+        detail: {
+            request,
+        }
+    });
+    window.dispatchEvent(event);
+};
+
+export const ProfilerPage = ({ contextManager }: ProfilerPageProps): JSX.Element => {
```
aarondr77, line 30: When we're doing this with users, I imagine we might just ask them to send us this output and then we will dig into it later. To make that a tenable workflow, I think we want a global copy button that copies everything displayed in this taskpane in a well formatted manner (ie: message numbers, etc)  so they don't need to go through and copy and paste each one.

mito-ai/src/Extensions/SettingsManager/SettingsWidget.tsx
```diff
@@ -24,18 +26,27 @@ const TABS_CONFIG = {
         label: 'Rules',
         component: RulesPage
     },
+    profiler: {
```
aarondr77, line 29: Thoughts on making this a little bit more hidden? 

Two ideas: 
1. We could move profiler down to the very button of this taskpane so its out of sight from the public sections
2. When you open the taskpane, it could ask for a password to unlock it? This might be overkill though :) 

Image

mito-ai/src/Extensions/SettingsManager/profiler/ContextManagerSection.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export const ContextManagerSection = ({ contextManager }: ContextManagerSectionProps): JSX.Element => {
+    const [showContents, setShowContents] = useState<boolean>(true);
+
+    const handleRefreshContextManager = (): void => {
+        // Force a re-render by toggling the state
+        setShowContents(false);
+        setTimeout(() => setShowContents(true), 0);
+    };
```
aarondr77: I think there's a react hook for this btw https://react.dev/reference/react/Component#forceupdate

mito-ai/src/Extensions/SettingsManager/profiler/ProfilerPage.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { IContextManager } from '../../ContextManager/ContextManagerPlugin';
+import { ContextManagerSection } from './ContextManagerSection';
+import { CapturedRequestsSection } from './CapturedRequestsSection';
+import '../../../../style/ProfilerPage.css';
+
+interface ProfilerPageProps {
+    contextManager: IContextManager;
+}
+
+// Simple event system for capturing completion requests
```
aarondr77: Can you just add a little bit more description here why were are using an event system for doing this.

mito-ai/src/Extensions/SettingsManager/profiler/ProfilerPage.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const captureCompletionRequest = (request: any): void => {
+    const event = new CustomEvent('mito-ai-capture-request', {
```
aarondr77: Let's make `'mito-ai-capture-request'` a constant somewhere so we're sure to always keep it in sync

mito-ai/src/Extensions/SettingsManager/profiler/ContextManagerSection.tsx
ngafar: The `forceUpdate` method is only available from class-based components.

mito-ai/src/Extensions/SettingsManager/profiler/ProfilerPage.tsx
ngafar: Saving this for another PR: https://github.com/mito-ds/mito/issues/1942

Want to keep this as close to a MVP as possible.

PR #1937, opened 2025-09-02
Image Uploads

ngafar:
# Description

Added specific logic for image uploads. By turning the uploaded image to a Base64 encoding, and passing it to the backend, we can format the prompt to have the LLM "see" the image. 

# Testing

Upload images, and ask questions about the images. Try different file types (jpg, png, gifs,). 

# Documentation

Yes, this is a divergence in the way uploads are handled, and should be documented.

### DIFF
--- a/mito-ai/mito_ai/completions/completion_handlers/utils.py
+++ b/mito-ai/mito_ai/completions/completion_handlers/utils.py
@@ -6,25 +6,33 @@ from mito_ai.completions.message_history import GlobalMessageHistory
 from mito_ai.completions.models import ThreadID
 from mito_ai.completions.providers import OpenAIProvider
 from openai.types.chat import ChatCompletionMessageParam
-from mito_ai.completions.prompt_builders.chat_system_message import create_chat_system_message_prompt
-from mito_ai.completions.prompt_builders.agent_system_message import create_agent_system_message_prompt
+from mito_ai.completions.prompt_builders.chat_system_message import (
+    create_chat_system_message_prompt,
+)
+from mito_ai.completions.prompt_builders.agent_system_message import (
+    create_agent_system_message_prompt,
+)
+
 
 async def append_chat_system_message(
-        message_history: GlobalMessageHistory,
-        model: str,
-        provider: OpenAIProvider,
-        thread_id: ThreadID
+    message_history: GlobalMessageHistory,
+    model: str,
+    provider: OpenAIProvider,
+    thread_id: ThreadID,
 ) -> None:
-    
+
     # If the system message already exists, do nothing
-    if any(msg["role"] == "system" for msg in message_history.get_ai_optimized_history(thread_id)):
+    if any(
+        msg["role"] == "system"
+        for msg in message_history.get_ai_optimized_history(thread_id)
+    ):
         return
-    
+
     system_message_prompt = create_chat_system_message_prompt()
 
     system_message: ChatCompletionMessageParam = {
         "role": "system",
-        "content": system_message_prompt
+        "content": system_message_prompt,
     }
 
     await message_history.append_message(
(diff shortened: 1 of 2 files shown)


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -614,6 +614,16 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+        // If the user has uploaded an image, we need to move the (base64 encoded) image 
+        // out of the additionalContext array and into agentExecutionMetadata.
+        additionalContext?.map((context) => {
+            if (context.type.startsWith('image/')) {
+                agentExecutionMetadata.base64EncodedUploadedImage = context.value
+            }
+        })
+        // Remove images from the additionalContext array.
+        additionalContext = additionalContext?.filter(c => !c.type.startsWith('image/'))
+
```
aarondr77: Let's turn this into a helper function since we use it twice?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -614,6 +614,16 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
             agentExecutionMetadata.index = messageIndex
         }
 
+        // If the user has uploaded an image, we need to move the (base64 encoded) image
```
aarondr77: small nitpick, but can you update this comment to explain the WHY instead of just the WHAT. In particular, I'm wondering why don't we just send the image in the additionalContext instead of the agentExeuctionMetadata?

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -232,15 +249,32 @@ const ChatInput: React.FC<ChatInputProps> = ({
...
+                result.push({
+                    type: contextItem.type,
+                    value: contextItem.value
+                });
+            } else if (contextItem.type.startsWith('image/')) {
+                // If the user uploaded an image, we:
+                // 1. Keep the original context item. This is the base64 encoded image 
+                //    that will be processed in ChatTaskpane.tsx.
+                // 2. Add a second item to the additionalContext array, which will
+                //    have the image's filename, and be used in the prompt.
```
aarondr77, line 265: So this is effectively just how we end up saying something like: "The user uploaded an image called XYZ for you"?

mito-ai/mito_ai/completions/completion_handlers/utils.py
```diff
@@ -62,10 +62,25 @@ async def append_agent_system_message(
...
+            {
+                "type": "text",
+                "text": text,
+            },
+            {
+                "type": "image_url",
+                "image_url": {"url": f"data:image/png;base64,{base64EncodedUploadedImage}"},
+            }
+       ]
+    elif base64EncodedActiveCellOutput is not None and base64EncodedActiveCellOutput != '':
```
aarondr77: This looks like it would prevent us in chat mode from being able to send the active cell output and share an image that the user uploaded, right?

mito-ai/mito_ai/completions/completion_handlers/utils.py
```diff
@@ -62,10 +62,25 @@ async def append_agent_system_message(
         thread_id=thread_id
     )
     
-def create_ai_optimized_message(text: str, base64EncodedActiveCellOutput: Optional[str] = None) -> ChatCompletionMessageParam:
+def create_ai_optimized_message(
+    text: str, 
+    base64EncodedActiveCellOutput: Optional[str] = None, 
+    base64EncodedUploadedImage: Optional[str] = None
```
aarondr77: Oooo one issue with this is that it doesn't let users upload multiple images. I tried uploading two screenshots because I couldn't fit everything in one screenshot and it was only able to see some of the data.

If we make this a list of strings, then we should be all set.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -66,15 +66,32 @@ const ChatInput: React.FC<ChatInputProps> = ({
...
-    const handleFileUploaded = (fileName: string): void => {
-        // Add the uploaded file to the additional context
-        setAdditionalContext(prev => [
-            ...prev, {
-                type: 'file',
-                value: fileName,
-                display: fileName
-            }
-        ]);
+    const handleFileUpload = (file: File): void => {
```
aarondr77, line 69: When the user adds images, we are still uploading the image to the current working directory, but maybe there's not any reason to do that here. That was just a hack we used since we didn't have access to the file path for writing python code. Thoughts on updating the `handleFileSelect` function so that it doesn't upload images and instead just adds them as context?

mito-ai/mito_ai/completions/completion_handlers/utils.py
aarondr77: Can you create a pytest to make sure this works as expected?

mito-ai/mito_ai/completions/completion_handlers/utils.py
aarondr77: Want to make a pytest for this too?

aarondr77 (commented):
Being able to upload an image is pretty sweet. I got it to interpret graphs, extract data from images, recreate existing graphs, etc. Awesome! 

I had a few code ideas that I left comments about here. Let me know what you think.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
ngafar: Yes, but it also includes the filename. This way if the user asks to open the image, we know what the file is called.

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
ngafar: I left this logic in (uploading the the current directory) in case the user wants to open the image. Thought I am not sure how much this will happen.

mito-ai/mito_ai/completions/completion_handlers/utils.py
ngafar: Going to save this for a follow up PR, it's not as straightforward as I thought.

ngafar: https://github.com/mito-ds/mito/issues/1941

PR #1947, opened 2025-09-04
Image Upload Size and Type Fix

ngafar:
# Description

This PR fixes two bugs:

1. Large images break the websocket connection - When a large image is uploaded (>1mb), the Base64 encoding is so large that sending it all in one request to the backend will cause a websocket disconnect. To fix this, we are encoding the image on the backend. 
2. File types hardcoded - When we are creating the final prompt message on the backend, we were prefixing everything with "image/png." For OpenAI requests, this was not issue. However, for Claude, you need to use the correct type.  

Also, because we've drastically simplified the encoding logic, I was able to add support for multiple image uploads. 

# Testing

Upload a few image, use different providers. Make sure your images are large, ideally >5mb. 

# Documentation

N/A, bug fix.

### DIFF
--- a/mito-ai/mito_ai/file_uploads/handlers.py
+++ b/mito-ai/mito_ai/file_uploads/handlers.py
@@ -35,13 +35,9 @@ def _check_image_size_limit(file_data: bytes, filename: str) -> None:
         return
 
     file_size_mb = len(file_data) / (1024 * 1024)  # Convert bytes to MB
-    truncated_filename = filename[0:5]
-    file_extension = filename.split(".")[-1]
 
     if file_size_mb > MAX_IMAGE_SIZE_MB:
-        raise ValueError(
-            f"{truncated_filename}[...].{file_extension} exceeds the 3MB limit for image uploads."
-        )
+        raise ValueError("Image exceeded 3MB limit.")
 
 
 class FileUploadHandler(APIHandler):
@@ -85,7 +81,7 @@ class FileUploadHandler(APIHandler):
             self.finish()
 
         except Exception as e:
-            self._handle_error(f"Failed to save file: {str(e)}")
+            self._handle_error(str(e))
 
     def _validate_file_upload(self) -> bool:
         """Validate that a file was uploaded in the request."""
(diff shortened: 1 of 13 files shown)


aarondr77:
When I upload this 5mb image, I'm getting this error. I wonder if there is just an aws setting we can update to make the max message size larger? 

whiteboard

In Chat Mode
Screenshot 2025-09-04 at 9 40 53 AM

In Agent Mode:
Screenshot 2025-09-04 at 9 39 27 AM

mito-ai/mito_ai/file_uploads/handlers.py
```diff
@@ -6,7 +6,40 @@
...
+        ".tif",
+        ".webp",
+        ".svg",
+    }
+    file_extension = os.path.splitext(filename)[1].lower()
+    return file_extension in image_extensions
+
+
+def _check_image_size_limit(file_data: bytes, filename: str) -> None:
+    if _is_image_file(filename):
```
aarondr77: Most minimal thing ever, but btw I think writing checks like this is a bit nicer because it removes a layer of nesting. No need to change now, but just a little pointer if you want it for the future. 

```python
if not _is_image_file(filename):
     return 

file_size_mb = ...
```

mito-ai/mito_ai/file_uploads/handlers.py
```diff
@@ -153,15 +189,22 @@ def _reconstruct_file(
...
-            # Reconstruct the file from chunks
+            # First, read all chunks to check total file size for images
+            all_file_data = b""
+            for i in range(1, total_chunks + 1):
+                chunk_filename = os.path.join(temp_dir, f"chunk_{i}")
+                print(f"DEBUG: Reading chunk {i} from {chunk_filename}")
+                with open(chunk_filename, "rb") as chunk_file:
+                    chunk_data = chunk_file.read()
+                    all_file_data += chunk_data
+                    print(f"DEBUG: Read {len(chunk_data)} bytes from chunk {i}")
```
aarondr77: Do we want to keep these prints?

aarondr77 (approved):
A few comments on the error message we are showing: 

1. It's getting displayed twice. That makes me think there is some underlying logic bug where we are raising the error twice? 
2. The formatting of the error message is a bit off. It shouldn't have the [...] with square brackets probably.

https://github.com/user-attachments/assets/ba477ae2-b325-4681-839f-df051ab941bd

Issue #1946, opened 2025-09-03
Don't let context buttons overflow chat input


aarondr77:
This is probably especially common for image uploads when users are uploading screenshots as they have long named on Mac. 

<img width="334" height="108" alt="Image" src="https://github.com/user-attachments/assets/bddb5911-09f2-4fe2-a6e8-1fd841b796ee" />

PR #1948, opened 2025-09-04
feat:fix for context buttons overflow chat input

himanshu-warulkar:
### Problem

Previously, the chat input UI component would overflow horizontally when a file with a long name was attached, causing the action buttons ("Attach File", "@ Add Context") and file pills to break the layout. This made the chat input difficult to use and visually inconsistent, especially with multiple context items or long file names.

### Solution

- Updated the CSS for the `.context-container` class in ChatInput.css:
  - Added `display: flex;`, `flex-wrap: wrap;`, and `gap: 8px;` to ensure that context buttons and file pills wrap to the next line instead of overflowing.
  - This change keeps the chat input layout clean and usable, regardless of the number or length of attached files and context items.

### Testing

- Verified in JupyterLab:
  - Attached files with long names and multiple context items.
  - Confirmed that buttons and pills wrap correctly and do not overflow.
  - Repeated the process multiple times in the same notebook and with large, real-world data.
- Verified in VS Code:
  - Ensured that the UI does not break and works as expected.
- Tested edge cases:
  - Multiple files, long file names, and many context items.
  - Dynamic addition and removal of context items and files.
- Confirmed that the fix does not affect other functionality or environments.

### Summary

This PR resolves GitHub issue #1946 by ensuring the chat input’s context buttons and file pills wrap to the next line, maintaining a consistent and user-friendly layout.

---

Let me know if you want to add or adjust any details!

### DIFF
--- a/mito-ai/style/ChatInput.css
+++ b/mito-ai/style/ChatInput.css
@@ -99,9 +99,3 @@
 .context-button:disabled:hover {
   background-color: var(--jp-layout-color2);
 }
-
-.context-container {
-  display: flex;
-  align-items: center;
-  gap: 8px;
-}
--- a/mito-ai/style/ChatInput.css
+++ b/mito-ai/style/ChatInput.css
@@ -99,3 +99,10 @@
 .context-button:disabled:hover {
   background-color: var(--jp-layout-color2);
 }
+
+.context-container {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 8px;
+  align-items: center;
+}
--- a/mito-ai/style/ChatInput.css
+++ b/mito-ai/style/ChatInput.css
@@ -102,7 +102,6 @@
 
 .context-container {
   display: flex;
-  flex-wrap: wrap;
-  gap: 8px;
   align-items: center;
+  gap: 8px;
 }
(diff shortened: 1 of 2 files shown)


mito-ai/style/ChatInput.css
```diff
@@ -98,4 +98,11 @@
 
 .context-button:disabled:hover {
   background-color: var(--jp-layout-color2);
+}
+
+.context-container {
```
aarondr77: A few small notes here: 

1. `SelectedContextContainer.css` file instead. That's where the rest of the `context-container` styling occurs. 
2. I'm still noticing that when the image name is long, the `selected-context-container` overflows the `context-container`. As you can see here: 

Image

aarondr77: I think to fix this, we need to set something like `overflow-x: hidden` on the `context-container` to make sure none of its children can overflow in the horizontal direction.

himanshu-warulkar: Thanks for the review, @aarondr77! 
I've moved the styles to SelectedContextContainer.css and applied the overflow fix as you suggested. I've also signed the CLA. Let me know if there's anything else!

PR #1950, opened 2025-09-05
Streamlit dashboard setup

aarondr77:
# Description

Move the deploy button into the Streamlit Preview toolbar. 

Screenshot 2025-09-04 at 11 36 03 PM

# Testing

Try it out.


mito-ai/mito_ai/app_builder/handlers.py
```diff
@@ -148,10 +148,12 @@ async def _handle_build_app(self, message: BuildAppRequest) -> None:
             
         try:
             notebook_path = str(notebook_path) if notebook_path else ""
+            print('HERE 1')
```
aarondr77: remove

PoojasreeDwarakanath:
Can we have the css of the deploy button to be the same as the preview button? - ie the purple background 
It was hard to know the it was a button. So users could potentially miss that.

PR #1956, opened 2025-09-06
Refresh app button

aarondr77:
# Description

When the preview is open, the user can decide to refresh the app! 

Also, in implementing this I made a few refactors:
1. Deploying the app no longer rebuilds the app, so you deploy what you see. 
2. Since the AppBuilder now just deploys the app, I renamed it accordingly 
3. I removed old app building code that we no longer use since the app is built with AI. 

# Testing

### Test 1 
1. Build an app preview
2. Edit the app.py file 
3. Click the refresh button and see that the app is replaced with the refreshed app

### Test 2
1. Build an app preview
2. Edit the app.py file
3. Click the deploy button and see that the edited version of the app is what is deployed

# Documentation

We should update the docs for app mode


PoojasreeDwarakanath:
I'm thinking if the name 'refresh app' a little misleading? Something like Recreate app is probably better? To give a sense that the button isnt just going to refresh the app's selections/ changes, but is going to rebuild an app

PR #1960, opened 2025-09-16
mito-ai: check if user is on mito-desktop

aarondr77:
# Description

Checks if the user is on `mito-desktop` by checking if they are using an electron based user agent. 

# Testing

To make testing easier, you can use (and review) [this PR which lets us side load](https://github.com/mito-ds/mito-desktop/pull/7) mito-ai into mito-desktop
- Make sure the value is False when you are not on `mito-desktop`
- Make sure the value is True when you are on `mito-desktop`

# Documentation

No

### DIFF
--- a/mito-ai/mito_ai/completions/handlers.py
+++ b/mito-ai/mito_ai/completions/handlers.py
@@ -133,11 +133,11 @@ class CompletionHandler(JupyterHandler, WebSocketHandler):
             # Extract environment information from the message
             environment = parsed_message.get('environment', {})
             if environment:
-                is_electron = environment.get('isElectron', False)
+                is_electron = environment.get('isElectron', None)
                 if is_electron is not None:
                     if is_electron != self.is_electron:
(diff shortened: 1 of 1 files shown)


mito-ai/src/websockets/completions/CompletionsWebsocketClient.ts
```diff
@@ -108,4 +112,18 @@ export class CompletionWebsocketClient extends BaseWebsocketClient<ICompletionRe
...
   }
+
+  // Override sendMessage to automatically add environment info
+  sendMessage<T extends ICompletionRequest, R extends CompleterMessage>(message: T): Promise<R> {
+    // Add environment info to all messages
+    const messageWithEnvironment = {
+      ...message,
+      environment: {
+        isElectron: this.isElectron,
+        userAgent: navigator.userAgent
```
aarondr77:
```suggestion

```

ngafar (approved):

PR #1968, opened 2025-09-22
changes to immediately show the app deploy notifications

PoojasreeDwarakanath:
# Description

On clicking the deploy app button, the in-progress notification wasnt immediate, causing some confusion as to whether the process is running. This PR fixes that by immediately showing an in-progress notification

# Testing

# Documentation
\-


aarondr77:
LGTM once tests pass

PR #1971, opened 2025-09-24
Stop Agent Immediately

ngafar:
# Description

Stops the agent immediately. This is done by breaking the connection on the frontend, and sending a `stop_agent` message to the backend. Then on the backend the last message is removed from the active thread. 

Addresses: https://github.com/mito-ds/mito/issues/1749

# Testing

Send a message and stop the agent. Try stopping at different times, for example:

- Before the first message arrives.
- Before the second message arrives. 
- Before the "thinking..." message is displayed. 

# Documentation

N/A - No explanation necessary, this should be default behavior when stopping the Agent.

### DIFF
--- a/mito-ai/mito_ai/completions/handlers.py
+++ b/mito-ai/mito_ai/completions/handlers.py
@@ -14,6 +14,7 @@ import tornado.web
 from jupyter_core.utils import ensure_async
 from jupyter_server.base.handlers import JupyterHandler
 from tornado.websocket import WebSocketHandler
+from openai.types.chat import ChatCompletionMessageParam
 from mito_ai.completions.message_history import GlobalMessageHistory
 from mito_ai.logger import get_logger
 from mito_ai.completions.models import (
@@ -228,11 +229,11 @@ class CompletionHandler(JupyterHandler, WebSocketHandler):
             if thread_id_to_stop:
                 self.log.info(f"Stopping agent, thread ID: {thread_id_to_stop}")
                 
-                ai_optimized_message = {
+                ai_optimized_message: ChatCompletionMessageParam = {
                     "role": "assistant",
                     "content": "The user made the following request: Stop processing my last request. I want to change it. Please answer my future requests without going back and finising my previous request."
                 }
-                display_optimized_message = {
+                display_optimized_message: ChatCompletionMessageParam = {
                     "role": "assistant",
                     "content": "Agent interupted by user "
                 }


mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -901,23 +946,57 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+        }
+        
+        // 2. Disconnect any active stream handlers
+        if (streamHandlerRef.current) {
+            websocketClient.stream.disconnect(streamHandlerRef.current, null);
+            streamHandlerRef.current = null;
+        }
+        
+        // 3. Clear streaming content
+        streamingContentRef.current = '';
```
aarondr77: The streamignContentRef is only used for chat mode, right? In agent mode we don't do any streaming. So does it make sense to do something with that here?

aarondr77: Same with the streamHanlderRef?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -901,23 +946,57 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+            });
+        } catch (error) {
+            console.error('Failed to send stop agent message to backend:', error);
+        }
+    }
+
+    const finalizeAgentStop = (): void => {
+        // This function is now handled by markAgentForStopping for immediate feedback
+        // Keeping it for backward compatibility with existing code that calls it
+        shouldContinueAgentExecution.current = false;
```
aarondr77: If we're not using this function anymore, then we can remove it. There's no backwards compatibility issues that we need to be worried about here.

aarondr77: But it does look like we are using it throughout the codebase still. What does this comment mean?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -789,7 +796,18 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+                    throw new Error('Request aborted');
+                }
+                
                 const aiResponse = await websocketClient.sendMessage<ICompletionRequest, ICompletionReply>(completionRequest);
+                
+                // Check if request was aborted after receiving response
+                if (abortController.signal.aborted) {
+                    throw new Error('Request aborted');
+                }
+
```
aarondr77: Since these are in the streaming case, we don't need them now. We don't have a stop button in chat mode right now.

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -801,6 +819,12 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
                     newChatHistoryManager.addAIMessageFromAgentResponse(agentResponse)
                 }
             } catch (error) {
+                // Check if this was an abort error
+                if ((error as any).message === 'Request aborted') {
+                    // Don't show error message for aborted requests
+                    return false;
+                }
```
aarondr77: And I guess we can remove this too

mito-ai/mito_ai/completions/handlers.py
```diff
@@ -222,7 +222,25 @@ async def on_message(self, message: str) -> None: # type: ignore
...
+            # Get the thread_id from metadata if available, otherwise use the newest thread
+            thread_id_to_stop = metadata_dict.get('threadId')
+
+            if not thread_id_to_stop:
+                # Get the newest thread ID
+                thread_id_to_stop = message_history._get_newest_thread_id()
+            
+            if thread_id_to_stop:
+                self.log.info(f"Stopping agent, thread ID: {thread_id_to_stop}")
+                message_history.remove_last_message(
```
aarondr77: Instead of calling `remove_last_message` let's call a function like: `add_thread_id_to_ignore_list`. 

Then, we just have to update the `message_history.append_message()` function so that it first checks the `thread_id_to_ignore_list`.

If the thread_is is present, 1) don't append the message 2) remove the thread_id from the list 3) and return. 

If the thread_is is not present: continue as we normally do.

mito-ai/mito_ai/completions/message_history.py
```diff
@@ -327,6 +327,39 @@ def truncate_histories(self, index: int, thread_id: ThreadID) -> None:
             self._update_last_interaction(thread)
             self._save_thread_to_disk(thread)
 
+    def remove_last_message(self, thread_id: ThreadID, new_message: Optional[ChatCompletionMessageParam] = None) -> bool:
```
aarondr77: Then, we can delete all of this. Nice!

aarondr77:
The agent seems to go back to the previous request once the user starts the conversation again. 

https://github.com/user-attachments/assets/14f60544-043c-48ba-8039-4bfdeb775766

mito-ai/mito_ai/completions/handlers.py
```diff
@@ -222,7 +222,34 @@ async def on_message(self, message: str) -> None: # type: ignore
                 )
                 self.reply(reply)
             return
-        
+
+        if type == MessageType.STOP_AGENT:
+            # Get the thread_id from metadata if available, otherwise use the newest thread
```
aarondr77: This is out of date. If the thread_id is not passed, then we don't process the request

mito-ai/mito_ai/completions/message_history.py
```diff
@@ -152,6 +152,9 @@ def __init__(self) -> None:
 
         # In-memory cache of all chat threads loaded from disk
         self._chat_threads: Dict[ThreadID, ChatThread] = {}
+        
+        # Set of thread IDs to ignore for message appending
+        self._thread_ids_to_ignore: set[ThreadID] = set()
```
aarondr77: Remove

PR #1976, opened 2025-09-26
Better MSSQL Connecting Instructions

ngafar:
# Description

Added explicit instructions on how the AI should connect to a Microsoft SQL Server database. 

# Testing

Add a MSSQL connection, and try to connect. I have the credentials for our AWS db, message me if you need. 

# Documentation

N/A

### DIFF
--- a/mito-ai/mito_ai/completions/prompt_builders/prompt_constants.py
+++ b/mito-ai/mito_ai/completions/prompt_builders/prompt_constants.py
@@ -125,6 +125,12 @@ If the user has requested data that you believe is stored in the database:
 connections[connection_name]["username"]
 ```
 
+- The user may colloquially ask for a "list of x", always assume they want a pandas DataFrame. 
+- When working with dataframes created from an SQL query, ALWAYS use lowercase column names. 
+- If you think the requested data is stored in the database, but you are unsure, then ask the user for clarification.
+
+## Additional MSSQL Rules
+
 - When connecting to a Microsoft SQL Server (MSSQL) database, use the following format:
 
 ```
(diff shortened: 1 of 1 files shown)


mito-ai/mito_ai/completions/prompt_builders/prompt_constants.py
```diff
@@ -125,6 +125,20 @@ def get_database_rules() -> str:
...
+encoded_password = urllib.parse.quote_plus(password) 
+conn_str = f"mssql+pyodbc://username:encoded_password@host:port/database?driver=ODBC+Driver+18+for+SQL+Server&TrustServerCertificate=yes"
+```
+
+- Always URL-encode passwords for MSSQL connections to handle special characters properly.
+- Include the port number in MSSQL connection strings.
+- Use "ODBC+Driver+18+for+SQL+Server" (with plus signs) in the driver parameter.
+- Always include "TrustServerCertificate=yes" for MSSQL connections to avoid SSL certificate issues.
+
 - When connecting to an Oracle database, use the following format:
```
aarondr77, line 150: Thoughts on updating the formatting of this a little bit so its more clear that these instructions only apply to Microsoft SQL Server?

Maybe something like 

### Additional Microsoft SQL Server Rules
.... 


### Additional Oracle Rules
.... 



And then we can move the general rules below up so its more clear it applies to everything?

aarondr77 (approved):

aarondr77 (approved):
Awesome. Just one thought about the system prompt. Feel free to do with it what you think. Then LGTM.

PR #1977, opened 2025-09-26
User Sign Up - Mito AI

ngafar:
# Description

Added a signup screen to the chat taskpane. 

# Testing

Remove your email from the `user.json` file. 

# Documentation

N/A - should be self-explanatory.

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
@@ -303,14 +303,14 @@ const ChatInput: React.FC<ChatInputProps> = ({
             return 'Agent is stopping...';
         } else if (agentModeEnabled) {
             return 'Ask agent to do anything';
+        } else if (isEditing) {
+            return 'Edit your message';
         } else if (displayOptimizedChatHistoryLength < 2) {
             return `Ask question (${operatingSystem === 'mac' ? '⌘' : 'Ctrl'}E), @ to mention`;
         } else {
             return `Ask followup (${operatingSystem === 'mac' ? '⌘' : 'Ctrl'}E), @ to mention`;
         }
     };
-
-
 
     // Automatically add active cell context when in Chat mode and there's active cell code
     useEffect(() => {


mito-ai/mito_ai/user/handlers.py
```diff
@@ -0,0 +1,33 @@
...
+    def put(self, key: str) -> None:
+        data = json.loads(self.request.body)
+        if "value" not in data:
+            self.set_status(400)
+            self.finish(json.dumps({"error": "Value is required"}))
+            return
+
+        set_user_field(key, data["value"])
+        identify() # Log the new user 
+        self.finish(json.dumps({"status": "updated", "key": key, "value": data["value"]}))
```
aarondr77: Maybe the status should be `success` instead of `updated`? I think that's kinda what we use elsewhere if I remember correctly

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1650,6 +1676,7 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
                                     chatInput.focus();
                                 }
                             }}
+                            disabled={isSignedUp === false}
```
aarondr77: These fields are not disabled or opaque either. One implementation idea this made me think of. Do you think it's possible to just disable the pointer events and add the opacity to the entire chat taskpane and all its children so we don't need to remember to do it everywhere? 

I think if we do something like, it would work? 

```css
.chat-taskpane *:not(.signup-form-container):not(.signup-form-container *) {
  color: blue;
  margin: 10px;
}
```

Image

mito-ai/src/Extensions/AiChat/SignUpForm.tsx
```diff
@@ -0,0 +1,53 @@
...
+            onSignUpSuccess?.();
+        } catch (error) {
+            console.error('Failed to set user email:', error);
+        }
+    };
+
+    return (
+        <div className="signup-form-container">
+            <div className="signup-form-message" data-testid="signup-form-message">
+                Get started with Mito, no credit card required, just an email.
```
aarondr77: I like the language that Cursor uses here. Maybe we could try something like: 

"To avoid abuse of our free AI credits, we ask that you login to use the AI features. No credit card required, just an email"

Image

mito-ai/src/Extensions/AiChat/SignUpForm.tsx
```diff
@@ -0,0 +1,53 @@
...
+                    className="signup-form-input"
+                    required
+                />
+                <button
+                    type="submit"
+                    className="button-base button-purple signup-form-button"
+                >
+                    Sign Up
+                </button>
+            </form>
```
aarondr77, line 58: I think we could make this pop a little bit more in dark mode. The disabled colors are less clear and the signup form itself is a little not standing out to me. 

What if we: 
1. Make the Sign Up Button a darker purple color?
2. Make the "Sign Up for Mito" title text a stronger color? In dark mode it could be white. In light mode it could be black. There must be a Jupyter color variable we can use for that. 

Image

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -148,6 +150,7 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
     websocketClient,
 }) => {
 
+    const [isSignedUp, setIsSignedUp] = useState<boolean>(false);
```
aarondr77, line 153: When the email is not set, the user is able to bypass it by using the Notebook Footer. And if they do so, it's not great because then they don't see the signup modal anymore, but they also can't use the taskpane. 

Can we do something about that? Let me know if you want to discuss different options

aarondr77 (changes requested):
Nice work on this! It looks very polished. I especially like all of the disabled and opacity changes you made here. A few changes here and then LGTM. Want to also hop on a quick call with @jake-stack to get his thoughts before you make any changes. He might have some good ideas. 

### Additional Logging

Can we just add a log when the Chat Taskpane opens called `opened_ai_chat_taskpane` you should be able to do this through the `logEvent` rest api endpoint very easily. 

That way we can calculate the drop-off rate of the signup requirement. 

### One big bug: 

For existing users that upgrade, instead of displaying the signup screen, they are taken to their last chat, but then all of the taskpane buttons are disabled and they don't have a way to sign in. 

We need to let these users sign in. 

Image

aarondr77:
Looks a bit funky on small screens. Can we maybe:

1.  add some vertical spacing between the input and button on this mode
2. make the button take up the full width, so that it is equally sized with the input field

Screenshot 2025-10-01 at 11 37 27 AM

mito-ai/src/Extensions/AiChat/SignUpForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return (
+        <div className="signup-form-container">
+            <div className="signup-form-header-logo">
+                <MitoLogo width="60" height="30" />
+            </div>
+            <span className="signup-form-header">
+                Sign Up for Mito
+            </span>
+            <div className="signup-form-message" data-testid="signup-form-message">
+                To avoid abuse of our free AI credits, we ask that you login to use the AI features. No credit card required, just an email.
```
aarondr77:
```suggestion
                To avoid abuse of our free AI credits, we ask that you login to use Mito AI. No credit card required, just an email.
```

aarondr77:
Can we visually disable the attach icon too?

Screenshot 2025-10-01 at 11 39 55 AM
Screenshot 2025-10-01 at 11 40 03 AM

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1470,7 +1470,7 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
     };
 
     return (
-        <div className="chat-taskpane">
+        <div className={`chat-taskpane ${isSignedUp ? '' : 'disabled'}`}>
```
aarondr77: You can use the `classNames` utility if you want

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1453,7 +1468,7 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
     };
 
     return (
-        <div className="chat-taskpane">
+        <div className={`chat-taskpane ${isSignedUp  || displayOptimizedChatHistory.length > 0 ? '' : 'disabled'}`}>
```
aarondr77: Can you just document the why here.

aarondr77 (commented):
Nice. LGTM with tiny polish updates and running by Jake.

PR #1990, opened 2025-10-03
mito-ai: fix common header issues

aarondr77:
# Description

# Testing

# Documentation


mito-ai/mito_ai/streamlit_conversion/agent_utils.py
```diff
@@ -189,7 +191,26 @@ def fix_diff_headers(diff: str) -> str:
...
-    return '\n'.join(lines)
+    corrected_diff = '\n'.join(lines)
+    corrected_diff = corrected_diff.lstrip()
+    
+    # If there is no diff, just return it without fixing file headers
+    if len(corrected_diff) == 0:
+        return corrected_diff
+    
+    # Remove known problametic file component headers that the AI sometimes returns
+    problamatic_file_header_components = ['--- a/app.py +++ b/app.py']
```
PoojasreeDwarakanath, line 202: Is it always a.py and b.py which shows up?

Issue #1991, opened 2025-10-03
Notebook Footer disabled for existing users without email


aarondr77:
Existing users who have previous chats (so they do not see the signup screen) have the notebook footer disabled. They won't see the signup screen until they create a new chat. 

<img width="1538" height="1028" alt="Image" src="https://github.com/user-attachments/assets/f4930591-95d1-440b-9b40-bbef01739385" />

PR #1993, opened 2025-10-06
Chat History REST API

ngafar:
# Description

Added two new REST endpoints:

1. `mito-ai/chat-history/threads` - returns a list of all threads.
2. `mito-ai/chat-history/threads/<thread_id>` - returns a specific thread. 

This makes it easier to retrieve a user's chat history, without the need to setup a webhook connection, and allows us to resolve https://github.com/mito-ds/mito/issues/1991.

# Testing

- Test the endpoints above.
- Delete your email from the `user.json` and make sure the footer is not disabled. 

# Documentation

N/A - behind the scenes change.

### DIFF
--- a/mito-ai/mito_ai/chat_history/handlers.py
+++ b/mito-ai/mito_ai/chat_history/handlers.py
@@ -2,39 +2,59 @@
 # Distributed under the terms of the GNU Affero General Public License v3.0 License.
 
 import tornado
-from typing import List
+from typing import List, Any
 from jupyter_server.base.handlers import APIHandler
 from mito_ai.completions.message_history import GlobalMessageHistory
 from mito_ai.completions.models import ChatThreadMetadata
 
 
-class ThreadHandler(APIHandler):
+class ChatHistoryHandler(APIHandler):
     """
     Endpoints for working with chat history threads.
     """
 
-    # @tornado.web.authenticated
-    def get(self) -> None:
-        """Get all chat history threads."""
+    @tornado.web.authenticated
+    def get(self, *args: Any, **kwargs: Any) -> None:
+        """Get all chat history threads or a specific thread by ID."""
         try:
             # Get the global message history instance
             message_history = GlobalMessageHistory()
             
-            # Get the list of threads
-            threads: List[ChatThreadMetadata] = message_history.get_threads()
+            # Check if a specific thread ID is provided in the URL
+            thread_id = kwargs.get("thread_id")
             
-            # Convert to dict format for JSON serialization
-            threads_data = [
-                {
-                    "thread_id": thread.thread_id,
-                    "name": thread.name,
-                    "creation_ts": thread.creation_ts,
-                    "last_interaction_ts": thread.last_interaction_ts,
-                }
-                for thread in threads
-            ]
-            
-            self.finish({"threads": threads_data})
+            if thread_id:
+                # Get specific thread
+                if thread_id in message_history._chat_threads:
+                    thread = message_history._chat_threads[thread_id]
+                    thread_data = {
+                        "thread_id": thread.thread_id,
+                        "name": thread.name,
+                        "creation_ts": thread.creation_ts,
+                        "last_interaction_ts": thread.last_interaction_ts,
+                        "display_history": thread.display_history,
+                        "ai_optimized_history": thread.ai_optimized_history,
+                    }
+                    self.finish(thread_data)
+                else:
+                    self.set_status(404)
+                    self.finish({"error": f"Thread with ID {thread_id} not found"})
+            else:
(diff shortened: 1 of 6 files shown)


mito-ai/mito_ai/chat_history/handlers.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    Endpoints for working with chat history threads.
+    """
+
+    @tornado.web.authenticated
+    def get(self, *args: Any, **kwargs: Any) -> None:
+        """Get all chat history threads or a specific thread by ID."""
+        try:
+            # Get the global message history instance
+            message_history = GlobalMessageHistory()
```
aarondr77: Hmmm I think there is a subtle logic issue here. 

There is a comment in the handlers.py that says the following: 

The GlobalMessageHistory is responsible for updating the message histories stored in the .mito/ai-chats directory. We create one GlobalMessageHistory per backend server instance instead of one per websocket connection so that the there is one manager of the locks for the .mito/ai-chats directory. This is my current understanding and it might be incorrect!

So I think by having multiple message histories, we can get into issues where two message histories are trying to edit the same file at the same time. I'n not 100% clear if the locks are per file on the computer or per global message history.

mito-ai/mito_ai/chat_history/handlers.py
```diff
@@ -0,0 +1,62 @@
...
+                # Convert to dict format for JSON serialization
+                threads_data = [
+                    {
+                        "thread_id": thread.thread_id,
+                        "name": thread.name,
+                        "creation_ts": thread.creation_ts,
+                        "last_interaction_ts": thread.last_interaction_ts,
+                    }
+                    for thread in threads
+                ]
```
aarondr77, line 57: I think we already have a function for this in the message_history that we can use. It's called `get_threads`. Would def be nice to reuse the existing logic.

mito-ai/src/restAPI/RestAPI.tsx
```diff
@@ -160,4 +160,42 @@ export const setUserKey = async (key: string, value: string): Promise<string> =>
...
+    }
+    return resp.data?.threads || [];
+}
+
+export const getChatHistoryThread = async (threadId: string): Promise<ChatThread> => {
+    const resp = await requestAPI<ChatThread>(`chat-history/threads/${threadId}`)
+    if (resp.error) {
+        throw new Error(resp.error.message);
+    }
+    return resp.data!;
```
aarondr77, line 200: It actually looks like we're not using this anywhere. If so, let's get rid of that whole flow for now. Better to not have unused code paths around I think.

mito-ai/mito_ai/chat_history/handlers.py
aarondr77: I think in order to fix this we could do the following: 

1. Currently, the global message history is created in the handlers.py file of the completion handler. This makes it hard to actually pass the global message history to the rest api endpoints because its not created yet in the init.py file
2. So can we move the creation of the global message history into the init.py file and then we can pass it down to the completion handlers and whatever rest endpoints need it? 

This would allow it to actually be global!

mito-ai/src/Extensions/NotebookFooter/NotebookFooter.tsx
```diff
@@ -27,24 +27,47 @@ const NotebookFooter: React.FC<NotebookFooterProps> = ({ notebookTracker, app })
...
+            // If user has email, they're signed up
+            if (hasEmail) {
+                setIsSignedUp(true);
+                return;
+            }
+            
+            // If no email, check if they have any chat history threads.
+            // Existing users will have a chat history, but no email,
+            // in this case, we consider them "signed up" and won't disable the footer. 
+            const hasThreads = await checkChatHistoryThreads();
```
aarondr77: This logic is inconsistent with the ChatTaskpane. In the chat taskpane we check if there is a message and it has at least one message in it: `isSignedUp || displayOptimizedChatHistory.length > 0`

I think because of the mismatch, we end up with this scenario where:

1. The last message I had open was empty
2. So the chat taskpane shows the signup, but the notebook footer still lets me send a message. 

Can we make a helper function that we use in both places to make sure the logic is consistent?

https://github.com/user-attachments/assets/1267faeb-826e-47ab-bdfc-82d6736bf8ef

aarondr77 (changes requested):

mito-ai/mito_ai/chat_history/handlers.py
ngafar: Slightly confused, is that not what's happening here. Going to the `get_threads()` definition, redirects you to `message_history.py`.

PR #2001, opened 2025-10-08
Cache anthropic

aarondr77:
# Description

Add caching for anthropic. We use 2 cache breakpoints.

1. System prompt (static content) is always cached with a breakpoint, so it's never reprocessed after the first request.
2. Conversation history gets a cache breakpoint on the last message that has passed the trim_old_message threshold, enabling incremental caching. Messages are stable after they are more than MESSAGE_HISTORY_TRIM_THRESHOLD old. At this point, the messages are not edited anymore, so they will not invalidate the cache. If we included the messages before the boundary in the cache, then every time we send a new  message, we would invalidate the cache and we would never get a cache hit except for the system prompt.

Check out this PR too which lets us track cache performance: https://github.com/mito-ds/ai-completion-server/pull/15

# Testing

Try it out, an run the ai-completion-server locally so you can observe the cache hit metrics. 

# Documentation

No. 

---

> [!NOTE]
> Adds Anthropic request caching (system prompt + stable history) and switches message trimming to a global threshold, with supporting type updates and tests.
> 
> - **Anthropic provider**:
>   - Add caching utilities: `add_cache_control_to_message` and `get_anthropic_system_prompt_and_messages_with_caching` (applies `cache_control: {type: "ephemeral"}` to system prompt and the stable-history boundary based on `MESSAGE_HISTORY_TRIM_THRESHOLD`).
>   - Use caching-aware messages in `AnthropicClient.request_completions` and `.stream_completions`.
> - **Message history**:
>   - `trim_old_messages` now uses `MESSAGE_HISTORY_TRIM_THRESHOLD` (removed `keep_recent` param) and preserves recent messages; still trims only user content and simplifies mixed-content trimming.
> - **Utils/Types**:
>   - Update Anthropic utils to accept `system` as `str | List[TextBlockParam] | anthropic.Omit` and propagate to server/stream params.
> - **Constants**:
>   - Add `MESSAGE_HISTORY_TRIM_THRESHOLD = 3`.
> - **Tests**:
>   - Update message history tests to threshold-based API and expand scenarios.
>   - Add caching tests for system prompt and history boundary; adjust provider tests and imports.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bd4f732cc10570deae0044f617a4bdf4625bd909. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/mito-ai/mito_ai/anthropic_client.py
+++ b/mito-ai/mito_ai/anthropic_client.py
@@ -5,10 +5,9 @@ import json
 import anthropic
 from typing import Dict, Any, Optional, Tuple, Union, Callable, List, cast
 
-from anthropic.types import Message, MessageParam
-from mito_ai.completions.models import CompletionError, ResponseFormatInfo, CompletionReply, CompletionStreamChunk, CompletionItem, MessageType
+from anthropic.types import Message, MessageParam, TextBlockParam
+from mito_ai.completions.models import ResponseFormatInfo, CompletionReply, CompletionStreamChunk, CompletionItem, MessageType
 from mito_ai.constants import MESSAGE_HISTORY_TRIM_THRESHOLD
-from mito_ai.utils.mito_server_utils import ProviderCompletionException
 from openai.types.chat import ChatCompletionMessageParam
 from mito_ai.utils.anthropic_utils import get_anthropic_completion_from_mito_server, stream_anthropic_completion_from_mito_server, get_anthropic_completion_function_params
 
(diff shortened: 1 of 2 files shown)


.cursorrules
```diff
@@ -0,0 +1,5 @@
+# Cursor Rules for Mito Project
+
+## Virtual Environment
+- ALWAYS activate the virtual environment before running any Python commands: `cd mito-ai && source venv/bin/activate`
```
ngafar, line 4: This rule is a great idea. I can't count the number of times Cursor has installed packages to my base env.

PR #2025, opened 2025-10-16
Drag-and-Drop File Uploads

ngafar:
# Description

Lets users drag files into the ChatInput to upload.

# Testing

Drag a file in.

# Documentation

Maybe. Drag-and-drop upload has become so widespread, that I'd assume this is expected behavior. But it might be good to draw the users attention to this feature.

### DIFF
--- a/mito-ai/style/ChatInput.css
+++ b/mito-ai/style/ChatInput.css
@@ -23,7 +23,7 @@
 }
 
 .chat-input-container.drag-over {
-  border: 2px dashed var(--jp-brand-color1);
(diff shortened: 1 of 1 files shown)


mito-ai/style/ChatInput.css
```diff
@@ -22,6 +22,11 @@
   margin-top: 0;
 }
 
+.chat-input-container.drag-over {
+  border: 2px dashed var(--purple-500);
```
aarondr77, line 26: Looks super good!

PR #2027, opened 2025-10-17
Give Agent Active Cell ID

ngafar:
# Description

We now send the active cell ID as part of the metadata to the backend. This ID is then used in the agent execution prompt, which already includes an array of `aiOptimizedCells`. By using the ID with this array, they agent now knows what cell the user is on, and the cell's content.  

# Testing

Ask the agent to make edits to a specific cell. Keep your requests vague, and look to make sure that only the selected cell is altered. 

# Documentation

N/A

### DIFF
--- a/mito-ai/mito_ai/completions/prompt_builders/agent_system_message.py
+++ b/mito-ai/mito_ai/completions/prompt_builders/agent_system_message.py
@@ -469,4 +469,4 @@ REMEMBER, YOU ARE GOING TO COMPLETE THE USER'S TASK OVER THE COURSE OF THE ENTIR
 
 OTHER USEFUL INFORMATION:
 1. When importing matplotlib, write the code `%matplotlib inline` to make sure the graphs render in Jupyter
-"""
+2. The active cell ID is shared with you so that when the user refers to "this cell" or similar phrases, you know which cell they mean. However, you are free to edit any cell that you see fit."""


mito-ai/src/Extensions/AiChat/ChatHistoryManager.tsx
```diff
@@ -203,6 +203,7 @@ export class ChatHistoryManager {
 
         const agentExecutionMetadata: IAgentExecutionMetadata = {
             promptType: 'agent:execution',
+            activeCellId: getActiveCellID(this.notebookTracker) || '',
```
aarondr77: Hmm I think this is going to return the wrong results when the user switches notebooks while the agent is working. So instead of using the notebookTracker, let's use the notebookPanel that we pass to the function. The notebookPanel is the agentTargetNotebook which we store whenever a new agent execution starts.

mito-ai/mito_ai/completions/prompt_builders/agent_execution_prompt.py
```diff
@@ -40,6 +41,9 @@ def create_agent_execution_prompt(md: AgentExecutionMetadata) -> str:
 {STREAMLIT_APP_STATUS_SECTION_HEADING}
 {streamlit_status_str}
 
+{ACTIVE_CELL_ID_SECTION_HEADING}
```
aarondr77, line 44: Thoughts on adding something to the prompt in the `OTHER USEFUL INFORMATION` section at the bottom that just says something like: "The active cell id is shared with you so that if the user refers to something like 'this' cell, you know what they are talking about, but you are free to edit any cell that you see fit". Something to that effect..

aarondr77 (approved):
Small changes and then LGTM

PR #2030, opened 2025-10-20
Agent Review 

ngafar:
# Description

This PR is adds:

1. _Review Changes_ button after the agent has completed.
2. Diff view that shows all of the changes the agent made.
3. Per cell accept/reject buttons. 
4. Auto-accept any remaining diffs before processing the next message.

# Testing

Create a notebook with lots of cells. Have the agent update and add new cells. Then, play around with the diffing, accepting and rejecting some cell. Finally, continue to work with the agent to ensure changes are permanent. 

There are lots of potential cases to test, here are just a few:

1. No review.
2. Review, but don't accept/reject any diffs.
3. Review, but accept some diffs.

# Documentation

N/A - Users should see button, and the uses cases are clearly labeled. 

---

> [!NOTE]
> Introduces an agent code review workflow with diff visualization and accept/reject (all or per-cell), refactors execution/review into hooks, updates chat input/message APIs, and adds comprehensive tests.
> 
> - **Agent Review UX**
>   - Adds review flow with "Review Changes" controls and panel; supports Accept all/Reject all and Undo via checkpoints.
>   - Enables per‑cell Accept/Reject via toolbar, auto-scrolls to next diff, and updates status across pre/in/post review states.
> - **Diffing & Notebook Utilities**
>   - Snapshots notebook before/after execution; computes unified diffs per changed cell; applies CodeMirror diff stripes via `applyCellEditorExtension`.
>   - New helpers: `turnOffDiffsForCell`, `shouldShowDiffToolbarButtons`, `scrollToNextCellWithDiff`, `runCellByIDInBackground`.
> - **Architecture & Hooks**
>   - Extracts `useAgentExecution` (agent loop, stop handling, checkpoints) and `useAgentReview` (change detection, accept/reject all or single).
>   - Auto-accepts any pending diffs before starting a new agent run.
> - **Chat Components & API Changes**
>   - Replaces `onSave`/`onUpdateMessage` with `handleSubmitUserMessage`; separates display vs AI-optimized context; updates NextSteps callback to receive `setAgentReviewStatus`.
>   - Keyboard shortcuts/command visibility adapt to chat vs agent review; centralized toolbar visibility logic.
> - **Tests**
>   - Updates unit tests for new props/signatures and behaviors.
>   - Adds UI tests for reviewing/accepting/rejecting all or single changes and diff stripe assertions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ee768008130b7e00f25a5816efedffab0471efc0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
@@ -218,7 +218,6 @@ const ChatInput: React.FC<ChatInputProps> = ({
             } else if (option.type === 'rule') {
                 setAdditionalContext(prev => [...prev, { type: 'rule', value: option.rule }]);
             } else if (option.type === 'db') {
-                console.log('db', option);
                 setAdditionalContext(prev => [
                     ...prev,
                     {
@@ -255,7 +254,6 @@ const ChatInput: React.FC<ChatInputProps> = ({
             } else {
                 contextChatRepresentation = `\`${option.variable.variable_name}\``
             }
-            console.log('variable', option);
         } else if (option.type === 'file') {
             // For files, add them as both back-ticked elements and the additional context container
             contextChatRepresentation = `\`${option.file.variable_name}\``
@@ -267,9 +265,7 @@ const ChatInput: React.FC<ChatInputProps> = ({
             // delete the context from there if they want to. 
             contextChatRepresentation = option.rule
             setAdditionalContext([...additionalContext, { type: 'rule', value: option.rule }]);
-            console.log('rule', option);
         } else if (option.type === 'db') {
-            console.log('variable', option);
             // For databases, add them as back-ticked elements
             contextChatRepresentation = `\`${option.variable.variable_name}\``
             setAdditionalContext([


mito-ai/src/utils/notebook.tsx
```diff
@@ -339,4 +340,122 @@ export const scrollToCell = (
...
+    const currentEditIndex = agentEdits.findIndex(edit => edit.cellId === currentCellId);
+    if (currentEditIndex < 0) {
+        return; // Current cell not found in edits
+    }
+
+    // Find the next cell that still has a diff
+    const nextCellWithDiff = agentEdits
+        .slice(currentEditIndex + 1)
+        .find(edit => changedCells.some(change => change.cellId === edit.cellId));
+
```
aarondr77: 1. If you have the agent create 3 cells
2. Click Review
3. Instead of resolving the first cell, you click on the last of the 3 new cells
4. Click Accept
5. This code doesn't identify that it should now take you back to the first cell. 

I think to implement this, we can:
1.  Remove the cells from the `changedCellsRef` as we Accept/Reject them because that is what we are feeding to this function as the `changedCells` it looks like. 
2. When we reach the end of the list, if there are still cells left in `changedCells`, we go back to the start of the list

https://github.com/user-attachments/assets/44d11d4e-8f9c-4d95-9613-2e3fba0995ac

mito-ai/src/utils/notebook.tsx
```diff
@@ -339,4 +340,122 @@ export const scrollToCell = (
...
+        });
+    } else {
+        // Reconfigure existing compartment
+        editorView.dispatch({
+            effects: compartment.reconfigure(extension),
+        });
+    }
+}
+
+export const runCellByIDInBackground = async (notebookPanel: NotebookPanel | null, cellId: string): Promise<void> => {
```
aarondr77, line 446: Nice. Should we be using this helper function in other places now too? Like in the `acceptAndRunCellUpdate`

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1300,6 +1366,33 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+            if (originalCode === undefined) return;
+            
+            // Remove from tracking and restore original code
+            cellStatesBeforeDiff.current.delete(activeCellId);
+            writeCodeToCellByID(notebookTracker, originalCode, activeCellId);
+            turnOffDiffsForCell(notebookTracker, activeCellId, codeDiffStripesCompartments.current);
+            updateCellToolbarButtons();
+
+            // Re-run cell in background
+            void runCellByIDInBackground(notebookTracker.currentWidget, activeCellId);
```
aarondr77: Oh interesting. It's not clear to me if we should or shouldn't be rerunning the cells when the user rejects the AI code. I guess you're right that rerunning makes sense. Otherwise the user is going to have to rerun the cells so that the agent has the correct version of all of the variables. Can we document that here?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1300,6 +1366,33 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
     }
 
     const rejectAICode = (): void => {
+        const activeCellId = notebookTracker.activeCell?.model.id;
```
aarondr77: Thoughts on organizing the code like this: 

1. AcceptAllCellEdits 
2. RejectAllCellEdits
3. AcceptSingleCellEdit
4. RejectSingleCellEdit

And then the Accept and Reject All functions can just call the single cell accept and reject ones? 

I think that might be more clear than relying on these checks in the function to determine if we are in AgentMode or Chat Mode and how we want to handle them. 

Also, for the sake of this file, maybe we can move those four helper functions into a new file called "CellEditAcceptAndRejectUtils.ts" or something like that?

mito-ai/src/utils/codeDiff.tsx
```diff
@@ -211,4 +220,42 @@ export const getCodeDiffsAndUnifiedCodeString = (originalCode: string | undefine
...
+export const shouldShowDiffToolbarButtons = (
+    notebookTracker: INotebookTracker,
+    cellStateBeforeDiff: ICellStateBeforeDiff | undefined,
+    cellStatesBeforeDiff: Map<string, string>
+): boolean => {
+    try {
+        const activeCellId = notebookTracker.activeCell?.model.id;
+        if (!activeCellId) return false;
+        
+        // Check both single-cell mode and multi-cell mode
```
aarondr77, line 252: Can you expand on this comment to explain why we need to be doing this? This feels like something we're going to look back on in two months and be pretty confused.

mito-ai/src/utils/codeDiff.tsx
```diff
@@ -211,4 +220,42 @@ export const getCodeDiffsAndUnifiedCodeString = (originalCode: string | undefine
...
+
+export const turnOffDiffsForCell = (
+    notebookTracker: INotebookTracker,
+    cellId: string,
+    codeDiffStripesCompartments: Map<string, Compartment>
+): void => {
+    applyDiffStripesToCell(notebookTracker, cellId, null, codeDiffStripesCompartments);
+}
+
+export const shouldShowDiffToolbarButtons = (
```
aarondr77, line 243: How did you fix that error we were seeing where if the cell was already the active cell when you clicked the `Review Changes` button it would not show the Accept/Reject buttons in the cell?

mito-ai/src/utils/notebook.tsx
```diff
@@ -339,4 +340,122 @@ export const scrollToCell = (
...
+    const nextCellWithDiff = agentEdits
+        .slice(currentEditIndex + 1)
+        .find(edit => changedCells.some(change => change.cellId === edit.cellId));
+
+    if (!nextCellWithDiff) {
+        return; // No more cells with diffs
+    }
+
+    // Scroll to and select the next cell with diff
+    setActiveCellByID(notebookTracker, nextCellWithDiff.cellId);
```
aarondr77: I think we should be using the notebookPanel here. Reminder, pretty much whenever we are using the Agent mode, let's use the notebookPanel which we store at the beginning of agent mode in the ref `agentTargetNotebookPanelRef`

mito-ai/src/utils/notebook.tsx
```diff
@@ -339,4 +340,122 @@ export const scrollToCell = (
...
+    const cell = notebook.widgets.find(widget => widget.model.id === cellId);
+    if (!cell || cell.model.type !== 'code') return;
+
+    // Set the cell as active temporarily
+    const originalActiveCellIndex = notebook.activeCellIndex;
+    const cellIndex = notebook.widgets.findIndex(widget => widget.model.id === cellId);
+    notebook.activeCellIndex = cellIndex;
+
+    try {
+        // Run the cell without awaiting - this makes it run in the background
```
aarondr77, line 462: Hmm can you explain why we want it to run in the background? Is this introducing any race conditions we should be aware of?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1523,6 +1591,90 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+
+        if (!notebookSnapshot || !currentNotebookSnapshot) {
+            return;
+        }
+
+        // Clear and populate the map of original cell states
+        cellStatesBeforeDiff.current.clear();
+
+        // Find cells that have changed between snapshots
+        const changedCells: { cellId: string, originalCode: string, currentCode: string }[] = [];
```
aarondr77: Can we create a type for `{ cellId: string, originalCode: string, currentCode: string }` and use that across the codebase? Right now we have to type this out in a bunch of places

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -235,6 +245,8 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
 
     // Track if checkpoint exists for UI updates
     const [hasCheckpoint, setHasCheckpoint] = useState<boolean>(false);
+    const [notebookSnapshot, setNotebookSnapshot] = useState<AIOptimizedCell[] | null>(null);
```
aarondr77: A few notes here: 

1. Can we call this `notebookSnapshotPreAgentExecution` and then maybe `notebookSnapshotAfterAgentExecution`? I wasn't sure what notebookSnapshot referred to
2. Do we need notebookSnapshotPreAgentExecution to be a state variable? instead, can it just be a ref?

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1523,6 +1591,90 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
+            }
+        });
+
+        if (changedCells.length === 0) {
+            console.log('No changes detected between snapshots');
+            return;
+        }
+
+        // For each changed cell, calculate and apply diff stripes
+        changedCells.forEach(change => {
```
aarondr77: Sweet!

aarondr77 (changes requested):
Awesome. Thanks for griding on this. I personally will use this feature when I'm using agent mode! 

I think it's working pretty well, but there might be a little room for code cleanup, which I think is worth doing just cause the frontend code is getting kinda complex! 

Also, can we add some basic tests for things?

mito-ai/src/utils/codeDiff.tsx
ngafar: I added `updateCellToolbarButtons();` to the end of `reviewAgentChanges`.

mito-ai/src/utils/notebook.tsx
ngafar: Adding this because we want to quickly remove the diffs, then send the user to the next cell. If there is a cell with a long running task, users will reject, then we re-run the cell. The user then has to wait for the cell to complete before they are taken to the next diff. 

My guess is that users will take this delay as a sign that there are no more diffs, and continue to prompt. 

I'm not sure about race conditions, but it's a good point. What did you have in mind?

mito-ai/src/utils/notebook.tsx
ngafar: I think we're good to leave things as-is for now. The "run in background" functionality is something we should probably use sparingly until we figure out if there are in fact any race condition issues.

mito-ai/src/Extensions/AiChat/CellEditAcceptAndRejectUtils.ts
```diff
@@ -0,0 +1,208 @@
...
+import { runCellByIDInBackground } from '../../utils/notebook';
+import { getCodeBlockFromMessage } from '../../utils/strings';
+import { ChatHistoryManager } from './ChatHistoryManager';
+import { ChangedCell } from './ChatTaskpane';
+import { AIOptimizedCell } from '../../websockets/completions/CompletionModels';
+
+/**
+ * Accepts a single cell edit in agent review mode
+ */
+export const AcceptSingleCellEdit = (
```
aarondr77: These functions should be in camel case. So `acceptSingleCellEdit`

mito-ai/src/Extensions/AiChat/ChatTaskpane.tsx
```diff
@@ -1285,24 +1287,15 @@ const ChatTaskpane: React.FC<IChatTaskpaneProps> = ({
...
-            
-            // Write the final code to the cell and turn off diffs
-            writeCodeToCellByID(notebookTracker, edit?.code || '', activeCellId);
-            turnOffDiffsForCell(notebookTracker, activeCellId, codeDiffStripesCompartments.current);
-            updateCellToolbarButtons();
-            
-            // Scroll to the next cell with a diff if in agent mode
-            scrollToNextCellWithDiff(
-                notebookTracker,
+            AcceptSingleCellEdit(
```
aarondr77: wow nice. A lot of code gone from ChatTaskpane

PR #2032, opened 2025-10-23
Fallback for E-Mail Signup

ngafar:
# Description

If we’re unable to set the user’s email, we now log the error and store a flag in local storage. This flag tells the chat task pane whether the user can start using Mito AI. When the flag is present, the user is allowed to use Mito AI.

# Testing

Try to get an error. You can break the URL in `RestAPI.tsx` for `setUserKey`. 

# Documentation

N/A

### DIFF
--- a/mito-ai/src/utils/userSignupState.ts
+++ b/mito-ai/src/utils/userSignupState.ts
@@ -5,13 +5,6 @@
 
 import { getUserKey, getChatHistoryThreads, getChatHistoryThread } from '../restAPI/RestAPI';
 
-export interface UserSignupState {
-    isSignedUp: boolean;
-    hasEmail: boolean;
-    hasChatHistory: boolean;
-    hasSoftSignup: boolean;
-}
-
 /**
  * Determines if a user should be considered "signed up" based on:
  * 1. Having an email address, OR
@@ -21,7 +14,7 @@ export interface UserSignupState {
  * This ensures consistent behavior across all components that need to check
  * if a user should have access to Mito AI features.
  */
-export const checkUserSignupState = async (): Promise<UserSignupState> => {
+export const isUserSignedUp = async (): Promise<boolean> => {
     try {
         // Check for soft signup flag first (for cases where email setting failed)
         const hasSoftSignup = localStorage.getItem('mito_ai_soft_signup') === 'true';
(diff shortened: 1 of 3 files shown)


mito-ai/src/Extensions/AiChat/SignUpForm.tsx
```diff
@@ -25,7 +26,16 @@ const SignUpForm: React.FC<SignUpFormProps> = ({ onSignUpSuccess }) => {
             // Emit signup success event for other components to listen to
             userSignupEvents.emitSignupSuccess();
         } catch (error) {
+            // If we can't set the email, we still want the user to be able to use Mito AI.
+            // In this case, we log the error, and emit the signup success event.
             console.error('Failed to set user email:', error);
+            void logEvent('mito_ai_failed_to_set_user_email', { 'error': error, 'email': email });
```
ngafar: I'm not sure we'll see a lot of these events in the log. Chances are if the user can't reach the `setUserKey` endpoint, they won't be able to reach the `logEvent` endpoint. However, that's just a theory and this is only one line.

aarondr77: Hmm does that mean that this is going to raise an exception also then? So we actually won't get down to the setting localStorage part?

mito-ai/src/utils/userSignupState.ts
```diff
@@ -39,9 +45,10 @@ export const checkUserSignupState = async (): Promise<UserSignupState> => {
 
         if (!hasThreads) {
             return {
-                isSignedUp: false,
+                isSignedUp: hasSoftSignup,
                 hasEmail: false,
-                hasChatHistory: false
+                hasChatHistory: false,
+                hasSoftSignup: hasSoftSignup
```
aarondr77: It doesn't look like we are ever using these different fields. Should we just simplify this to: 

1. Calling the function `isUserSignedUp`
2. It just returns True or False

mito-ai/src/Extensions/AiChat/SignUpForm.tsx
aarondr77: @ngafar ^

ngafar: Good catch. I've moved the `logEvent` to the bottom.

ngafar:
@cursoragent review

PR #2059, opened 2025-11-18
Modifying manage apps

PoojasreeDwarakanath:
# Description
This PR contains the changes to the manage apps pane to allow the user to login in if they have not already. It also adds auto refresh of the pane on logout.

# Testing

# Documentation
\-

### DIFF
--- a/mito-ai/src/Extensions/AppManager/AppsList.tsx
+++ b/mito-ai/src/Extensions/AppManager/AppsList.tsx
@@ -47,6 +47,10 @@ export const AppsList: React.FC<AppsListProps> = ({ appManagerService }) => {
     }
   }
 
+  React.useEffect(() => {
+      void refreshApps();
+    }, []);
+
   const handleLogin = async (): Promise<void> => {
     try {
       await showAuthenticationPopup();


mito-ai/src/Extensions/AppManager/AppsList.tsx
```diff
@@ -21,44 +23,41 @@ export const AppsList: React.FC<AppsListProps> = ({ appManagerService }) => {
...
   const [error, setError] = React.useState<string | null>(null);
 
-  // Fetch apps on component mount
-  React.useEffect(() => {
-    const loadApps = async (): Promise<void> => {
-      try {
-        console.log('[AppsList] Starting to load apps...');
-        setLoading(true);
-        setError(null);
+  const refreshApps = React.useCallback(async (): Promise<void> => {
```
aarondr77: Is there any reason we need to have these useCallback and useEffects? I think we could just write this code like this:

```typescript

// HERE WE GET RID OF THE USECALLBACK
const refreshApps = async (): Promise<void> => {
    try {
      console.log('[AppsList] Refreshing apps...');
      setLoading(true);
      setError(null);

      const response: GetAppsResponse = await fetchUserApps(appManagerService);
      console.log('[AppsList] fetchUserApps response:', response);

      if (isGetAppsSuccess(response)) {
        setApps(response.apps);
      } else {
        setError(response.errorMessage || 'Failed to load apps');
        setApps([]);
      }
    } catch (err) {
      console.error('[AppsList] Error loading apps:', err);
      setError(err instanceof Error ? err.message : 'An unexpected error occurred');
      setApps([]);
    } finally {
      setLoading(false);
    }
  }


// HERE WE GET RID OF THE DEPENDENCY ON THE REFRESH APP BEING CREATED
  React.useEffect(() => {
    void refreshApps();
  }, []);


```

mito-ai/src/Extensions/AppManager/AppsList.tsx
```diff
@@ -113,9 +112,17 @@ export const AppsList: React.FC<AppsListProps> = ({ appManagerService }) => {
...
               console.log('Logout clicked');
-              void logoutAndClearJWTTokens();
+              try {
+                await logoutAndClearJWTTokens();
+                // Refresh the apps list after logout to show "User not authenticated" message
+                await refreshApps();
+              } catch (err) {
+                console.error('[AppsList] Error during logout:', err);
+                // Still refresh even if logout had an error
+                await refreshApps();
```
aarondr77: Since we want to do the refreshApps in both cases, maybe we can just move it to a finally clause?

PR #2066, opened 2025-11-24
Haiku 4.5

aarondr77:
# Description

Adds Claude Haiku 4.5 Support. Try it out. 

If it passes our first round of vibes test, then we can do more thorough testing with CI. 

Note: Anthropic only supports the beta extended context window with Sonnet, not haiku. So to avoid users running into the context window limit issues, I added some token estimation code. it's just a rough estimate, but if it estimates that we would exceed the 200k context window limit, then it automatically bumps the model to Sonnet 4.5 instead. 

# Testing

Each model we add, we generally have to do some extra system prompt changes to clarify the expected outputs or validate the correct results when we get them back. 

Give this new model a spin and see if you catch any. So far I have not fwiw

# Documentation

---

> [!NOTE]
> Adds Claude Haiku 4.5, introduces automatic model selection with rough token-based Sonnet fallback for large contexts, updates UI/tests, and raises Anthropic max tokens.
> 
> - **Anthropic backend**:
>   - **Auto model selection**: New `select_correct_model` uses rough token estimation (`does_message_exceed_max_tokens` via `get_rough_token_estimatation_anthropic`) to pick `claude-haiku-4-5-20251001` for fast tasks or switch to `claude-sonnet-4-5-20250929` when context is large; applied in `get_anthropic_completion_function_params` and streaming.
>   - **Token/limits**: Increase `MAX_TOKENS` to `64_000`; include `betas: ["context-1m-2025-08-07"]` in requests.
> - **Utils**:
>   - Add `mito_ai/utils/tokens.py` with `get_rough_token_estimatation_anthropic`.
> - **UI**:
>   - Update `ModelSelector` to include Haiku 4.5; adjust display names and keep Sonnet as default. Model constants added in `src/utils/models.ts`.
> - **Tests**:
>   - Extend LLM provider tests to run against Claude Haiku 4.5.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e3f40a2abb47f60f4f7d9cb31418c76df1e46697. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/mito-ai/mito_ai/anthropic_client.py
+++ b/mito-ai/mito_ai/anthropic_client.py
@@ -277,7 +277,7 @@ class AnthropicClient:
                               reply_fn: Callable[[Union[CompletionReply, CompletionStreamChunk]], None]) -> str:
         try:
             anthropic_system_prompt, anthropic_messages = get_anthropic_system_prompt_and_messages_with_caching(messages)
-            model = select_correct_model(model, message_type, anthropic_system_prompt, messages)
+            model = select_correct_model(model, message_type, anthropic_system_prompt, anthropic_messages)
 
             accumulated_response = ""
 


mito-ai/mito_ai/utils/tokens.py
```diff
@@ -0,0 +1,26 @@
...
+    General rule of thumb: 1 token is about 4 characters. 
+    To be safe we use: 1 token is about 5 characters
+    """
+    
+    try:
+        stringified_system_message = str(system_message)
+        stringified_messages = str(messages)
+        total_stringified_context = stringified_system_message + stringified_messages 
+        
+        return len(total_stringified_context) / 5 + 1
```
ngafar: Are we dividing by 6 (`5 + 1`) because we assume this is how many characters are in a token? 

If so, maybe we should store this in a variable for clarity.

mito-ai/mito_ai/utils/anthropic_utils.py
```diff
@@ -10,14 +10,37 @@
...
 
 __user_email: Optional[str] = None
 __user_id: Optional[str] = None
 
 ANTHROPIC_TIMEOUT = 60
 max_retries = 1
 
-FAST_ANTHROPIC_MODEL = "claude-3-5-haiku-latest"
+FAST_ANTHROPIC_MODEL = "claude-haiku-4-5-20251001" # This should be in sync with ModelSelector.tsx
+LARGE_CONTEXT_MODEL = "claude-sonnet-4-5-20250929" # This should be in sync with ModelSelector.tsx
```
ngafar, line 22: I actually did not know that Haiku was the faster model, and had to Google it. This is something I suspect will also confuse users. I wonder what's a good way to convey this info to users, so that they can decide between speed vs. deep thought.

mito-ai/src/components/ModelSelector.tsx
```diff
@@ -20,6 +20,7 @@ interface ModelMapping {
 const MODEL_MAPPINGS: ModelMapping[] = [
   { displayName: 'GPT 4.1', fullName: 'gpt-4.1' },
   { displayName: CLAUDE_SONNET_DISPLAY_NAME, fullName: CLAUDE_SONNET_MODEL_NAME },
+  { displayName: CLAUDE_HAIKU_DISPLAY_NAME, fullName: CLAUDE_HAIKU_MODEL_NAME },
```
ngafar, line 22: Might just be a me thing, but in my head the faster (or older) model is always on top.

mito-ai/mito_ai/utils/anthropic_utils.py
```diff
@@ -10,14 +10,37 @@
...
+
+def does_message_exceed_max_tokens(system: Union[str, List[TextBlockParam], anthropic.Omit], messages: List[MessageParam]) -> bool:
+    token_estimation = get_rough_token_estimatation_anthropic(system, messages)
+
+    if token_estimation is not None and token_estimation > 200_000:
+        return True
+    return False
+
+def select_correct_model(default_model: str, message_type: MessageType, system: Union[str, List[TextBlockParam], anthropic.Omit], messages: List[MessageParam]) -> str:
```
ngafar, line 31: Couple of questions:

- So we're deciding which model to use on the backend? 
- Does this override the chose the user makes on the frontend? 
- This is needed to make sure we don't blow up the context window?

aarondr77: So the reason we are doing this is: Anthropic only supports the beta extended context window with Sonnet, not haiku. So to avoid users running into the context window limit issues, I added some token estimation code. it's just a rough estimate, but if it estimates that we would exceed the 200k context window limit, then it automatically bumps the model to Sonnet 4.5 instead.

That 200k limit was what was causing that spike in errors where we would tell people to switch to GPT 4.1 instead. 

So to answer your questions..

> So we're deciding which model to use on the backend?

yeah

> Does this override the chose the user makes on the frontend?

yeah

> This is needed to make sure we don't blow up the context window?

yeah

mito-ai/src/components/ModelSelector.tsx
aarondr77: We can switch the order. but technically I think sonnet 4.5 was released before haiku 4.5

mito-ai/mito_ai/utils/anthropic_utils.py
aarondr77: This is a great callout. I think what we should do is just add some symbols or text that guides people in the right direction. ie: lightning bolt icon for fast, brain icon for deep thinker. And maybe when you hover over it, it could say something like "best for your hardest problems"

mito-ai/mito_ai/utils/tokens.py
aarondr77: The general rule of thumb is 1 token is about 4 characters. To be safe we're using estimating 1 token is about 3 (thanks for questioning this, made me realize my logic was wrong) characters.

mito-ai/mito_ai/utils/anthropic_utils.py
aarondr77: Going to save this for another PR btw

PR #2072, opened 2025-12-02
App Streaming

ngafar:
# Description

When creating the initial app, we now go cell-by-cell. This let's us use a faster model (Haiku), and opens up the possiblity of incrementally showing the app as it is being built for the first time.  

# Testing

Create a notebook with lots of cells, build an app. In the terminal you should see the steps for each cell. 

# Documentation

N/A

### DIFF
--- a/mito-ai/mito_ai/streamlit_conversion/streamlit_agent_handler.py
+++ b/mito-ai/mito_ai/streamlit_conversion/streamlit_agent_handler.py
@@ -73,7 +73,7 @@ async def generate_new_streamlit_code(notebook: List[dict], streamlit_app_prompt
             edit_prompt = f"Incorporate the following notebook cell into the existing Streamlit app, maintaining the app's structure and adding the new functionality:\n\n{cell_source}"
             streamlit_code = await update_existing_streamlit_code([cell], streamlit_code, edit_prompt)
     
-    return streamlit_code
+    return streamlit_code if streamlit_code else ""
 
 
 async def update_existing_streamlit_code(notebook: List[dict], streamlit_app_code: str, edit_prompt: str) -> str:


mito-ai/mito_ai/streamlit_conversion/agent_utils.py
```diff
@@ -18,7 +18,7 @@ def extract_todo_placeholders(agent_response: str) -> List[str]:
 async def get_response_from_agent(message_to_agent: List[MessageParam]) -> str:
     """Gets the streaming response from the agent using the mito server"""
     model = STREAMLIT_AI_MODEL
-    max_tokens = 64000 # TODO: If we move to haiku, we must reset this to 8192
+    max_tokens = 64000
```
ngafar, line 21: According to the Anthropic [docs](https://platform.claude.com/docs/en/about-claude/models/overview#latest-models-comparison) Haiku 4.5 supports a max output of 64k tokens.

mito-ai/mito_ai/streamlit_conversion/streamlit_agent_handler.py
```diff
@@ -32,31 +43,39 @@ async def generate_new_streamlit_code(notebook: List[dict], streamlit_app_prompt
     ]
     agent_response = await get_response_from_agent(messages)
     converted_code = extract_code_blocks(agent_response)
-    
-    # Extract the TODOs from the agent's response
-    todo_placeholders = extract_todo_placeholders(agent_response)
-    
-    for todo_placeholder in todo_placeholders:
-        print(f"Processing AI TODO: {todo_placeholder}")
-        todo_prompt = get_finish_todo_prompt(notebook, converted_code, todo_placeholder)
```
aarondr77, line 41: Now that we are going cell by cell I think it might be reasonable to not have the todo's anymore, but to delete it, I think we need to:

1. Let's delete the `get_finish_todo_prompt` and `extract_todo_placeholders` functions since we're not using them anymore 
2. Update the `get_streamlit_app_creation_prompt` function so it no longer tells the model to use TODOs

Then, let's make sure to test this [with this notebook](https://github.com/mito-ds/mito/blob/dev/mito-ai/example-notebooks/geospatial/bike-violations.ipynb) (its in the mito-ai/example-notebooks/geospatial folder) and confirm that it still moves over the entire `violation_data_list`

PR #2079, opened 2025-12-05
Re-add Desktop Version Check

ngafar:
# Description

Desktop users now see notification when there is a new version of the mito-ai package. I've also updated the upgrade links.

# Testing

Trick the server into thinking there is an update, and test the links. 

# Documentation

N/A - Users will see.

### DIFF
--- a/mito-ai/src/utils/version_check.ts
+++ b/mito-ai/src/utils/version_check.ts
@@ -6,10 +6,12 @@
 import { ServerConnection } from '@jupyterlab/services';
 import { Notification } from '@jupyterlab/apputils';
 import * as semver from 'semver';
+import { isElectronBasedFrontend } from './user';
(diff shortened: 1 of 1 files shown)


mito-ai/src/utils/version_check.ts
```diff
@@ -22,7 +24,10 @@ export function showVersionOutdatedNotification(currentVersion: string, latestVe
       {
         label: 'Learn how to update',
         callback: () => {
-          window.open(DOCS_UPDATE_URL, '_blank');
+          window.open(
+            isElectronBasedFrontend() ? DESKTOP_DOCS_UPDATE_URL : PIP_DOCS_UPDATE_URL,
```
aarondr77, line 28: Smart!

PR #2081, opened 2025-12-08
Spreadsheet Footer Button

ngafar:
# Description

Added a spreadsheet button to the notebook footer. When clicked it creates a new Mitosheet. 

# Testing

Use the button, make sure the Mitosheet works as expected.

# Documentation

No, user should be able to click and see.

### DIFF
--- a/mito-ai/src/tests/NotebookFooter/notebookFooter.test.tsx
+++ b/mito-ai/src/tests/NotebookFooter/notebookFooter.test.tsx
@@ -90,7 +90,8 @@ const createMockNotebook = (widgetCount = 3) => {
 // Create mock app
 const createMockApp = () => ({
     commands: {
-        execute: jest.fn().mockResolvedValue(undefined)
+        execute: jest.fn().mockResolvedValue(undefined),
+        hasCommand: jest.fn().mockReturnValue(false)
     }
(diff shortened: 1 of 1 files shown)


mito-ai/src/Extensions/NotebookFooter/NotebookFooter.tsx
```diff
@@ -60,13 +61,15 @@ const NotebookFooter: React.FC<NotebookFooterProps> = ({ notebookTracker, app })
...
-    const addCell = (cellType: 'code' | 'markdown' = 'code'): void => {
+    const addCell = (cellType: 'code' | 'markdown' | 'spreadsheet' = 'code'): void => {
         if (notebook.widgets.length && notebook.widgets.length > 0) {
             notebook.activeCellIndex = notebook.widgets.length - 1;
         }
 
         if (cellType === 'code') {
             NotebookActions.insertBelow(notebook);
+        } else if (cellType === 'spreadsheet') {
+            void app.commands.execute('mitosheet:create-empty-mitosheet');
```
aarondr77, line 78: If they don't have the mitosheet package installed, then this won't work. So we can we conditionally show this button depending on if they have it installed. We can just check by seeing if the mitosheet:create-empty-mitosheet command exists

PR #2096, opened 2025-12-11
Chat History Optimization 

ngafar:
# Description

-  Moved examples into system prompt.
- Trimming active cell code from old user messages.

The code preview that is shown in the `UserCodeBlock` component should not be affected. This is because we are making edits to the AI optimized messages, not the display optimized messages. 

# Testing

Using chat mode, send some messages, make sure things are working. Trimming happens after three messages, so make sure to hit that threshold. 

# Documentation

N/A

---

> [!NOTE]
> Moves examples into the system prompt and enhances message trimming to remove code sections while preserving notebooks for recent messages via a new threshold.
> 
> - **Prompting**:
>   - Move illustrative examples out of `chat_prompt.py` and into `create_chat_system_message_prompt()` with a new detailed Example 3 and citation formatting guidance.
>   - Simplify `create_chat_prompt()` to only include `files`, `variables`, `active_cell_id`, and `code` sections plus task.
> - **Message history trimming**:
>   - Extend `trim_sections_from_message_content` to trim `CODE_SECTION_HEADING` and optionally `JUPYTER_NOTEBOOK_SECTION_HEADING`.
>   - Add notebook preservation logic in `trim_old_messages` controlled by `NOTEBOOK_PRESERVATION_THRESHOLD` (recent messages keep notebook; older messages trim it).
>   - Handle mixed content by keeping only trimmed text content.
> - **Constants**:
>   - Add `NOTEBOOK_PRESERVATION_THRESHOLD: int = 6`.
> - **Tests**:
>   - Update tests to expect code section trimming and verify notebook preservation/mixed content behavior across prompt builders.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ead6b362666a9607138cdc182818a39b004ad5e4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/mito-ai/mito_ai/utils/message_history_utils.py
+++ b/mito-ai/mito_ai/utils/message_history_utils.py
@@ -6,19 +6,10 @@ from typing import List
 from mito_ai.constants import MESSAGE_HISTORY_TRIM_THRESHOLD, NOTEBOOK_PRESERVATION_THRESHOLD
 from openai.types.chat import ChatCompletionMessageParam
 from mito_ai.completions.prompt_builders.prompt_constants import (
-    ACTIVE_CELL_ID_SECTION_HEADING,
-    ACTIVE_CELL_OUTPUT_SECTION_HEADING,
-    GET_CELL_OUTPUT_TOOL_RESPONSE_SECTION_HEADING,
-    FILES_SECTION_HEADING,
-    STREAMLIT_APP_STATUS_SECTION_HEADING,
-    VARIABLES_SECTION_HEADING,
-    JUPYTER_NOTEBOOK_SECTION_HEADING,
     CONTENT_REMOVED_PLACEHOLDER,
-    CODE_SECTION_HEADING,
-    SELECTED_CONTEXT_SECTION_HEADING,
-    TASK_SECTION_HEADING,
-    ERROR_TRACEBACK_SECTION_HEADING,
-    ERROR_PRODUCING_CELL_ID_SECTION_HEADING,
+    JUPYTER_NOTEBOOK_SECTION_HEADING,
+    iter_prompt_section_headings,
+    get_prompt_section_boundary_headings,
 )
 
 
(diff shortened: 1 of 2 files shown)


mito-ai/mito_ai/completions/prompt_builders/chat_system_message.py
```diff
@@ -83,6 +85,62 @@ def create_chat_system_message_prompt() -> str:
 ==== 
 {CHAT_CODE_FORMATTING_RULES}
 
+
+<Example 1>
```
aarondr77: We already have <Example 1> and <Example 2> in this file. It looks like the Example 1 that you added here should be kept and changed into Example 3, and the Example 2 that you copied here should be deleted because its duplicated with the original Example 2

aarondr77: Also, make sure to group this with the existing examples

mito-ai/mito_ai/utils/message_history_utils.py
```diff
@@ -33,7 +34,8 @@ def trim_sections_from_message_content(content: str) -> str:
         GET_CELL_OUTPUT_TOOL_RESPONSE_SECTION_HEADING,
         ACTIVE_CELL_OUTPUT_SECTION_HEADING,
         ACTIVE_CELL_ID_SECTION_HEADING,
-        STREAMLIT_APP_STATUS_SECTION_HEADING
+        STREAMLIT_APP_STATUS_SECTION_HEADING,
+        CODE_SECTION_HEADING,
```
aarondr77: Currently we have `MESSAGE_HISTORY_TRIM_THRESHOLD=3` which means that the previous active code cell or notebook will only be preserved for 2 user messages. I kinda feel like that is not enough. What if we updated this code slightly so that we trimmed all of the variables at 3, but we preserved the notebook for a few more conversation turns. Sometimes I will say things like "Can you put the code back to where it was at X"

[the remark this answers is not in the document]
aarondr77: This is going to be a problem because code will have empty lines in it. It's also possible for the active cell output to have blank lines in it too. We need to fix this and then add tests to make sure it is working properly.

PR #2113, opened 2025-12-17
Paste Image into Chat

ngafar:
# Description

Users can paste images into the ChatInput. 

# Testing

- Copy a picture and paste it into chat. 
- Try different formats (jpg, png, svg).
- Make sure normal copy-and-paste behavior is unaffected.  

# Documentation

N/A

### DIFF
--- a/mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
+++ b/mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
@@ -147,20 +147,34 @@ const ChatInput: React.FC<ChatInputProps> = ({
     };
 
     const handlePaste = async (e: React.ClipboardEvent<HTMLTextAreaElement>): Promise<void> => {
-        const mostRecentClipboardItem = e.clipboardData.items[0];
-        if (!mostRecentClipboardItem) return;
+        const items = e.clipboardData.items;
+        const clipboardItem = items?.[items.length - 1];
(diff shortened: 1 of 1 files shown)


mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -146,6 +146,31 @@ const ChatInput: React.FC<ChatInputProps> = ({
         }
     };
 
+    const handlePaste = async (e: React.ClipboardEvent<HTMLTextAreaElement>): Promise<void> => {
```
aarondr77, line 149: If I paste one image in and then paste another image in, the AI knows that it is receiving two images, but it thinks they are the same image. Only the first one is seen by the AI. 

Unlike if I drag and drop two images in, then both get sent

mito-ai/src/Extensions/AiChat/ChatMessage/ChatInput.tsx
```diff
@@ -146,6 +146,31 @@ const ChatInput: React.FC<ChatInputProps> = ({
...
+        const mostRecentClipboardItem = e.clipboardData.items[0];
+        if (!mostRecentClipboardItem) return;
+
+        if (mostRecentClipboardItem.type.startsWith('image/')) {
```
aarondr77: Do we want to expand this so it does the same behavior for csv, etc files?

PR #2144, opened 2026-01-13
Chart Wizard Converter

ngafar:
# Description

Added the ability to convert charts that don't have the Chart Wizard config section. This lets users edit existing charts with the wizard. 

# Testing

Create a chart without the config section, launch the chart wizard, use the convert tool. 

Converting a chart should not change it's visual appearance. 

# Documentation

Yes, we should include a small blurb on conversions in the same place we write about the chart wizard.

---

> [!NOTE]
> Enables converting matplotlib charts without a config section into the Chart Wizard format and streamlines the widget logic.
> 
> - Backend: registers `POST /mito-ai/chart-wizard` handler using the LLM (FAST_ANTHROPIC_MODEL) with a new prompt builder `create_chart_conversion_prompt`; updates prompt constants (imports must be before config); wires endpoint in `__init__.py`.
> - Frontend: adds a "Convert" flow in `ChartWizardWidget` with loading state and error notifications; extracts code from markdown blocks and updates the active cell; introduces `useChartConfig` and `useDebouncedNotebookUpdate` hooks for parsing config and debounced cell execution; updates parser imports and REST client (`convertChartCode`).
> - Styling: adds disabled button states and minor layout tweaks for the no-config state and loading dots.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 84f9047bf123e74e6dca52588ca9569ccfd5508d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/mito-ai/mito_ai/chart_wizard/handlers.py
+++ b/mito-ai/mito_ai/chart_wizard/handlers.py
@@ -6,8 +6,8 @@ import tornado
 from typing import List
 from jupyter_server.base.handlers import APIHandler
 from openai.types.chat import ChatCompletionMessageParam
-from mito_ai.openai_client import OPENAI_MODEL_FALLBACK
 from mito_ai.completions.providers import OpenAIProvider
+from mito_ai.utils.anthropic_utils import FAST_ANTHROPIC_MODEL
 from mito_ai.completions.models import MessageType
 from mito_ai.completions.prompt_builders.chart_conversion_prompt import create_chart_conversion_prompt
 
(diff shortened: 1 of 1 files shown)


mito-ai/mito_ai/chart_wizard/handlers.py
```diff
@@ -0,0 +1,52 @@
...
+            code = data.get('code', '')
+            
+            # Create prompt using the prompt builder
+            prompt = create_chart_conversion_prompt(code)
+            
+            # Call LLM
+            messages: List[ChatCompletionMessageParam] = [{"role": "user", "content": prompt}]
+            converted_code = await self._llm.request_completions(
+                messages=messages,
+                model=OPENAI_MODEL_FALLBACK,
```
aarondr77: In an ideal world, we want to be using whatever model is the selected model (or at least the selected provider). But that would require a bigger refactor because it would have to go through the handler.py instead of the provider.py (like it is now). 

Let's not do that now. 

But for now, can we use the ANTHROPIC_FAST_MODEL since it's hardcoded for the streamlit work already. That way, we're only relying on one hardcoded model provider.

PR #2154, opened 2026-01-20
Chart Wizard - Active Cell Warning

ngafar:
# Description

When the active cell is not the chart wizard cell, changes made in the chart wizard are not applied.

We should keep the wizard open even if the active cell changes. Users may move around the notebook for many legitimate reasons, and we shouldn’t interrupt or reset the wizard unless the user explicitly indicates they are finished. For example, a user might switch to another cell to copy a hex color code and then return to the wizard.

# Testing

Open the chart wizard, then switch to another cell. You should:

1. See the warning message.
2. Be able to click the button to get to the active cell.

# Documentation

No, users will automatically see when applicable. 

---

> [!NOTE]
> Introduces an in-widget warning and quick navigation when edits may target the wrong cell.
> 
> - Detects active cell changes via `notebookTracker.activeCellChanged`; shows a banner when it differs from the chart’s `cellId`, with a "Go to Chart Cell" action that activates the correct notebook and scrolls to the cell
> - Extends `ChartWizardData` with `app` and `notebookPanelId`; `AugmentedImageRenderer` now stores `app` and passes both to the widget for navigation
> - Adds corresponding styles for the warning UI in `ChartWizardWidget.css`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7a4aeb99d72651914a30291245af94bcd249883b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/mito-ai/src/Extensions/ChartWizard/ChartWizardWidget.tsx
+++ b/mito-ai/src/Extensions/ChartWizard/ChartWizardWidget.tsx
@@ -20,7 +20,7 @@ import {
     isHexColor
 } from './inputs';
 import { useChartConfig, useDebouncedNotebookUpdate } from './hooks';
-import { getActiveCellID, setActiveCellByID } from '../../utils/notebook';
+import { getActiveCellID, scrollToCell } from '../../utils/notebook';
 import '../../../style/ChartWizardWidget.css';
 
 interface ChartWizardContentProps {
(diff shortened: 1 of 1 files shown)


mito-ai/src/Extensions/ChartWizard/ChartWizardWidget.tsx
```diff
@@ -244,6 +271,25 @@ const ChartWizardContent: React.FC<ChartWizardContentProps> = ({ chartData }) =>
...
+            {isActiveCellMismatch && (
+                <div className="chart-wizard-warning">
+                    <div className="chart-wizard-warning-content">
+                        The active cell is no longer the chart cell. Changes may not be applied correctly.
+                    </div>
+                    <button
+                        className="chart-wizard-warning-button"
+                        onClick={() => {
+                            if (chartData) {
+                                setActiveCellByID(chartData.notebookTracker, chartData.cellId);
```
aarondr77: If the user switches to a different notebook (which we know they do), then the `go to chat cell` button does not work. Can we make sure we keep track of the notebook panel and make it the active one.

aarondr77: Use case I have in mind is: Creating a graph in one notebook, but want to look in a different notebook to figure out the colors you previously used, for example.

aarondr77 (approved):

PR #2203, opened 2026-03-07
fix: prevent duplicate error messages on connection failure

darshan1328:
# Description

This PR fixes a bug in   ChatTaskpane.tsx    where error messages were being displayed twice in the chat UI. I consolidated the error handling logic in the _sendMessageAndSaveResponse and initializeChatHistory functions.

## Issue Fixed:
Fixed the duplicate "Upgrade" / Connection Error message bug.

## Motivation & Context:
Previously, the code was calling addAIMessageFromResponseAndUpdateState twice back-to-back in catch blocks—once for the error title and once for the hint. I combined these using logical OR (||) operators to ensure a single, clean message is displayed to the user while still capturing the most descriptive error text available. I applied this fix to all three catch blocks in the file to ensure consistent behavior across streaming, non-streaming, and initialization phases.

# Testing

I tested this by running a local developer build of the JupyterLab extension.

I disabled my internet connection and attempted to send a message. Only a single "Network Error" message appeared instead of two.

# Documentation

No new documentation is required as this is a logic fix for an existing UI component.


ngafar:
Hi @darshan1328, can you please provide some more context on recreating this issue. I've tried:

1. Triggering the "Free Trial Limit Reached" message.
2. Turning off my internet connection.

In both cases I only received a single error message.

ngafar:
Unable to recreate issue, closing this PR.

PR #2219, opened 2026-04-01
Dataframe to Mito AI Context

ngafar:
# Description

Users can now select a range of cells in a dataframe output, and send the selection to Mito AI.

# Testing

**Setup**:

1. Run Mito AI -- `jlpm watch` and `jupyter lab`.
2. Run Mitosheet -- `jlpm watch`.

Next, create a dataframe and select a range. You should see the "Ask Mito AI" button. Click on it, and ask the AI a question. 

# Documentation

Yes, we should call out in our release notes, and a newsletter. 

I don't think we have a section in our docs for our improved default Dataframe outputs, so there might not be a spot in the docs for this feature. 

---

> [!NOTE]
> **Medium Risk**
> Adds a new cross-extension command/bridge that injects DataFrame viewer selections into chat context and auto-focuses the chat input, which could affect message composition/state and prompt content formatting. UI-only changes are low risk, but the new context type flows from frontend to backend prompt generation end-to-end.
> 
> **Overview**
> Enables sending a selected cell-range from the `mitosheet` DataFrame output viewer into Mito AI as *additional context*, via a new Jupyter command (`COMMAND_MITO_AI_ADD_DATAFRAME_VIEWER_SELECTION`) and a lightweight in-memory event bridge.
> 
> When a selection is emitted, the chat taskpane auto-opens/focuses the input, appends the selection as a new `dataframe_viewer_selection` context item (with a brief attention glow), displays it in the selected-context UI, and includes it in backend prompt rendering (`selected_context.py`) under a new “DataFrame output viewer” section.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3cce5464e651fafe0f153433e8bdc9a46833248b. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


mito-ai/src/dataframeViewerContextBridge.ts
```diff
@@ -0,0 +1,33 @@
...
+  return () => {
+    listener = null;
+  };
+}
+
+export function emitDataframeViewerSelection(
+  item: DataframeViewerSelectionContextItem
+): void {
+  listener?.(item);
+}
```
aarondr77: I don't think we need this listener. The reasons we would need a listener is if: 1) We wanted the selection to automatically be applied to the chat without the user pressing the button, 2) We wanted the selection to automatically update if the user changed the selected values. 

We don't do either of these now. So as a result, I think we can simplify this by just sending the selection along with the COMMAND_MITO_AI_ADD_DATAFRAME_VIEWER_SELECTION and using whatever data is there.

mitosheet/src/viewer/dataframeViewerAiContext.ts
```diff
@@ -0,0 +1,67 @@
...
+    bounds: SelectionBounds,
+    sortedRows: any[][],
+    columns: ColumnMetadata[]
+): { display: string; value: string } {
+    const { minRow, maxRow, minCol, maxCol } = bounds;
+    const rowCount = maxRow - minRow + 1;
+    const colCount = maxCol - minCol + 1;
+    const headers: string[] = [];
+    for (let c = minCol; c <= maxCol; c++) {
+        headers.push(columnHeaderLabel(columns[c]));
```
aarondr77, line 46: Nice adding the column headers is a good call!

mito-ai/mito_ai/completions/prompt_builders/prompt_section_registry/selected_context.py
```diff
@@ -82,6 +87,12 @@ def get_selected_context_str(additional_context: Optional[List[Dict[str, str]]])
                 + "\n\n".join(line_selection_strs)
             )
 
+    if len(selected_dataframe_viewer) > 0:
+        context_parts.append(
+            "The user has selected the following cells in a pandas DataFrame. When answering the user's question, plase pay special attention to the data in the selection. In most cases, your answer should be based on the data in the selection, and not the entire DataFrame.\n"
```
aarondr77, line 92: It might be helpful to also include which dataframe it is, which cell it is, etc.

mito-ai/src/Extensions/AiChat/AiChatPlugin.ts
```diff
@@ -76,6 +80,23 @@ const AiChatPlugin: JupyterFrontEndPlugin<WidgetTracker> = {
     let widget = newWidget();
 
     // Add an application command
+    app.commands.addCommand(COMMAND_MITO_AI_ADD_DATAFRAME_VIEWER_SELECTION, {
```
aarondr77: I think if we want to get rid of the listener, we would: 

1. Move this addCommand call to the ChatInput file
2. IN the execute function, instead of emitting the event, we just call the `setAdditionalContext` function directly

aarondr77 (commented):
Sweet looks great. I think there's one simplification we can make to the code and otherwise LGTM!

mito-ai/src/dataframeViewerContextBridge.ts
aarondr77: See this comment for implementation ideas https://github.com/mito-ds/mito/pull/2219#discussion_r3028822987

PR #2245, opened 2026-04-15
Mitosheet - AI Notes

ngafar:
# Description

Adds a new AI Notes/annotation feature. Look for the button in the upper right hand corner. 

# Testing

Use the button, look at some annotations, make sure the fixes work.

# Documentation

Yes

---

> [!NOTE]
> **Medium Risk**
> Adds new LLM/network call paths and JSON parsing, and can apply auto-generated dataframe-mutating code via suggested actions, which increases the chance of incorrect annotations/actions or unexpected data changes.
> 
> **Overview**
> Adds a new **AI Notes** feature that fetches LLM-generated column/cell annotations for the active sheet and exposes them via new API events `get_ai_notes_annotations` and `get_ai_notes_action_code`.
> 
> Introduces a dedicated backend completion path (`ai_notes_backend_completion`) supporting BYO LLM URL, `OPENAI_API_KEY`, or Mito-hosted fallback, plus robust JSON extraction/validation and automatic supplementation of missing-value notes; also adds deterministic pandas code generators for recommended fixes (drop duplicates, drop/fill missing, remove IQR outliers, remove row).
> 
> Updates the UI styling to surface annotations (amber corner markers on cells/headers, popover + taskpane styles, and a toolbar AI actions group) and gates availability to notebook/Streamlit/VS Code via `ai_notes_enabled()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a351ed9532f3b1f14bb97d43bf1793d7d200e9d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ngafar:
bugbot run

mitosheet/mitosheet/api/get_ai_notes_annotations.py
```diff
@@ -0,0 +1,385 @@
...
+    lines.append("Null counts per column:")
+    for col in df.columns:
+        lines.append(f"  - {repr(col)}: {int(nulls[col])}")
+    sample = df.head(_CONTEXT_ROWS)
+    lines.append(f"First {len(sample)} rows (string preview, 0-based row indices):")
+    lines.append(sample.to_string(max_rows=_CONTEXT_ROWS))
+    return "\n".join(lines)
+
+
+def _build_prompt(df_context: str, primary_df: str) -> str:
```
aarondr77, line 46: I think this prompt needs work. It completely omits any description of what we want to be alerting on. If I were the AI and was given this task, I would just guess that I should some of these "outlier|missing|invalid_domain|inconsistency|duplicate|other", but I don't even know what all of them mean. 

For example, what is an `invalid_domain`? Does `duplicate` mean I should flag every time the value appears multiple times? 

Proof that this isn't a good enough prompt is that this is an AI alert I am getting:
1. It is alerting that this is valid -> This isn't a useful alert. It's just clutter! 
2. The recommended action is to remove the row -> If the alert is that its valid data, then the recommended action should not be "remove the row"!

Image

mitosheet/src/mito/components/toolbar/Toolbar.tsx
```diff
@@ -191,34 +192,56 @@ export const Toolbar = (
...
+                                        ...prev,
+                                        currOpenTaskpane: {type: TaskpaneType.SUGGESTED_VISUALIZATIONS},
+                                        selectedTabType: 'data'
+                                    }));
+                                }}
+                            >
+                                <span style={{transform: 'scale(0.85)', display: 'flex'}}><HexagonAIIcon outlineColor='var(--mito-highlight)' fillColor='var(--mito-highlight-very-light)' /></span>
+                                Suggest Graphs
+                            </div>
+                            <div className='mito-ai-actions-group'>
```
aarondr77, line 211: I don't think it makes sense to group these two buttons together. They aren't like a toggle the way that the "Chat and Agent" button is in the AI taskpane. 

Instead, I'd vote we just have three separate buttons
1. Suggested Graphs
2. AI Alerts -> Renamed from AI Notes
3. AI Chat -> Renamed from Ask AI

Image

mitosheet/src/mito/components/toolbar/Toolbar.tsx
```diff
@@ -191,34 +192,56 @@ export const Toolbar = (
...
-                                ...prev,
-                                currOpenTaskpane: {type: TaskpaneType.AITRANSFORMATION},
-                                selectedTabType: 'data'
-                            }));
-                        }}
-                    >
-                        <AIIcon />
-                        AI
-                    </div>
+                    {props.userProfile.mitoConfig.MITO_CONFIG_FEATURE_DISPLAY_AI_TRANSFORMATION && (
```
aarondr77, line 195: Nice! Let's make sure we keep using this if we refactor a bit. It should be used for both the AI alerts and AI Chat

mitosheet/src/mito/components/endo/AINotesPopover.tsx
```diff
@@ -0,0 +1,197 @@
...
+import { MitoAPI } from '../../api/api';
+import { AINotesAnnotation, EditorState, GridState, SheetData, UIState } from '../../types';
+import { getDisplayColumnHeader } from '../../utils/columnHeaders';
+import {
+    getSuggestedAINotesActions,
+    selectAINotesTargetInGrid,
+    useAINotesApply,
+} from '../../utils/aiNotesUtils';
+
+export const AINotesPopover = (props: {
```
aarondr77, line 18: If you select the Column Alert on the final column in a mitosheet that has enough columns to fill the sheet, the alert opens like this. 

Image

mitosheet/css/endo/AINotesPopover.css
```diff
@@ -0,0 +1,190 @@
+/*
+ * Copyright (c) Saga Inc.
+ * Distributed under the terms of the GNU Affero General Public License v3.0 License.
+ */
+
+.mito-ai-notes-popover {
```
aarondr77, line 6: Did we change anything that is effecting the AI generated column hint style? It's looking kinda strange to me now (I'm on Safari). I think before the dotted borders were a lot thinner. 

Edit: Actually, it looks fine on Chrome, so let's ignore this issue for now. 

Image

mitosheet/src/mito/components/endo/AINotesPopover.tsx
```diff
@@ -0,0 +1,197 @@
...
+                                role="button"
+                                tabIndex={0}
+                                title="Select this column in the sheet"
+                                onClick={(e) => { e.stopPropagation(); selectAndDismiss('column'); }}
+                                onKeyDown={(e) => {
+                                    if (e.key === 'Enter' || e.key === ' ') { e.preventDefault(); e.stopPropagation(); selectAndDismiss('column'); }
+                                }}
+                            >
+                                {colLabel}
+                            </span>
```
aarondr77, line 125: In addition to this actually selecting the column, it is opening the AI Alerts Taskpane for me. I think we don't need that

mitosheet/src/mito/components/taskpanes/AINotes/AINotesTaskpane.tsx
```diff
@@ -0,0 +1,276 @@
...
+        | { type: 'error'; message: string }
+        | { type: 'ready' }
+    >({ type: 'loading' });
+
+    const { applyingActionId, actionError: applyError, apply, clearState: clearApplyState } = useAINotesApply(props.mitoAPI);
+    const [applyingAnnotationId, setApplyingAnnotationId] = useState<string | undefined>(undefined);
+    const [errorAnnotationId, setErrorAnnotationId] = useState<string | undefined>(undefined);
+
+    const fetchAnnotations = useCallback(async (focusPreference?: string): Promise<void> => {
+        setState({ type: 'loading' });
```
aarondr77, line 60: Can we improve the loading state? Maybe it can: 

1. Say Fetching AI Alerts with the loading indicator
2. Have some space between the text and the disabled refresh button

Image

mitosheet/src/mito/components/taskpanes/AINotes/AINotesTaskpane.tsx
```diff
@@ -0,0 +1,276 @@
...
+                                            })}
+                                        >
+                                            <button
+                                                type="button"
+                                                className="ai-notes-item"
+                                                onClick={() => {
+                                                    handleAnnotationSummaryClick(a);
+                                                }}
+                                            >
+                                                <div className="ai-notes-item-column">{columnDisplayName(a)}</div>
```
aarondr77, line 207: It's kinda hard to read these alerts and know exactly what it's talking about. In particular, is it a cell alert or a column alert? 

Some ideas:
1. We could update this to say "Alert for: <Column Name> or <ColumnName:CellRow>

Image

aarondr77 (changes requested):
I think the Alerts are a cool idea, but need some love here. The most important change is that we need to improve the quality of the alerts! Right now, they are pretty silly IMO. I shared some examples here. 

The alert popups and cell icons themselves look cool! If we can improve the quality of the alerts then I think this could be a useful feature.

aarondr77:
I think we should update this to be a better CTA and make sure it's inline with the current rate limit strategy. 

- I think the limits reset every month?? 
- I'm not sure that the limit is 100
- Users should be prompted to upgrade to Pro for unlimited completions. 

Nice that we are blocking this here though! 

Screenshot 2026-04-29 at 4 08 27 PM

aarondr77:
I think we need to make sure we give the AI an opt-out of creating cell level alerts. It seems to very often do things like this when there are not good cell alerts. It already has column level alerts (the appropriate alert type) for this

Screenshot 2026-04-29 at 4 11 32 PM

aarondr77:
There seems to be some issue with scrolling when I:

0. Grab the dataset from the mito-ai/testing/datasets/large-data folder.
1. Select a cell level alert (in this video I selected one for a cell that isn't rendered because we only render the first 1500 rows). 
2. Try to scroll back up the dataframe

Notice that it keeps jumping back to the end of the dataframe. 

A similar thing happens if I click on a column alert to scroll to that column, then I can't scroll away from that column. 

https://github.com/user-attachments/assets/acecf8a7-adce-4fb9-9772-8ab549301acd

aarondr77:
The action here doesn't line up with the alert. The alert talks about a specific cell having a formatting issue, but the alert is for dropping missing values. When I use the alert action, it doesn't fix the alert. 


https://github.com/user-attachments/assets/9f908fd8-fc45-4469-a922-5beecf974f4f

aarondr77:
Hmm this alert is kinda confusing because it talks about a range of values, but it calculated this alert based only on the first few rows instead of looking at all of them.

1.  So its calling this out as a outlier, but in the context of other salaries lower in the column, its not
2. The range that it reports is inaccurate. 

One thing we could do is instead of asking the AI to determine issues like this, we could run some existing profiling tools, or create our own. This one is pretty interesting. https://github.com/patelheet30/lintdata

Here are a few ideas (not mutually exclusive)
1. We could create a standard set of tests that we run automatically and feed the results to the AI to pick the most important ones
2. We could do the above, but skip the AI step all together and just use our curated linting results. Then, we could let the user resolve the issues with AI, instead of using AI to also alert
3. We could let the AI write custom python code lints that we execute on its behalf and build the alerts off of. aka: let the AI write code to build alerts instead of using its ability to just look at the data and identify alerts. 

Screenshot 2026-04-29 at 4 22 59 PM

PR #2247, opened 2026-04-15
Fix: Mitosheet not showing graphs

ngafar:
# Description

As of the most recent release (April 15, 2026), creating graphs sometimes results in no chart being shown. The following error appears in the terminal:

> Failed to execute graph function ReferenceError: Plotly is not defined

According to Cursor:

> Root cause: graph execution depended on Plotly being available globally, but Plotly loading was asynchronous and environment-dependent (CDN/RequireJS paths could race or fail), so graph scripts sometimes ran before Plotly existed.

The solution is to add Plotly to the `package.json`, instead of depending on a remote CDN script. 

# Testing

Create a chart with the graph, the suggest graphs and visualize selection buttons

# Documentation

N/A

---

> [!NOTE]
> **Medium Risk**
> Adjusts how Plotly JS is embedded/loaded for graph rendering and updates a large generated Streamlit component bundle, which could affect runtime graph execution if the embed mode or bundle contents are incorrect.
> 
> **Overview**
> Fixes intermittent missing charts by making `get_html_and_script_from_figure` accept Plotly’s `include_plotlyjs` modes (`bool` or script path) when generating the HTML/script payload for the frontend.
> 
> Also updates the built Streamlit component `component.js` (minified bundle) to reflect the new behavior/build output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d371bf1d09b4244232b44425547cdd29e777b3b6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ngafar:
bugbot run

mitosheet/src/mito/components/taskpanes/ControlPanel/SummaryStatsTab/ColumnSummaryGraph.tsx
```diff
@@ -49,16 +50,44 @@ function ColumnSummaryGraph(props: ColumnSummaryGraphProps): JSX.Element {
...
-        }
-        try {
-            const executeScript = new Function(graphObj.script);
-            executeScript()
-        } catch (e) {
-            console.error("Failed to execute graph function", e)
-        }
+        let isCancelled = false;
 
+        const renderGraph = async () => {
```
aarondr77, line 55: Looks like this code is duplicated, can we just make it a function?

mitosheet/package.json
```diff
@@ -33,6 +33,7 @@
     "@types/fscreen": "^1.0.1",
     "@types/react-dom": "^18.3.0",
     "fscreen": "^1.1.0",
+    "plotly.js-dist-min": "^3.5.0",
```
aarondr77, line 36: Ohhh I think this was left over from our testing... We think we don't need this anymore I believe.

mitosheet/src/mito/utils/plotly.tsx
```diff
@@ -4,55 +4,32 @@
...
- * in the standard way does not work - and if you try and use a script you'll get
- * a variety of very confusing errors. 
- * 
- * Thus, we need to take an approach that involves using requirejs, which comes
- * in the AMD context anyways. 
- * 
+ * Load Plotly into `window.Plotly` from the local bundle so graph rendering does
+ * not depend on network access in notebook environments.
  */
-const loadPlotly = (): void => {
```
aarondr77, line 44: Want to try reverting this change and seeing if it still works? I think maybe the only thing we needed to change was the retry loading when plotly was missing logic

aarondr77 (commented):
Okay, wait can we just try updating this one more time to isolate the changes we actually need. I think maybe all we needed was the retry plotly when its still missing logic in the ColumnSummaryGraph.tsx and GraphSidebar.tsx code instead of also bundling plotly

PR #2252, opened 2026-04-16
Fix/delete empty cells on reject 

### LINKED ISSUE #2035: Delete Cell When Rejecting Code
When reviewing code in Agent mode, a user can reject a cell’s changes. However, if rejecting leaves the cell empty, it (the empty cell) currently remains in the notebook. We should automatically delete empty cells after rejection to keep the notebook clean.

ace2016:
# [Improvement] – Delete agent-created cells on rejection instead of leaving empty cells in notebook

## Description

refrence: #2035 

**Context:**
When operating in Agent mode, the Mito AI agent can create new cells in a Jupyter notebook. During the review phase, users can reject individual cell changes or reject all changes at once. However, when a user rejected a cell that the agent had *created* (as opposed to *modified*), the cell was reverted to its `originalCode` — which was an empty string `''` since the cell didn't exist before the agent ran. This left **empty, useless cells cluttering the notebook**, degrading the user experience and requiring manual cleanup.

The root cause was in `AgentReviewUtils.ts`: both `rejectSingleCellEdit` and `rejectAllCellEdits` unconditionally wrote `changedCell.originalCode` back to the cell without distinguishing between "agent modified an existing cell" and "agent created a brand-new cell."

**Approach:**
- **Added an explicit `isNewCell: boolean` flag** to the `ChangedCell` interface in `ChatTaskpane.tsx`. This cleanly distinguishes agent-created cells from pre-existing cells, avoiding an ambiguous `originalCode === ''` heuristic that would incorrectly delete pre-existing empty cells.
- **Populated the flag in `detectChanges()`** in `useAgentReview.ts`:
  - `isNewCell: true` — cell exists in current snapshot but NOT in pre-agent snapshot (agent created it)
  - `isNewCell: false` — cell exists in both snapshots (agent modified it), or was removed
- **Added `deleteCellByIDInNotebookPanel()` utility** in `notebook.tsx`:
  - Deletes a cell by ID using the standard `NotebookActions.deleteCells` API
  - Guards against deleting the last cell in a notebook (JupyterLab requires ≥1 cell) — clears the cell content instead
  - Preserves the active cell index after deletion, clamping to valid range
- **Updated rejection logic** in `AgentReviewUtils.ts`:
  - `rejectSingleCellEdit`: if `isNewCell`, delete the cell entirely; otherwise restore original code and re-run in background (existing behavior preserved)
  - `rejectAllCellEdits`: same branching — delete new cells, restore modified cells

**Impact:**

| Area | Effect |
|---|---|
| **Functionality** | Rejecting agent-created cells now deletes them from the notebook instead of leaving empty cells behind |
| **Correctness** | Pre-existing empty cells are correctly restored (not deleted) on rejection, thanks to the explicit `isNewCell` flag |
| **Edge Cases** | Last-cell guard prevents notebook from becoming empty (JupyterLab invariant) |
| **Developer Experience** | New reusable `deleteCellByIDInNotebookPanel` utility available for future use |
| **Backward Compatibility** | No changes to the accept flow; modified-cell rejection behavior is unchanged |

## Files Changed

| File | Change |
|---|---|
| `ChatTaskpane.tsx` | Added `isNewCell: boolean` to `ChangedCell` interface |
| `useAgentReview.ts` | Set `isNewCell` flag in `detectChanges()` for all three change types |
| `notebook.tsx` | Added `deleteCellByIDInNotebookPanel()` utility with last-cell guard |
| `AgentReviewUtils.ts` | Updated `rejectSingleCellEdit` and `rejectAllCellEdits` to delete new cells |

**4 files changed, 77 insertions, 14 deletions**

## Visual Proof / Evidence

https://github.com/user-attachments/assets/9856862c-0eb7-4d7f-ad3f-bc3207aac557

**TypeScript type-check passes with zero errors:**
```
$ tsc --noEmit
Exit code: 0
(no output — clean build)
```

## Tests

- [x] Static type verification (`tsc --noEmit` — 0 errors)
- [x] Manual verification in JupyterLab (requires running dev instance)

### Test Scenarios:

| # | Scenario | Expected Result |
|---|---|---|
| 1 | Agent creates new cell → Reject single cell | Cell is **deleted** from notebook |
| 2 | Agent modifies existing cell → Reject single cell | Cell **reverts to original code** |
| 3 | Mix of new + modified cells → Reject All | New cells **deleted**, modified cells **reverted** |
| 4 | Agent creates only cell in notebook → Reject | Cell is **cleared** (not deleted — last-cell guard) |
| 5 | Agent modifies pre-existing empty cell → Reject | Cell **restored to empty** (not deleted — `isNewCell: false`) |

Closes #2035


mito-ai/src/Extensions/AiChat/AgentReviewUtils.ts
```diff
@@ -118,13 +127,22 @@ export const rejectAllCellEdits = (
...
+            // Cell was created by the agent — delete it entirely instead of leaving it empty
+            deleteCellByIDInNotebookPanel(notebookPanel, changedCell.cellId);
+        } else {
+            // Cell existed before — restore original code
+            writeCodeToCellByIDInNotebookPanel(notebookPanel, changedCell.originalCode, changedCell.cellId);
+
+            // Re-run the rejected cell in background. We want to make sure that the agent has the 
+            // most up-to-date version of every variable.
+            void runCellByIDInBackground(notebookPanel, changedCell.cellId);
+        }
```
aarondr77: It looks like this code is shared with lines 64-74 above. Can we turn this into a helper function and reuse it in nothing places?

mitosheet/src/jupyter/MitoInstanceRegistry.ts
```diff
@@ -0,0 +1,148 @@
...
+ * Distributed under the terms of the GNU Affero General Public License v3.0 License.
+ */
+
+import { MitoAPI } from '../mito/api/api';
+
+/**
+ * Represents a single entry in the cross-cell undo stack.
+ * Tracks which analysis (cell) performed an edit and when.
+ */
+export interface UndoEntry {
```
aarondr77: It looks like this PR has the UNDO registry PR in here as well. Can you separate this out, so it just has the delete empty cell on reject code in it. Let's keep these PRs separate for reviewing purposes.

aarondr77 (changes requested):
Hey @ace2016 thanks for the contribution here. This is sweet! Appreciate the video showing how it works as well. A little bit of housecleaning here and then I'll get this merged! Appreciate it!

ace2016:
Hello @aarondr77, Both have been addressed,  extracted a `revertCellChanges()` helper for the duplicated logic and separated out the undo registry into its own PR: #2216. Would appreciate a review on that one too when you get a chance 🤙🏾

PR #2275, opened 2026-04-23
Mito AI - MCP Client

ngafar:
# Description

Adds an MCP client to Mito AI. 
 
# Testing

Go to the settings page, add a new MCP server. 

As an example:

```
command: uvx
arguments: mcp-server-time
```

Then ask the Agent to get the time anywhere in the world, it should use the [time](https://github.com/modelcontextprotocol/servers/tree/main/src/time) MCP server.

# Documentation

Yes, we definitely need to update the docs, making clear the MCP server vs client distinction. 

---

> [!NOTE]
> **High Risk**
> Adds new authenticated REST endpoints and agent/tooling paths that can spawn and call external MCP stdio servers, which is security- and stability-sensitive. Also changes strict OpenAI JSON-schema generation, which could impact completion parsing across providers if mismatched.
> 
> **Overview**
> Adds **Model Context Protocol (MCP) integration** end-to-end: the agent can now choose a new `mcp_tool_call` response type, see *available MCP tools* in prompts, and dispatch calls via the `ToolExecutor.execute_mcp_tool` contract.
> 
> Introduces a backend MCP client and config store (`~/.mito/mcp/servers.json`) plus authenticated Jupyter server endpoints to **list/add/delete MCP servers**, verifying connectivity by listing tools and exposing those tools in the Settings UI.
> 
> Tightens OpenAI `json_schema.strict` handling by normalizing object nodes (even when `type` is missing) and recursively forcing `additionalProperties: false`; updates models/tests accordingly, and adds `mcp>=1.0.0` as a dependency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit beb2136b3207946e1ee825fbf0e8ee81a55431bd. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mito-ai-core/mito_ai_core/completions/prompt_builders/agent_execution_prompt.py
+++ b/mito-ai-core/mito_ai_core/completions/prompt_builders/agent_execution_prompt.py
@@ -1,11 +1,18 @@
 # Copyright (c) Saga Inc.
 # Distributed under the terms of the GNU Affero General Public License v3.0 License.
 
-from typing import List
+import json
+from typing import Any, Dict, List, Optional
 
 from mito_ai_core.agent.types import AgentContext
 from mito_ai_core.completions.prompt_builders.prompt_section_registry import SG, Prompt
 from mito_ai_core.completions.prompt_builders.prompt_section_registry.base import PromptSection
+
+
+def _format_available_mcp_tools(mcp_tools: Optional[List[Dict[str, Any]]]) -> str:
+    if not mcp_tools:
+        return "No MCP tools are currently available."
+    return json.dumps(mcp_tools, indent=2)
 
 
 def create_agent_execution_prompt(context: AgentContext, user_input: str) -> str:
@@ -17,6 +24,10 @@ def create_agent_execution_prompt(context: AgentContext, user_input: str) -> str
         SG.Files(context.files),
         SG.Variables(context.variables),
         SG.SelectedContext(context.additional_context),
+        SG.Generic(
+            "Available MCP Tools",
+            _format_available_mcp_tools(context.mcp_tools),
+        ),
         SG.ActiveCellId(context.active_cell_id),
         SG.Notebook(context.cells),
         SG.Task(f"{user_input}"),
--- a/mito-ai-core/mito_ai_core/completions/prompt_builders/agent_execution_prompt.py
+++ b/mito-ai-core/mito_ai_core/completions/prompt_builders/agent_execution_prompt.py
@@ -1,18 +1,12 @@
 # Copyright (c) Saga Inc.
 # Distributed under the terms of the GNU Affero General Public License v3.0 License.
 
-import json
-from typing import Any, Dict, List, Optional
+from typing import List
 
 from mito_ai_core.agent.types import AgentContext
+from mito_ai_core.completions.prompt_builders.mcp_tools import format_available_mcp_tools
 from mito_ai_core.completions.prompt_builders.prompt_section_registry import SG, Prompt
 from mito_ai_core.completions.prompt_builders.prompt_section_registry.base import PromptSection
-
-
-def _format_available_mcp_tools(mcp_tools: Optional[List[Dict[str, Any]]]) -> str:
-    if not mcp_tools:
-        return "No MCP tools are currently available."
-    return json.dumps(mcp_tools, indent=2)
 
 
 def create_agent_execution_prompt(context: AgentContext, user_input: str) -> str:
@@ -26,7 +20,7 @@ def create_agent_execution_prompt(context: AgentContext, user_input: str) -> str
         SG.SelectedContext(context.additional_context),
         SG.Generic(
             "Available MCP Tools",
-            _format_available_mcp_tools(context.mcp_tools),
+            format_available_mcp_tools(context.mcp_tools),
         ),
         SG.ActiveCellId(context.active_cell_id),
         SG.Notebook(context.cells),
--- a/mito-ai-core/mito_ai_core/completions/prompt_builders/agent_execution_prompt.py
+++ b/mito-ai-core/mito_ai_core/completions/prompt_builders/agent_execution_prompt.py
@@ -4,7 +4,6 @@
 from typing import List
 
 from mito_ai_core.agent.types import AgentContext
-from mito_ai_core.completions.prompt_builders.mcp_tools import format_available_mcp_tools
 from mito_ai_core.completions.prompt_builders.prompt_section_registry import SG, Prompt
 from mito_ai_core.completions.prompt_builders.prompt_section_registry.base import PromptSection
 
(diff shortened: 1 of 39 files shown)


ngafar:
Two backend tests are failing, both because the `mcp` package [does not support](https://github.com/modelcontextprotocol/python-sdk/blob/main/pyproject.toml#L6) Python 3.9

ngafar:
bugbot run

mito-ai-core/mito_ai_core/tests/providers/test_anthropic_client.py
```diff
@@ -16,7 +16,7 @@
     "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mP8/wcAAgMBAp9l9AAAAABJRU5ErkJggg=="
 )
 
-def test_mixed_text_and_image():
+def test_mixed_text_and_image() -> None:
```
ngafar, line 19: Doubt these changes are related this PR. These are only to resolve the mypy errors.

mito-ai/mito_ai/completions/agent_loop.py
```diff
@@ -147,6 +148,7 @@ async def start_agent_loop(
         files=metadata.files,
         is_chrome_browser=metadata.isChromeBrowser,
         additional_context=metadata.additionalContext,
+        mcp_tools=await get_available_mcp_tools(),
```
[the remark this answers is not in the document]
ngafar: This seems like a semi-involved lift, I will create a ticket for this, and tackle in a separate PR.

mito-ai-core/mito_ai_core/completions/prompt_builders/agent_tool_result_prompt.py
```diff
@@ -49,6 +50,10 @@ def create_agent_tool_result_prompt(context: AgentContext, tool_result: ToolResu
         SG.Rules(context.additional_context),
         SG.Files(context.files),
         SG.Variables(context.variables),
+        SG.Generic(
```
aarondr77: Let's look up the best practices, but my hunch is that we instead of adding the mcp to the agent-tool-result and the agent-execution-prompt, we should instead just add them all into the system prompt, just like where we define all of the other tools. 

This means: 
1. All of the agent's tools are defined in one place
2. We don't add the mcp definintions into every tool call

The downside is if users add new mcp in the middle of the chat they won't appear. But that is okay I think. When a user adds an mcp let's just have a message appear that says "You need to start a new chat to access it" or something like that.

mito-ai-core/mito_ai_core/completions/prompt_builders/agent_system_message.py
```diff
@@ -333,6 +333,30 @@ def create_agent_system_message_prompt(include_cell_output_tool: bool) -> str:
...
+    "type": "mcp_tool_call",
+    "message": "<string>",
+    "mcp_tool_call": {{
+        "mcp_server_id": "<string>",
+        "tool_name": "<string>",
+        "arguments": "{{}}"
+    }}
+}}
+
+Important information:
```
aarondr77, line 361: There must be a standard or best practices list of instructions to use along with the `mcp` package. Want to see if you can find a few to compare to so we can make sure we're solving for any common failure modes?

mito-ai-core/mito_ai_core/completions/prompt_builders/agent_system_message.py
```diff
@@ -333,6 +333,30 @@ def create_agent_system_message_prompt(include_cell_output_tool: bool) -> str:
...
+1. Only use MCP tools listed in the "Available MCP Tools" section of the current prompt.
+2. The mcp_server_id must exactly match one of the listed server IDs.
+3. The tool_name must exactly match a listed tool for that server.
+4. The arguments field must be a JSON string whose parsed object follows that tool's input_schema. If no arguments are needed, use "{}".
+5. Use this tool for tasks outside notebook mutation when an MCP tool is available (e.g. weather lookup, web APIs, external systems).
+6. If no suitable MCP tool is available, use ASK_USER_QUESTION or continue with notebook tools as appropriate.
```
aarondr77: Number 6 here feels like it's privileging mcp tools ahead of our prebuilt tools. It makes it sound like our tools should only be used when an mcp tool is not available for it. Did we add this to address a specific failure mode?

mito-ai/src/Extensions/SettingsManager/mcp/MCPForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+        e.preventDefault();
+        setIsLoading(true);
+        try {
+            await onSubmit(e);
+        } finally {
+            setIsLoading(false);
+        }
+    };
+
+    return (
```
aarondr77, line 38: Can we have some documentation in here that explains what sorts of MCP tools are supported right now. Maybe we could even list some popular ones as well to help people brainstorm use cases.

mito-ai/mito_ai/completions/agent_loop.py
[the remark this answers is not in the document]
aarondr77: One simple thing we could do is:

1. When the user loads the server for the first time we are already reading all of the tools to display them to the user
2. We could just save the tools to a json file and use that to make it available to the  system prompt.

mito-ai-core/mito_ai_core/completions/prompt_builders/agent_execution_prompt.py
```diff
@@ -17,6 +18,10 @@ def create_agent_execution_prompt(context: AgentContext, user_input: str) -> str
         SG.Files(context.files),
         SG.Variables(context.variables),
         SG.SelectedContext(context.additional_context),
+        SG.Generic(
```
aarondr77: Instead of passing the MCP tools in every agent execution prompt and agent tool result prompt, can we just put them in the system prompt where we define the MCP tool? That way we're treating them just like every other tool, and we don't have to pass them in every turn of the conversation, which seems like major overkill.

The only downside is that if a user adds a new MCP in the middle of a conversation, it won't be usable yet, but I think that's totally okay.

aarondr77: We can also probably read the MCP tools directly from the agent system message builder instead of passing it through as agent context. This is what we do for the default rules. You can see that example here in the agent_system_prompt

```
default_rules = get_default_rules_content()
```

aarondr77 (changes requested):
Sweet this is very cool! I'd love to think of a bunch of use cases and then create blogs/documentation/videos, etc. Would be great marketing! 

There's a bit of cleanup I think we can do and then LGTM

mito-ai/src/Extensions/SettingsManager/mcp/MCPForm.tsx
ngafar: Added a basic description. I didn't want to add too much into to an already narrow form, instead we could do a page in the docs, and link to it.

PR #2280, opened 2026-04-23
Add code and output comments as AI chat context

aarondr77:
# Description

Adds a Google Docs-style commenting system for Jupyter notebooks that integrates with the Mito AI chat. Users can leave comments on code (via a CodeMirror gutter speech-bubble icon) and on cell outputs (via a cell toolbar button). Comments are ephemeral and appear as context pills in the AI chat input, following the same `additionalContext` pattern used by DataFrame viewer selections.

**Code comments**: A CM6 gutter extension shows a speech-bubble icon on the current cursor line. Clicking it opens a popover where the user types a comment. The comment, along with the cell ID, line range, and selected code, is added as a `code_comment` context pill.

**Output comments**: A new cell toolbar button ("Comment on output for AI") appears on cells with output. Clicking it opens the same popover. The comment plus a base64 snapshot of the output (or text fallback) is added as an `output_comment` context pill.

# Documentation

No new documentation needed at this stage. This is a new UI feature with self-explanatory tooltips.

---

> [!NOTE]
> **Medium Risk**
> Adds a new JupyterLab/CodeMirror plugin that injects UI into notebook editors/outputs and wires new commands into chat context, which could affect notebook rendering/interaction if edge cases aren’t handled.
> 
> **Overview**
> Adds **ephemeral notebook comments** that can be attached to Mito AI chat as `additionalContext`.
> 
> Users can add *code comments* from a CodeMirror selection tooltip and *output comments* from a hover button on cell outputs; comments appear as context pills (with a new `CommentIcon`), can be clicked to scroll/highlight the referenced cell/lines, and can be edited/deleted via gutter/output indicators.
> 
> Chat input now registers new comment add/remove commands, de-dupes comments per range/cell, and pushes comment state to the notebook via `COMMAND_MITO_AI_UPDATE_COMMENT_INDICATORS`. Backend prompt building (`selected_context.py`) now renders `code_comment`/`output_comment` context into the final prompt, and Chart Wizard output button styling is shifted to avoid overlapping the new output comment button.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8788b3d5f5ca94327055e068f8838d65af4a2892. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mito-ai/src/Extensions/Comments/AddCommentBubble.ts
```diff
@@ -0,0 +1,147 @@
+/*
```
aarondr77, line 1: Are we even using this file anymore? We aren't using the Gutter, so either we don't need this or the file needs to be renamed

mito-ai/src/Extensions/Comments/CommentsPlugin.tsx
```diff
@@ -0,0 +1,395 @@
...
+    backdrop.className = 'comment-popover-backdrop';
+
+    const popover = document.createElement('div');
+    popover.className = 'comment-popover';
+    popover.style.top = `${rect.bottom + 4}px`;
+
+    // Align the popover's right edge with the button's right edge
+    popover.style.right = `${window.innerWidth - rect.right}px`;
+
+    const textarea = document.createElement('textarea');
```
aarondr77, line 82: Can we make this a react component that we render instead of creating all of the individual components? 

Write a very concise style guide for this in the top-level .cursorrules and agent.md file of mito/ so we don't do this again in the future. They should be identical in both files and when writing these new sections, follow the style of the existing sections!

mito-ai/src/Extensions/Comments/CommentsPlugin.tsx
```diff
@@ -0,0 +1,395 @@
...
+        if (cell instanceof CodeCell) {
+            applySelectionExtensionToCell(cell);
+        }
+    }
+}
+
+/**
+ * Get the cell number (1-indexed) for a cell by its ID.
+ */
+function getCellNumber(notebookPanel: NotebookPanel, cellId: string): number {
```
aarondr77: I think we already have a version of this probably called getCellIndex in the notebook.tsx file. Let's call that function instead of creating a new one here

mito-ai/src/Extensions/Comments/CommentsPlugin.tsx
```diff
@@ -0,0 +1,395 @@
...
+): void {
+    const outputWrapper = cell.node.querySelector('.jp-Cell-outputWrapper') as HTMLElement | null;
+    if (!outputWrapper) {
+        return;
+    }
+
+    // Don't add if already present
+    if (outputWrapper.querySelector('.output-comment-button-container')) {
+        return;
+    }
```
aarondr77, line 223: Is there a more react way of doing this? Like a state variable that keeps track if the comment button is displayed?

It's possible that you think that won't work because we are using code mirror extensions as far as I can tell, and maybe there is some interface between the code mirror and our react code that makes it hard to control / understand if the fields  are rendered. 

Either keep this code if you think it is best, or move to a more react-style if you think that is best.

mito-ai/src/Extensions/Comments/CommentsPlugin.tsx
```diff
@@ -0,0 +1,395 @@
...
+        showCommentPopover(
+            btnRect,
+            async (comment: string) => {
+                const truncatedDisplay = comment.length > 30
+                    ? comment.substring(0, 30) + '...'
+                    : comment;
+
+                let outputSnapshot: string | undefined;
+                try {
+                    outputSnapshot = await getCellOutputByID(notebookTracker, cellId);
```
aarondr77: What is the purpose of this? Why do we need to get the cell output? What do we do with it?

mito-ai/src/Extensions/Comments/CommentsPlugin.tsx
```diff
@@ -0,0 +1,395 @@
...
+                    endLine,
+                    selectedCode,
+                    comment,
+                });
+
+                void commands.execute(COMMAND_MITO_AI_ADD_CODE_COMMENT, {
+                    value,
+                    display: truncatedDisplay,
+                });
+            });
```
aarondr77, line 548: I expected that this code would be part of the onSubmit function of the comment input field Add button. Why is it here?

aarondr77:
Re: [comment on commentGutterExtension.ts](https://github.com/mito-ds/mito/pull/2280#discussion_r2926460898) — Good catch. The file is still used (exports `commentSelectionExtension`, `dismissCommentTooltip`, etc.) but the name was stale. Renamed to `commentSelectionExtension.ts`.

aarondr77:
Re: [comment on CommentsPlugin.tsx L196](https://github.com/mito-ds/mito/pull/2280#discussion_r2926476577)

Good question. The imperative DOM approach here is intentional — this code runs in a JupyterLab plugin `activate` function, which is outside the React tree. JupyterLab owns the output cell DOM nodes, so there's no React parent we can add state to. A React state approach would require either a top-level context provider wrapping the notebook (not feasible) or React portals that still need imperative DOM discovery for mount points.

This matches the established pattern in the codebase — `ChartWizardPlugin` does the exact same thing (`createRoot` + `appendChild` into JupyterLab-owned DOM). So I think keeping the current approach is the right call.

mito-ai/src/Extensions/Comments/AddCommentBubble.ts
```diff
@@ -0,0 +1,147 @@
...
+ * The detail contains the bounding rect for positioning the comment popover.
+ */
+export const COMMENT_TOOLTIP_CLICK_EVENT = 'mito-ai-comment-tooltip-click';
+
+export interface CommentTooltipClickDetail {
+    rect: DOMRect;
+}
+
+// Speech bubble SVG used in the button
+const COMMENT_SVG = `<svg width="12" height="12" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M21 15a2 2 0 0 1-2 2H7l-4 4V5a2 2 0 0 1 2-2h14a2 2 0 0 1 2 2z" fill="currentColor"/></svg>`;
```
ngafar, line 24: I think this might be duplicated in CommentIcon.tsx

aarondr77: For the code mirror extension we need just the svg, so we keep it separate

PR #2299, opened 2026-05-04
Fixed duplicate kernel listeners in context manager

### LINKED ISSUE #2135: Variable Manager Request Execution
When I open this [large-bitcoin-analysis.ipynb](https://github.com/user-attachments/files/24419078/large-bitcoin-analysis.ipynb) and press run All, the cells finish executing, but the kernel never returns to the idle state. 

Steps to replicate: 
1. Open this notebook in Jupyter with Mito AI
2. Run All Cells
3. Hover over the status icon in the top right and notice that there are 105 execution requests (it says running 105 cells). 

**There are 105 execution requests, but only 35 cells**. 35 * 3 = 105, btw :). This happens because inside of the context manager we watch for execute_input messages, which indicate is a request to execute code and then we request a kernel execution to find the content of each variable. 

Either what is happening is: 
1. We are connecting a kernel listener too many times and the same message is getting sent a bunch of times, and/or,
2. We are sending 2 message per cell execution

ace2016:
# Description

Fixes #2135.

This change fixes a duplicate kernel-listener registration bug in the Mito AI context manager. We were registering `iopubMessage` and `statusChanged` listeners more than once for the same `NotebookPanel`, which caused extra silent variable-inspection executions to be triggered during `Run All`.

Before this fix, each real cell execution could trigger multiple context-manager callbacks, which inflated the execution request count and contributed to the kernel appearing stuck after all visible cells had finished.

What changed:
- made context-manager listener registration idempotent per notebook panel
- added regression tests to verify we only register one listener set per notebook
- added regression coverage to verify `execute_input` only triggers one variable refresh callback after re-activating the same notebook

Relevant files:
- `mito-ai/src/Extensions/ContextManager/ContextManagerPlugin.ts`
- `mito-ai/src/tests/ContextManager/ContextManagerPlugin.test.tsx`

Manual verification on the large bitcoin notebook now shows the kernel returning to `Idle` after `Run All`, with the old tripled execution pattern no longer reproducing.

Confirmation screenshot:
image

# Testing

Automated:
- [x] `npx jest src/tests/ContextManager/ContextManagerPlugin.test.tsx --runInBand`

Manual:
- [x] Opened `large-bitcoin-analysis (1).ipynb` in JupyterLab with Mito AI enabled
- [x] Ran `Run All Cells`
- [x] Confirmed all visible cells completed
- [x] Hovered the kernel status indicator and confirmed the kernel returned to `Idle`
- [x] Confirmed the old `3x` duplicate-listener execution pattern no longer reproduced

Notes:
- Manual verification screenshot shows `Kernel status: Idle` and `Execute 68 cells` after notebook completion.
- `68` is consistent with notebook execution plus silent variable inspection, but the kernel-idle bug from #2135 no longer reproduces.

# Documentation

No documentation changes required for this fix.

### DIFF
--- a/mito-ai/src/Extensions/ContextManager/ContextManagerPlugin.ts
+++ b/mito-ai/src/Extensions/ContextManager/ContextManagerPlugin.ts
@@ -38,6 +38,7 @@ export interface IContextManager {
 export class ContextManager implements IContextManager {
     private notebookContexts: Map<string, NotebookContext> = new Map();
     private notebookTracker: INotebookTracker;
+    private initializedNotebookPanels: WeakSet<NotebookPanel> = new WeakSet();
 
     constructor(app: JupyterFrontEnd, notebookTracker: INotebookTracker) {
         this.notebookTracker = notebookTracker;
@@ -88,6 +89,12 @@ export class ContextManager implements IContextManager {
         // which files are available.
         const updatedFiles = await getFiles(app, notebookPanel);
         this.updateNotebookFiles(notebookPanel.id, updatedFiles);
+
+        // Avoid registering duplicate listeners when the same notebook panel becomes active again.
+        if (this.initializedNotebookPanels.has(notebookPanel)) {
+            return;
+        }
+        this.initializedNotebookPanels.add(notebookPanel);
     
         // Listen for kernel restart or shut down events and clear the variables for this notebook
         notebookPanel.context.sessionContext.statusChanged.connect((sender, status) => {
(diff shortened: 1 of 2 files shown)


ace2016:
hello @aarondr77 pls could you take a look at this for a review?

It should fix #2135 by preventing duplicate kernel listeners from getting registered. I also added a regression test and manually verified that `large-bitcoin-analysis.ipynb` now finishes and the kernel returns to `Idle`.

I added a screenshot in the PR for confirmation too.

aarondr77:
Hey @ace2016 thanks for picking up this issue. This is a great contribution and a meaningful improvement. Appreciate you!

PR #2340, opened 2026-05-29
Added Skills - Excel to Python 

ngafar:
# Description

Created a new `read_skill` tool, and added the first skill: `excel_to_python`. This let's us move all of this specialized instructions out of the system prompt, and only add it to context as needed. 

This removes ~2k tokens from the system prompt. For next steps we can remove the database instructions. 

# Testing

Send a generic message, and look at the system prompt. You should only notice that only the skill names and descriptions are included. 

Next, ask the Agent to convert a Excel file. It should load in the skill, and perform the conversion as expected. 

# Documentation

N/A -- Behind-the-scenes change. 

---

> [!NOTE]
> **Low Risk**
> Additive agent/prompt change with in-process skill loading; Jupyter frontend delegates read_skill to the backend, so behavior depends on server-side executor wiring staying correct.
> 
> **Overview**
> Introduces an on-demand **`read_skill`** agent tool so specialized instructions load into context only when needed, instead of living in the default system prompt.
> 
> The agent loop, **`AgentResponse`** schema, parsers, and **`ToolExecutor`** implementations (Python/MCP/JupyterLab) now support **`read_skill`** with a **`skill_name`** field and validation for empty or unknown skills. A new **`mito_ai_core.skills`** package registers skills (first: **`excel_to_python`**) and returns full skill text as a tool result; the system prompt lists only skill names and descriptions via **`format_available_skills()`**, documents the tool, and **removes** the always-on Excel-to-Python rules block (~2k tokens). Jupyter’s TypeScript **`AgentResponse`** type is updated; the frontend tool handler treats **`read_skill`** as unsupported because the backend resolves it in-process.
> 
> Tests cover registry utilities, runner dispatch, and executor behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aba4ae30b3ee460b7b6c17aae0f76a568b1853e4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mito-ai-python-tool-executor/mito_ai_python_tool_executor/executor.py
+++ b/mito-ai-python-tool-executor/mito_ai_python_tool_executor/executor.py
@@ -445,3 +445,14 @@ class PythonToolExecutor:
             output=STREAMLIT_FUNCTIONALITY_DISABLED_MESSAGE,
             variables=vars_,
         )
+
+    async def read_skill(
+        self,
+        ctx: AgentContext,
+        skill_name: str,
+        message: str,
+    ) -> ToolResult:
+        del ctx, message
+        from mito_ai_core.skills.utils import read_skill as read_skill_content
+
+        return read_skill_content(skill_name)
--- a/mito-ai-core/mito_ai_core/agent/agent_runner.py
+++ b/mito-ai-core/mito_ai_core/agent/agent_runner.py
@@ -60,6 +60,7 @@ class AgentRunner:
             "create_streamlit_app",
             "edit_streamlit_app",
             "mcp_tool_call",
+            "read_skill",
         }
     )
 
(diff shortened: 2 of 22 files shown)


mito-ai-python-tool-executor/mito_ai_python_tool_executor/executor.py
```diff
@@ -445,3 +445,14 @@ async def edit_streamlit_app(
             output=STREAMLIT_FUNCTIONALITY_DISABLED_MESSAGE,
             variables=vars_,
         )
+
+    async def read_skill(
```
aarondr77, line 449: Nice work remembering we need to add the skill implementation to all of the tool executors!

aarondr77 (approved):
Sweet! This is working super well. LGTM

